PR #31, opened 2013-09-09
Fix Latex output of don't-care variables.

JoeOsborn:
Forces the unification module's `$dontcare` values--wildcard syntax that would unify against anything--to output in Latex identically to the underscores that K uses for don't-care values.

Sorry for the lousy branch name, but it seems like Github chose it for me and I'm not sure how to change it.


traiansf:
Thanks Joe.  It all seeems fine.

PR #51, opened 2013-10-01
remove an unused (and seemingly unpractical) class

osa1:


osa1:
added one more commit that removes a method with an already existing implementation in one of our dependencies (ApacheCommons library)

Issue #50, opened 2013-09-28
-xml parameter doesn't work


osa1:
while trying to fix some other bug I realized that -xml parameter fails at runtime with the following error:

```
➜  lesson_1 git:(master) ✗ rm -rf lambda-kompiled .k
➜  lesson_1 git:(master) ✗ kompile -xml lambda.k
Exception in thread "main" java.lang.NoClassDefFoundError: org/xmlpull/v1/XmlPullParserException
        at com.thoughtworks.xstream.XStream.<init>(XStream.java:336)
        at org.kframework.parser.DefinitionLoader.loadDefinition(DefinitionLoader.java:96)
        at org.kframework.kompile.KompileFrontEnd.genericCompile(KompileFrontEnd.java:282)
        at org.kframework.kompile.KompileFrontEnd.kompile(KompileFrontEnd.java:250)
        at org.kframework.main.Main.main(Main.java:68)
Caused by: java.lang.ClassNotFoundException: org.xmlpull.v1.XmlPullParserException
        at java.net.URLClassLoader$1.run(URLClassLoader.java:366)
        at java.net.URLClassLoader$1.run(URLClassLoader.java:355)
        at java.security.AccessController.doPrivileged(Native Method)
        at java.net.URLClassLoader.findClass(URLClassLoader.java:354)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:423)
        at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:308)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:356)
        ... 5 more
➜  lesson_1 git:(master) ✗ 
```

maybe some classes are missing in jar files?

traiansf:
Unfortunately -xml has not been maintained lately, and some jars it
depended on were removed from the lib directory.
If we are to still keep this option, we should get back those jars (can be
traced somewhere here: http://xstream.codehaus.org/download.html), or
change the serializer.

best wishes,
Traian

2013/9/28 Ömer Sinan Ağacan notifications@github.com

osa1:
is it unused now or do we need to fix it at some point?

traiansf:
I am not sure.  It was originally supposed to be our intermediate
representation. However, we switched to a binary format and moreover, we
now want to use KORE as the textual representation of the intermediate
language.  Thus being said, the XML contains a lot more information then
Kore will probably do, so it might be good for debugging purposes.
Moreover, Radu said he would like it maintained, but apparently even
[...]

PR #56, opened 2013-10-03
remove xml backend ( #50 )

osa1:
issue #50


traiansf:
Shouldn't we also remove xstream from the dependent libraries?

osa1:
Right .. Added one more commit to remove xstream library. But I'm still not sure it it's complete. `DefinitionReader` still does some xml reading(using `XmlLoader`) .. are those also related with XML backend?

traiansf:
I don't think so.  It seems to be using DOM to parse configuration patterns.

osa1:
OK, this patch should be done, then.

PR #64, opened 2013-10-04
Add a change-log

daejunpark:
I just created a change-log file in order to notice the options changes to the users of nightly build versions.

Please let me know if there is a better place to put this change-log.

### DIFF
--- a/dist/CHANGES
+++ b/dist/CHANGES
@@ -1,0 +1,6 @@
+CHANGELOG
+
+2013-10-04  Daejun Park  <dpark69@illinois.edu>
+
+    * The command-line options are totally changed. Refer to the following page
+      for details: https://github.com/kframework/k/wiki/Command-line-options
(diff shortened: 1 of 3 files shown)


traiansf:
I think this is nice;  however, how do we mmake sure it is maintained?

grosu:
The fact a link to a wiki is provided makes more maintainable.  We will keep updating the wiki as needed.  However, I have two observations/suggestions:

1) Instead of a new file, CHANGES, we should add the one line comment at the top of the README file.  Something like this:

"WARNING: The kompile, krun, kast and ktest command line options have recently (starting with version X) changed!  A list of changes can be found at https://github.com/kframework/k/wiki/Command-Line-Option-Changes."

2) Note that I also changed the title of the wiki above.  The previous title was both inconsistent (the other wikis use capitalized words in their titles) and misleading (because "changes" did not appear in it).

daejunpark:
I did it as you said.

daejunpark:
Fix the problem of PDFs generation. I checked that 'ant release' includes PDFs.

PR #75, opened 2013-10-08
fix issue 67

osa1:


osa1:
I put back the removed method.

PR #167, opened 2013-11-22
Experimental load-cfg option

chathhorn:
This patch adds an experimental "--load-cfg" option that loads a configuration saved using the "--output binary" option into the $PGM configuration variable. This is intended for passing the final configuration from one semantic phase to the next, e.g., from the "static" to the "dynamic" phases of the C semantics.

Could someone please take a look at this, especially the changes to MaudeFilter.java?


grosu:
Denis, you in particular should be interested in this, because we discussed about this feature in our yesterday meeting.

Grigore

chathhorn:
If there are no objections to this, I'm just going to merge it tomorrow.

src/javasources/KTool/src/org/kframework/backend/maude/MaudeFilter.java
```diff
@@ -799,7 +801,7 @@ public void visit(FreezerHole hole) {
 	@Override
 	public void visit(KInjectedLabel kInjectedLabel) {
 		Term term = kInjectedLabel.getTerm();
-        String sort = term.getSort();
+        String sort = term.getSort().equals("K") ? "KList" : term.getSort();
```
traiansf, line 804: Why casting to KList?  I have a bad feeling this might break some implicit assumptions, especially in the Java rewrite engine.

chathhorn, line 804: I think the reason is because there is no "K2KLabel" op defined in [k-prelude.maude](https://github.com/kframework/k/blob/master/lib/maude/lib/k-prelude.maude). Would it make better sense to add that to the maude file instead of filtering it out here? Also, I was thinking that this wouldn't affect the Java backend -- mostly because of the directory structure, I guess (this being in "backend/maude").

PR #237, opened 2014-01-29
Documentation Backend

Multiball:
This is a prototype for a documentation backend for kompile  to generate a reference manual for builtins. Currently, it essentially duplicates the functionality of the latex backend, ignoring rules and recursively exploring imported nodes. Rules can be added to the resulting document by tagging them with the [documentation] attribute or by passing additional attributes to the backend by using the option --documentation.


traiansf:
Good start.  Looking forward to seeing documentation in-between syntax items, too.

radumereuta:
For that we need support from the basic parser.

traiansf:
Yes, I know that.  But I'm still looking forward :-)

PR #239, opened 2014-02-03
Links to Modules in Documentation Backend

Multiball:
This version of the documentation backend adds links to all modules it contains at the beginning of the file, after the filename header. Because Kompile chokes if autoinclude() returns false from the backend, auto-include modules are always included; this is the next problem to fix.


traiansf:
Looks good to me so far.

PR #243, opened 2014-02-10
Added sample documentation comments 

Multiball:
These comments were added to to include/builtins/map.k, include/builtins/set.k, and include/builtins/list.k. As autoinclude cannot currently be turned off, these modules will be present in any file kompiled with the option --backend-doc. 

These comments represent what I think would be helpful to users looking for general-purpose tools and information. The documentation for each builtin module should include a general description of the module, as well as information on usage that may not be mentioned explicitly in the module, such as user-defined lists. For any container sort, information on adding, removing, and updating elements should be present. 

These comments are based on what existing comments there were, which were sometimes cryptic. As I'm not an expert in internal implementations, these comments are spare, and should probably include information about potential misuse or common errors. These comments are meant to be an example and a starting point, not as authoritative and sufficient, and they should be updated by persons more knowledgeable about implementation and use.

### DIFF
--- a/include/builtins/list.k
+++ b/include/builtins/list.k
@@ -15,20 +15,20 @@ module LIST
 
   /*@ \section{User-Defined Lists} It is very common in \K to define a shorthand
    for lists of user-defined sorts. \K 's builtin way of doing this is to use
-   List\left{K,"Separator"\right}, where "Separator" is any valid character or sequence of
+   List\{K,"Separator"\}, where "Separator" is any valid character or sequence of
    characters used to separate distinct elements.
    For example, after defining \\
    syntax K ::= Elt\\ 
    a user could then define
-   syntax Elts ::= List\left{Elt,","\right} \\
+   syntax Elts ::= List\{Elt,","\} \\
    which would be a comma-separated list whose elements are all of sort Elt. A
    user could just as well define \\
-   syntax Elts ::= List\left{Elt,"and"\right} \\
+   syntax Elts ::= List\{Elt,"and"\} \\
    which would be a list containing elements of sort Elt that are separated with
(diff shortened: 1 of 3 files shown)


include/builtins/list.k
```diff
@@ -6,11 +6,43 @@ module LIST
...
+   user could just as well define \\
+   syntax Elts ::= List\left{Elt,"and"\right} \\
+   which would be a list containing elements of sort Elt that are separated with
+   the word "and". If only one argument is given, the separator is asumed to be
+   commas. So, \\
+   syntax Elts = List\left{Elt\right} \\
+   sould define a comma-separated list containing elements of sort Elt.
+  */
+
+
```
traiansf, line 37: You should mention that, unlike lists in functional languages, Lists in K are associative, meaning you can easily access list elements at both ends and concatenate lists.  This also allows one item (MyLisItem)  to be an entire list (without requiring the list terminator_

include/builtins/map.k
```diff
@@ -6,25 +6,38 @@ module MAP
   imports BAG
   imports SET
 
+  /*@\section{Description} The Map represents a generalized associative array;
+   arbitrary keys can be paired with arbitrary values; the values can later
+   be referenced using the key.
+  (TODO: What are all these "MyMap"s doing? how are they different from Maps?)
```
traiansf: MyMap, MyList, MySet are replacements for Map, List, Set.  In the near future they will be renamed as such

traiansf:
Nice,  Could you please take care of the Latex errors, though? (latex compilation just goes through because it is in the nonstopmode, but if you take a look at the log you'll see some problems.

dwightguth:
What are the next steps here?

PR #247, opened 2014-02-25
add DependsOn annotation to denote compiler passes that must run after another compiler pass

dwightguth:


radumereuta:
lgtm

I think putting it in package 'compile' is a good choice for now. Could we need that class for something else? I am trying to justify moving it to a more general package.

andreistefanescu:
lgtm

grosu:
Shouldn't we call it kompile instead of compile, that is,

  package org.kframework.kompile.utils

for consistency with the kompile tool?

dwightguth:
Grigore, you're right; but that's a much larger change. It doesn't belong
here.

radumereuta:
I wouldn't
kompile is the tool, which contains compilation steps. I would still like to make this distinction in the code.

dwightguth:
Okay, I thought about it some more and changed the design a bit. Otherwise we find ourselves in a situation where something that outputs a term that is already flattened doesn't know how to tell anything that that term is flattened.

adamsmd:
What is "NO_CONCRETE_SYNTAX"?  I can't find documentation for it.

adamsmd:
Does this attribute actually enforce DependsOn (if so where is the code to do that?) or is it just unchecked documentation?

radumereuta:
Dwight just added it. But you are right Michael, all the options in that enum should have it's own javadoc description.

adamsmd:
Radu, actually I mean that I can't find where NO_CONCRETE_SYNTAX is defined.

radumereuta:
It's defined in src/javasources/KTool/src/org/kframework/compile/utils/KilProperty.java
Which hasn't been commited yet. It's defined only in this pull request. (I think)

Valid question about enforcing the check at runtime/compile time. I'm curios about the answer. It would be great to be able to force it somehow. It would be great if we can have contracts on our transformers.

adamsmd:
Ah, I missed the fact that this was an enum instead of a class which made the code not make sense.

dwightguth:
This is currently not annotating anything or enforcing anything. But I am adding it because of a change Traian made which caused a particular transformer to depend on this KilProperty, and I want him to take the opportunity to annotate the transformer with @DependsOn(KilProperty.NO_CONCRETE_SYNTAX) and to annotate FlattenSyntax with @Provides(KilProperty.NO_CONCRETE_SYNTAX).

Once we have gone through all the transformers that are part of compile and explicitly declared their dependencies, we will be able to do interesting things like write integration tests which randomly reorder them to see if the output is equivalent. Or allow users to choose which transformations to apply to any given file.

traiansf:
Looks like it could do the job.  Thanks Dwight!

PR #248, opened 2014-02-26
make directory structure slightly more sane

dwightguth:
This is going to be controversial, but honestly, src/java/src and src/sdf is already bad enough. We don't need src/javasources/KTool/src and src/javasources/parser/Concrete. It just adds noise.

I am open to renames though as long as they aren't unnecessarily long.


andreistefanescu:
I am not opposed to this change, but you should have given notice to allow people to commit and push their code.

dwightguth:
I'm not sure I follow. The way I would have expected this to work would be
for me to be responsible for merging stuff that gets pushed before I push
this upstream, and users to be responsible for merging their local changes
following my commit. That's pretty standard procedure. Is there some reason
why that would be a pain point in this particular case?

On Wed, Feb 26, 2014 at 3:06 PM, Andrei Stefanescu <notifications@github.com

andreistefanescu:
I do not want to have to manually merge this with my local changes, since
it's annoying. I would prefer to push my changes, since it is easier for
you to merge, and in order to do that, I need to know when you plan to push.

dwightguth:
I can push whenever people want me to, as long as there is a reasonable
turnaround (i.e. this week sometime).

On Wed, Feb 26, 2014 at 3:21 PM, Andrei Stefanescu <notifications@github.com

radumereuta:
Dwight, we just had a refactoring of the folder structure a few months ago.
Pretty much everybody was involved, and we agreed on how things should look like. We had to take into consideration the workflow, and setting up. We decided to keep the Eclipse/InteliJ porject settings on the repository so we can checkout faster.

I won't agree on this commit until you present the idea behind this change. I can't really understand what you did from the commit.

For the future, modifications that big that affects everybody, I would like a heads up so we can discuss the consequences.

dwightguth:
This is the heads up. We are discussing the consequences now. I don't
intend to push this through unilaterally, so sorry if it came off that way.

I didn't understand the purpose of the javasources folder which contained a
folder containing no java code. So I took the two folders under it, which
were javasources/parser/Concrete and javasources/KTool, which contained
eclipse/IntelliJ projects under them, and put them directly under src. I
will go ahead and update the Eclipse/IntelliJ-specific files.

radumereuta:
Can you present a comparison from the old tree and the new tree? Something like:
root
 |- /bin
 |- /include...
Or a figure, or anything. Or even better. Maybe you can present it at the next live meeting on the white board.

dwightguth:
It's pretty simple

src/
    javasources/KTool/ -> java/
    javasources/parser/Concrete/ -> sdf

dwightguth:
To make things absolutely clear: nothing is going to get reorganized until
we've had a chance for people to support or oppose this change, and until
we're in a state where it can be done effectively with a minimum of
developer disruption. That said, some change does need to be made because
having non-java code under a folder named "javasources" is very clearly
incorrect.

cos:
If a major change will be made, I propose using the Maven directory structure as a guideline: https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html

It is a very well established convention within the Java community.

dwightguth:
Funny story actually. I was looking into what would be required to create bindings for the Gnu MPFR library in Java with a minimum of effort and Maven came up. I'm going to close this pull request in favor of talking about whether we should try to redo the entire build system to use Maven. It's kind of a complete mess with things all over the place right now, and I feel like maybe we should take the opportunity to clean it up.

PR #250, opened 2014-02-27
moved isPredicateLabel from MetaK to KLabelConstant

traiansf:
@dwightguth: is this what you proposed?

### DIFF
--- a/documentation/to-be-processed.txt
+++ b/documentation/to-be-processed.txt
@@ -588,3 +588,10 @@ function: is a label has this attribute, then it is as if it is a builtin functi
 
 anywhere: these are just rules which can apply anywhere they match, so they are expected to be comparatively very slow.  Their application counts as computational steps, so they are interleaved etc. in formal analyses.  It is hard to reason with them in reachability logic.  One should avoid such rules by any means if possible, and use strictness and normal rules instead.
 
+
+KOMPILE - isSort predicates
+---------------------------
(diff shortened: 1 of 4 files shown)


dwightguth:
Looks good, my only question is, is it correct to assume that all klabels that start with "is" are predicates? Will we have any kind of odd behavior if for example the user specifies a klabel "issue"?

traiansf:
Not entirely sure about this.... but the original assumption about this was that regular users will use concrete syntax (which prefixes labels with ') while more advanced users should be aware of the issue.  But I agree in principle we should be more careful here.

dwightguth:
Okay. I will approve this if you take the opportunity to make sure that this piece of information is in the k manual.

grosu:
This is a good opportunity to remind everybody about this file:

  https://github.com/kframework/k/blob/master/documentation/to-be-processed.txt

which is currently supposed to collect items that we need to eventually discuss in the manual.  So for now we can assume that something "is in the k manual" if it is listed as an item in this file.

traiansf:
@dwightguth: would this be satisfactory?

dwightguth:
lgtm. Can someone else who works with kompile weigh in to make sure they understand the change?

radumereuta:
lgtm. I think we agreed that only Dwight will merge the pull requests, so I won't do it for now.

PR #252, opened 2014-02-27
run junit tests on every "ant test" invocation

dwightguth:
We will eventually also add more unit tests, which is why I am putting them as part of the integrated build.


grosu:
Are these unit tests to be invoked with "ant test" going to be different from the tests we run using ktest?  If no, then why bother having two different ways of testing the same thing?  If yes, what are they going to test that is different from ktest?  I know you Dwight want to add a different kind of tests, I'm just curious what kind.  Ones that skip parsing?  Or real junit tests that test each method of combination of related ones?  That is going to be a lot of work.

Grigore

dwightguth:
The goal of using junit is to have unit tests. Ktest is a completely
end-to-end test framework. You run the tools from beginning to end and
compare their output.

Conversely, junit tests are designed to be fairly localized. A junit test
will mock out its dependencies so that the code under test is manageable in
size, and assert that its functionality is as expected. This allows them to
run more quickly and test a wider variety of functionality than end-to-end
tests, which are constrained by what cases within the code real programs
typically invoke. The downside of unit tests is that they are not good for
testing interactions between components, which is why we will continue to
have ktest. However, I am hoping that we can eventually decrease the
end-to-end tests we run as part of our developer workflow to a core case
which test a wide variety of functionality, instead of what we have now
which is a very large test suite that takes a long time and frequently
tests the same features over and over again with only one or two test cases
which test functionality not present in other tests.

traiansf:
JUnit tests are unit tests, in the sense that they test a Java class in isolation.  Our ktest tests are more of integration tests, in the sense that they test that everything goes well together.

One of the main advantages of unit tests is that they localize a problem much more precisely than other tests.  They are also usually incredibly fast :-)

grosu:
Got it, makes sense and it is definitely interesting and useful.

Grigore

dwightguth:
Traian, can I get your feedback on the change itself? You are already familiar with JUnit since these are your tests, so I want to make sure you're okay with the way I have everything configured before I merge this.

traiansf:
The changes look good and seem to achieve the proposed task.  Two (potentially minor) things:
1. I don't know what hamcrest is supposed to be doing.  
2. We should make sure hamcrest and junit are not packed in the distribution (ant pack), as they are probably not needed outside our unit-test.  Looking at this I realise it is probably a bad idea we have the sources to the JIdea K plugin packed in the editor-support dir)

dwightguth:
1. Hamcrest is a dependency of junit 4.11, the latest version which I included in the jar.
2. These are both put in src/javasources/KTool/lib instead of lib/java in order to solve this particular use case.

@denis-bogdanas Can you please submit a change which puts the K idea plugin under src/ and then copies the jar into the editor-support directory?

traiansf:
OK, then I'm fully satisfied.  please feel free to merge it from my part.

2014-02-27 22:35 GMT+02:00 dwightguth notifications@github.com:

PR #255, opened 2014-02-28
write some sample unit tests

dwightguth:
I created these unit tests as some basic examples of what you can do to unit test different types of code. When we institute our testing policy, I will expect people to be familiar with these concepts so that they can test their changes.


traiansf:
looks good to me.

This mockito stuff seems quite interesting.

2014-02-28 18:37 GMT+02:00 dwightguth notifications@github.com:

PR #256, opened 2014-02-28
add checkstyle with 1 check (tabs) and fix it so it passes

dwightguth:
Tabs are bad. They mess up editors and formatting. So we are going to disallow them.


dwightguth:
A quick note: you might find it more effective to review this change if you spot-check a couple modified java files and then download the diff onto your local machine and run git diff -w. That will exclude all whitespace changes which will otherwise make this diff very hard to read.

radumereuta:
At a first glance, it looks ok.

PR #258, opened 2014-03-01
Renamed and upated the tutorial README files (now README.md)

grosu:
Lots of low-level and boring changes to the README files in the tutorial.  Now they are all in the MD format, so they should be easier to process.


dwightguth:
I wasn't able to go through every change here but I looked at some of it and it looks like you pretty consistently made the same types of changes. Since this only touches README files, I'm going to go ahead and merge it.

PR #264, opened 2014-03-04
Added many documentation comments; added a "nodoc" tag.

Multiball:
In this update, I have added comments for several builtin modules, listing them in include/builtins/k-reference.k . Kompile this file with the option --backend=doc to generate k-reference-doc.tex. The comments I have added are far from complete, and there are many sections labeled "TODO" that include some question about functionality that should be filled in by people more knowledgeable than I.

I have added a "nodoc" tag to indicate productions that should not be included in the documentation. This is used to hide productions that are essential to internal mechanics but are never used by end users, especially hooking rules.

The greatest difficulty I've encountered is attempting not to display output from modules not included in the original file (which, thusfar, has been k-reference.k). By recursively importing files, many modules are imported that do not include functionality of interest to the end user, and including them in the reference manual would be a mistake. However, determining which modules are imported from the main file and subsequently not producing or deleting latex output from modules imported from outside the main file has proven to be fairly difficult by the time the definition is passed through the BackendFilter. If anyone has any ideas about this, I would like to hear them.

There are some other things I would like to know:
-What are the available color options for configuration cells? As an end user, I've often wanted to have a complete list.
-Where are the functions of the family asValue()? Those are important to know about.
-How much do we want to say about things not directly included in .k files, like KResults and Configurations? Should talking about those be relegated to a piece of documentation separate from this "reference manual," or is it relevant enough to include?


traiansf:
We used to have a pdf file with all available colors:  
http://k-framework.googlecode.com/svn/trunk/docs/available-colors.pdf?p=6703
(this was the last version) but it was removed from the distribution during some restructuring
I guess it would be nice to have this automatically generated from sources, be them either K or .tex

include/builtins/bool.k
```diff
@@ -9,6 +9,13 @@ endmodule
 module BOOL-HOOKS
   imports BOOL-SYNTAX-HOOKS
 
+  /*@ \section{Bool} Bools are used not only as a builtin Boolean type, but are 
+   also used for side conditions. Any \K statement that evaluates to a Bool can
+   be used as the condition for a side condition, even if it uses a user-defined
+   class or function. This can be useful for, say, defining a particular data
+   structure within \K and then defining a function returning a Bool that checks
+   for membership in that structure.*/
+
```
traiansf, line 21: It should be noted, though, that in order to be usable in a condition, the syntax declaration of a boolean function (i.e., a predicate) needs to be tagged with "predicate" (or "function")

include/builtins/float.k
```diff
@@ -34,6 +37,7 @@ module FLOAT-HOOKS
 
   rule F1:Float =/=Float F2:Float => notBool(F1 ==Float F2)
 
+  /*@ cast from Int to Flaot or vice versa: */
```
traiansf: Flaot => Float

include/builtins/id.k
```diff
@@ -1,6 +1,7 @@
 require "string.k"
 
 module ID-SYNTAX-HOOKS
+  /*@ This is the default definition used for variable names.*/
```
traiansf: Maybe we could add here (or refer to somewhere else) that the notInRules tag instructs the kompiler not to use this definition when parsing rules / .k files

include/builtins/int.k
```diff
@@ -9,6 +9,13 @@ module INT-HOOKS
   imports BOOL-HOOKS
   imports K-EQUAL-HOOKS
 
+  /*@ The following is a list of integer operations in K. \\
+   TODO: I don't know how best to reformat this to give explanations for unusual
+   operators (~Int, \char`\^Int) without destroying the priorities here. I can't just
+   put everything in a separate syntax declatarion like I did for Sets. Do we
+   change the parser to allow for between-syntax-declaration comments? Is there a
+   way to assign priority besides using the > operator?*/
```
traiansf, line 17: Yes, we can use syntax priorities.  There are used in a few places like https://github.com/kframework/k/blob/master/tutorial/2_languages/3_fun/1_untyped/2_substitution/fun-untyped.k
(lines 104-111).  This would deserve some documentation on its own...

include/builtins/int.k
```diff
@@ -52,6 +59,11 @@ module INT-HOOKS
   // rule absInt(I:Int) => I when I >=Int 0
   // rule absInt(I:Int) => 0 -Int I when I <Int 0
 
+  /*@ The following is a list of boolean operators on Integers. These follow C-style conventions
+   Except for  ``is not equal to", which is ``=\=Int" instead of ``!=Int".
+   ``=Int" is equivalent to ``==Int"; both check for equality. 
+   ``dividesInt" performs modulus and checks to see if the result is 0.*/
```
traiansf, line 65: There are some distinctions between =Int and ==Int, but I don't feel the most qualified to talk about them.  Andrei, Grigore?

include/builtins/k-equal.k
```diff
@@ -6,7 +6,11 @@ module K-EQUAL-HOOKS
   imports K-SET-EXTRAS
   imports K-MAP-EXTRAS
 
+  /*@\section{Equality}
+   This module defines common helper functions used to test equality, inequality, and inclusion between some builtin \K constructs.*/
```
traiansf: Less characters per line, pls.  at least less then 120 would help with github.

include/builtins/k-equal.k
```diff
@@ -14,28 +18,27 @@ module K-EQUAL-HOOKS
   syntax priorities equalEqualK notEqualEqualK > boolOperation
   rule K1:K =/=K K2:K => notBool(K1 ==K K2)
 
-
-  syntax K ::= "#if" Bool "#then" K "#else" K "#fi"   [function, hook(KEqual:ite)]
-  syntax KItem ::= "#if" Bool "#then" KItem "#else" KItem "#fi"   [function, hook(KEqual:ite)]
-  syntax Int ::= "#if" Bool "#then" Int "#else" Int "#fi"   [function, smtlib(ite)]
+  syntax K ::= "#if" Bool "#then" K "#else" K "#fi"   [nodoc, function, hook(KEqual:ite)]
```
traiansf, line 23: if should be documented, as it is available to users (and actually used in some definitions).

On thing to make the users aware of here is that #if_#then_#else_ evaluates all its arguments before it chooses the branch.

include/builtins/k-equal.k
```diff
@@ -58,32 +61,33 @@ module K-EQUAL-HOOKS
...
+  /*@ inequality over maps */
   syntax Bool ::= Map "=/=Map" Map   [function]
   rule Map1:Map =/=Map Map2:Map => notBool(Map1 ==Map Map2)
 
 
-  // equality over KList
+  /*@ equality over KList */
   syntax Bool ::= KList "==KList" KList    [function]
 
+  /*@ TODO: What's this? Is it important? */
```
traiansf, line 74: This is how ==KList is implemented. I think it's safe to have a nodoc tag for it.

include/builtins/k-equal.k
```diff
@@ -92,10 +96,12 @@ module K-EQUAL-HOOKS
   rule ListItem(_:K) _:List ==List .List => false
   rule .List ==List .List => true
 
-  // disequality over lists
+  /*@ inequality over lists */
   syntax Bool ::= List "=/=List" List [function]
   rule L1:List =/=List L2:List => notBool(L1 ==List L2)
 
+  /*@ These declarations define ``=" as an alias for ``==" for the relevant types.*/
```
traiansf, line 105: Again, these are not precisely aliases.  They have something to do with symbolic calculus.

include/builtins/k-reference.k
```diff
@@ -0,0 +1,105 @@
...
+at the beginning of the document. This manual is a work in progress; if there's a 
+question whose solution you were unable to find in this manual, please send an 
+email to (TODO: get email address) detailing the nature of the question and a 
+suggestion as to where it might best be placed.
+*/
+
+
+
+/*@ \section{Builtin Literals}
+K has predefined the lexer rules for the most common syntactic categories that are directly accessible through subsorting:
```
traiansf: maybe shorter line?

include/builtins/k-reference.k
```diff
@@ -0,0 +1,105 @@
...
+
+\begin{verbatim}
+syntax #Id ::= Lexer{[a-zA-Z][a-zA-Z0-9]*}
+syntax #Id -/- [a-zA-Z0-9]
+\end{verbatim}
+
+Note that this works for simple expressions that have a repeated character set
+at the end. If you want to specify something more complex, you will have to use
+\verb^Lexer{...}^ and specify the follow restriction.
+
```
traiansf, line 86: Maybe add here the discussion about notInRules / notInPrograms.  Radu, what were all the options here?

include/builtins/list.k
```diff
@@ -13,7 +13,7 @@ module LIST
    Mylist,Mylist construct in combination with the MyListItem construct. It's 
    worth noting that \K lists aren't exactly like lists in functional languages;
    they're associative, which means that it's easy to access elements at both
-   ends of the lists and concatenate them. For example, L:Mylist,E accesses
+   ends of the lists and concatenate them. For example, ``L:Mylist,E" accesses
```
traiansf: Tehnically speaking, you would have to elaborate E a little here, something like
MyListItem(E)

include/builtins/map.k
```diff
@@ -59,7 +62,7 @@ module MAP
 //  rule update(M1:MyMap, K1:K |-> K2:K M2:MyMap) => update(M1[K1 <- K2], M2)
 //  rule update(M:MyMap, .MyMap) => M
 
-  /*@ map domain restriction (TODO: what does this mean? Remove key from map? This will aslo remove the resulting value, right?) */
+  /*@ map domain restriction (TODO: what does this mean? Remove key from map? This will aslo remove the associated value, right?) */
```
traiansf: yes, that's the idea.  it's the k version of Map.remove

include/builtins/random.k
```diff
@@ -3,7 +3,8 @@ require "int.k"
 module RANDOM-HOOKS
   imports #RANDOM-INTERFACE
   imports INT-HOOKS
-  
+
+  /*@ TODO: What is this? A random number generator with a seed?  */
```
traiansf, line 7: actually, the seed needs to be initialized at execution time. this just request the ith number in the sequence.

include/builtins/set.k
```diff
@@ -1,12 +1,13 @@
...
 module SET
   imports K-EQUAL-HOOKS
   imports INT-HOOKS
+  imports DEFAULT-CONFIGURATION
 
- /*@ \section{Description} The Set construct behaves like a mathematical
+ /*@ \section{The K Set} The Set construct behaves like a mathematical
   set: it is a collection of unique items. The builtin \K Set offers support
   for some common set operations.
   (TODO: How does one add an element to a Set? How does one remove an element
```
traiansf: maybe we should have explicit functions for these, but right now we count on rules to help us here:
matching a set in LHS and mentioning it with an additional element in the RHS will add the element.
rule <barrier> true </barrier> <k> wait ...</k> <id> X </id>  <waiting> W:MySet (. => MySetItem(X)) </waiting>  when notBool(X in W)
matching a set with an explicit element in LHS and mentioning it without that element in the RHS will remove the element.
rule <barrier> false </barrier> <k> wait => . ...</k> <waiting> W:MySet (MySetItem(X) => .) </waiting>

include/builtins/string.k
```diff
@@ -11,47 +11,70 @@ module STRING-HOOKS
...
+  syntax Bool ::= String "=/=String" String      [function, left, hook(#STRING:_=/=String_)]
+  rule S1:String =/=String S2:String => notBool(S1 ==String S2)
 
+  /*@ String length: */
   syntax Int ::= "lengthString" "(" String ")" [function, klabel(lengthString), hook(#STRING:lengthString)]
+
+  /*@ Get the character at the given index (TODO: check this): */
   syntax String ::= "chrChar" "(" Int ")"   [function, klabel(chrChar), hook(#STRING:chrChar)]
+
+  /*@  TODO: What's this? */
```
traiansf: the ASCII code for a character. defined only if the string is an ASCII character.

include/builtins/string.k
```diff
@@ -11,47 +11,70 @@ module STRING-HOOKS
...
+  /*@ String equality: */
   syntax Bool ::= String "==String" String [function, left, hook(#STRING:_==String_)]
+  /*@ String inequality */
+  syntax Bool ::= String "=/=String" String      [function, left, hook(#STRING:_=/=String_)]
+  rule S1:String =/=String S2:String => notBool(S1 ==String S2)
 
+  /*@ String length: */
   syntax Int ::= "lengthString" "(" String ")" [function, klabel(lengthString), hook(#STRING:lengthString)]
+
+  /*@ Get the character at the given index (TODO: check this): */
```
traiansf: The ASCII character corresponding to the given ASCII code

include/builtins/string.k
```diff
@@ -11,47 +11,70 @@ module STRING-HOOKS
...
   syntax String ::= "chrChar" "(" Int ")"   [function, klabel(chrChar), hook(#STRING:chrChar)]
+
+  /*@  TODO: What's this? */
   syntax Int ::= "ordChar" "(" String ")"  [function, klabel(ordChar), hook(#STRING:ordChar)]
 
+  /*@ Get a substring given by the start and stop indices, respectively: */
   syntax String ::= "substrString" "(" String "," Int "," Int ")" [function, klabel(substrString), hook(#STRING:substrString)]
+  /*@ TODO: I don't know what these do, and it's not obvious. */
```
traiansf: look at find and rfind in the Maude Manual
I'm not sure about the Char versions, though

include/builtins/string.k
```diff
@@ -11,47 +11,70 @@ module STRING-HOOKS
...
   syntax Int ::= "String2Int" "(" String ")"	 [function, klabel(String2Int), hook(#CONVERSION:string2int)]
   syntax String ::= "Int2String" "(" Int ")"	 [function, klabel(Int2String), hook(#CONVERSION:int2string)]
-
-
   syntax String ::= "KLabel2String" "(" KLabel ")"  [function, klabel(KLabel2String)]
   syntax KLabel ::= "String2KLabel" "(" String ")" [function, cons(KLabel1String2KLabelSyn)]
 
+  /*@ In the first String, replace all occurences of the second string with the third: */
   syntax String ::= "replaceAll" "(" String "," String "," String ")" [function]
+  /*@ In the first String, replace the Int'th occurrence of the second String with the Third:*/
```
traiansf: Are you sure about this?

include/builtins/string.k
```diff
@@ -11,47 +11,70 @@ module STRING-HOOKS
...
+  /*@ Trim whitespace from the left side of a String: */
   syntax String ::= "ltrim" "(" String ")" [function]
+  /*@ Trim whitespace from the right side of a String: */
   syntax String ::= "rtrim" "(" String ")" [function]
   
-  syntax Bool ::= String "=/=String" String      [function, left, hook(#STRING:_=/=String_)]
-  rule S1:String =/=String S2:String => notBool(S1 ==String S2)
 
+
+  /*@ TODO: Figure out what these do. Some sort of inclusion, probably. */
```
traiansf: String comparison (lexicographic, I believe)

include/builtins/string.k
traiansf: Apparently the char versions are for locating the first character of the needle.  I wonder why

grosu: Also, it would be nice to use the "functional" notation for such operations, that is,

  syntax String ::= substrString(String,Int,Int) [function, klabel(substrString), hook(#STRING:substrString)]

instead of

  syntax String ::= "substrString" "(" String "," Int "," Int ")" [function, klabel(substrString), hook(#STRING:substrString)]

Grigore

include/builtins/k-reference.k
grosu: We should stick to 80-character lines of text, everywhere.

Grigore

include/builtins/int.k
grosu, line 65: = is the "mathematical" equality, which reduces to theorem proving, while == is the structural one.  Andrei, can you please send a pertinent paragraph explaining the difference between them, so Kevin can incorporate it in the documentation?

Grigore

include/builtins/string.k
```diff
@@ -11,47 +11,70 @@ module STRING-HOOKS
...
   syntax Bool ::= String "<String" String [function, hook(#STRING:_<String_)]
                 | String "<=String" String [function, hook(#STRING:_<=String_)]
                 | String ">String" String [function, hook(#STRING:_>String_)]
                 | String ">=String" String [function, hook(#STRING:_>=String_)]
 
+  /*@ TODO: Figure out what these do. I have no idea. */
   syntax String ::= categoryChar(String) [function, hook(#STRING:categoryChar)]
```
dwightguth, line 94: This gives you the unicode General Category and Bidirectional Character Type of a character.

src/javasources/KTool/src/org/kframework/backend/latex/DocumentationFilter.java
```diff
@@ -23,17 +29,33 @@ public void visit(Module mod) {
...
+        /*
+        String currName = imp.getFilename().trim();
+        String mainName = GlobalSettings.mainFile.getAbsolutePath().trim();
+        if(currName.equals(mainName)){
+            System.out.println( "Adding " + imp.getName() + " to set." );
+            importedFromMainFile.add(imp.getName());
+        } else{
+            System.out.println("Imported " + imp.getName() + " from " + currName + ", which is not main file " + mainName);            
+        }
+        */
```
dwightguth: If this is commented out, does it need to be checked in at all?

src/javasources/KTool/src/org/kframework/backend/latex/DocumentationFilter.java
```diff
@@ -46,4 +68,23 @@ public void visit(Rule rule) {
...
         if(process) super.visit(rule);
     }
+
+    @Override
+    public void visit(Syntax syn){
+        super.visit(syn);
+        if(firstProduction){
+            //because no productions were processed, delete the syntax block declaration
+            int start = result.lastIndexOf("\\begin{syntaxBlock}");
+            result.delete(start, result.length() -1);
```
dwightguth, line 68: Can I see the output of a file that uses this line of code? I want to be sure that it works correctly,

dwightguth:
Traian had a lot of good comments, so I just added a couple things. I would like to see the output of this filter though, to make sure that it looks reasonable and doesn't have any presentational bugs.

Multiball:
I've revised the comments with respect to the new information I've 
received. There are some things I still need to correct, like the 
difference between the == operator and the = operator, pending 
information on the difference. I'm still not sure how "syntax 
priorities" are supposed to work; the fun.k example Traian linked me 
seems to use the same Syntax syntax as int.k, using the '>' character to 
define priority.

I've nodoc'd all the productions containing the '#' character 
indiscriminately, thinking these constructs to be important only to 
internal mechanics. Is this not true? What is this '#' character? When 
would the end user need and/or want to use it?

The pdf of all available colors is exactly the kind of thing I wanted 
when using K. It mentions that some palates are available under 
"options;" what are these "options?" They look like latex package names. 
Is this relevant to the end user? Similarly, does .tex source exist for 
these? It'd be nice if there were something I could more or less 
copy/paste. Restructuring that thing in latex seems like a lot of tedium 
with a good chance of getting the colors wrong.

To check the output of the filter, run
$kompile --backend=doc k-reference.k
in include/builtins. The K-EQUALS-HOOKS module contains some productions 
that are tagged "nodoc".

traiansf:
W.r.t. xcolor:
1. we load all those options with the xcolor package when we compile the
latex file, so the user should have all of them available.
2, the pdf I sent you is cut from this larger pdf documenting xcolor:
http://ftp.univie.ac.at/packages/tex/macros/latex/contrib/xcolor/xcolor.pdf

I guess the sources for it can be found following the installation
instructions:
http://www.ctan.org/tex-archive/macros/latex/contrib/xcolor
(usually the latex package contains sources for documentation, too.

w.r.t. syntax priorities, I also gave you line numbers
if you look there you will see something like:

  syntax priorities '@_
                  > '__
                  > arith
                  > '_:=_
                  > 'let_in_ 'letrec_in_ 'if_then_else_
                  > '_;_
                  > 'fun_
                  > 'datatype_=__

which orders productions by things (klabels, tags) identifying them.
Radu can tell you more about it.

best wishes,
Traian

2014-03-04 22:24 GMT+02:00 Multiball notifications@github.com:

include/builtins/k-reference.k
radumereuta, line 86: There are 3 parser generated at kompile time:
1. program parser. It takes the exact syntax of the syntax module and generates parser from that
2. rule parser. It takes all the syntax declarations from all included modules and mix it with K (cells, meta-variables, lists, bags...etc), except for Ids
3. ground parser. Just like the above, except for meta-variables, which are replaced with Ids

Now, one can annotate a production with one of the following:
notInRules, notInGround, notInPrograms to specify to komple that you don't want those productions to be included in a specific parser. This is usually useful for avoiding ambiguities. Most common case is Ids with meta-variables.

Something else to add would be 'onlyLabel' which should be considered deprecated and has the same value as notInRules and notInGround, if used at the same time.

radumereuta:
The deal with priorities is at follows.
In SDF one can specify a priority relation between two productions in order to specify how to disambiguate ambiguous trees.

A typical use is for arithmetic operations like \* and +. Without a priority between the two, the string 1+2_3 would create the following ambiguous parse forest:
amb('___('_+_(1, 2), 3),
        '_+_(1,'___(2, 3)))
Assuming the normal operator priority, the second parse is the one we are interested in. To solve this, we can specify that '___ > '_+_ and SDF would reject any branch where \* is a parent of +.
Note 1: that the filter is applied to the parent only when there is an outermost non-terminal.
In the case of Exp "[" Exp "]" (arrays), the filter is applied only on the left expression.
Note 2: priorities are transitively closed a > b, b > c => a > c

There are two ways to specify priorities:
1. Inline. This means you are restricted to only one sort category.
syntax Exp ::= Exp "_" Exp [mul]
                    > Exp "+" Exp [add]
                    |  Exp "-" Exp [sub]
2. with tags
syntax priority mul > add sub
syntax priorities '___ > '_-_ add

Associativity. The principle is the same for left and right associativity.
syntax Exp ::= Exp "*" Exp [mul]
                    > left:
                       Exp "+" Exp [add]
                    |  Exp "-" Exp [sub]
                    > Exp "=" Exp [right]
Block priority (for + and -) specify how to handle ambiguities between + and -, while production priority specify how to disambiguate when having the same type of production.
Note 3: productions inherit the associativity from the block, but can be overridden.

---

Is this enough to continue with the documentation?
If you have any more questions I can answer. I prefer that someone who is a native English speaker would write the final documentation. I don't feel that confident in my writing abilities.

grosu:
Radu,

Please add this to the "manual" as well.

Grigore

include/builtins/k-reference.k
grosu, line 86: Radu:

1) Please add this to the manual as well

2) Can you remind us why we need the ground parser?  Can't that just be a special case of the rule parser?  Please explain that in the manual, too.

Grigore

dwightguth:
ok to test

radumereuta:
JENKINS please test this.

radumereuta:
JENKINS: retest this please

radumereuta:
Hmm... funny, I've just tested the program that fails, on my local machine, and it appears to be working correctly.
Does someone else have any idea why it's failing?

dwightguth:
I will investigate what Jenkins says about this test on Monday. If you want
to get to it sooner, the Jenkins URL is http://fslwork.cs.illinois.edu:8080/

Look for the build corresponding to that pull request, and pull up its
console log.

grosu:
What is this?  Can somebody explain us how this works?  Sorry if it was explained and I missed it.

What happens if you forget to say "please"?  Will JENKINS reply saying "Please be polite!"?  :)

Grigore

dwightguth:
ok to test

dwightguth:
test this please

dwightguth:
Hey, can you fix your merge conflicts and then see if this issue with ktest is still happening?

cos:
Will this ever get merged?

PR #265, opened 2014-03-04
my first pull request; addressed comments from code review

yilongli:


dwightguth:
lgtm. I need to fix Jenkins to make sure that it demonstrates that it passes all the tests, and then it can be pushed as long as Andrei S approves.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/PatternMatcher.java
```diff
@@ -250,7 +259,6 @@ public void match(Term subject, Term pattern) {
                "kind mismatch between " + subject + " (" + subject.kind() + ")"
                + " and " + pattern + " (" + pattern.kind() + ")";
 
-        assert !subject.isSymbolic() : "the subject term shall be ground";
```
andreistefanescu, line 253: If you want, you can put "assert subject.isGround()"; the "assert !subject.isSymbolic()" should be true not only for the main term, but also for every subterm. However, I don't think the assertion is necessary here.

dwightguth:
Okay, I'm still fiddling around with Jenkins to try to set things up correctly for testing, but in the meantime, Yilong, you can merge this as long as you've run ant test first.

PR #267, opened 2014-03-04
New parser. Adding NullabilityCheck. (First pull request).

radumereuta:
Should be reviewed by Michael.


src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
```diff
@@ -0,0 +1,125 @@
...
+            } while (changed);
+
+            return nullable;
+        }
+
+        /**
+         * Recursive DFS that traverses all the states and returns a set of all reachable {@Link NonTerminal}. 
+         * @param start The state from which to run the collector.
+         * @return A set of all reachable {@Link NonTerminal}.
+         */
```
adamsmd: The algorithm as you have written it is quadratic  (ret.addAll() is the culprit.)  Instead of returning a Set<NonTerminal>, pass that as parameter and have each call just add to it.  You should do the same for "visited" so it is not being "globally" modified.

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
```diff
@@ -0,0 +1,125 @@
...
+                        if (st instanceof ExitState) {
+                            nullable.add(ee.getKey());
+                            dependencies.remove(ee.getKey());
+                            changed = true;
+                            break;
+                        } else { // nextable states
+                            assert st instanceof NextableState : "Expected nextable states here!";
+                            if (st instanceof PrimitiveState) {
+                                PrimitiveState ss = (PrimitiveState) st;
+                                Set<MatchResult> matchResults = ss.matches("", 0, null);
```
adamsmd: Note that this fails for regex containing lookahead.  I'm not sure if we want to do it now, but the eventual "correct" solution is to have "PrimitiveState" contain a method that returns whether it is nullable.  Then each implementation can determine that for itself.  For RegExState, we might just require the user to pass a flag to the constructor.

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
radumereuta: What's wrong with 'global' fields? I have a specialized class that handles this case without passing arguments at each call.

I want to solve the return quadratic issue the same way I did for 'visited'.

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
```diff
@@ -0,0 +1,125 @@
...
+        Helper hp = new Helper();
+        return hp.checkNullability2(startNt);
+    }
+
+    /**
+     * Helper class that holds temporary information for the nullability check.
+     * @author Radu
+     *
+     */
+    private static class Helper {
```
dwightguth: Why have an instantiable inner class? Why not just instantiate the outer class and put all the functionality there?

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
adamsmd: They make the code harder to reason about and introduce invarients that could be violated.  For example, at line 48, you have to make sure to clear the "visited" otherwise things break.  Passing "visited" as a parameter prevents the problem of that invariant.

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
```diff
@@ -0,0 +1,125 @@
...
+        public Set<NonTerminal> checkNullability2(NonTerminal startNt) {
+            Set<NonTerminal> all = getReachableNT(startNt.entryState);
+            all.add(startNt);
+            visited.clear();
+
+            // insert all the non-terminals and their start state
+            for (NonTerminal nt : all)
+                dependencies.put(nt, new LinkedList<State>(Arrays.asList(nt.entryState)));
+
+            boolean changed = false;
```
dwightguth: boolean changed;

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
radumereuta: I want to be able to call this function without instantiating any class, like: NullabilityCheck.getReachableNullableNT(exp)

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
```diff
@@ -0,0 +1,125 @@
...
+            do {
+                changed = false;
```
adamsmd: The control flow of this function is hard to understand (e.g., breaks inside while loops, etc.).  Is there a simpler way to write it?  I think this function, would be easier to implement in recursive style instead of work-list style.  (Recur on the "next" and "child".)

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
radumereuta: I know, but we're not sure yet how to implement the PrimitiveStates. Maybe we want to implement them directly in the parser. I don't know... I'll leave it for you to decide.

dwightguth: If I understand correctly, the issue is that a regex that contains nothing but lookahead doesn't match the empty string but in fact should be treated here as if it were empty because that's what it actually captures? Can you give me an example of a K grammar that would capture this problem?

src/javasources/KTool/src/org/kframework/parser/concrete2/Parser.java
```diff
@@ -399,12 +399,12 @@ public int hashCode() {
...
-                    value = new NonTerminalCall(key);
-                    ntCalls.put(key, value);
-            }
-            return value;
+        NonTerminalCall value = ntCalls.get(key);
+        if (value == null) {
+                value = new NonTerminalCall(key);
+                ntCalls.put(key, value);
+        }
+        return value;
```
dwightguth, line 407: If this is just a whitespace change, can you revert it? We try not to do too much rewriting of code that isn't directly affected by the change we're in the middle of.

src/javasources/KTool/test/org/kframework/parser/concrete2/ParserTest.java
```diff
@@ -1,14 +1,20 @@
 package org.kframework.parser.concrete2;
```
dwightguth, line 1: Looks to me like you're using junit 3... Can you upgrade to the annotation-based way of specifying test cases?

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
adamsmd: Yes, Dwight, that understanding is correct.  When we get to the point of using "first sets" to get LL style performance, these will happen all the time because there will be an analysis that adds those.  I am not sure if there is a current K grammar that has this problem.  (Thus why I say we don't necessarily need to fix this now, but it does eventually need to change to the correct solution.)

src/javasources/KTool/test/org/kframework/parser/concrete2/ParserTest.java
```diff
@@ -41,7 +47,7 @@ public void testEmptyGrammar() throws Exception {
         Term expected = amb(klist(amb(KList.EMPTY)));
 
         assertEquals("Empty Grammar check: ", expected, result);
-
+        assertEquals("Expected Nullable NTs", 1, NullabilityCheck.getReachableNullableNT(nt1).size());
```
dwightguth: Please assert what the nullable state is.

src/javasources/KTool/src/org/kframework/parser/concrete2/Parser.java
radumereuta, line 407: I'll have to revert my entire commit.
But how should I handle whitespace fixes like this? Make a separate commit just for this?

src/javasources/KTool/test/org/kframework/parser/concrete2/ParserTest.java
```diff
@@ -164,6 +173,7 @@ public void testListOfTokens() throws Exception {
                 System.out.println("Time: " + ((end - start) / 1000000.0));
             }
         }
+        assertEquals("Expected Nullable NTs", 1, NullabilityCheck.getReachableNullableNT(nt1).size());
```
dwightguth: Please assert what the nullable nonterminal is.

src/javasources/KTool/test/org/kframework/parser/concrete2/ParserTest.java
```diff
@@ -201,6 +211,7 @@ public void testNestedNonTerminals1() throws Exception {
             Term expected = amb(klist(amb(klist(Token.kAppOf("K", "x"), amb(klist(Token.kAppOf("K", "x"), amb(KList.EMPTY), Token.kAppOf("K", "y"))), Token.kAppOf("K", "y")))));
             assertEquals("x^ny^n check: ", expected, result);
         }
+        assertEquals("Expected Nullable NTs", 1, NullabilityCheck.getReachableNullableNT(nt1).size());
```
dwightguth: Please assert what the nullable nonterminal is.

src/javasources/KTool/test/org/kframework/parser/concrete2/ParserTest.java
```diff
@@ -232,6 +243,7 @@ public void testNestedNonTerminals2() throws Exception {
             Term expected = amb(klist(amb(klist(amb(klist(amb(KList.EMPTY), Token.kAppOf("K", "y"))), Token.kAppOf("K", "y")))));
             assertEquals("y^n check: ", expected, result);
         }
+        assertEquals("Expected Nullable NTs", 1, NullabilityCheck.getReachableNullableNT(nt1).size());
```
dwightguth: Please assert what the nullable nonterminal is.

src/javasources/KTool/test/org/kframework/parser/concrete2/ParserTest.java
```diff
@@ -263,6 +275,7 @@ public void testNestedNonTerminals3() throws Exception {
             Term expected = amb(klist(amb(klist(Token.kAppOf("K", "x"), amb(klist(Token.kAppOf("K", "x"), amb(KList.EMPTY)))))));
             assertEquals("x^n check: ", expected, result);
         }
+        assertEquals("Expected Nullable NTs", 1, NullabilityCheck.getReachableNullableNT(nt1).size());
```
dwightguth: Please assert what the nullable nonterminal is.

src/javasources/KTool/test/org/kframework/parser/concrete2/ParserTest.java
```diff
@@ -333,8 +353,8 @@ public void testNestedNonTerminals5() throws Exception {
 
         for (int i = 2; i < 10; i++) {
             Grammar.NonTerminal nt = new Grammar.NonTerminal(new Grammar.NonTerminalId("NT"+i),
-                    new Grammar.StateId("NT"+1+"Entry"), new Grammar.State.OrderingInfo(-i),
-                    new Grammar.StateId("NT"+1+"Exit"), new Grammar.State.OrderingInfo(2*i-1));
+                    new Grammar.StateId("NT"+i+"Entry"), new Grammar.State.OrderingInfo(-i),
```
dwightguth, line 366: Can you please help me understand why this was changed from 1 to i? Like, was it wrong before? If it was wrong before, why didn't the test fail? Etc.

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
adamsmd: That can be done with something like:

... getReachableNullableNT(exp) {
  new NullabilityCheck().checkNullability2(startNt);
}

However, I think the instance fields can be better handled as parameters in which case, everything becomes static.

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
dwightguth: I'm sorry, maybe I'm not following. What I wanted to know was under what circumstances in a K definition a RegexState would have a lookahead pattern. Are you saying that feature is currently not required by any features of K, or that the current implementation doesn't do this because you will be implementing a step later that adds lookahead patterns to certain RegexStates?

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
dwightguth: Okay, but that doesn't answer the question. You could still have the following:

public class NullabilityCheck {
  public static Set<NonTerminal> getReachableNullableNT(NonTerminal starrtNt) {
    NullabilityCheck check = new NullabilityCheck();
    return check.checkNullability(startNt);
  }

  private NullabilityCheck() {
  }
}

src/javasources/KTool/src/org/kframework/parser/concrete2/Parser.java
dwightguth, line 407: Hmm. I need to think about this further. Question: did the unusual whitespace in this method significantly impair readability of the method, or did you just make the change in order to be more "correct"?

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
adamsmd: I am saying three things:
1. I am not sure if Token{"abc"} or anything like that that _currently_ allows "abc" to include a lookahead.
2. But in the future it will.
3. Certain transformations on grammars that we will do in the future to "optimize" the grammar, will definitely add certain RegExStates that contain lookahead patterns.

src/javasources/KTool/test/org/kframework/parser/concrete2/ParserTest.java
radumereuta, line 366: Michael?

src/javasources/KTool/src/org/kframework/parser/concrete2/Parser.java
radumereuta, line 407: Both.

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
dwightguth: Sorry, I'm confused. maybe I'm missing a line of code in here somewhere, but as far as I can tell getReachableNT gets called by itself and by the first line of checkNullability2. checkNullability2 gets called by only the entry point. So in point of fact line 48 is completely superfluous and could be deleted. Right?

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
radumereuta: Even right now you may have Regular Expressions that have follow restrictions. Even Ids have follow restrictions.
When you write
syntax Id ::= Token{[a-zA-Z][a-zA-Z0-9]*}
The front end will generate automatically a follow restriction (lookahead) that checks to see if the current location is not followed by any of the [a-zA-Z0-9].

The problem in this location is when you have such a lookahead for an empty string: "" -/- [a-z]
Something like that.

src/javasources/KTool/test/org/kframework/parser/concrete2/ParserTest.java
adamsmd, line 366: It was wrong before.  It probably passed because the "StateId" is used only for minor things like printing/debugging and (possibly) the lowest priority element of the compare operator.

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
dwightguth: Okay. I suggest you go ahead and make the change that Michael is suggesting now. Go ahead and keep the naive solution in RegexState.isNullable right now, but please make the refactoring. Also include a TODO at that code location.

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
adamsmd: Yeah, I noticed that, but it comes down to what invarients hold in the code.  Line 48 can be deleted as long as getReachableNT and checkNullability2 isn't called anywhere else.  But of course, code evolves.  Passing as a parameter and avoiding the global is a good way to protect against against that.  My point isn't about line 48 in particular, but rather why the unnecessary use of globals is bad.

src/javasources/KTool/test/org/kframework/parser/concrete2/ParserTest.java
dwightguth, line 366: So duplicate StateIds are legal then?

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
radumereuta: I forgot to check for visited nodes on checkNullability2. Which is wrong. It could go into infinite loops. I will fix it.

src/javasources/KTool/src/org/kframework/parser/concrete2/Parser.java
dwightguth, line 407: Okay. Well you changed it from having 8-space indents to having a mixture of 8-space indents and 4-space indents. I'm okay with having a code policy of only 4-space indents, but in that case you need to change it the rest of the way, and then we need to change the rest of the codebase and implement a checkstyle policy. Otherwise we're going to be having stuff like this happen all the time. Personally I'd prefer to avoid going to that extreme, but there's no middle ground here. Either we insist on an indentation policy or we don't.

radumereuta, line 407: It's not a mixture of 8 and 4. We want only 4, but it was 8. So it was a correction. Now it's more readable and clear that it's part of a single block. (you have to look over the original file to see the difference)

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
dwightguth: Okay, I concede your point here. Use class-based state for fields in a data type or configuration, but use local state for intermediate results in an algorithm.

Go ahead and do what Michael suggests.

src/javasources/KTool/src/org/kframework/parser/concrete2/Parser.java
dwightguth, line 407: I talked with Andrei S and he talked me into this. As long as we only make whitespace changes to 4-space indents, I am okay with what you did here. But I note that the if block is still indented 8 spaces...

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
dwightguth: Also: note I'm not very happy with the idea of just passing a flag in because I'm not sure how that really solves the problem, it just pushes it back a layer to another part of the code that has to decide what to pass in. But we can worry about that later, because as I said, I suggest we leave the naive solution for now, as long as it's appropriately refactored right now.

src/javasources/KTool/src/org/kframework/parser/concrete2/Parser.java
radumereuta, line 407: Ah, I missed that. It wasn't that visible. I will fix it in the next commit.

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
dwightguth: I think it might be possible to continue to do this iteratively and still simplify it a lot. Maybe I'm misunderstanding the algorithm, but can't you get rid of the changed field and just iterate until you've removed everything from the map as having been completely processed? and then in the inner loop you only break if you don't need to know the rest of the states to know the answer. And then use an iterator to iterate over the map and call iterator.remove so that you don't have to break out of the for loop every time you change it.

radumereuta: How do you iterate over a map?

dwightguth:
```
Iterator<Entry<NonTerminal, Queue<State>>> iter = dependencies.entrySet().iterator();
while (iter.hasNext()) {
    Entry<NonTerminal, Queue<State>> ee = iter.next();
    ...
    iter.remove();
}
```

radumereuta: But that iterates over the Set, not the Map.

dwightguth: It iterates over a java.util.HashMap$EntrySet object which represents a view into the original map which supports updates. The iterator behaves the same way.

radumereuta:
I case it didn't send a notification, I've pushed another commit to address some of your issues.
Since your reviews were quite helpful, I would appreciate another look.

radumereuta:
I've added another commit with some more comments in the hope the code is better explained this way.
I prefer the work-list approach rather than the recursive one. It's closer to the way I think. I find it more intuitive.

I've cleaned up some unnecessary complications in the implementation thanks to your suggestions.
If there isn't anything else, could you merge this so I can continue with the next steps of the parser?

grosu:
You should never wait for a merge in order to continue to work on your thing.  I've had a similar problem with the tutorial.  Read the wiki that Dwight wrote.  Create a branch and work there, then merge that into your master branch (in your clone) after the previous changes have been merged by Dwight into the main master.  In other words, we should not put pressure on Dwight to merge our thing asap, because otherwise we cannot continue our work.  These are orthogonal.

Grigore

src/javasources/KTool/src/org/kframework/parser/concrete2/Parser.java
grosu, line 407: Guys, please also clarify these things in the wiki, because that is our basis for these conventions.

Grigore

grosu, line 407: I think we should.  And definitely we should use spaces and not tabs; all IDEs allow you use use spaces instead of tabs.  Also, is anybody updating our src/.idea as we change the conventions?

I see that the instructions in the src/README file are for Eclipse.  Shouldn't we change them to be for ItelliJ?  I think we have no Eclipse developers at this moment.  Even I started to use IntelliJ (two days ago :)).

Grigore

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
grosu: Is the Token{...} production explained in the manual?  If not, it should.  Radu, can you please add a quick description of it, as a reminder that we should describe it thoroughly eventually?

Grigore

dwightguth:
I'm not sure I can approve this diff until I understand it better. I did some looking into the problem of nullability in a grammar and found the following algorithm which as far as I can tell is n^2*m where n is the number of productions and m is the number of symbols per production.

```
repeat
    for each production X → Y1 Y2 · · · Yk
        if Y1 . . . Yk are all nullable (or if k = 0)
            then nullable[X ] ← true
until nullable did not change in this iteration.
```

Unless I'm mistaken your algorithm has a worse running time than this one. Is there some subtle nuance of the general case of CFGs that makes this algorithm not work correctly?

dwightguth:
JENKINS: retest this please

grosu:
Dwight, it would be nice to display the output of ktest in this message.  That is more informative than just that the tests passed.  I'd like to know each time how long it took to run all the tests, how many languages were tested, how many programs, etc.  That is meaningful info.  If we want more statistics, we can add them to ktest's output.  Omer offered to continue to help with the development of ktest, so we can make feature requests.

Thanks,
Grigore

dwightguth:
Grigore: unfortunately I don't have a way to configure what is posted onto github as a comment. The build in Jenkins was misconfigured though; eventually I intend to send out email notifications like the other jenkins build does. I think it has all that information already.

JENKINS: retest this please.

dwightguth:
JENKINS: retest this please

dwightguth:
Okay apparently the regex doesn't support line breaks. Good to know.

dwightguth:
Congratulations, Radu, you have tabs. I suggest you reconfigure your IDE to automatically convert tab characters to spaces as you type them, then fix your commit.

radumereuta:
Can you tell me where exactly? I've just tested my editor and it inserts only spaces.

dwightguth:
If you look at the build log from Jenkins you'll see the first instance in
the file is on line 345. But there might be others because it only reports
1 per file. I suggest you run "ant checkstyle".

src/javasources/KTool/src/org/kframework/parser/concrete2/NullabilityCheck.java
```diff
@@ -0,0 +1,139 @@
...
+     * Starting from the start symbol, find all the non terminals that can match on the empty string.
+     * @param startNt The start non terminal.
+     * @return A set of all reachable non terminals that are nullable.
+     */
+    public static Set<NonTerminal> getReachableNullableNT(NonTerminal startNt) {
+        NullabilityCheck nc = new NullabilityCheck();
+        return nc.checkNullability2(startNt);
+    }
+
+    Set<NonTerminal> nullable = new HashSet<>();
```
dwightguth, line 34: Can these be moved into local state of the method if they're intermediate products of the algorithm?

dwightguth, line 34: Actually before you do that please address my larger concern with the algorithm you chose.

dwightguth:
I see a couple small things, but I'm still concerned about the larger issue that I don't understand why this algorithm was used instead of the standard algorithm in compiler textbooks for computing Nullability sets.

grosu:
Thanks, Dwight, for getting into the details of this!  I wish I also had time to do it, because I think that the front-end parser is one of the most visible parts of the framework right now, which can have a big first impact on our users.  Omer, you said that you are also interested in parsing issues.  Please feel free to also step in and take a look at what Michael and Radu are doing these days.

The idea is to implement a super-duper parser, to eliminate our need and dependence upon SDF.  BTW, our new rewrite engine is almost at the stage where the dependence to Maude is removed, too.

Grigore

radumereuta:
Dwight, what do you mean it's not standard? It pretty much is. It's basically a topological sorting, trying to find the dependency of nullable NonTerminals.

It may appear that the implementation is not standard, because the data structures I'm running the algorithm is a bit different. I'm running it on the state machine instead of productions. This way I'm avoiding the dependency on the front end structures that are frequently changed.

I'll try to address some of your other concerns shortly.

dwightguth:
Hmm okay I see what you're trying to do. Basically you want to try to sort the graph into as close to topological order as possible so that you can process all the nodes in close to linear time.

This is not at all clear from the algorithm however, and it is considerably worsened by the fact that you seem to be mixing together the two components of the algorithm.

I did some looking into the problem of sorting graphs with cycles and I came up with the feedback vertex/arc problems. I assume this is where your exponential behavior arises from?

Still, I think this could be made a lot clearer. I suggest you break this in half. The first half does the sorting and produces a linear list of productions/nonterminals/whatever. And the second half performs the simple and easy-to-understand algorithm that I posted in a previous comment.

radumereuta:
Regarding your proposed algorithm. That might become exponential for the example that I just give you, just because of the fact that you want to produce that linear list of "productions/nonterminals/whatever".

I admit. The algorithm I implemented is not the most efficient, but it's less to write. The complexity is something like O(nr_prod \* longest_nullability_chain).
The best that I can think of would be O(nr_prod + longest_nullability_chain).
You can change from \* to + if you keep a reference to what productions are being blocked by the currently discovered nullable NT. Basically, in the worst case scenario I, if I have the dependency A < B < C < D ... < X, and X is at the end in my mapping, I have to go through every production for each relation.

If you really really insist that my chosen solution is that horrible, I could spend a few more hours to try and implement this algorithm as efficient as possible, but I doubt it would make it more readable.

dwightguth:
Hmm. And there really isn't a way to make a division of labor between
conceptually distinct pieces of the algorithm without sacrificing
asymptotic running time in the complexity of the cycle? Like, can you
demonstrate to me in terms that I can understand that we literally must
choose one or the other? Because if that's the case I'm fine with the more
unreadable algorithm, you just need to comment it differently to make it
more clear what it is that it's doing in order to achieve the running time
that it has.

radumereuta:
1. I don't know how to make it more readable.
2. It is good enough for now.
3. I have other things to worry. I have delayed the new parser for far too long. I don't have time to make every single piece of code perfect. If this becomes an issue later on I will come back.

Sorry

grosu:
Radu, your attitude is inappropriate here and we do not want to set such a precedent.  You spend a lot more time making some code work than polishing it, so polishing it and making it understandable by others, including by yourself later, is worth the time.  So please stop saying "I have other things to worry".  Yes, I agree that the parser has been delayed for way too long, but that is not because you write high quality code.  That never slows things down overall.

Grigore

dwightguth:
Some concrete suggestions that might help:
1. Find ways to make methods that do one specific thing and then make the algorithm composed of such methods. For example, one of the reasons I am so unhappy with this code is because you are mixing a topological sorting algorithm with something else entirely. If you can demonstrate to me that that's an unavoidable necessity in order to achieve the correct running time, I will be satisfied. Otherwise this code needs to be made more modular.
2. When you have state variables that serve roles in the computation, document clearly what they do. For example, you might comment precisely what states are stored in the values of the map that you are constructing at any time.
3. (I just thought of this, so sorry for not telling you earlier) Name your variables. For example nullableNonTerminals, allNonTerminalsReachableFromStartState, statesToPutBackInMap. And of course rename variables like ee.getValue(), st, ss, etc. Which I can't even tell you what might be good names because I have no idea what information they represent.
4. Work on a separate change while waiting for code review on the change you started now.
5. Try to write comments from the perspective of someone unfamiliar with the code. For example, I have no idea what <><><><> is supposed to represent even after you explained it to me, so it can't possibly explain anything on its own.
6. Whenever you write a comment, ask yourself whether the code could be modified somehow to make the information the comment provides unnecessary. (This doesn't count for stuff like javadocs, etc).

Go ahead and try to make another pass and see if you can construct something that I can understand without needing to have the algorithm explained to me by you.

dwightguth:
Also wow I just thought of this also, so sorry to not bring it up earlier.

Why exactly is the following algorithm exponential?
1. approximate a topological sorting of the graph. (i.e. come up with something that would be topologically sorted if you removed just enough arcs to make the graph a DAG). Polynomial time.
2. for (fixed point, linear in number of productions because each iteration except the last causes at least one production to be completely processed).
3. ->for each production
4. ->->for each symbol in the production
5. ->->->if the symbol is not nullable, the production is not nullable
6. ->->if each symbol is nullable, the production is nullable

Looks polynomial to me.

dwightguth:
An addendum to Grigore's point: Industry has a conception of what is called "technical debt". You incur technical debt by making decisions now that will cost time later. By making these decisions, you speed up development in the short term, but over longer periods of time it causes the opposite to happen. The cheap fast solutions you chose are hard for others to understand, or are full of bugs that you need to correct later. Over time this causes project progress to stall as increasingly large amounts of time have to be spent fixing bugs and trying to understand code that someone else wrote after that person leaves the team. Or worse, the bugs never get fixed and the code never gets understood, in which case the code quality decreases and the code gets increasingly messy as developers make changes to code they don't fully understand. This creates a downward spiral that will eventually make your stakeholders very unhappy with you, which is the situation we are in right now.

The solution is to pay off the technical debt. By spending small amounts of time right away to ensure that each piece of code is in the best possible state for maintenance later, you eventually reach a point where old portions of the code don't need to be touched at all, freeing up very substantial amounts of time for the development of new functionality. Unfortunately we are simply not there right now. But in order to get to that state, it's essential that we don't check in anything further that is hacky or unclean unless we can't possibly avoid it.

radumereuta:
It's been pointed out to me that I didn't explain well enough the context in which I wrote this code.

I assume that the point of having nullability check is understood. Usually this is done at the production level, but considering the dynamics of K, and how often we change the way productions look, I don't see it feasible to write the algorithm by analyzing productions. Instead I'm trying to apply a reachability algorithm on the state automaton graph generated from the productions. That contains only 4 basic elements which will not change too much.

That's why viewing it as a standard topological sorting algorithm is not enough. I'm sorry for not making that clear earlier. I will try and rewrite the documentation in order to better explain the context in which this has to run.

If there is still something unclear, please let me know.

dwightguth:
Wait a minute. The complexity of the algorithm derives from the fact that
it's not being done on a grammar? I assumed that it was the same algorithm
either way, just adapted differently.

If that's the case then we need to change that design assumption. create a
representation of a grammar in this package if you don't want to use kil
(which is a choice I agree with). But don't make the algorithm vastly more
complex.

I believe what we should have is the following (although you are free to
correct me if you think alterations neef to be made) :
1. A basic parser that takes a set of strings and parses everything except
   the rules into the kore/kil format. We already have this I think.
2. A piece of the parser which transforms kil of the grammar into a
   suitable internal representation of the grammar. Some adjustments for
   modularity can be made to this piece at a later date. Its output would be a
   very basic grammar data structure. Production, Terminal (with regex) and
   nonterminal. That's it.
3. A piece of the parser which constructs the parsing table with states and
   so forth from the grammar. This piece will adjust itself based on whether
   the grammar is LL, LR, etc. For now we assume all grammars require full
   generality. I don't know what work has already been done, if any, for this
   piece.
4. The parser itself, which takes the parse table and the string to parse
   and returns kil/kore. This is Michael's component.

This class belongs in piece #3 when we still have access to the original
grammar. From what you say that should make the algorithm much more
readable. Is that correct?

Also, a quick note. In the future, to avoid this type of problem, please
submit a design of all future components in the parser to me for a design
review. that way we don't have to change problematic designs at the last
minute, which is very expensive in developer time.

radumereuta:
1. The need for intermediate representation for the grammar is questionable. Yet another layer to maintain when the transformation from grammar to states is straightforward.
2. Regarding future designs, I doubt that's even possible. We are still figuring stuff out. Anyway, probably Michael would have a better perspective on this.

Anyway, I think it's pointless trying to lay down in writing every aspect of the parser. I had many meetings with Michael in front of the white board trying to grasp as many details as possible.
If you still want to review the parsing code, we should have a few live meetings in which we go through every aspect of the design. When will you reach Urbana?

grosu:
They also had this notion of "technical debt" at Microsoft when I did my sabbatical there.  Microsoft was quite serious about it, serious enough to post the problem to their researchers.  Some researchers who I know well were considering the idea of developing an account for each programmer, and to have Microsoft put $ in it that earn interest, but that $ could only be accessed after years of work.  However, in time, as bugs are found in their code, technical debt would be paid from that account.  If they don't have bugs, then they get more $ after a few years.  If they end up with a negative balance for too long, they are fired.  Not sure if they ever implemented that ...

Grigore

adamsmd:
Dwight, the nullability check is one of the later stages of pre-processing of a grammar.  Therefore it may not have any relation to a K production.  In addition, we have future extensions to K planned that make K level grammars _not_ be production based.

adamsmd:
Radu, this algorithm should not involve topological sorting.  The following is the correct (untested) algorithm.  This algorithm is linear.  (You may have gotten confused with the orderingInfo algorithm which does involve topological sorting.)

Nullability of a state is based on the following two implications:
 A. EntryState state => nullable(state)
 B. nullable(state) && childNullable(state) => nullable(state.next)
Where childNullable(state) is true when it is possible to get from the start of the state to the end of the state without consuming input.

The following algorithm is a least fixed-point algorithm for solving those implications.  mark(state) is called when we discover an implication implying nullable(state).  We can discover this implication one of three ways:
1. A state is an entry state (see rule A)
2. The nullable(state) in rule B becomes true (in which case we check childNullable).)
3. The childNullable(state) in rule B becomes true (in which case we check nullable(state).)

(ChildNullable(state) becomes true when an exit state becomes nullable.)

```
// A state is nullable iff the *start* of it is reachable from the entry of its nt without consuming input
// A non-terminal is nullable if its exit state is nullable
Set<State> nullableStates;

foreach (nt : nts) { mark(nt.entryState); }

// marks a state as nullable if it is not already, and calls mark on any
// states that should be nullable as a result.
void mark(state) {
  if (!nullableStates.contains(state)) {
    nullableStates.add(state);
    if (state instanceof NextableState) {
      if (childNullable(state)) {
        for (s : state.next) {
          mark(s);
        }
      }
    } else {
      assert state instanceof ExitState;
      // previous calls to childNullable would have returned False
      // so now we restart those recursions
      for (s : state.nt.callers) {
        if (nullableStates.contains(s)) {
          for (s : state.next) {
            mark(s);
          }
        }
      }
    }
  }
}

// returns true if the state being nullable should imply that the next states are nullable
boolean childNullable(state) {
  EntryState -> True
  PrimitiveState -> state.null
  NTState -> nullableStates.contains(state.child.exitState);
}
```

grosu:
Well, I think that having meetings with Dwight to convince him that what you are doing makes sense, or that you are making the right compromises, is cheating.  The implementation should be clean and well documented so that he and others understand it without having to be explained in front of the whiteboard.  If your code is very prototype-ish because you are not sure yet what you are doing, then that should be in a branch until you figure it out.  If the code is good enough to be merged, then you should also take the time to polish and document it (again, the latter should take much less than the code itself, so even if you end up throwing it away because you change the algorithm, it is still OK).

That being said, I think it is of course OK to meet and talk to Dwight :-)

Grigore

dwightguth:
I am in the process of drafting a longer response but this discussion is
evolving quickly and I want to address some points more briefly to avoid
confusing my larger draft.

To Michael:

You said "K level grammars may not be production based". But you use the
word "grammar" which is a very clearly well-defined concept referring to a
set of productions. This is quite confusing to me. Are you imagining having
productions anywhere in your internal representation of the parser? Because
if so, then you will need the layer I suggested which converts KIL into a
grammar. And if not, I am quite skeptical and need to be persuaded that
this is a good idea, because you will be inventing large amounts of new
algorithms which are all incredibly well-described and solved problems when
dealing with grammars of the form that computer science has expected us to
describe languages in for a number of decades now.

dwightguth:
To Grigore: the purpose of having design reviews is not to make sure you
have the right design, or to explain your design. It's to attempt to
address design-level concerns as much as possible before beginning
programming. This is relevant here because we seem to be in a situation
where a lot of developer time has been spent writing code that represents a
design that may need to be abandoned. That is very inefficient and wastes a
lot of time. That's why I want to have design reviews before each component
is written, in order to avoid this inefficiency as much as possible.

dwightguth:
In defense of my design proposal:

I don't understand why maintenance on this layer would be needed. A context
free grammar is an incredibly well defined concept in computer science. The
kil of all the productions is not a grammar, it's an AST of a language
specification language. the actual grammar data structure should never have
to change because it has a 1:1 mapping with the mathematical meaning of a
formal grammar. This modularizes parametric lists and also the parametric
maps and sets we want to eventually add also, and frees us to later specify
syntax any way we want to in K.

You need this anyway: conversation of the grammar into a set of states will
be extremely complex if you're required to try to deal with modularity and
the module kil class at the same time. This also creates a separation of
concerns which allows our new parser to parse languages specified in
grammars not written in K. Which could make this parsing framework much
more desirable to outsiders. Also you will need this if you want to perform
any kind of predictive analysis on grammars, such as is required by the
parser table generator. All of these things mean changing the design now
will considerably speed up feature implementation down the road.

Furthermore, if we don't want to have normal grammar productions in K then
we will need to create a grammar internally in order to allow us to perform
the algorithms that are typically performed on grammars. I will need an
incredibly large amount of persuasion before I am ready to believe that we
should discard the concept of a grammar entirely.

What's the downside you are imagining here?

As far as design reviews go, the process can be flexible. But design
reviews mitigate inefficiency by making us think through the design as best
we can before writing a single line of code. As such, I /will/ be drafting
a design review policy.

In any event, we definitely need to have this design discussion before
pushing this change. It can wait until Monday when I'm on site if you want
though. But development on this piece of code should stop until we're able
to address this. Find things that don't require me to sign off on a design
and work on them until then, if you don't want to have the discussion now.

grosu:
OK, design review makes sense.  I thought that the purpose of the meeting was to explain you their code, which would be cheating.  But I misunderstood, Radu said it correctly (to review the design).  My mistake.

I actually think that we should all understand the overall design, not only you, Dwight.  We had a few meetings about the parsing algorithm, but somehow it still feels mysterious.  I asked Michael to write it up, but he said that it takes too long.  I actually disagreed, but for the sake of making (any) progress I gave up and accepted it.  What I specifically disagree with is that something well-understood is complicated to write up or explain.  My fear is that it is not well-understood (but I do hope by any means that I am wrong here).  I think that we should not even attempt to implement something complicated if it is not already written up and discussed with and understood by others.

Grigore

dwightguth:
Alright. I propose Michael and Radu explain the design to everyone's
satisfaction at the meeting on Monday, then. And have prepared before the
meeting a design diagram showing the intended inputs and outputs of each
(modular) piece of the parser, from the basic parser all the way to parsed
kore.

grosu:
Michael,

Thanks for these details!  Would it be possible that you and Radu, now that you spent the time writing up this pseudocode, make it more rigorous and write it up for us (and for you)?  No need to be very formal, but we need enough detail to understand what's going on.  I think we all need a round of discussions about the overall design and algorithm here.

Thanks,
Grigore

grosu:
Our messages crossed.  Actually, what I meant in my previous message by asking for more
details on the algorithm was more details on the overall design.  If the design and the rationale
behind it are clear, then the details of the algorithm should also be easier to digest.  So yes,
focusing on explaining us the overall design first takes priority.  Ideally, it would be nice to have
both by Monday: a few slides describing the design with some example(s), and a high-level
algorithm.

Grigore

dlucanu:
The subject is of interest for all of us. I propose to have a joint meeting.

Dorel

dwightguth:
Okay, based on our design discussions I am persuaded by this algorithm as long as it's cleaned up a little.
1. Add a comment explaining the dependencies variable.
2. Rename ee and st
3. Comment ee.
4. Change your comments at the top to more closely mirror the explanation of the algorithm that Michael provided.

Then get Michael to sign off on it, and I will merge.

src/javasources/KTool/src/org/kframework/parser/concrete2/Grammar.java
```diff
@@ -194,6 +195,13 @@ public MatchResult(int matchEnd, Function matchFunction) {
         public PrimitiveState(StateId stateId, NonTerminal nt, OrderingInfo orderingInfo) {
             super(stateId, nt, orderingInfo, true);
         }
+
+        public boolean isNullable() {
+            Set<MatchResult> matchResults = this.matches("", 0, null);
+            if (matchResults.size() != 0)
+                return true;
+            return false;
```
adamsmd, line 203: You should probably peplace lines 201-201 with "return matchResults.size() == 0".

(1) You got the polarity of the test flipped.  (2) There is no need for that extra "if".

adamsmd:
My understanding is that Radu is working on rewriting this to use the algorithm described in https://github.com/kframework/k/pull/267#issuecomment-36816529.  Until that happens, I don't think this is ready for merge.

radumereuta:
You are right. We should forget about this pull request. I'm starting from scratch with the best ideas from this thread. This discussion got out of hand and it would be hard to follow in the future.

dwightguth:
Okay. In that case please close this without merging.

PR #268, opened 2014-03-04
added support for iteration over sets and maps; fixed set concatenation ...

andreistefanescu:
...in the right-hand side


src/javasources/KTool/src/org/kframework/backend/java/builtins/BuiltinSetOperations.java
```diff
@@ -15,6 +16,21 @@
  */
 public class BuiltinSetOperations {
 
+    public static BuiltinSet construct(BuiltinSet term1, BuiltinSet term2, TermContext context) {
+        Variable frame = null;
+        if (term1.hasFrame() && term2.hasFrame()) {
+            throw new IllegalArgumentException("both map arguments have frames");
```
dwightguth: The reason this is a problem is because frames are variables and you can't have a frame with more than one variable, correct? Can you please comment this here?

src/javasources/KTool/src/org/kframework/backend/java/kil/MapKeyChoice.java
```diff
@@ -0,0 +1,103 @@
...
+        if (!(map instanceof BuiltinMap)) {
+            return this;
+        }
+
+        if (!((BuiltinMap) map).getEntries().isEmpty()) {
+            return ((BuiltinMap) map).getEntries().keySet().iterator().next();
+        } else if (((BuiltinMap) map).isEmpty()) {
+            return new Bottom(kind);
+        } else {
+            return this;
```
dwightguth, line 46: Can you explain in a one line comment what case this happens in?

src/javasources/KTool/src/org/kframework/backend/java/kil/SetElementChoice.java
```diff
@@ -0,0 +1,103 @@
...
+        if (!(set instanceof BuiltinSet)) {
+            return this;
+        }
+
+        if (!((BuiltinSet) set).elements().isEmpty()) {
+            return ((BuiltinSet) set).elements().iterator().next();
+        } else if (((BuiltinSet) set).isEmpty()) {
+            return new Bottom(kind);
+        } else {
+            return this;
```
dwightguth, line 46: Same comment as above.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
```diff
@@ -365,33 +365,46 @@ public ASTNode transform(org.kframework.kil.ListBuiltin node) throws Transformer
...
-            return new BuiltinSet(entries);
+            ArrayList<Term> baseTerms = new ArrayList<>(node.baseTerms().size());
+            for (org.kframework.kil.Term term : node.baseTerms()) {
+                baseTerms.add((Term) term.accept(this));
+            }
+            baseTerms.add(new BuiltinSet(entries));
+
+            Term result = baseTerms.get(0);
+            for (int i = 1; i < baseTerms.size(); ++i) {
+                result = new KItem(
```
dwightguth, line 397: Can I get a brief explanation of why this is being done?

src/javasources/KTool/src/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
```diff
@@ -421,7 +429,12 @@ public ASTNode transform(org.kframework.kil.MapBuiltin node) throws TransformerE
...
         } else {
+            ArrayList<Term> baseTerms = new ArrayList<>(node.baseTerms().size());
+            for (org.kframework.kil.Term term : node.baseTerms()) {
+                baseTerms.add((Term) term.accept(this));
+            }
             baseTerms.add(new BuiltinMap(entries));
+
             Term result = baseTerms.get(0);
             for (int i = 1; i < baseTerms.size(); ++i) {
                 result = new KItem(
```
dwightguth, line 440: same here?

src/javasources/KTool/src/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
```diff
@@ -599,10 +612,17 @@ public ASTNode transform(org.kframework.kil.Rule node) throws TransformerExcepti
...
+                } else {
+                    lookups.add(new SetLookup(base, key), BoolToken.TRUE);
+                }
             } else {
                 Term value = (Term) lookup.value().accept(this);
                 if (lookup instanceof org.kframework.kil.MapLookup) {
+                    if (lookup.choice()) {
+                        lookups.add(new MapKeyChoice(base), key);
+                    }
                     lookups.add(new MapLookup(base, key, kind), value);
```
dwightguth, line 626: why isn't this an else?

src/javasources/KTool/src/org/kframework/backend/java/symbolic/LocalEvaluator.java
```diff
@@ -45,6 +45,11 @@ public ASTNode transform(ListLookup listLookup) {
     }
 
     @Override
+    public ASTNode transform(SetElementChoice setElementChoice) {
+        return setElementChoice.evaluateChoice();
```
dwightguth, line 49: Should we maybe have some kind of Evaluatable interface, and then do something like:

```
public ASTNode transform(Evaluatable evaluatable) {
    return evaluatable.evaluate();
}
```

?

src/javasources/KTool/src/org/kframework/backend/java/kil/MapKeyChoice.java
andreistefanescu, line 46: This is a general approach: is a function (any function) cannot be evaluated, it is left as it is. I don't think such a comment belongs here, as it would be repeated in a lot of places.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
andreistefanescu, line 626: Because in the case of a map, you first choose the key, and then you use the key to get the value.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/LocalEvaluator.java
andreistefanescu, line 49: I guess we could.

src/javasources/KTool/src/org/kframework/compile/transformers/DataStructureToLookupUpdate.java
```diff
@@ -172,30 +203,51 @@ public ASTNode transform(Rule node) throws TransformerException {
...
-            return node;
-        }
-
-        status = Status.RHS;
-        Term rhs = (Term) rewrite.getRight().accept(this);
-        Term requires = node.getRequires();
-        if (requires != null) {
-            status = Status.CONDITION;
-            requires = (Term) requires.accept(this);
+        /* remaining lookups must be iterations over builtin data structures */
```
dwightguth: I'm having trouble following this code which makes me have trouble following why this is true. Can some kind of overall explanation of what's going on in each phase of this method be put in here somewhere?

src/javasources/KTool/src/org/kframework/backend/java/kil/MapKeyChoice.java
dwightguth, line 46: okay now I'm confused then. What's the difference between not being able to be evaluated, and returning Bottom?

src/javasources/KTool/src/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
dwightguth, line 626: Oh never mind I see now.

src/javasources/KTool/src/org/kframework/backend/java/kil/MapKeyChoice.java
andreistefanescu, line 46: .Map[0] => Bottom (you know for sure there is no mapping for 0)
M:Map[0] => M:Map[0](you don't know whether there is a mapping for 0 or not, or what that mapping is)

dwightguth, line 46: Okay, sounds good. I'll sign off on this.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
andreistefanescu, line 397: This is done because of how the data structure classes during execution look like. Currently, they can only have one "frame" (component which is not an element), and since in these case there are more, the issue is resolved by using the concatenation . We could change the classes, of course.

dwightguth, line 397: Hmm. Why was it done this way? It seems to me like it makes it more confusing and less performance-efficient to use a giant tree of labels instead of a single builtin map/set object. Am I missing some element of the design?

andreistefanescu, line 397: Well, initially I was hoping to have some efficient, in-place rewriting when possible, and this case would have helped. As it is right now, in essence it is a suboptimal design.

src/javasources/KTool/src/org/kframework/backend/java/builtins/BuiltinSetOperations.java
andreistefanescu: Yes, this is correct. I think this follows from the design of the data structure classes, but if you want I can add a comment here.

src/javasources/KTool/src/org/kframework/compile/transformers/DataStructureToLookupUpdate.java
```diff
@@ -51,7 +52,10 @@
...
 
     private interface VariableCache {
-        public Set<Variable> variables();
+        /**
+         * Returns a {@code Set} of {@link Variable} instances that are not matched yet.
+         */
+        public Set<Variable> unmatchedVariables();
     }
 
     private class ExtendedListLookup extends ListLookup implements VariableCache {
```
yilongli, line 61: Could you also add some Javadoc for this class and the similar classes below?

src/javasources/KTool/src/org/kframework/backend/java/builtins/BuiltinSetOperations.java
grosu: Please ALWYAS add a coment explaining what you've done.  This will help many people later on.  Think about it this way: you've already spent hours getting it to work.  Why not spend 2 more minutes to make it easily accessible to others and even yourself later on?  I personally think that we should reject code which is not commented.  Even code of experimental nature.  We should learn good habits and this is good practice for them.

Grigore

dwightguth: Grigore: I asked Andrei to elucidate what exactly was going on here and he did in a follow-up commit. I agree with your philosophy though; always better to just take a moment to add a comment than to deliberate over whether one is necessary.

PR #269, opened 2014-03-05
test commit that should fail tests when run on a pull request

dwightguth:
This is a test pull request I created to make sure that the Jenkins build correctly picks up the diff from the pull request and fails on the compile error I introduced.


dwightguth:
Worked as expected. Closing.

PR #270, opened 2014-03-07
able to run environment-based fun-untyped with Java engine

yilongli:
Sorry, looks like I have completely messed up the commit history... I tried to sync my fork with the upstream repository and somehow I merged my local repo with my fork. Then a lot up duplicate commits show up. I guess that's because I accidentally changed a commit that has been pushed.


dwightguth:
Yeah I suggest you save your changes somewhere, create a clean repository that contains only the upstream commit history, and then cherry-pick only the changes you want reviewed onto the fork, and push it again. I can't approve a pull request that contains changes belonging to someone else which have already been pushed.

yilongli:
I have rewritten the history and the `ant test` has passed. It should be ready for review now.

src/javasources/KTool/src/org/kframework/backend/java/builtins/BuiltinFloatOperations.java
```diff
@@ -29,7 +29,7 @@ public static Term sub(Term term1, Term term2, TermContext context) {
             return new KItem(
                     KLabelConstant.of("'--Float_",context1),
                     new KList(ImmutableList.<Term>of(term2),null),
-                    context1).evaluate(context);
```
dwightguth, line 32: I'm not sure I'm comfortable with KItem and KLabelConstant taking different Context classes.

src/javasources/KTool/src/org/kframework/backend/java/builtins/MetaK.java
```diff
@@ -36,13 +45,35 @@ public static BoolToken unifiable(Term term1, Term term2, TermContext context) {
         SymbolicUnifier unifier = new SymbolicUnifier(constraint, context);
         try {
             unifier.unify(term1, term2);
-        } catch (MatcherException e) {
+        } catch (UnificationFailure e) {
```
dwightguth, line 47: why are these two different exception types?

src/javasources/KTool/src/org/kframework/backend/java/builtins/MetaK.java
```diff
@@ -36,13 +45,35 @@ public static BoolToken unifiable(Term term1, Term term2, TermContext context) {
...
+     * @param context
+     *            the term context
+     * @return {@link BoolToken#TRUE} if the two terms can be matched;
+     *         otherwise, {@link BoolToken#FALSE}
+     */
+    public static BoolToken matchable(Term subject, Term pattern, TermContext context) {
+        PatternMatcher matcher = new PatternMatcher(context);
+        try {
+            matcher.match(subject, pattern);
+        } catch (PatternMatchingFailure e) {
```
dwightguth: Same question, why a third Exception class?

src/javasources/KTool/src/org/kframework/backend/java/indexing/KLabelIndex.java
```diff
@@ -14,9 +14,16 @@ public KLabelIndex(KLabelConstant kLabel) {
...
+        return kLabel;
+    }
+    
     @Override
     public boolean isUnifiable(Index index) {
-        return index instanceof TopIndex || equals(index);
+        return index instanceof TopIndex
+                || equals(index)
+                || (index instanceof TokenIndex && 
+                        kLabel.sorts().contains(((TokenIndex) index).sort()));
```
dwightguth: what exactly is this doing?

src/javasources/KTool/src/org/kframework/backend/java/indexing/TokenIndex.java
```diff
@@ -10,10 +11,17 @@
...
+        return sort;
+    }
 
     @Override
     public boolean isUnifiable(Index index) {
-        return index instanceof TopIndex || equals(index);
+        return index instanceof TopIndex
+                || equals(index)
+                || (index instanceof KLabelIndex && 
+                        ((KLabelIndex) index).kLabel().sorts().contains(sort));
```
dwightguth: same question

src/javasources/KTool/src/org/kframework/backend/java/kil/KItem.java
```diff
@@ -67,7 +73,21 @@ public KItem(Term kLabel, Term kList, Context context) {
             List<Production> productions = kLabelConstant.productions();
             if (productions.size() != 0) {
                 Set<String> sorts = new HashSet<String>();
-
+                
+                if (!K.do_kompilation) {
+                    /* YilongL: user-defined sort predicate rules are interpreted as overloaded productions at runtime */
```
dwightguth, line 87: can you reword this to make it clearer? something like: sort checks are not implemented as rewrite rules in Java, so we need to precompute the sort of terms.

src/javasources/KTool/src/org/kframework/backend/java/builtins/BuiltinFloatOperations.java
yilongli, line 32: I agree. I tend to change KLabelConstant to receive TermContext as well.

src/javasources/KTool/src/org/kframework/backend/java/builtins/MetaK.java
yilongli, line 47: MatcherException is from the generic kil class hierarchy. It is relatively expensive to construct because it needs to fill in the stacktrace. Since the Java engine only uses this exception for control-flow, I wrote the UnificationFailure/PatternMatchingFailure which implements the singleton pattern and has no stacktrace inside.

src/javasources/KTool/src/org/kframework/backend/java/builtins/MetaK.java
yilongli: I can combine UnificationFailure and PatternMatchingFailure but I prefer to keep it this way.

dwightguth: I'm fine with separate exceptions for these two if we make a common base
class so we can catch them together if we ever want to.

src/javasources/KTool/src/org/kframework/backend/java/kil/KItem.java
```diff
@@ -181,8 +201,32 @@ public KItem(Term kLabel, Term kList, Context context) {
...
     }
+    
+    public boolean isEvaluable(TermContext context) {
+        if (evaluable != null) {
+            return evaluable;
+        }
+
+        evaluable = false;
+        if (!(kLabel instanceof KLabelConstant)) {
+            return evaluable;
```
dwightguth: return false

src/javasources/KTool/src/org/kframework/backend/java/kil/KItem.java
```diff
@@ -181,8 +201,32 @@ public KItem(Term kLabel, Term kList, Context context) {
...
+        }
+        KLabelConstant kLabelConstant = (KLabelConstant) kLabel;
+        
+        if (!(kList instanceof KList)) {
+            return evaluable;
+        }
```
dwightguth, line 230: return false

src/javasources/KTool/src/org/kframework/backend/java/kil/KItem.java
```diff
@@ -181,8 +201,32 @@ public KItem(Term kLabel, Term kList, Context context) {
...
+        
+        if (kLabelConstant.label().startsWith("is")
+                || !context.definition().functionRules().get(kLabelConstant).isEmpty()
+                || BuiltinFunction.isBuiltinKLabel(kLabelConstant)) {
+            return evaluable = true;
```
dwightguth: evaluable = true

src/javasources/KTool/src/org/kframework/backend/java/indexing/KLabelIndex.java
yilongli: Consider the following productions and the lookup rule in fun-untyped:
`syntax Name ::= Token{[a-z][\_a-zA-Z0-9]*}       [notInRules]`
`syntax Name ::= "$h"`

```
  rule <k> X:Name => V ...</k>
       <env>... X |-> L ...</env>
       <store>... L |-> V ...</store>
```

Name is a token sort thanks to the first production and, therefore, the lookup rule has a `TokenIndex`. However, `$h` has a `KLabelIndex` and will get stuck when it appears on top of the k cell.

src/javasources/KTool/src/org/kframework/backend/java/kil/KLabelConstant.java
```diff
@@ -43,6 +50,12 @@ private KLabelConstant(String label, Context context) {
         this.label = label;
         productions = ImmutableList.copyOf(context.productionsOf(label));
         
+        Set<String> set = new HashSet<>();
+        for (Production prod : productions) {
+            set.add(prod.getSort());
+        }
+        sorts = ImmutableSet.copyOf(set);
```
dwightguth: is there an ImmutableSet builder?

src/javasources/KTool/src/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
```diff
@@ -141,7 +141,7 @@ public ASTNode transform(org.kframework.kil.KApp node) throws TransformerExcepti
         if (kList instanceof Variable) {
             kList = new KList((Variable) kList);
         }
-        return new KItem(kLabel, kList, this.context);
+        return new KItem(kLabel, kList, TermContext.of(definition));
```
dwightguth, line 172: is this method guaranteed to return the same object each time?

src/javasources/KTool/src/org/kframework/backend/java/symbolic/SubstitutionTransformer.java
```diff
@@ -32,7 +32,29 @@ public SubstitutionTransformer(Map<Variable, ? extends Term> substitution, TermC
         postTransformer.addTransformer(new LocalSubstitutionTransformer());
         postTransformer.addTransformer(new VariableUpdaterTransformer());
     }
+    
+    /**
+     * Stops further traversing the term when it contains no variable that needs
+     * to be substituted.
+     * 
+     */
+    private class LocalVariableChecker extends LocalTransformer {
```
dwightguth, line 41: why was this class moved?

src/javasources/KTool/src/org/kframework/compile/transformers/AddInjections.java
```diff
@@ -192,7 +192,7 @@ public ASTNode transform(TermCons node) throws TransformerException {
         if (sort.equals(KSorts.K) || sort.equals(KSorts.KLABEL) || sort.equals(KSorts.KLIST)) {
             transformedNode.setSort(KSorts.KITEM);
             // TODO (AndreiS): remove special case
-            if (node.getProduction().getLabel().equals("#if_#then_#else_#fi")) {
+            if (node.getProduction().getLabel().equals("#if_#then_#else_#fi") && !sort.equals(KSorts.KLIST)) {
```
dwightguth, line 195: I'm kind of uncomfortable with this being special cased. is there a more uniform way of doing this?

tests/regression/java-rewrite-engine/smt_model/program.test.out
```diff
@@ -1,3 +1,3 @@
 <k>
-    _0!0:Int |-> 1
```
dwightguth, line 2: this doesn't seem connected? why is this a part of this diff?

src/javasources/KTool/src/org/kframework/backend/java/indexing/KLabelIndex.java
dwightguth: Okay, but this doesn't feel scalable. Are we going to have an or clause for
every pairwise set of kinds? Can't we handle this more generically somehow?

Also, do we even want to support this use case? It feels wrong to me to
have the ability to overload token sorts with arbitrary productions.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
yilongli, line 172: Yes, TermContext objects are cached.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/SubstitutionTransformer.java
yilongli, line 41: I just reordered the classes to fit the pre/post transformers order.

src/javasources/KTool/src/org/kframework/compile/transformers/AddInjections.java
yilongli, line 195: I agree. I am not very sure about the special case here, so I just try to write whatever I need without breaking existing code.
@andreistefanescu Could you shed some light to this case.

tests/regression/java-rewrite-engine/smt_model/program.test.out
yilongli, line 2: `_0!0` is a generated fresh variable. The name often gets changed as we change the implementation of the Java engine. In this case, it got changed because I fixed a bug related to detecting unbound variables on the RHS of a rule.

src/javasources/KTool/src/org/kframework/backend/java/builtins/MetaK.java
yilongli: Can you come up with a good name for the base class?

src/javasources/KTool/src/org/kframework/backend/java/indexing/KLabelIndex.java
yilongli: For now, it is just between TokenIndex and KLabelIndex. I am not sure what would be a better way to handle this but I will keep that in mind.

I am afraid I cannot answer your second question. I would like to get rid of all these tricky cases and keep my code clean of course. @grosu

src/javasources/KTool/src/org/kframework/backend/java/builtins/MetaK.java
dwightguth: UnificationOrMatchingFailure

tests/regression/java-rewrite-engine/smt_model/program.test.out
dwightguth, line 2: Okay, this is a failure in test design. If the name of the key of this map
is unspecified, then we shouldn't specify it in the test output. is there a
way to specify a k pattern that the output of this program should match?

src/javasources/KTool/src/org/kframework/backend/java/indexing/KLabelIndex.java
dwightguth: No it's not though. You've already special cased TopIndex as well. That's
what I'm concerned about here: we seem to be writing an algorithm that
scales quadratically in the number of kinds. And for all I really know
there may be other corner cases you're missing also, since you apparently
only just now thought of this one. so I'm not really very pleased with this
method even if we can remove the corner case in question. If we're doing
things based on sorts but having to do cases kind by kind, what have we
done to solve this in the past? This sounds very similar to something we
should have had to deal with for the unification algorithm itself. How did
we handle it there?

src/javasources/KTool/src/org/kframework/compile/transformers/AddInjections.java
andreistefanescu, line 195: The desired signature of if_then_else_fi is forall t . bool -> t -> t -> t. Unfortunately, the K frontend does not support that, so it has the more generic signature bool -> klist -> klist -> klist. This would result in if_then_else_fi being always injected. I don't know how else to avoid this problem.

src/javasources/KTool/src/org/kframework/backend/java/indexing/KLabelIndex.java
andreistefanescu: I think
`syntax Name ::= "$h"``

should actually be
`syntax Name ::= Token{$h}``

We should probably have a check in compile for this.

dwightguth: we could actually make this even more general:

syntax A ::= "string"
syntax A ::= Token{regex}

If we wanted to get really crazy, even something like:

syntax A ::= "string"
syntax A ::= `regex`
syntax B ::= "string terminal" NonTerminal `regex terminal`

The semantics would be that if a terminal is all by itself in a production,
the terminal is captured and put in a token, otherwise it's discarded.

Just a thought.

src/javasources/KTool/src/org/kframework/backend/java/indexing/TokenIndex.java
```diff
@@ -10,10 +11,31 @@
...
+                 * therefore, it's possible to match a TokenIndex with a
+                 * KLabelIndex. For instance, consider the following productions
+                 * and rule:
+                 * 
+                 * syntax Name ::= some lexical production
+                 *               | "$h"
+                 *               
+                 * rule X:Name => .      
+                 * 
+                 * The above rule should be able to match when "$h" appears
```
dwightguth: this doesn't address my original concern that these methods require a quadratic number of code changes.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
```diff
@@ -12,13 +12,40 @@
...
 import java.util.Set;
 
-import com.google.common.collect.Iterables;
 import org.kframework.backend.java.builtins.BoolToken;
 import org.kframework.backend.java.builtins.Int32Token;
 import org.kframework.backend.java.builtins.IntToken;
 import org.kframework.backend.java.builtins.StringToken;
 import org.kframework.backend.java.builtins.UninterpretedToken;
-import org.kframework.backend.java.kil.*;
+import org.kframework.backend.java.kil.BuiltinList;
```
dwightguth, line 20: I didn't mean remove all star imports. kil is different. this is actually now less readable.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/PatternMatcher.java
```diff
@@ -205,7 +205,7 @@
         return result;
     }
 
-    private PatternMatcher(TermContext context) {
```
dwightguth, line 208: why was this done?

src/javasources/KTool/src/org/kframework/compile/transformers/AddInjections.java
dwightguth, line 195: what's stopping me from writing my own if then else fi and having it break because this special case doesn't apply?

dwightguth:
I still can't merge this until I get answers to the rest of my questions.

src/javasources/KTool/src/org/kframework/compile/transformers/AddInjections.java
andreistefanescu, line 195: Technically, it won't break, it will just not work how you might expect. Again, if you don't like it, propose a solution.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/PatternMatcher.java
yilongli, line 208: I needed to provide an interface for matching two terms besides a term and a rule. So I took the fastest way to do this by exposing the constructor. You are right. I should have made another static method for this functionality. Thanks!

tests/regression/java-rewrite-engine/smt_model/program.test.out
yilongli, line 2: @andreistefanescu I need your help here. Is there any way to modify this test to address Dwight's last comment? Or could you somehow rewrite the test?

src/javasources/KTool/src/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
yilongli, line 20: Okay, I can configure my Eclipse settings to do that.

tests/regression/java-rewrite-engine/smt_model/program.test.out
dwightguth, line 2: I'm sorry I was imprecise. I don't need this fixed necessarily. I need the
answer to the specific question I asked: is there a pattern you could write
which this test would match, but would nonetheless express the correctness
of this result?

src/javasources/KTool/src/org/kframework/backend/java/indexing/TokenIndex.java
yilongli: This commit was made before your comment regarding quadratic number of code changes. For now, the right way to do this is to use `Token{"$h"}` instead of `"$h"` and write a kompiler pass to check it. The special case for KLabelIndex and TokenIndex won't be necessary then. The only special case is TopIndex and, thus, there will be no quadratic number of code changes.

yilongli:
Right now, I don't have time to work on the code. I will let you know when it's ready for review again. Thanks.

tests/regression/java-rewrite-engine/smt_model/program.test.out
yilongli, line 2: No, I get your point. I was asking AndreiS if there is way to rewrite the test or specify a pattern in the ktest config file to get rid of this for good. This test is failing from time to time and I have to modify the output many times.

andreistefanescu, line 2: This name is returned by Z3, so I have no control over it.

src/javasources/KTool/src/org/kframework/compile/transformers/AddInjections.java
dwightguth, line 195: @andreistefanescu , can you please clarify me in specific terms what the difference between an if_then_else_fi that we define and one the user would define would be?

tests/regression/java-rewrite-engine/smt_model/program.test.out
dwightguth, line 2: No, I get that we have an undefined variable name here. What I'm wondering is if it's possible to specify this test in such a way that we don't specify what the name of the variable needs to be?

yilongli, line 2: Fixed by modifying the test to be insensitive w.r.t. the name of generated fresh variable.

yilongli:
Ready for code review again.

src/javasources/KTool/src/org/kframework/compile/sharing/TokenSortCollector.java
```diff
@@ -18,20 +25,82 @@
...
+     * Checks if a specified production is valid w.r.t. its sort: namely, a
+     * lexical production should have a token sort and a non-lexical production
+     * should have a non-token sort. Throws a K exception when one of the two
+     * restrictions is violated. Currently, this check is only enabled in the
+     * Java backend.
+     * 
+     * @param production
+     *            the specified production
+     */
+    private void checkIllegalProduction(Production production) {
```
dwightguth, line 75: Sorry just to make sure I understand this, we're not concerned about the case when a token is part of a sort that is a subsort of a sort containing KLabels? That doesn't cause anything to break if it happens?

dwightguth:
This looks good, and Andrei answered my remaining questions. I am merging this.

PR #271, opened 2014-03-07
Moved Indexing to kompile

owolabileg:
Moved the creation of rule index table to happen at kompile time


src/javasources/KTool/src/org/kframework/backend/java/indexing/BottomIndex.java
```diff
@@ -6,6 +6,7 @@
  */
 public class BottomIndex implements Index {
```
dwightguth, line 7: None of the changes in this file are needed for a singleton class.

src/javasources/KTool/src/org/kframework/backend/java/indexing/TopIndex.java
```diff
@@ -6,6 +6,7 @@
  */
 public class TopIndex implements Index {
```
dwightguth, line 9: Same comment here as BottomIndex

src/javasources/KTool/src/org/kframework/backend/java/indexing/IndexingTable.java
```diff
@@ -0,0 +1,149 @@
...
+
+        /* populate the table of rules rewriting the top configuration */
+        List<Index> indices = new ArrayList<Index>();
+        indices.add(TopIndex.TOP);
+        indices.add(BottomIndex.BOTTOM);
+        for (KLabelConstant kLabel : definition.kLabels()) {
+            indices.add(new KLabelIndex(kLabel));
+            indices.add(new FreezerIndex(kLabel, -1));
+            if (!kLabel.productions().isEmpty()) {
+                for (int i = 0; i < kLabel.productions().get(0).getArity(); ++i) {
```
dwightguth: Are all the productions of a KLabel guaranteed to have the same arity? I'm not sure that's necessarily a safe assumption.

dwightguth: Also do we need to add a FreezerIndex for every klabel or only for the klabels that get put into freezers in rules in the definition?

dwightguth: Oh never mind I see what's going on here, the indices list is an intermediate value that we iterate through. But I'm still not convinced that we're safe here. What if there's another production with a greater arity than the one at index 0? Wouldn't we find ourselves potentially missing a value in the index?

dwightguth: @andreistefanescu I believe this was originally your code? Can you verify whether this is a safe assumption?

src/javasources/KTool/src/org/kframework/backend/java/indexing/IndexingTable.java
```diff
@@ -0,0 +1,149 @@
...
+                if (rule.containsAttribute("heat")) {
+                    if (index.isUnifiable(rule.indexingPair().first)) {
+                        heatingListBuilder.add(rule);
+                    }
+                } else if (rule.containsAttribute("cool")) {
+                    if (index.isUnifiable(rule.indexingPair().second)) {
+                        coolingListBuilder.add(rule);
+                    }
+                } else if(rule.containsAttribute("alphaRule")){
+                    if(index.isUnifiable(rule.indexingPair().first)) {
```
dwightguth, line 84: if we remove rules from the main rule table if they're tagged with "alphaRule", and then don't actually query the simulationRuleTable during execution, then those rules will never be applied in ordinary rewriting. Can someone familiar with the alphaRule attribute please verify if that's correct? @liyili2 ?

src/javasources/KTool/src/org/kframework/backend/java/kil/Definition.java
```diff
@@ -41,6 +42,7 @@
     private final Set<KLabelConstant> kLabels;
     private final Set<KLabelConstant> frozenKLabels;
     private final Context context;
+    private IndexingTable index;
```
dwightguth: I'm uncomfortable with this being a class. Please create an interface for this dependency.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -14,28 +14,19 @@
...
-import com.google.common.collect.Sets;
 import org.kframework.backend.java.builtins.IntToken;
-import org.kframework.backend.java.indexing.BottomIndex;
-import org.kframework.backend.java.indexing.FreezerIndex;
-import org.kframework.backend.java.indexing.Index;
-import org.kframework.backend.java.indexing.IndexingPair;
-import org.kframework.backend.java.indexing.KLabelIndex;
-import org.kframework.backend.java.indexing.TokenIndex;
-import org.kframework.backend.java.indexing.TopIndex;
+import org.kframework.backend.java.indexing.*;
```
dwightguth: Please don't import *. Just the classes you want. This will eventually be a code style guideline.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -61,165 +50,38 @@
...
+
             phase2PluggableKastChecker = new PluggableKastStructureChecker();
             phase2PluggableKastChecker.register(new CheckingLeftAssocConstructs(definition));
         } else {
             phase1PluggableKastChecker = null;
             phase2PluggableKastChecker = null;
         }
-
-        // Index may be built with or without measurement
-        if (K.get_indexing_stats){
```
dwightguth, line 101: I'm not comfortable with this code being deleted entirely. We should still be able to measure how long indexing takes to be created after this change goes through.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -294,11 +156,9 @@ public ConstrainedTerm rewrite(ConstrainedTerm constrainedTerm) {
         }
 
         if (K.do_indexing) {
```
dwightguth: Yeah, this shouldn't be like this. You need an interface defining the common core of what it means to be an index, then all indexes implement this interface. You use the command line flag to set which class instantiates the interface at run time, but then following that, no code depends on that command line flag.

dwightguth:
Overall looks okay. A few small things, plus I see some stuff that looks weird that seems to have already been there before this change went out, but I want to make sure that we address my questions before committing this change nonetheless.

src/javasources/KTool/src/org/kframework/backend/java/indexing/IndexingTable.java
andreistefanescu: The production arity is encoded in the klabel string by the number of '_'s. So, unless there is some weird use of the klabel, this assumption is safe.

dwightguth: Okay, but nothing actually stops me from writing:

```
syntax Foo ::= Bar "*" Bar [klabel(Foo)]
syntax Foo ::= Bar "*" [klabel(Foo)]
```

right now, does it? Because unless that's a compiler error if I try to do that with a nice error message saying all KLabels must have the same arity, then this is still a problem. Now maybe having that nice compiler error is the way to go, I don't know. But it still needs to be enforced, otherwise it's a latent bug.

andreistefanescu: I agree with you that it is a potentially latent bug...I supposed this could be fixed by just taking the maximum arity over all the productions.

dwightguth: That's reasonable, and it would still be linear time, just linear in the
number of productions instead of the number of klabels. Go ahead and make
the change, since it shouldn't be too hard.

andreistefanescu: Do you mean me or Owolabi?

src/javasources/KTool/src/org/kframework/backend/java/indexing/BottomIndex.java
andreistefanescu, line 7: There is an issue with serializing: the serialized constant and the unserialized constant are not the same. An alternative is do write code that replaces the deserialized object with the constant. You can see how that is done for KLabelConstant.

src/javasources/KTool/src/org/kframework/backend/java/indexing/IndexingTable.java
dwightguth: Owolabi, it's his change.

src/javasources/KTool/src/org/kframework/backend/java/indexing/BottomIndex.java
dwightguth, line 7: Good catch. Owolabi, please go ahead and revert the changes to TopIndex and BottomIndex, but nevertheless implement the readResolve method for each of them.

owolabileg, line 7: Fixed.

src/javasources/KTool/src/org/kframework/backend/java/indexing/IndexingTable.java
owolabileg: Fixed. @andreistefanescu please take a look.

src/javasources/KTool/src/org/kframework/backend/java/indexing/IndexingTable.java
owolabileg, line 84: @liyili2 we should also talk about what needs to happen when another form of indexing, apart from the rule table is to be used

src/javasources/KTool/src/org/kframework/backend/java/indexing/TopIndex.java
owolabileg, line 9: Fixed.

src/javasources/KTool/src/org/kframework/backend/java/kil/Definition.java
owolabileg: Fixed.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/SymbolicRewriter.java
owolabileg: Fixed.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/SymbolicRewriter.java
owolabileg, line 101: I added a TODO for this, I'll see to this later.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/SymbolicRewriter.java
owolabileg: Fixed.

owolabileg:
I have have attempted to address all of Dwight's comments. Please re-review.

src/javasources/KTool/src/org/kframework/backend/java/indexing/IndexingTable.java
andreistefanescu: Looks good.

src/javasources/KTool/src/org/kframework/backend/java/indexing/BottomIndex.java
```diff
@@ -1,6 +1,8 @@
 package org.kframework.backend.java.indexing;
 
 
+import java.util.HashMap;
+
```
dwightguth: this looks unused to me

src/javasources/KTool/src/org/kframework/backend/java/indexing/pathIndex/PathIndex.java
```diff
@@ -51,16 +50,14 @@ public PathIndex(Definition definition) {
         this.definition = definition;
         this.indexedRules = new LinkedHashMap<>();
         termVisitor = new TermVisitor(definition.context());
-        multiCellInfoHolder =
```
dwightguth, line 54: is this supposed to be in a different pull request?

dwightguth:
it looks to me like you pushed some unrelated changes inro this pull request. can you please rebase to keep things clean and everything on its own branch?

liyili2:
where is these changes? I did not modify anything about the files. I don't
even know these files

dwightguth:
Git blame shows you worked on the simulation rules, is that not correct?

liyili2:
What do you mean by the simulation rule. All my work is in the ksimulation.
I did not do anything about the file you are pointing at

dwightguth:
Look at the larger diff. This code was moved from another file, which you
/did/ modify.

liyili2:
which file are you talking about?

liyili2:
If you are talking about the files in the indexing package. I can tell you
that i never did that. Maybe it is the problem when I first used eclipse to
modify the k master. I think you can help me to change it back.

dwightguth:
As I said, it's in the diff. Do you know how to read a diff? Take the lines
I referenced you in and follow them back to the file they were originally
taken from, in a different package. I can tell you if I have to, but I'd
feel better if you took the opportunity to learn this yourself because it's
a very important coding skill.

liyili2:
I am sorry. Actually, I do not know what is a pull request? I think I never
had anything related to a pull request till now. In addition, after I saw
the code, I found that the only places which are related to me is the
siimulation rules and alpha rules. So What so I change?

src/javasources/KTool/src/org/kframework/backend/java/indexing/IndexingTable.java
liyili2, line 84: Yes, I need the alphaRule here. When user wrote it, they suppose the alphaRules are not a part of their semantics. They are the rules for simulation only. The user wrote these rules to indicate that they are actually interested in the cell or variables by applying the rules. For example, let's say the whole configuration is w, then the alphaRule is a, then they user actually are interested in the cell or data in a(w). 
So, the basic idea of the simulation tool is two steps rewriting. First, I need to take two programs (a specification and a implementation), and use the rule set without alphaRule to rewrite the two programs to next step. Second, After I got the rewriting result as a Term, I apply alphaRule to the term to get the data which the users are insterested in and compare the specification side with the implementation side in order to determine if the two side are equivalent. 

I mean I did not need the rule during the execution of a program with its semantics. However, I need a way to distinguish the alpha rule and other semantics rules and I can use these two different things on different purpose.

src/javasources/KTool/src/org/kframework/backend/java/indexing/BottomIndex.java
owolabileg: Fixed.

src/javasources/KTool/src/org/kframework/backend/java/indexing/pathIndex/PathIndex.java
owolabileg, line 54: No, it's related to this pull request. Since I was making a generic interface for indexing, I also took the opportunity to also cleanup some previously not-so-good code. Also, without this change, I would have needed to serialize more files than necessary when using the path indexing.

owolabileg:
I addressed the comments from the previous round of review. Kindly review again.

src/javasources/KTool/src/org/kframework/backend/java/indexing/IndexingTable.java
dwightguth, line 84: Okay, this seems fine then. @liyili2, please comment this next time you commit something pertaining to simulation though.

dwightguth:
Looks good to me. If you can get someone to sign off that they understand this change (Andrei or Yilong probably since they're most familiar with the indexing and the rewrite engine, if I had to guess), you can merge this.

yilongli:
Yes, it looks good to me. I will sign off on this.

liyili2:
Thank you Owolabi. It is good. I will remember next time.

dwightguth:
Okay I can't merge this for you because you need to resolve conflicts first. Please resolve your conflicts by rebasing and then re-push your change to your fork. Then I will merge.

dwightguth:
JENKINS: add to whitelist

owolabileg:
I think everything is now set to be merged. Can I go ahead and merge now? In particular, note that after rebasing and resolving conflicts, I had to force push my changes. Otherwise, git would not let me push, even to my own fork. With help from @bmmoore, I eventually realized that this was happening because rebasing meant that git could not do a "non-fastforward" update. I looked over the history now and the history seems to have been preserved despite the force push. If everything looks good to you, then @dwightguth may want to add a description of how to handle this case to the code review wiki, for the benefit of others. 

Thank you.

dwightguth:
I will go ahead and update the wiki for this case tomorrow.

PR #272, opened 2014-03-08
fixed idea project

andreistefanescu:


traiansf:
Looks good to me.  Thanks Andrei!
Dwight, please merge :-)

grosu:
What is this?  The title "fixed idea" doesn't tell anything.  In general, please have a short description of everything that you make a pull request for.  Not a big deal to write a sentence or two.

traiansf:
But it's exactly what it says :-)
It fixed the idea project which was failing due to dependencies introduced
by Dwight when adding the unit testing libraries.

2014-03-08 10:46 GMT+02:00 Grigore Rosu notifications@github.com:

grosu:
Well, it can also mean what I initially understood, namely that he started a new project called "fixed idea".

Again, github gives us a form to add a sentence or two describing the pull request.  Considering that only takes a few seconds, which is nothing compared to how long it takes to do the actual thing, for the benefit of others in the team please add that sentence or two description.  In this case, what you just said could have been a good description and it would have saved us all this time and messages.

Grigore

dwightguth:
I actually agree with Grigore here. Give a more detailed description of the
change in the pull request.

PR #275, opened 2014-03-10
added option to kompile and krun controlling whether the java backend th...

andreistefanescu:
...rows a failed assertion when function definitions are not deterministic


dwightguth:
This looks like a reasonable feature, but I'm concerned that we just added a second parameter that has to be specified on both kompile and krun. Is there a way around this other than the more involved change we discussed in the meeting on Monday? Maybe a flag in the serialized definition object?

andreistefanescu:
Sure, but that would still be hacking. I think we should go ahead an implement the general solution as soon as possible.

dwightguth:
Alright. I will put this on my list of things to try to do ASAP. Go ahead and merge this if you can get Yilong to approve.

yilongli:
I understand this change. I will sign off on this.

PR #278, opened 2014-03-10
KTEST: Support 'unixOnly' attributes in config.xml

daejunpark:
I implemented the 'unixOnly' feature as Dwight asked as follows. It would be great if Omer also reviews this. 

On Feb 27, 2014, at 2:20 PM, Dwight Guth dwight.guth@runtimeverification.com wrote:

It recently came to my attention that some of the ktest tests for the Java rewrite engine got commented out. One in particular I am unable to add back in because a feature is missing from ktest. I'm not sure who deleted it, but it needs to be added back in.

Take a look at the tests/regression/java-rewrite-engine/io/test1 folder. It contains a file prepare.sh. Because of limitations of git, it's necessary that this script run on unix before these tests, in order for it to succeed. And because of limitations on IO portability between linux and windows, it's necessary that these tests only run on unix-based OSes.

Can you please add back in the functionality which allows me to specify this information in the config xml file?


src/javasources/KTool/src/org/kframework/ktest/Test/TestSuite.java
```diff
@@ -159,6 +163,46 @@ public void dryRun() {
...
+    private List<TestCase> runUnixOnlySteps(List<TestCase> tests) throws InterruptedException {
+        int len = tests.size();
+        List<TestCase> successfulTests = new ArrayList<>(len);
+        List<Proc<TestCase>> ps = new ArrayList<>(len);
+
+        System.out.format("Running initial scripts...%n");
+        startTpe();
+        for (TestCase tc : tests) {
+            if (tc.hasUnixOnly()) {
+                assert GlobalSettings.isUnixLikeOS();
```
osa1, line 180: Why not move this assertion to the beginning of this method? Currently assertion does not fail if tests list is empty but this method is called while running on Windows system.

src/javasources/KTool/src/org/kframework/ktest/Test/TestSuite.java
```diff
@@ -94,11 +95,14 @@ public TestSuite(List<TestCase> tests, CmdArg cmdArg) {
     public boolean run() throws InterruptedException, TransformerException,
             ParserConfigurationException, IOException {
         boolean ret = true;
-        List<TestCase> successfulTests = tests;
+
+        List<TestCase> successfulTests = runUnixOnlySteps(tests);
```
osa1, line 99: So currently `runUnixOnlySteps` method is also called when running on Windows.. Maybe put a conditional here to make sure this method is only called when running on Unix-like systems.

src/javasources/KTool/src/org/kframework/utils/general/GlobalSettings.java
```diff
@@ -101,6 +101,10 @@ public static boolean isWindowsOS() {
         return os() == OS.WIN;
     }
 
+    public static boolean isUnixLikeOS() {
```
osa1, line 104: Maybe call this method `isPosix`, as in http://en.wikipedia.org/wiki/Posix ? (not very important, just thought Mac OS may not be _Unix-like_ but this preparation scripts should work fine on Mac OS since they use very basic shell commands/programs only)

src/javasources/KTool/src/org/kframework/ktest/Config/ConfigFileParser.java
```diff
@@ -261,7 +272,7 @@ private TestCase parseTestCase(Element testNode) throws InvalidConfigError {
         Map<String, List<PgmArg>> pgmSpecificKRunOpts = parsePgmSpecificKRunOpts(childNodes);
 
         TestCase ret = new TestCase(definition, programs, extensions, excludes, results,
-                kompileOpts, krunOpts, pgmSpecificKRunOpts, skips);
+                kompileOpts, krunOpts, pgmSpecificKRunOpts, skips, unixOnly);
```
osa1, line 275: (not related with this patch) This kind of parameter passing in ktest gets only worse with time.. ktest becomes more complex and we need to add more parameters(I know I'm responsible for this). I think widely-used solution to this problem is to use some kind of builder class with methods for settings object properties but that leads to more verbosity and more code to maintain... Or maybe we can share some settings using some kind of immutable ktest setting object. What do you guys think about that?

dwightguth, line 275: The way to do this generally is to have a constructor with a small number of parameters and then call setters for all the various pieces of configuration. It's slightly more verbose, yes, but that's not intrinsically a bad thing if it makes the code easier to read.

src/javasources/KTool/src/org/kframework/ktest/Test/TestCase.java
```diff
@@ -95,7 +102,7 @@ public static TestCase makeTestCaseFromK(CmdArg cmdArgs) throws InvalidConfigErr
 
         return new TestCase(targetFile, programs, cmdArgs.getExtensions(),
```
dwightguth, line 103: okay yeah see this is why you call setters. That way you can specify default values when you want to. Can I see that as part of this change? It shouldn't be too difficult to manage.

src/javasources/KTool/src/org/kframework/ktest/Test/TestSuite.java
```diff
@@ -159,6 +163,46 @@ public void dryRun() {
...
+        int len = tests.size();
+        List<TestCase> successfulTests = new ArrayList<>(len);
+        List<Proc<TestCase>> ps = new ArrayList<>(len);
+
+        System.out.format("Running initial scripts...%n");
+        startTpe();
+        for (TestCase tc : tests) {
+            if (tc.hasUnixOnly()) {
+                assert GlobalSettings.isUnixLikeOS();
+                Proc<TestCase> p = new Proc<>(tc, tc.getUnixOnlyCmd(), tc.toUnixOnlyLogString(),
```
dwightguth, line 181: I'm not really comfortable with all of these parameters just getting blindly passed along. Can't we pass along the test case and the test suite and have the Proc object access what it needs?

src/javasources/KTool/src/org/kframework/utils/general/GlobalSettings.java
dwightguth, line 104: Good idea. Make it so. And change the name of the setting to "posixInitScript"

dwightguth:
Pretty good, just try to address some of these pieces of feedback, and we should be good to go.

src/javasources/KTool/src/org/kframework/ktest/Test/TestSuite.java
osa1, line 181: These are not blindly passed. The idea was to have general `Proc` class that's capable of running some commands and comparing stdout/stderr streams to passed values using custom comparator methods.

So actually `Proc` is not coupled with KTest at all, it's a more general class. This is why we need to pass "program command", "expected output", "expected error message", "timeout" etc. parameters. It can't read it from KTest classes because it does not know anything about KTest classes.

dwightguth, line 181: See I could believe that if the parameters you were passing resembled a process class, but that doesn't seem to be the case here. You have a "log string", whatever that is (I have no idea because the javadoc just says "string to be used for logging" which doesn't really help. You have a verbose flag, a color flag. And I have no idea what updateOut and generateOut are supposed to do.

As far as I can tell what you should have are the following:

```
public interface Executable {
    public String[] getCommand();
    public long getTimeout();
    public TimeUnit getTimeUnit();
    public String getStdin();
    public boolean matchesStdout(String actual);
    public boolean matchesStderr(String actual);
    public boolean matchesReturnCode(int actual);
}
public class Proc<T extends Executable> {
    T executable;
    public Proc(T executable) { this.executable = executable; }
}
```

osa1, line 181: `log string` is used to give a name to that program while printing log messages. Like

```
Finished running <<<LOG STRING>>> took N seconds to run .....
```

I agree that documentation is a bit weak, apologies for that.

`updateOut` and `generateOut` are not added by me, they're used to write outputs of programs to files instead of writing to stdout. This is used to generate correct output files using correct programs and then by using once generated correct outputs, testing updates. Maybe whoever implemented this should add documentation.

Feel free to change the class however you want, I'm just explaining how it's implemented and why it's implemented that way. The parts that you think don't resembles a process class were probably added later, so some refactoring might be good here.

dwightguth, line 181: Okay, thanks for the clarification. Here's what I want:
1. An interface abstracting the information about how to execute a process and evaluate its output.
2. TestCase implementing this interface
3. A Proc class that doesn't take every single piece of information in its constructor.
4. All non-process-specific code removed from the Proc class.

dwightguth:
JENKINS: add to whitelist

dwightguth:
it's been six days to implement what I estimate to be fairly minor refactorings. can I get an estimate here? are you running into problems?

osa1:
Nope, all changes should be trivial. I guess @daejunpark  is busy these days and I can't edit this pull request because it's done from Daejun's repository ...

If there's a way for me to edit this commits please tell me and I'll make required changes this weekend.

dwightguth:
Just pull the change into your own fork and make a separate pull request.
It's tedious, but should suffice.

dwightguth:
Closing as duplicate of #301.

PR #279, opened 2014-03-11
K buitlin visitor refactoring

andreistefanescu:


src/javasources/KTool/src/org/kframework/backend/java/symbolic/BuiltinFunction.java
```diff
@@ -27,6 +31,17 @@
  */
 public class BuiltinFunction {
 
+    private static final String hookPropertiesFileName = "hooks.properties";
+
+    /**
+     * Set of hook module names excluded from evaluation during compilation.
```
dwightguth: What does that mean, excluded from evaluation?

dwightguth:
Overall looks fine, if you can just clarify that one point. @yilongli ?

Please remember not to push until you've gotten the other pull request this depends on approved.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/BuiltinFunction.java
yilongli: Right now, we evaluate ground functions that appear on the right-hand side of a rule during the kompilation process. However, there are certain functions we would like to exclude.

yilongli:
Looks good to me.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/BuiltinFunction.java
dwightguth: Okay, if this comment is updated to make what you just said more clear, Andrei can merge this.

yilongli: I think it will be sufficient to put a `@see` annotation that points to the method in `KILtoBackendKILTransformer` that performs the partial evaluation in the RHS of a rule.

PR #282, opened 2014-03-12
NewParser - added labels to the generated trees (worked with Michael).

radumereuta:
@mdmkolbe (Michael) should review this.

### DIFF
--- a/src/javasources/KTool/src/org/kframework/parser/concrete2/Grammar.java
+++ b/src/javasources/KTool/src/org/kframework/parser/concrete2/Grammar.java
@@ -9,6 +9,7 @@ import org.kframework.kil.Ambiguity;
 import org.kframework.kil.KApp;
 import org.kframework.kil.KLabel;
 import org.kframework.kil.KList;
+import org.kframework.kil.KSorts;
 import org.kframework.kil.Term;
 import org.kframework.kil.Token;
 
@@ -157,7 +158,7 @@ public class Grammar {
                 }
             }
             if (this instanceof ExitState) {
-                return new HashSet<KList>(Arrays.asList(new KList(Arrays.asList((Term)new Ambiguity("K", new ArrayList<Term>(result))))));
+                return new HashSet<KList>(Arrays.asList(new KList(Arrays.asList((Term)new Ambiguity(KSorts.K, new ArrayList<Term>(result))))));
             } else {
                 return result;
             }
@@ -258,7 +259,7 @@ public class Grammar {
             matcher.useAnchoringBounds(false);
             Set<MatchResult> results = new HashSet<MatchResult>();
             if (matcher.lookingAt()) {
-                results.add(new MatchResult(matcher.end(), Function.constant(Token.kAppOf("K", matcher.group())))); //matchFunction(text, matcher.start(), matcher.end())));
+                results.add(new MatchResult(matcher.end(), Function.constant(Token.kAppOf(KSorts.K, matcher.group())))); //matchFunction(text, matcher.start(), matcher.end())));
             }
             return results;
         }
--- a/src/javasources/KTool/src/org/kframework/parser/concrete2/Grammar.java
+++ b/src/javasources/KTool/src/org/kframework/parser/concrete2/Grammar.java
@@ -141,6 +141,7 @@ public class Grammar {
             int key;
             public OrderingInfo(int key) { this.key = key; }
             public int compareTo(OrderingInfo that) { return Integer.compare(this.key, that.key); }
+            public static final OrderingInfo ZERO = new OrderingInfo(0);
         }
 
         public State(StateId stateId, NonTerminal nt, OrderingInfo orderingInfo, KLabel label) {
--- a/src/javasources/KTool/src/org/kframework/parser/concrete2/Grammar.java
+++ b/src/javasources/KTool/src/org/kframework/parser/concrete2/Grammar.java
@@ -6,6 +6,8 @@ import java.util.HashSet;
 import java.util.Set;
 
 import org.kframework.kil.Ambiguity;
+import org.kframework.kil.KApp;
+import org.kframework.kil.KLabel;
 import org.kframework.kil.KList;
 import org.kframework.kil.Term;
 import org.kframework.kil.Token;
@@ -69,8 +71,8 @@ public class Grammar {
                            StateId entryStateId, State.OrderingInfo entryOrderingInfo,
                            StateId exitStateId, State.OrderingInfo exitOrderingInfo) {
             this.nonTerminalId = nonTerminalId;
-            this.entryState = new EntryState(entryStateId, this, entryOrderingInfo);
-            this.exitState = new ExitState(exitStateId, this, exitOrderingInfo);
+            this.entryState = new EntryState(entryStateId, this, entryOrderingInfo, null);
+            this.exitState = new ExitState(exitStateId, this, exitOrderingInfo, null);
         }
 
         public int compareTo(NonTerminal that) { return this.nonTerminalId.compareTo(that.nonTerminalId); }
@@ -97,6 +99,7 @@ public class Grammar {
         final StateId stateId;
         final NonTerminal nt;
         final OrderingInfo orderingInfo;
+        final KLabel label;
 
         static class OrderingInfo implements Comparable<OrderingInfo> {
             int key;
@@ -104,15 +107,24 @@ public class Grammar {
             public int compareTo(OrderingInfo that) { return Integer.compare(this.key, that.key); }
         }
 
-        public State(StateId stateId, NonTerminal nt, OrderingInfo orderingInfo) {
-            this.stateId = stateId; this.nt = nt; this.orderingInfo = orderingInfo;
+        public State(StateId stateId, NonTerminal nt, OrderingInfo orderingInfo, KLabel label) {
+            this.stateId = stateId; this.nt = nt; this.orderingInfo = orderingInfo; this.label = label;
         }
 
         public Set<KList> runRule(Set<KList> input) {
-            if (this instanceof ExitState) {
-                return new HashSet<KList>(Arrays.asList(new KList(Arrays.asList((Term)new Ambiguity("K", new ArrayList<Term>(input))))));
+            Set<KList> result;
+            if (this.label == null) {
+                result = input;
             } else {
-                return input;
+                result = new HashSet<>();
+                for (KList kl : input) {
+                    result.add(new KList(Arrays.<Term>asList(new KApp(label, kl))));
+                }
+            }
+            if (this instanceof ExitState) {
+                return new HashSet<KList>(Arrays.asList(new KList(Arrays.asList((Term)new Ambiguity("K", new ArrayList<Term>(result))))));
+            } else {
+                return result;
             }
         }
 
@@ -146,22 +158,22 @@ public class Grammar {
     }
 
     public static class ExitState extends State {
-        public ExitState(StateId stateId, NonTerminal nt, OrderingInfo orderingInfo) {
-            super(stateId, nt, orderingInfo);
+        public ExitState(StateId stateId, NonTerminal nt, OrderingInfo orderingInfo, KLabel label) {
+            super(stateId, nt, orderingInfo, label);
         }
     }
 
     public abstract static class NextableState extends State {
         public final Set<State> next = new HashSet<State>();
-        NextableState(StateId stateId, NonTerminal nt, OrderingInfo orderingInfo, boolean intermediary) {
-            super(stateId, nt, orderingInfo);
+        NextableState(StateId stateId, NonTerminal nt, OrderingInfo orderingInfo, boolean intermediary, KLabel label) {
+            super(stateId, nt, orderingInfo, label);
             if (intermediary) { nt.intermediaryStates.add(this); }
         }
     }
 
     public static class EntryState extends NextableState {
-        public EntryState(StateId stateId, NonTerminal nt, OrderingInfo orderingInfo) {
-            super(stateId, nt, orderingInfo, false);
+        public EntryState(StateId stateId, NonTerminal nt, OrderingInfo orderingInfo, KLabel label) {
+            super(stateId, nt, orderingInfo, false, label);
         }
     }
 
@@ -171,8 +183,8 @@ public class Grammar {
 
         public NonTerminalState(
                 StateId stateId, NonTerminal nt, OrderingInfo orderingInfo,
-                NonTerminal child, boolean isLookahead) {
-            super(stateId, nt, orderingInfo, true);
+                NonTerminal child, boolean isLookahead, KLabel label) {
+            super(stateId, nt, orderingInfo, true, label);
             nt.intermediaryStates.add(this);
             this.child = child;
             this.isLookahead = isLookahead;
@@ -191,16 +203,16 @@ public class Grammar {
 
         abstract Set<MatchResult> matches(CharSequence text, int startPosition, Function context);
 
-        public PrimitiveState(StateId stateId, NonTerminal nt, OrderingInfo orderingInfo) {
-            super(stateId, nt, orderingInfo, true);
+        public PrimitiveState(StateId stateId, NonTerminal nt, OrderingInfo orderingInfo, KLabel label) {
+            super(stateId, nt, orderingInfo, true, label);
         }
     }
 
     public static class RegExState extends PrimitiveState {
         private final java.util.regex.Pattern pattern;
 
-        public RegExState(StateId stateId, NonTerminal nt, OrderingInfo orderingInfo, java.util.regex.Pattern pattern) {
-            super(stateId, nt, orderingInfo);
+        public RegExState(StateId stateId, NonTerminal nt, OrderingInfo orderingInfo, java.util.regex.Pattern pattern, KLabel label) {
+            super(stateId, nt, orderingInfo, label);
             this.pattern = pattern;
         }
 
--- a/src/javasources/KTool/src/org/kframework/parser/concrete2/Grammar.java
+++ b/src/javasources/KTool/src/org/kframework/parser/concrete2/Grammar.java
@@ -12,6 +12,41 @@ import org.kframework.kil.KList;
 import org.kframework.kil.Term;
 import org.kframework.kil.Token;
 
+
+/**
+ * The classes used by the parser to represent the internal structure of a grammar.
+ * A grammar represents a NFA generated from EBNF style grammar (in this case the K syntax declarations).
+ * The main object is the {@link NonTerminal}, containing a unique {@link NonTerminalId},
+ * and two states: entry and exit.
+ *
+ * There are four main types of states:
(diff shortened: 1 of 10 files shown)


adamsmd:
This code looks good to me.

src/javasources/KTool/src/org/kframework/parser/concrete2/Grammar.java
```diff
@@ -104,15 +107,24 @@ public int hashCode() {
...
         }
 
         public Set<KList> runRule(Set<KList> input) {
+            Set<KList> result;
+            if (this.label == null) {
+                result = input;
+            } else {
+                result = new HashSet<>();
+                for (KList kl : input) {
+                    result.add(new KList(Arrays.<Term>asList(new KApp(label, kl))));
```
dwightguth, line 159: Can you maybe add:

```
public KList(Term... terms);
```

as a constructor, to clean this line of code up a bit?

dwightguth, line 159: Hmm never mind I guess that wouldn't work with the presence of a copy constructor. I'll have to think about this a bit more. Never mind for now.

src/javasources/KTool/src/org/kframework/parser/concrete2/Grammar.java
```diff
@@ -104,15 +107,24 @@ public int hashCode() {
...
+                }
+            }
             if (this instanceof ExitState) {
-                return new HashSet<KList>(Arrays.asList(new KList(Arrays.asList((Term)new Ambiguity("K", new ArrayList<Term>(input))))));
+                return new HashSet<KList>(Arrays.asList(new KList(Arrays.asList((Term)new Ambiguity("K", new ArrayList<Term>(result))))));
```
dwightguth: please break this line into multiple statements and use a constant to represent the K sort.

src/javasources/KTool/src/org/kframework/parser/concrete2/Grammar.java
```diff
@@ -191,16 +203,16 @@ public MatchResult(int matchEnd, Function matchFunction) {
...
+            super(stateId, nt, orderingInfo, true, label);
         }
     }
 
     public static class RegExState extends PrimitiveState {
         private final java.util.regex.Pattern pattern;
 
-        public RegExState(StateId stateId, NonTerminal nt, OrderingInfo orderingInfo, java.util.regex.Pattern pattern) {
-            super(stateId, nt, orderingInfo);
+        public RegExState(StateId stateId, NonTerminal nt, OrderingInfo orderingInfo, java.util.regex.Pattern pattern, KLabel label) {
```
dwightguth, line 257: also break this line.

src/javasources/KTool/src/org/kframework/parser/concrete2/Parser.java
```diff
@@ -442,7 +442,7 @@ public static void main(String[] args) {
 
         Grammar.NonTerminal nt1 = new Grammar.NonTerminal(ntistart, stistart, new Grammar.State.OrderingInfo(0), stiend, new Grammar.State.OrderingInfo(100));
 
-        Grammar.RegExState res1 = new Grammar.RegExState(new Grammar.StateId("RegExStid"), nt1, new Grammar.State.OrderingInfo(1), Pattern.compile("[a-zA-Z0-9]"));
+        Grammar.RegExState res1 = new Grammar.RegExState(new Grammar.StateId("RegExStid"), nt1, new Grammar.State.OrderingInfo(1), Pattern.compile("[a-zA-Z0-9]"), null);
```
dwightguth, line 446: and this one.

src/javasources/KTool/src/org/kframework/parser/concrete2/Parser.java
```diff
@@ -504,7 +504,7 @@ public Term parse(NonTerminal nt, int position) {
         NonTerminal startNt = new NonTerminal(new NonTerminalId("<start>"),
                                               new StateId("<start-entry>"), new State.OrderingInfo(Integer.MIN_VALUE),
                                               new StateId("<start-exit>"), new State.OrderingInfo(Integer.MAX_VALUE));
-        State state = new NonTerminalState(new StateId("<start>"), startNt, new State.OrderingInfo(Integer.MAX_VALUE - 1), nt, true);
+        State state = new NonTerminalState(new StateId("<start>"), startNt, new State.OrderingInfo(Integer.MAX_VALUE - 1), nt, true, null);
```
dwightguth, line 512: and this one

src/javasources/KTool/test/org/kframework/parser/concrete2/ParserTest.java
```diff
@@ -51,7 +51,7 @@ public void testSingleToken() throws Exception {
 
         Grammar.NonTerminal nt1 = new Grammar.NonTerminal(ntistart, stistart, new Grammar.State.OrderingInfo(0), stiend, new Grammar.State.OrderingInfo(100));
 
-        Grammar.RegExState res = new Grammar.RegExState(new Grammar.StateId("RegExStid"), nt1, new Grammar.State.OrderingInfo(1), Pattern.compile("[a-zA-Z0-9]+"));
+        Grammar.RegExState res = new Grammar.RegExState(new Grammar.StateId("RegExStid"), nt1, new Grammar.State.OrderingInfo(1), Pattern.compile("[a-zA-Z0-9]+"), null);
```
dwightguth, line 57: and this one

src/javasources/KTool/test/org/kframework/parser/concrete2/ParserTest.java
```diff
@@ -66,14 +66,15 @@ public void testSingleToken() throws Exception {
...
+        // A ::= #token{"[a-zA-Z0-9]+ +"} #token{"[a-zA-Z0-9]+"} [klabel(seq)]
         Grammar.NonTerminalId ntistart = new Grammar.NonTerminalId("StartNT");
         Grammar.StateId stistart = new Grammar.StateId("StartState");
         Grammar.StateId stiend = new Grammar.StateId("EndState");
 
         Grammar.NonTerminal nt1 = new Grammar.NonTerminal(ntistart, stistart, new Grammar.State.OrderingInfo(0), stiend, new Grammar.State.OrderingInfo(100));
 
-        Grammar.RegExState res1 = new Grammar.RegExState(new Grammar.StateId("RegExStid"), nt1, new Grammar.State.OrderingInfo(1), Pattern.compile("[a-zA-Z0-9]+ +"));
-        Grammar.RegExState res2 = new Grammar.RegExState(new Grammar.StateId("RegExStid2"), nt1, new Grammar.State.OrderingInfo(2), Pattern.compile("[a-zA-Z0-9]+"));
+        Grammar.RegExState res1 = new Grammar.RegExState(new Grammar.StateId("RegExStid"), nt1, new Grammar.State.OrderingInfo(1), Pattern.compile("[a-zA-Z0-9]+ +"), null);
```
dwightguth, line 81: and these

src/javasources/KTool/src/org/kframework/parser/concrete2/Grammar.java
adamsmd: I agree with using a constant.

Breaking it into multiple _lines_ makes sense to me, but breaking it into multiple statements would be silly as this is a pure, constructive expression.

grosu: I guess Dwight's point is that it is hard to read as it stands, not that it is impure.  I would personally write a different assignment for which `new` which makes sense, e.g.,

```
  Ambiguity amb = new Ambiguity("K", new ArrayList<Term>(input));
```

Maybe I would go even further, like one assignment per `new`.

Dwight, if this is a style that we want people to use, then we should make it clear in the wiki page.  The point is that we want the code to be readable and understandable by others easily.

radumereuta: On that line you have one concept. Create a new object containing the results of parsing a specific NonTerminal. Now you want to split that single line into a block of statements. Yes, it makes it more readable, but here I think the bigger picture is more important. Adding more than one line doesn't help readability when you are looking at the overall structure of the code.

I'm not sure we can come up with a general rule that can satisfy all of these concerns.

dwightguth: The problem here is that it's very hard for me to understand what this term represents. If you want it to be a single statement, then that statement needs to not intersperse meaningful information with boilerplate function calls, and what function calls it does make need to be clearly understandable in their purpose. Essentially it's the difference between:

```
return new HashSet<KList>(Arrays.asList(new KList(Arrays.asList((Term)new Ambiguity("K", new ArrayList<Term>(result))))));
```

and:

```
KList term = KList.of(Ambiguity.of("K", result));
return Collections.singleton(term);
```

or at least:

```
 KList term = KList.of(Ambiguity.of("K", result));
 Set<KList> s = new HashSet<>();
 s.add(term);
 return term;
```

...if you don't want an unmodifiable set.

You see how one is much easier to read?

Grigore: this isn't a clearly definable style rule, so I can't make a style guideline. But I can comment in the Code Review page that code containing complex, indecipherable expressions may not be accepted.

dwightguth:
Okay, in addition to the explicit comments I made, I'd like the following.
1. Lines modified in this commit should not wrap in github. I think that means ~115 characters per line maximum?
2. You should not modify constructor calls to add a new parameter. If you need a new constructor, add a new constructor with the extra parameter, and make the former constructor call the latter with the null parameter. This minimizes the lines changed by a change to the constructor call.

grosu:
115?  That's way too much! We have already agreed on 99 characters long ago; see Item 4 in the wiki (https://github.com/kframework/k/wiki/Style-Guidelines).  I remember that several of you at the time said that 99/100 is also the limit that Google and other companies enforce in their code.

adamsmd:
Regarding line wrapping.  For reference, in my browser, the lines wrap(*) at 80 characters per source line on the "conversation" page and 118 on the "Files Changed" page. (I'm not counting the "+"/"-" at the start.)  Are browsers/screens different?

(In any case, I agree that those lines could do with being broken down.)

(*) Technically not "warp" so much as get clipped where I need to use the horizontal scrollbar.

Regarding constructor arguments, I disagree in this case.  We should not compromise the interface for the sake of a smaller diff.  This is especially the case now, as very little code calls this code, changing the interface has a low cost.

src/javasources/KTool/src/org/kframework/parser/concrete2/Grammar.java
adamsmd: If KList and Ambiguity had "of" methods, we could certainly use them.  Currently they don't.

dwightguth:
But you're not compromising the interface. you're still able to have no
label for a state, it just doesn't require an explicit null. This is
standard practice for method calls with multiple optional parameters. For
exactly this reason, that changing every use every time doesn't scale.
Already it's made this diff harder to read. That's more cost than I want to
pay .

src/javasources/KTool/src/org/kframework/parser/concrete2/Grammar.java
dwightguth: Gosh, far be it from me to ask you to program things you need...

dwightguth: Sorry, that was way too sarcastic. I apologize. But there is a point here:
code becomes unreadable if we're not willing to take minuscule amounts of
effort to make it less so, and having to do something that's not been done
before is a hideously bad reason not to do something. As is "because it
would be a little bit more work". So I can't help but feel troubled that
this is your mindset ...

grosu: Code quality is not negotiable in this project anymore, and Dwight is explicitly in charge of it.  We should all make an effort to be disciplined in order for this to work out well. That means on the one hand that Dwight should be less sarcastic and focus on the actual issue and propose the best solution in his opinion, and on the other hand that we, the others, should not oppose minor or stylistic requested changes to our code only because we can and want to.  Neither of these is constructive.  The code in this project belongs to the "K Team" and we want it to look good, as if written by one, excellent programmer.  This will make our own life, before anybody's else, easier in the long term.

dwightguth: Again, I'm sorry I was out of line. I have a tendency to think and speak
impulsively and that often works fine but sometimes gets me into trouble. I
will try to remember to be better about this, and I will accept whatever
responsibility Grigore feels appropriate to ensure we have a good culture
of communication and code quality, even if it means being disciplined
myself.

That said, I will now update the Code Reviews page on the wiki to make the
foundational principles above clear. These are vital and non-negotiable
priorities:
1. Changes to make code cleaner which take only a couple days' extra work
   should be done prior to code being pushed upstream.
2. Users must be willing to modify whatever classes are necessary in order
   to cleanly commit their own changes.

The constructors which take varargs are literally one line each. Also, a
single concept may require many statements if it is composed of pieces
distinct in meaning. E.g. a term and a set.

Make it so.

adamsmd: Dwight, thank you for the apology.  For the future, you should know that, due to events in my past, I have no tolerance for insults or rudeness.

> So I can't help but feel troubled that this is your mindset ...

Please don't make assumptions or impute my motives.  Personally, I think KIL would be cleaner to use in general if it had those factory methods (though perhaps they should be variadic constructors).  It is, however, a change to the KIL interface (a core component that is not part of the parser), which is not something to be done willy-nilly and thus why I brought it up.

adamsmd:
> This is standard practice for method calls with multiple optional parameters.

The point is that "label" is _not_ intended to be optional.  State is designed as an algebraic data type.  In that context, a "null" label has a specific semantic meaning that the developer should think about when constructing a State.

> For exactly this reason, that changing every use every time doesn't scale.

I am not proposing changing "every use every time".  That is a straw man.  The code should be written so that it has the correct interface.  That should not be compromised for the sake of making the diff easier to read.

src/javasources/KTool/src/org/kframework/parser/concrete2/Grammar.java
grosu: Micahel, but you started it, by saying that what Dwight proposed was 'silly'.
That is itself insulting.  It is best if we take these discussions aside, they do not fit in github.  So if you want to continue this discussion, do it in person or by separate messages, please.  But it is best to stop it.

Grigore

Sent from phone

-------- Original message --------
From: mdmkolbe
Date:03/12/2014 17:52 (GMT-06:00)
To: kframework/k
Cc: "Rosu, Grigore"
Subject: Re: [k] NewParser - added labels to the generated trees (worked with Michael). (#282)

In src/javasources/KTool/src/org/kframework/parser/concrete2/Grammar.java:

> ```
>          if (this instanceof ExitState) {
> ```
> -                return new HashSet<KList>(Arrays.asList(new KList(Arrays.asList((Term)new Ambiguity("K", new ArrayList<Term>(input))))));
> -                return new HashSet<KList>(Arrays.asList(new KList(Arrays.asList((Term)new Ambiguity("K", new ArrayList<Term>(result))))));

Dwight, thank you for the apology. For the future, you should know that, due to events in my past, I have no tolerance for insults or rudeness.

So I can't help but feel troubled that this is your mindset ...

Please don't make assumptions or impute my motives. Personally, I think KIL would be cleaner to use in general if it had those factory methods (though perhaps they should be variadic constructors). It is, however, a change to the KIL interface (a core component that is not part of the parser), which is not something to be done willy-nilly and thus why I brought it up.

dwightguth: Okay. Well sorry if I misunderstood your thought process. But in the
future, please show more ownership of the code base. If a change is good,
easy to make, and makes the code cleaner, never second guess making it out
of a concern for code evolving or modifying a different part of the code
base. Just make it and get the owner of the code to review it.

Also, a quick note. I understand it's not good communication practices, and
rude, and I will try to avoid doing it, but in the future you should also
be aware that if I make comments like this I generally don't mean them at
all personally, they just represent confusion as to how a thought process
makes sense to someone. If the misunderstanding can be cleared up, I will
promptly forget the whole thing happened.

dwightguth:
Either label is or isn't nullable. If it's not, all states should have
labels, and a null argument should be a failed assertion. If it is, I
should not have to specify a null parameter in a constructor. Of course
being null has a meaning. That's what a nullable field is. But you still
don't make unscalable changes to the code base. Imagine if later we wanted
the default label to be a constant instead of null. Then your way of doing
things would require yet more unscalable changes.

This becomes a third foundational principle of code design which I will put
on the wiki tomorrow.
1. The ultimate objective of writing code is to encode each piece of
   information about how the system should behave in as few places as
   possible, and to avoid requiring boilerplate changes to methods that are
   not directly a part of a change.

Thus I speak to thee a second time: make it so.

dwightguth:
You've shoved at least 3, maybe more, conceptually distinct changes into this branch. I can no longer merge this, even if you had made the changes which I request five days ago now, which you haven't.

This is unacceptable. You need to learn how to use and merge branches in git, and you need to treat completing code reviews for your highest-priority task as highest priority, instead of continuing on to other changes without making the changes I requested or else persuading me I am wrong to want them.

dwightguth:
@mdmkolbe @radumereuta We spoke on Monday regarding this pull request. Once it has been cleaned up and the extra commits removed it should be extremely trivial to get it into a state that I can merge. I am a little confused as to why this hasn't happened yet because it should take about an hour and has been two days, in addition to the 5 days it had been before that. This is our highest priority change, so I'd like to see it merged right away.  I thought that we clearly agreed on how we should prioritize our work. In general, such plans should not be changed without prior agreement, because I need to be able to know that progress is being made towards the milestones that we agreed we would try to achieve. 

You should continue to work on your highest priority task until nothing further can be done on it (i.e. you are waiting for me to code review it, or it has already been merged into master). Only then do you work on the next-highest priority task. Thus we should never be in a situation with more than one change per person in progress which has not been submitted for code review.

radumereuta:
@dwightguth Sorry for not amending your comments.
I'm trying to focus on getting something that can be tested by the end of the week.
Right now this is my top priority.

You should ignore this pull request until next week.

dwightguth:
No, I shouldn't. I told you, your top priority is to get the code that labels ASTs merged into master. That is higher priority than preparing the next code review. Which is the comments on Grammar.java, which you also have partially completed and then moved onto another thing for. And then after that is the four things we discussed needed completion in order to use the parser out of the box, which is the nullability check, the ordering info, and the conversion to the grammar format from K. One of which is also "completed" and then ignored for review. 

So no, working on whatever you're focused on now isn't your highest priority. As I said, you should never start anything having to do with a lower priority change unless there is nothing at all you can do for a higher priority change. In particular what this means is that since you have outstanding work to be done on the labeling change, you should be working on that before anything else. A change is only completely done when it is merged into master. You should never begin a lower priority change if there is still work to be done on a higher priority change, and in particular there is always work to be done unless you are either waiting for me to reply to code review comments, or it is already merged into master. Which is not the case for three of the four things you have worked on this past week.

radumereuta:
I'm going to close this pull request and restart a new one by cherry picking the right order of commits so I preserve the modularity of changes.

PR #283, opened 2014-03-13
machine integers implemented with (java) primitives in the java backend

andreistefanescu:


src/javasources/KTool/src/org/kframework/backend/java/builtins/BitVector.java
```diff
@@ -0,0 +1,134 @@
...
+        this.value = value;
+        this.bitwidth = bitwidth;
+    }
+
+    /**
+     * Returns a {@code BitVector} representation of the given integer value on the given
+     * bit width.
+     */
+    public static BitVector of(BigInteger value, int bitwidth) {
+        switch (bitwidth) {
```
dwightguth, line 41: can you assert that the bitwidth is positive?

src/javasources/KTool/src/org/kframework/backend/java/builtins/PrimitiveIntToken.java
```diff
@@ -0,0 +1,101 @@
...
+import org.kframework.backend.java.symbolic.Visitor;
+import org.kframework.kil.ASTNode;
+
+/**
+ * A primitive integer token. Generic type T may be instantiated with one of
+ * {@code Byte}, {@code Short}, {@code Integer}, or {@code Long}.
+ *
+ * @author AndreiS
+ */
+public final class PrimitiveIntToken<T extends Number> extends BitVector<T> {
```
dwightguth, line 16: PrimitiveIntToken and BitVector feel like very similar classes to me. What exactly is this class providing by extending BitVector, and is there a way to do it without duplicating so much code?

src/javasources/KTool/src/org/kframework/backend/java/builtins/BitVector.java
```diff
@@ -0,0 +1,134 @@
...
+            case Byte.SIZE:
+                return PrimitiveIntToken.of(value.byteValue());
+            case Short.SIZE:
+                return PrimitiveIntToken.of(value.shortValue());
+            case Integer.SIZE:
+                return PrimitiveIntToken.of(value.intValue());
+            case Long.SIZE:
+                return PrimitiveIntToken.of(value.longValue());
+            default:
+                return new BitVector<>(value, bitwidth);
```
dwightguth, line 51: I'm concerned that this case in the code is going to require a bunch of extra logic. Is there some way we can design this which separates the logic of performing mathematical operations from the logic of what the semantics of those mathematical operations are?

PR #285, opened 2014-03-15
add negative contracts on KilProperty

dwightguth:
Technically the KilProperty contracts are ternary, not binary: a kilproperty can either be required/ensured, or it can be forbidden/destroyed, or it can be unspecified. Trying to implement a ternary EnumSet with an undefined state seemed overkill for this, so I just created negative annotations. If someone has a better solution, I'm open to it.


dwightguth:
@andreistefanescu or @traiansf Can you review this?

traiansf:
looks good.  But I would like to get more insight on what is this good for and patterns of usage.

andreistefanescu:
lgtm

dwightguth:
@traiansf does this additional comment address your concern?

traiansf:
@dwightguth Yes, it does. please feel free to merge.

PR #289, opened 2014-03-19
Fixing an issue with Tokens creating ambiguities with variables.

radumereuta:
@andreistefanescu requested this fix.
Now, if you write Id ::= "Main", whenever you will use the token 'Main' in rules or programs, it will be transformed into a token.
The general rule is: if you declare Token{...}, that sort will become a lexer sort. For lexer sorts, if you declare another production with a single terminal, that production will produce a generic token instead of a klabel.


dwightguth:
Needs an end to end test. Otherwise looks good.

yilongli:
I am afraid a test like `syntax Id ::= "Main"` would fail due to the check I put in `TokenSortCollector` before the transformation here.

dwightguth:
I don't understand. That's the exact case that this change is supposed to cover. Are you saying the change doesn't work?

yilongli:
I haven't checked but I suppose the check in `TokenSortCollector` would be preformed before this transformation so an error will be thrown before this transformation can take place.

andreistefanescu:
The following should be sufficient (please test with both java and maude):

module TEST
  syntax Id ::= "Main"
  configuration <k> Id2String(Main) </k>
endmodule

radumereuta:
You are right, it fails for the java backend.
The idea of this change is to allow the users to write
Id ::= "Main" as a short form of
Id ::= Token{"Main"}
Should I go ahead and add this case as an exception to that check?

yilongli:
Yes, please. Thank you, Radu.

dwightguth:
Still needs a test, then this can be merged.

dwightguth:
Actually a test for both backends really.

src/javasources/KTool/src/org/kframework/compile/transformers/RemoveBrackets.java
```diff
@@ -15,6 +20,28 @@ public RemoveBrackets(Context context) {
...
+    /**
+     * Fixing an issue where syntax like Id ::= "Main" should be automatically be transformed
+     * into constants (#token). I'm letting it parse as klabel at first so I don't modify the SDF parser
+     * which is unstable on modifications. After it parses I use the RemoveBrackets filter to do the post
+     * processing that transforms the TermCons element into a constant. Putting it here ensures me that
+     * the filter will be applied in every parser call.
+     */
+    @Override
+    public ASTNode transform(TermCons tc) throws TransformerException {
+        if (context.getTokenSorts().contains(tc.getSort())) {
```
andreistefanescu: please rename this class from RemoveBrackets to something like PostParserFilter

dwightguth: Makes sense. Do it.

radumereuta:
I need some help with this.
The tests seem to pass on my machine, even after merging with 'master'.

andreistefanescu:
There seem to be some problem with Jenkins. Some of the same tests are broken in my pull request as well, and they pass on my machine.

dwightguth:
retest this please

PR #291, opened 2014-03-19
machine integers implemented with (java) primitives or big integers in the java backend

andreistefanescu:
A test if failing due to a problem with Id2String. Fixing it requires #289. Please review the rest in the meantime.


include/builtins/mint.k
```diff
@@ -1,280 +1,310 @@
...
+               | uvalueMInt(MInt)     [function, hook(#MINT:uvalue)]
+
+  /* Checks whether a machine integer is zero */
+  syntax Bool ::= zeroMInt(MInt)    [function, hook(#MINT:zero)]
+
+  /*
+   * Functions for signed and unsigned minimum and maximum values of a machine
+   * integer on a given bit width.
+   */
+  syntax Int ::= sinfMInt(Int)    [function]
```
dwightguth: this doesn't look like an infimum/supremum to me. Should we maybe call this sMinValue, sMaxValue, etc? Or am I conceptually misunderstanding somehow?

include/builtins/mint.k
```diff
@@ -1,280 +1,310 @@
...
-                | "uleqMInt" "(" MInt "," MInt ")"    [function]
-                | "gtMInt" "(" MInt "," MInt ")"      [function]
-                | "ugtMInt" "(" MInt "," MInt ")"     [function]
-                | "geqMInt" "(" MInt "," MInt ")"     [function]
-                | "ugeqMInt" "(" MInt "," MInt ")"    [function]
-                | MInt "==MInt" MInt                  [function]
-                | MInt "=/=MInt" MInt                 [function]
+  syntax MInt ::= andMInt(MInt, MInt)   [function, hook(#MINT:and)]
+                | orMInt(MInt, MInt)    [function, hook(#MINT:or)]
+                | xorMInt(MInt, MInt)   [function, hook(#MINT:xor)]
```
dwightguth, line 122: complement?

include/builtins/mint.k
```diff
@@ -1,280 +1,310 @@
...
+
+  /*
+   * Division and reminder
+   * sdiv/srem  interprets operands as signed; undefined if the second argument
+   *            is 0 or if the result overflows (it happens when the first
+   *            operand is the minimum value and the second operand is -1)
+   * udiv/urem  interprets operands as unsigned; undefined if the second
+   *            argument is 0 
+   */
+  syntax MInt ::= sdivMInt(MInt, MInt)    [function, hook(#MINT:sdiv)]
```
dwightguth: should we consider a +MInt, -MInt, *MInt, /sMInt, /uMInt, etc. syntax maybe?

include/builtins/mint.k
```diff
@@ -1,280 +1,310 @@
...
-  rule ufMInt(resMInt(_, B:Bool, _)) => B
-  rule ofMInt(resMInt(_, _, B:Bool)) => B
+    I <Int sinfMInt(N) orBool I >Int ssupMInt(N)
+  rule
+    uoverflowMInt(N:Int, I:Int)
+  =>
+    I <Int uinfMInt(N) orBool I >Int usupMInt(N)
 
+  /* Projection functions for results of operations with overflow. */
+  syntax MInt ::= miMInt(MyList)          [function]
```
dwightguth, line 69: can you put a comment here referencing where in the code this is used?

include/builtins/mint.k
```diff
@@ -1,280 +1,310 @@
...
-        iMInt(sMInt(mi(N, I1))) +Int iMInt(sMInt(mi(N, I2))) <Int infMInt(N),
-        iMInt(sMInt(mi(N, I1))) +Int iMInt(sMInt(mi(N, I2))) >Int supMInt(N))
-  rule
-    uaddMInt(mi(N:Int, I1:Int), mi(N:Int, I2:Int))
-  =>
-    resMInt(
-        uMInt(mi(N, I1 +Int I2)),
-        false,
-        iMInt(uMInt(mi(N, I1))) +Int iMInt(uMInt(mi(N, I2))) >Int usupMInt(N))
+  syntax MyList ::= digitsOfMInt(MInt, Int)   [function, hook(#MINT:toDigits)]
```
dwightguth, line 145: can you put a comment here referencing where in the code this is used?

include/java/hooks.properties
```diff
@@ -43,27 +43,27 @@ KEqual\:ite : org.kframework.backend.java.builtins.TermEquality.ite
...
-\#INT32\:minInt32 : org.kframework.backend.java.builtins.BuiltinInt32Operations.min
-\#INT32\:maxInt32 : org.kframework.backend.java.builtins.BuiltinInt32Operations.max
-\#INT32\:absInt32 : org.kframework.backend.java.builtins.BuiltinInt32Operations.abs
-\#INT32\:_\=\=Int32_ : org.kframework.backend.java.builtins.BuiltinInt32Operations.eq
-\#INT32\:_\=/\=Int32_ : org.kframework.backend.java.builtins.BuiltinInt32Operations.ne
-\#INT32\:_>Int32_ : org.kframework.backend.java.builtins.BuiltinInt32Operations.gt
-\#INT32\:_>\=Int32_ : org.kframework.backend.java.builtins.BuiltinInt32Operations.ge
-\#INT32\:_<Int32_ : org.kframework.backend.java.builtins.BuiltinInt32Operations.lt
-\#INT32\:_<\=Int32_ : org.kframework.backend.java.builtins.BuiltinInt32Operations.le
+\#INT32\:_+Int32_ : org.kframework.backend.java.builtins.BuiltinBitVectorOperations.add
```
dwightguth: This looks like unused code...?

src/javasources/KTool/src/org/kframework/backend/java/builtins/BigIntegerBitVector.java
```diff
@@ -0,0 +1,284 @@
...
+        return BigInteger.ZERO;
+    }
+
+    private BigInteger unsignedMax() {
+        return BigInteger.ONE.shiftLeft(bitwidth).subtract(BigInteger.ONE);
+    }
+
+    @Override
+    public BigInteger signedValue() {
+        return value
```
dwightguth, line 64: What exactly does value represent if it's not the signed /or/ unsigned value?

src/javasources/KTool/src/org/kframework/backend/java/builtins/BigIntegerBitVector.java
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    @Override
+    public BitVector<BigInteger> mul(BitVector<BigInteger> bitVector) {
+        return BigIntegerBitVector.of(value.multiply(bitVector.value), bitwidth);
+    }
+
+    @Override
+    public BitVector<BigInteger> sdiv(BitVector<BigInteger> bitVector) {
+        if (!bitVector.signedValue().equals(BigInteger.ZERO)
```
dwightguth: I'd feel better if this just tried naively to divide the integers and let it throw an exception if there were divide by zero or over/underflow.

src/javasources/KTool/src/org/kframework/backend/java/builtins/BigIntegerBitVector.java
```diff
@@ -0,0 +1,284 @@
...
+                    bitwidth);
+        } else {
+            return null;
+        }
+    }
+
+    @Override
+    public BuiltinList sadd(BitVector<BigInteger> bitVector) {
+        BigInteger result = signedValue().add(bitVector.signedValue());
+        return new BuiltinList(ImmutableList.<Term>of(
```
dwightguth: is there not a BuiltinList.of?

src/javasources/KTool/src/org/kframework/backend/java/builtins/BigIntegerBitVector.java
```diff
@@ -0,0 +1,284 @@
...
+            return BigIntegerBitVector.of(
+                    signedValue().remainder(bitVector.signedValue()),
+                    bitwidth);
+        } else {
+            return null;
+        }
+    }
+
+    @Override
+    public BitVector<BigInteger> urem(BitVector<BigInteger> bitVector) {
```
dwightguth, line 125: same for the rest of these

src/javasources/KTool/src/org/kframework/backend/java/builtins/BigIntegerBitVector.java
dwightguth: also can this be made a helper function?

src/javasources/KTool/src/org/kframework/backend/java/builtins/BigIntegerBitVector.java
```diff
@@ -0,0 +1,284 @@
...
+        BigInteger result = signedValue().add(bitVector.signedValue());
+        return new BuiltinList(ImmutableList.<Term>of(
+                BitVector.of(result, bitwidth),
+                BoolToken.of(checkSignedOverflow(result))));
+    }
+
+    @Override
+    public BuiltinList uadd(BitVector<BigInteger> bitVector) {
+        BigInteger result = signedValue().add(bitVector.signedValue());
+        return new BuiltinList(ImmutableList.<Term>of(
```
dwightguth: same with this?

src/javasources/KTool/src/org/kframework/backend/java/builtins/BitVector.java
```diff
@@ -0,0 +1,207 @@
...
+    protected BitVector(T value, int bitwidth) {
+        this.value = value;
+        this.bitwidth = bitwidth;
+    }
+
+    /**
+     * Returns a {@code BitVector} representation of the given big integer value on the given
+     * bit width.
+     */
+    public static BitVector of(BigInteger value, int bitwidth) {
```
dwightguth, line 50: will this create an unchecked warning?

src/javasources/KTool/src/org/kframework/backend/java/builtins/BitVector.java
```diff
@@ -0,0 +1,207 @@
...
+        switch (bitwidth) {
+            case Integer.SIZE:
+                return Int32Token.of(value.intValue());
+            default:
+                return BigIntegerBitVector.of(value, bitwidth);
+        }
+    }
+
+    /**
+     * Returns a {@code BitVector} representation of the given big integer value on the given
```
dwightguth: s/big integer/long/

src/javasources/KTool/src/org/kframework/backend/java/builtins/BigIntegerBitVector.java
```diff
@@ -0,0 +1,284 @@
...
+
+import com.google.common.collect.ImmutableList;
+
+
+/**
+ * Implementation of a bit vector using {@link java.math.BigInteger}.
+ *
+ * @author AndreiS
+ */
+public class BigIntegerBitVector extends BitVector<BigInteger> {
```
dwightguth: final

src/javasources/KTool/src/org/kframework/backend/java/builtins/BuiltinBitVectorOperations.java
```diff
@@ -0,0 +1,274 @@
...
+import java.util.List;
+
+
+/**
+ * Table of {@code public static} methods on builtin fixed precision integers.
+ *
+ * @author AndreiS
+ */
+@SuppressWarnings("unchecked")
+public class BuiltinBitVectorOperations {
```
dwightguth: uninstantiable classes should be final and have a private constructor.

src/javasources/KTool/src/org/kframework/backend/java/builtins/BuiltinBitVectorOperations.java
```diff
@@ -0,0 +1,274 @@
...
+            }
+            return BitVector.fromDigits(digits, bitwidth.intValue());
+        } else {
+            throw new IllegalArgumentException(digitList + " contains list variables");
+        }
+    }
+
+    /**
+     * Throws {@link IllegalArgumentException}.
+     * @return does not return anything; it is not void as a convenience.
```
dwightguth: I think the better pattern here is to have a method that returns an exception and then to throw that exception. Like:

```
public static BoolToken eq(BitVector term1, BitVector term2, TermContext context) {
    if (term1.bitwidth() == term2.bitwidth()) {
        return term1.eq(term2);
    } else {
        throw failWithBitwidthMismatch(term1, term2);
    }
}

private static Exception failWithBitwidthMismatch(...) { return new IllegalArgumentException(...); }
```

dwightguth: although you should probably call the method bitwidthMismatchException() instead in this case.

src/javasources/KTool/src/org/kframework/backend/java/builtins/Int32Token.java
```diff
@@ -1,141 +1,241 @@
...
+        return Int32Token.of(value * bitVector.value);
+    }
 
-        Int32Token intToken = cache.get(value);
-        if (intToken == null) {
-            intToken = new Int32Token(value);
-            cache.put(value, intToken);
+    @Override
+    public Int32Token sdiv(BitVector<Integer> bitVector) {
+        if (!(bitVector.value == 0 || (value == Integer.MIN_VALUE && bitVector.value == -1))) {
```
dwightguth: I see why you did it in the place above the way you did it here, because it's needed here. I'd still rather it threw an exception than returned null though.

src/javasources/KTool/src/org/kframework/backend/java/builtins/BigIntegerBitVector.java
dwightguth: never mind about this, but I'd still like to see an exception instead of null.

src/javasources/KTool/src/org/kframework/backend/java/builtins/Int32Token.java
dwightguth: actually wait, hang on. Shouldn't it return the incorrect overflowed value and then set the overflow flag? You could potentially actually want the "incorrect" value of the arithmetic operation in your semantics. Division by zero part is fine to throw an exception though.

src/javasources/KTool/src/org/kframework/backend/java/builtins/MetaK.java
```diff
@@ -165,8 +167,8 @@ public static BuiltinMap variablesMap(Term term, TermContext context) {
      *            the term context
      * @return the K label
      */
-    public static KLabel getKLabel(KItem kItem, TermContext context) {
+    public static KItem getKLabel(KItem kItem, TermContext context) {
         // TODO(AndreiS): handle KLabel variables
-        return (KLabel) kItem.kLabel();
+        return new KItem(new KLabelInjection(kItem.kLabel()), new KList(), context);
```
dwightguth, line 172: is this part of a separate change?

src/javasources/KTool/src/org/kframework/backend/java/builtins/PrimitiveIntToken.java
```diff
@@ -0,0 +1,90 @@
...
+import org.kframework.backend.java.symbolic.Visitor;
+import org.kframework.kil.ASTNode;
+
+/**
+ * A primitive integer token. Generic type T may be instantiated with one of
+ * {@code Byte}, {@code Short}, {@code Integer}, or {@code Long}.
+ *
+ * @author AndreiS
+ */
+public abstract class PrimitiveIntToken<T extends Number> extends BitVector<T> {
```
dwightguth: Is this dead code?

src/javasources/KTool/src/org/kframework/backend/java/symbolic/JavaSymbolicBackend.java
```diff
@@ -1,5 +1,6 @@
 package org.kframework.backend.java.symbolic;
 
+import com.google.common.math.IntMath;
```
dwightguth: this looks unused

tests/regression/java-rewrite-engine/machine_integers/program5.test
```diff
@@ -0,0 +1,3 @@
+ MyListItem(miMInt(uaddMInt(mi(32, -1), mi(32, -1)))),
```
dwightguth: doesn't uaddMInt already return the same thing as this convoluted expression which calls it twice?

dwightguth:
Overall it looks pretty good. Tons of minor things. But I'd also like to see more tests. When we spoke earlier I told you we didn't need test coverage for copy/paste functionality that was identical. But there's for example logic specific to division which isn't tested here. I'd also like to see unit tests instead of just end to end tests. Preferably something that is capable of testing several different bit widths without additional duplication of code. BuiltinBitVectorOperations is the perfect candidate for a test that asserts a formal relation between input and output for all of the unique classes of hooked operations (according to how their code is similar to each other).

include/builtins/mint.k
andreistefanescu: I think /sMInt is a little bit weird, and then I wanted to be consistent (and avoid mixfix syntax altogether). I think add, sub, etc are fairly standard names.

include/builtins/mint.k
andreistefanescu, line 145: Actually, this one is not used anywhere in the library. It is provided as a fast way of converting integers to bytes (for storage in memory in languages which store things as bytes).

src/javasources/KTool/src/org/kframework/backend/java/builtins/BigIntegerBitVector.java
andreistefanescu, line 64: This is the signed value. I preferred to call the methods signedValue and unsignedValue for clarity.

src/javasources/KTool/src/org/kframework/backend/java/builtins/BitVector.java
andreistefanescu, line 50: I don't think so...why do you say that?

src/javasources/KTool/src/org/kframework/backend/java/builtins/MetaK.java
andreistefanescu, line 172: A minor change...I changed the order in which hooks and K rules are applied (now hooks come first), and I also fixed this issue. I think it is minor enough to go through.

include/builtins/mint.k
dwightguth: okay. Then we need eqMInt and neqMInt too.

src/javasources/KTool/src/org/kframework/backend/java/builtins/PrimitiveIntToken.java
andreistefanescu: Yes it is, thank you for noting.

include/builtins/mint.k
dwightguth, line 145: okay. Then put a comment to this effect instead.

src/javasources/KTool/src/org/kframework/backend/java/builtins/BigIntegerBitVector.java
dwightguth, line 64: if value is the signed value then signedValue() should return value without modification. But it doesn't.

src/javasources/KTool/src/org/kframework/backend/java/builtins/BitVector.java
dwightguth, line 50: BitVector is a raw type here, isn't it?

tests/regression/java-rewrite-engine/machine_integers/program5.test
andreistefanescu: Good catch...thanks!

src/javasources/KTool/src/org/kframework/backend/java/builtins/MetaK.java
dwightguth, line 172: yeah, that's fine

src/javasources/KTool/src/org/kframework/backend/java/kil/KItem.java
```diff
@@ -264,6 +264,25 @@ public Term evaluateFunction(SymbolicConstraint constraint, TermContext context)
         }
         KList kList = (KList) this.kList;
 
+        if (BuiltinFunction.isBuiltinKLabel(kLabelConstant)) {
```
dwightguth, line 267: this is a relatively small ancillary change and doesn't muddle up the diff too much, so I'm not that worried about keeping it in a separate diff, but you need to add a comment here explaining /why/ hooks are applied at this point and not the later point.

src/javasources/KTool/src/org/kframework/backend/java/builtins/BitVector.java
andreistefanescu, line 50: Yes, cut converting to a raw type is always allowed. Converting a raw type would create an unchecked warning.

andreistefanescu:
I also want to test the implementation of machine integers with K rules. These end to end tests can be run with both backends (or at least those which do not use MyList which is not supported in Maude yet), while unit tests can only be run with the java backend. I can add some more tests.

src/javasources/KTool/src/org/kframework/backend/java/builtins/BitVector.java
dwightguth, line 50: Okay. Can you give me a brief summary of where in the code this constructor
is called, then?

dwightguth:
Okay, that's reasonable reasoning. Just, as I said, make sure every unique
bit of logic in these classes is tested.

src/javasources/KTool/src/org/kframework/backend/java/builtins/BigIntegerBitVector.java
andreistefanescu, line 64: Sorry, I answered without paying attention. The value coincides with the signed value and the unsigned value on the last bitwidth bits (essentially, if you add and it overflows, it does not normalize the value until it is needed). Should I just keep it normalized (I don't think it matters much though).

src/javasources/KTool/src/org/kframework/backend/java/builtins/BitVector.java
andreistefanescu, line 50: It is called in BuiltinBitVectorOperations, and there it generates a lot of unchecked warnings. But I don't know how to avoid that. Do you have any suggestions?

src/javasources/KTool/src/org/kframework/backend/java/builtins/BigIntegerBitVector.java
dwightguth, line 64: That's fine as long as you make sure never to access value directly (it
should be protected but not public) and put a comment in BitVector.java
indicating that direct access to the value field is not guaranteed to be
safe.

andreistefanescu, line 64: It is protected already. I'll add a comment.

include/builtins/mint.k
andreistefanescu, line 122: xor 0

grosu:
Dwight,

These messages are quite annoying and useless for most us us.  Can they be just sent to you and the author of the pull request?

Grigore

src/javasources/KTool/src/org/kframework/backend/java/builtins/Int32Token.java
andreistefanescu: The problem is that div and rem are a combo: the reminder is undefined if the division overflows. I don't know if we should change this...

include/builtins/mint.k
dwightguth, line 122: at least a helper function then?

src/javasources/KTool/src/org/kframework/backend/java/builtins/Int32Token.java
dwightguth: I don't understand. anything % -1 == 0. Why does remainder come into this?

include/builtins/mint.k
```diff
@@ -1,280 +1,310 @@
...
+  rule
+    soverflowMInt(N:Int, I:Int)
+  =>
+    I <Int sminMInt(N) orBool I >Int smaxMInt(N)
+  rule
+    uoverflowMInt(N:Int, I:Int)
+  =>
+    I <Int uminMInt(N) orBool I >Int umaxMInt(N)
+
+  /* Projection functions for results of operations with overflow. */
```
dwightguth: I still want like an example usage pattern here.

include/builtins/mint.k
```diff
@@ -1,280 +1,310 @@
...
+                | sgeMInt(MInt, MInt)   [function, hook(#MINT:sge)]
+                | ugeMInt(MInt, MInt)   [function, hook(#MINT:uge)]
+                | eqMInt(MInt, MInt)    [function, hook(#MINT:eq)]
+                | neMInt(MInt, MInt)    [function, hook(#MINT:ne)]
+
+  /*
+   * Returns a list of digits representing the given machine integers in the
+   * given base. Each digit is represented as a machine integers. The base must
+   * be an integer greater than 1.
+   */
```
dwightguth, line 144: the comments on these two functions are good but don't explain why they exist.

src/javasources/KTool/src/org/kframework/backend/java/builtins/Int32Token.java
andreistefanescu: No. The reminder a % b is defined in terms of a / b. Thus, if a / b is undefined (because it overflows), then the reminder is also undefined

dwightguth: (declare-fun x () (_ BitVec 8))
(assert (not (= (bvsrem x #xff) #x00)))
(check-sat)
unsat

Doesn't look undefined to me. In general shouldn't we be implementing the
same semantics for our data types that smtlib does? Or at least using a
semantics that we can encode in smtlib? Otherwise our interpreter will be
unsound, right?

andreistefanescu: Dwight, we cannot simply mimic smtlib, because smtlib does not differentiate between singed and unsigned value.

src/javasources/KTool/src/org/kframework/backend/java/builtins/BigIntegerBitVector.java
```diff
@@ -0,0 +1,278 @@
...
+
+    @Override
+    public BuiltinList sadd(BitVector<BigInteger> bitVector) {
+        BigInteger result = signedValue().add(bitVector.signedValue());
+        return getBuiltinList(result, checkSignedOverflow(result));
+    }
+
+    @Override
+    public BuiltinList uadd(BitVector<BigInteger> bitVector) {
+        BigInteger result = signedValue().add(bitVector.signedValue());
```
dwightguth: shouldn't this be unsignedValue? And some other places in this file too...

src/javasources/KTool/src/org/kframework/backend/java/builtins/BigIntegerBitVector.java
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    @Override
+    public BuiltinList umul(BitVector<BigInteger> bitVector) {
+        BigInteger result = signedValue().multiply(bitVector.signedValue());
+        return getBuiltinList(result, checkUnsignedOverflow(result));
+    }
+
+    @Override
+    public BitVector<BigInteger> shl(IntToken intToken) {
```
dwightguth, line 172: I think we should consider making these undefined for negative Ints.

src/javasources/KTool/src/org/kframework/backend/java/builtins/BigIntegerBitVector.java
andreistefanescu: Yes, they should. Thanks for noticing the bug.

andreistefanescu:
retest this please

andreistefanescu:
retest this please

andreistefanescu:
Dwight, I think I addressed all the comments. Please let me know if there are any other issues. I am still waiting on the issue with tokens in order to pass my last test, but I would like to merge after that.

dwightguth:
I'm satisfied. You can merge when ktest passes.

andreistefanescu:
retest this please

PR #293, opened 2014-03-20
Removing pdmc from master

traiansf:
@dwightguth  I'm moving this work to another branch.  Will reintegrate it step by step (and test by test).


dwightguth:
it's hard to tell from this diff because github kind of sucks, but if this removes the entire backend/pdmc folder then you can merge this as soon as the tests pass. Otherwise I want a little detail about what you're removing and what you're not.

traiansf:
Removing pdmc completely (src and test) was all I did.  I'm merging this.

dwightguth:
Andrei brought something to my attention which is that you deleted the line
in the ant script which cleans up the generated javacc files. I didn't
catch the implications of that or I'd have stopped you from merging. You've
broken everyone's working copy when they merge because of it. Can you add
the line back in?

traiansf:
Are you sure? It was actually breaking on ant clean on my machine because I
didn't delete it.

I can put that back, but since it would have to be removed anyway once it
runs once, wouldn't it be better for people to run a
  git clean -f
?

2014-03-21 0:56 GMT+02:00 dwightguth notifications@github.com:

> Andrei brought something to my attention which is that you deleted the line
> in the ant script which cleans up the generated javacc files. I didn't
> catch the implications of that or I'd have stopped you from merging. You've
> broken everyone's working copy when they merge because of it. Can you add
> the line back in?

dwightguth:
Hmm. Well if the code doesn't work if the folder isn't there then yeah,
let's keep this as is. Please send out a general announcement on the klist
with a specific command to run if this breaks your stuff though, because
Andrei was stumped by it.

PR #294, opened 2014-03-20
Kore first2

traiansf:
The idea of this commit is to split FlattenSyntax into two, so that one of them can be gradually moved before others to reduce the definition to KORE as soon as possible.

I believe this is a bad commit for a code review, but I would like guidelines.

First, the code I'm modifying has tabs.
Next, IntelliJ Idea has the nice habit of compressing imports.


owolabileg:
--> Next, IntelliJ Idea has the nice habit of compressing imports.
I had to change the settings to avoid this: 
1) Go to File -> Settings -> Code Style -> Java
2) click on the imports tab
3) Set "Class count to use import with '*'" to higher number than the default (5). I set mine to 20 for now.

dwightguth:
My advice to you is to merge master into unified-builtins2 before doing this change. That should deal with the tabs because master already has removed all tabs. If you have troubles with the merge I suggest using a ignore-all-space merge strategy and then running a script to remove all the tabs that this strategy introduced. If you have trouble, let me know and I can help you out.

Then you should be able to make this change without worrying about style guidelines.

dwightguth:
What's the next step here?

radumereuta:
Fix the build first, when it passes, then merge.
All the transformation steps need to be revised and modified to work with this change.

dwightguth:
Okay but I don't even know what this code review is doing. Is it the same thing as #302? Both have the exact same description. These commits include a bunch of whitespace changes anyway that don't belong here. Can I just close this and have the changes be recreated elsewhere? Or what? @traiansf ?

traiansf:
Yes, I've reopened this as #302 Please review (and perhaps merge) that one.

PR #295, opened 2014-03-20
Refactor ktest to clean up some irregularities

dwightguth:
I am working on a change to do something different with the options to Kompile, and in order to make that work it was necessary to make some changes to ktest. So for simplicity I broke them off into their own pull request. To summarize:
- refactored the xsd to be more modular, and changed the syntax for configuration parameters in ktest xml
- regularized use of - and -- to be a simple pass-through to the underlying process
- fix some small bugs preventing kast from being used for the java backend
- make kompile and krun run out of the working directory of the definition
- remove "=" from most parameter invocations, because the new options framework will not support it.
- uncomment and fix some test cases that were inappropriately commented out

Note I am pretty sure this won't pass all the tests yet, I am just trying to run the test suite on jenkins because it's faster.


dwightguth:
retest this please

dwightguth:
@osa1 @daejunpark Can you review this please?

osa1:
Awesome! Thanks for getting your hands on config file parsing and fixing that hacky argument parsing parts. I'm currently going through the patch.

dwightguth:
@osa1 or @daejunpark can you give me an ETA on reviewing this? I would like to get this merged so it doesn't block my rollout of the JCommander change.

lib/ktest.xsd
```diff
@@ -1,5 +1,24 @@
...
+      <xs:attribute name="name" type="xs:string" use="required" />
+    </xs:extension>
+  </xs:complexContent>
+</xs:complexType>
+
+<xs:complexType name="option">
+  <xs:attribute name="name" type="xs:string" use="required" />
+  <xs:attribute name="key" type="xs:string" use="optional" />
+  <xs:attribute name="value" type="xs:string" use="optional" />
+</xs:complexType>
```
daejunpark, line 21: I really like this, which I wanted to do but I can't because of lack of the knowledge of XSD. It looks good much better!

samples/java_rewrite_engine/tutorial/1_k/2_imp/tests/config.xml
```diff
@@ -1,11 +1,11 @@
...
   <include file="../../../../../../tutorial/1_k/2_imp/lesson_5/tests/config.xml" >
-   <kompile-option name="-backend" value="java" />
+   <kompile-option name="--backend" value="java" />
     <all-programs>
-      <krun-option name="-backend" value="java" />
-      <krun-option name="-smt" value="none" />
-      <krun-option name="-color" value="off" />
+      <krun-option name="--backend" value="java" />
+      <krun-option name="--smt" value="none" />
+      <krun-option name="--color" value="off" />
```
daejunpark, line 8: I'm not sure if it works with the XML comment. The reason of using a single dash might be to simply comment out the entire `<test>` nodes. (Of course, I agree we shouldn't do that.) I can remember that there was a suggestion that we can specify the name of options without dashes. Personally, I have no preference, and @osa1 will have a better idea.

daejunpark:
@dwightguth I just leaved minor comments, but I think that @osa1 is definitely better than me to review this since he is the father of the current ktest. :)

samples/java_rewrite_engine/tutorial/1_k/2_imp/tests/config.xml
osa1, line 8: Ahh, right. Now I remember it. @dwightguth if we use `--` before arguments, we can't comment out test blocks in XML files because `--` in comments is invalid in XML. So in the past we thought maybe instead of putting `--` before arguments, why not just remove `--` altogether? Provided that we'll have long names for all arguments, this should not be a problem.

dwightguth, line 8: Yes, I'm aware of the issue with xml comments. I don't really know how to fix it though. The problem with what you propose is that it adds a very undesirable piece of complexity to ktest. What if later we want to make a way to call krun on more than one file? Then we would specify something like:

<krun-option name="some-file" />

and it would break. The reason I have it the way it is now is because ktest is now strictly pass-through: what it gets is exactly what it passes to the process to be run. This simplicity of design will prevent us from running into this type of problem later, not to mention it will greatly decrease confusion to users. Yeah, it interferes with comments, but /**/ can't be nested in most programming languages. Programmers are used to modifying code that's commented out if they really have to.

dwightguth:
@osa1 if you have other comments to make you need to tell me now or I'll merge this later today.

samples/java_rewrite_engine/tutorial/1_k/2_imp/tests/config.xml
osa1, line 8: OK, I think your arguments about ktest being strictly pass-through and calling kompile/krun/etc. with multiple files make sense.

osa1:
@dwightguth looks good to me.

PR #302, opened 2014-03-22
Splitted Flatten Syntax in two.

traiansf:
This pull request is just about splitting FlattenSyntax into two classes, one which only flattens the terms and the other which only flattens syntax declarations.
At the moment, FlattenSyntax calls the FlattenTerms transformer when transforming a definition.  To actually split them we need a discussion about how do we handle the succession of transformations for 7-8 backends we have, which all seem to differ one from another in the current steeps before FlattenSyntax.


traiansf:
So... what do we do about this?  The tests are failing because the  tests on branch unified-builtins have not yet been stabilized.

src/javasources/KTool/src/org/kframework/compile/transformers/FlattenSyntax.java
```diff
@@ -15,34 +15,32 @@
 import java.util.List;
 import java.util.Set;
 
+/**
+ * Transformer replacing concrete syntax declarations by KLabel declarations.
+ */
 @KilProperty.Ensures(KilProperty.NO_CONCRETE_SYNTAX)
```
dwightguth, line 21: If this transformer no longer processes terms then this KilProperty should be split into NO_CONCRETE_SYNTAX and NO_CONCRETE_SYNTAX_DECLARATIONS

dwightguth, line 21: oh wait I see it's still doing it for now, never mind.

dwightguth, line 21: Okay, yeah, this should ensure BOTH of the KilProperties I mentioned above.

dwightguth:
It's hard for me to read this diff, but if all you did was take all the code pertaining to flattening terms and put it in its own transformer more or less verbatim, this looks okay.

Can you explain to me what you mean by "tests on the branch have not yet been stabilized"? Because from what I see on the Jenkins dashboard it would seem to indicate tests in this branch should be passing.

dwightguth:
Can this be merged into master in addition to the unified-builtins branch? Liyi needs this also and I want this to be done only once.

radumereuta:
It should be able to do so, but fixing the failing tests might take a while.

dwightguth:
Hmm. We seem to have a lot of stuff waiting on different things. Handle it however you feel is appropriate as long as I don't have to code review anything twice, and we don't ever break tests in master. But if there's a legitimate reason why it would be easier to let tests break and wait for another piece of work to finish in order to fix the tests, then that's fine as long as the only things we break are feature branches like this one.

traiansf:
we could.  but I would propose Liyi work in the unified-builtins branch.
 There might be some things which need to be done differently with lists,
maps, and sets concerning KORE.

nevertheless, we can do a similar thing on master.  Actually, We could move
this push request on master and then merge it into the unified bultins
branch.

2014-03-25 0:00 GMT+02:00 dwightguth notifications@github.com:

dwightguth:
Oh okay. I misinterpreted from my discussion with Liyi earlier, then. Don't
worry about master then, please. You can take like a couple hours to try it
out if you want and if it works, okay, but don't spend lots of time on it
if it turns out to be complicated.

traiansf:
I'm closing this.  We'll merge changes from master once #305 is merged.

PR #305, opened 2014-03-25
Splitted Flatten Syntax in two.

traiansf:
This is the same change as #302, only this time applied on the master branch.  If this passes tests, you can merge it and I'll merge master into unified-builtins then (so we can remove the other pull request).

### DIFF
--- a/src/javasources/KTool/src/org/kframework/compile/transformers/FlattenSyntax.java
+++ b/src/javasources/KTool/src/org/kframework/compile/transformers/FlattenSyntax.java
@@ -14,7 +14,7 @@ import java.util.Set;
 /**
  * Transformer replacing concrete syntax declarations by KLabel declarations.
  */
-@KilProperty.Ensures(KilProperty.NO_CONCRETE_SYNTAX)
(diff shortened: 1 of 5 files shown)


traiansf:
Ok. good we did not commit it on the branch. There were some issues in krun, which run FlattenSyntax, but on terms not on definitions.
Now it's ready for review & approval

To answer the questions there, indeed FlattenSyntax&FlattenTerms are obtained by splitting the original FlatenSyntax in two, and added a call to FlattenTerms in the transformer for Definition of FlattenSyntax to avoid updating the list of Transformation for the 7-8 backends.

dwightguth:
If you fix the KilProperty thing that I talked about, and get @radumereuta or @liyili2 to sign off on this, you can merge it.

radumereuta:
That was surprisingly easy to fix.
lgtm

traiansf:
@dwightguth: I've modified the KilProperties, added Javadoc for them, added both to the FlattenSyntax for now (took a couple of minutes and google searches to figure how that is done) and all tests passed.  As @radumereuta gave his ok, I'm merging this.

PR #307, opened 2014-03-25
renamed theMain in Main and theObject in Object in kool; no need for Id2...

andreistefanescu:
...String definition for Main or Object


yilongli:
Thanks, Andrei. Btw, we may actually reuse the test programs for maude backend in the samples directory. As far as I know, theMain/theObject should be the only change I made to those test programs.

andreistefanescu:
Thank you Yilong, I missed that.

PR #308, opened 2014-03-25
fix bug where kompile crashes but ktest reports success

dwightguth:
This is a bug that causes serialization errors to report a 0 return value on the command line, resulting in ktest passing even though kompilation did not succeed.


dwightguth:
@andreistefanescu can you review this?

andreistefanescu:
lgtm

PR #309, opened 2014-03-25
use JCommander in kompile

dwightguth:
Please review only commit a8f9444. The others belong to pull request #295 and will be rebased out before merging.

This change:
- replaces the old Apache CLI parser for kompile with a new one using JCommander
- removes all kompile-specific settings from GlobalSettings and uses instance instead of static state to obtain the relevant settings
- serializes and deserializes kompile options to krun and merges the backend flag into krun's settings
- modifies the usage pattern of the Stopwatch to make it more compatible with the instance state
- breaks kcheck; we are deprecating this tool and it was completely impossible to fix it in the new architecture, because of its undesirable reliance on static state
- breaks code which attempts to access "--" options with only one "-"
- breaks code which attempts to specify arguments to a parameter with an "=" (except -cFOO=bar)
- renames experimental options to begin with an X
- possibly some other small stuff?

This isn't ready for review quite yet, I'm just creating the pull request to utilize Jenkins.


dwightguth:
I am working to figure out whether any performance optimizations can be made to speed up ktest so that this doesn't time out, but in the meantime I believe it is ready for review. I would like feedback from @traiansf @ciobaca @andreistefanescu @andreiarusoaie @liyili2 @osa1 @radumereuta . Basically if you own a tool or a kompile backend I need you to test your stuff and if I broke it tell me so I can add a regression test.

radumereuta:
I've tried it briefly and it worked for me.
You may want to do something with the length of the lines when you run --help. Many of them pass over the 80 character limit.

src/javasources/KTool/src/org/kframework/backend/symbolic/TagUserRules.java
```diff
@@ -28,25 +26,19 @@
  */
 public class TagUserRules extends CopyOnWriteTransformer {
 
-    public static final Set<String> notSymbolicTags;
```
traiansf, line 31: I'm trying to understand why a static object is bad in this instance, as notSymbolicTags is practically a constant.  One benefit I can see is that it is initialized only once and it can be accessed by other classes.

dwightguth, line 31: Because the code isn't guaranteed to work. We have no guarantee that a
static block will execute after any non-static code, such as for example
the code that would initialize GlobalSettings.nonSymbolicTags. If you want
to argue that the access to kompile settings isn't needed here, okay, I can
get rid of it. But static code can't access data that was initialized by
instance fields. It's part of the Java standard.

src/javasources/KTool/src/org/kframework/kompile/KompileOptions.java
```diff
@@ -0,0 +1,198 @@
...
+import org.kframework.parser.ExperimentalParserOptions;
+
+import com.beust.jcommander.IStringConverter;
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.ParameterException;
+import com.beust.jcommander.ParametersDelegate;
+
+public final class KompileOptions implements Serializable {
+    
+    public static enum Backend {
```
adamsmd, line 26: Add comment explaining why these constant are lower case.

src/javasources/KTool/src/org/kframework/main/GlobalOptions.java
```diff
@@ -0,0 +1,56 @@
...
+import org.kframework.utils.Stopwatch;
+import org.kframework.utils.errorsystem.KException.ExceptionType;
+import org.kframework.utils.errorsystem.KExceptionManager;
+import org.kframework.utils.general.GlobalSettings;
+
+import com.beust.jcommander.Parameter;
+
+public final class GlobalOptions {
+    
+    public static enum Warnings {
```
adamsmd, line 19: Add comment explaining why these constants are lower case.

src/javasources/KTool/src/org/kframework/kompile/KompileOptions.java
```diff
@@ -0,0 +1,198 @@
...
+        pdf(true, false),
+        latex(true, false),
+        doc(true, false),
+        html(true, false),
+        kore(false, false),
+        maude(false, false),
+        java(false, true),
+        unparse(false, false),
+        unflatten(false, false),
+        unflatten_java(false, true),
```
adamsmd: Is JCommander going to make this option have to be "unflatten_java" or can we still use "unflatten-java"?

dwightguth: It's going to have to be unflatten_java, I'm pretty sure. If you're
concerned, I can look further into customizing this behavior, it should be
possible. It will just require a bit more boilerplate code.

adamsmd: Well, I see two possible concerns.

(1) If the tutorial, documentation or some tool uses this option, then this would break the command line API presented by kompile.

(2) Most program use "-" instead of "_" in their command line options.  One reason for this is that "-" is easier to type and is a single keystroke instead of a shift-keystroke (at least on a US Qwerty keyboard).  But, without getting into debates about the aesthetics one versus the other, "-" is much more common and will be more familiar to users.

In the grand scheme of things, both of these concerns are relatively minor, so it comes down to a question of how much we allow the limitations of JCommander to dictate the command line interface.  If it is relatively easy to get JCommander to support "-", then I'd do that, but I understand if it takes a disproportionate amount of code.

(Would it be possible to tell JCommander to convert "-" to "_" and lower case to upper case before trying to match these enums?  That would solve both this issue and the "lower case constant" problem.)

dwightguth: I will investigate. Worst case I implement a one-line class for every
individual Enum type we do this with, not a huge deal.

dwightguth:
oops, I rebased wrong, hang on.

dwightguth:
Okay, this is ready. @traiansf @ciobaca @andreistefanescu @andreiarusoaie @liyili2 @osa1 Please review. If you own a kompile backend or a tool which does not have regression tests, and you want it to not be broke, you /MUST/ test this change. Remember that you are expected to review changes within 2 business days (i.e. in this case, by the end of the day Tuesday).

liyili2:
good. It looks good for me. I approve it.

dwightguth:
test this please

src/javasources/KTool/src/org/kframework/backend/html/HtmlBackend.java
```diff
@@ -27,8 +27,8 @@ public void run(Definition definition) throws IOException {
 
         String html = htmlFilter.getHTML();
 
-        FileUtil.save(GlobalSettings.outputDir + File.separator + FilenameUtils.removeExtension(new File(definition.getMainFile()).getName()) + ".html", html);
-        FileUtil.save(GlobalSettings.outputDir + File.separator + "k-definition.css",
+        FileUtil.save(options.directory.getPath() + File.separator + FilenameUtils.removeExtension(new File(definition.getMainFile()).getName()) + ".html", html);
+        FileUtil.save(options.directory.getPath() + File.separator + "k-definition.css",
```
radumereuta, line 31: I think outputDir was a better name than just simply 'directory'. It's too general.

src/javasources/KTool/src/org/kframework/backend/latex/DocumentationFilter.java
```diff
@@ -38,7 +38,7 @@ public void visit(Rule rule) {
         // termComment = false;
         Attributes atts = rule.getAttributes(); 
         boolean process = false;
-        for(String tag : GlobalSettings.doctags) {
+        for(String tag : options.experimental.documentation) {
```
radumereuta, line 41: Again, doctag is a better name.
The name 'documentation' was used at one point as a boolean if the --doc option was used.

src/javasources/KTool/src/org/kframework/backend/latex/LatexBackend.java
```diff
@@ -45,11 +46,11 @@ public void compile(Definition javaDef) throws IOException {
         String latexified = "\\nonstopmode" + endl +
                 "\\PassOptionsToPackage{pdftex,usenames,dvipsnames,svgnames,x11names}{xcolor}"+ endl +
                 "\\PassOptionsToPackage{pdftex}{hyperref}"+ endl +
-                "\\documentclass{article}" + endl + "\\usepackage[" + GlobalSettings.style + "]{k}" + endl;
+                "\\documentclass{article}" + endl + "\\usepackage[" + options.docStyle() + "]{k}" + endl;
         String preamble = lf.getPreamble().toString();
         latexified += preamble + "\\begin{document}" + endl + lf.getResult() + "\\end{document}" + endl;
 
-        File canonicalFile = GlobalSettings.mainFile.getCanonicalFile();
+        File canonicalFile = options.definition();
```
radumereuta: Better name for the main file of a definition please.

src/javasources/KTool/src/org/kframework/kast/KastFrontEnd.java
```diff
@@ -64,12 +69,14 @@ public static void kast(String[] args) {
...
-            GlobalSettings.verbose = true;
+            globalOptions.verbose = true;
         }
+        
+        globalOptions.initialize();
 
         // set fast kast option
         if (cmd.hasOption("fast-kast")) {
-            GlobalSettings.fastKast = !GlobalSettings.fastKast;
+            parserOptions.fastKast = true;
```
radumereuta, line 79: It's ok to change this, but I did this for a reason. It was easy to test. If I wanted to set fast-kast as the default option I would go into GlobalSettings and modify the option to true. If I wanted to temporarily disable it, I would just add the fast-kast option to kompile.

src/javasources/KTool/src/org/kframework/backend/html/HtmlBackend.java
dwightguth, line 31: I did this for a reason, to give one-to-one correlation between variable names and option names. If you think it's a bad name, then we need to discuss changing the name of the option that specifies this value. As is, it clearly identifies itself as a field containing the value of the option with this name, which wouldn't be clear under your proposal.

src/javasources/KTool/src/org/kframework/backend/latex/DocumentationFilter.java
dwightguth, line 41: see above.

src/javasources/KTool/src/org/kframework/utils/Stopwatch.java
```diff
@@ -1,17 +1,26 @@
+// Copyright (C) 2012-2014 K Team. All Rights Reserved.
 package org.kframework.utils;
 
-import org.kframework.utils.general.GlobalSettings;
+import org.kframework.main.GlobalOptions;
 
 import java.util.Formatter;
 
 public class Stopwatch {
```
radumereuta, line 11: This class is used intensively in many parts of the tool. Please add javadoc which explains the design decision and how to use it.

dwightguth:
test this please

src/javasources/KTool/src/org/kframework/backend/html/HtmlBackend.java
radumereuta, line 31: Fair enough, then be sure you add javadoc to the field.

radumereuta:
It looks like a welcomed refactoring of some global fields. If you make all tests pass, I'm ok with this change.
Also, don't forget to merge it into unified-builtins :)

traiansf:
I've tested it on the PDF and HTML backends and everything seems fine.
Ready to merge as far as I'm concerned.

Just a note that the help message does not fit into 80 characters/line.

dwightguth:
I fixed the issue with description line length. @andreiarusoaie @ciobaca This is your _last_ chance to review this change. If nobody has any other comments and all the tests pass I will merge this tomorrow. If that breaks the symbolic backend or kagreg, that is on your own heads. I already had to break kcheck to make this work, because it was completely unmaintainable code. It will be up to you to create a pull request which integrates your prover into `krun --prove`.

andreiarusoaie:
Sorry, I forgot to comment on this. It looks ok for me. You can merge.

dwightguth:
retest this please

PR #310, opened 2014-03-25
java 6 hasn't worked for a while now; deleting

dwightguth:
I'm not sure why nobody noticed, but a number of classes in the framework no longer compile on Java 6 (mostly because of the try(Object foo = bar()) {} syntax). So I am removing the build path for Java 6 since it's been broken for a while now. Apparently once upon a time we had users who couldn't upgrade to Java 7 but that is apparently no longer the case, or we'd have heard about it by now.


dwightguth:
@radumereuta these were your build files originally can you review?

dwightguth:
@radumereuta please review this, it's a very simple change and I want to get it off my plate.

radumereuta:
Ah, sorry, I've missed your first comment.
It looks ok. You can merge.

PR #312, opened 2014-03-26
Klabel strictness

traiansf:
Updated code to allow strict and seqstrict for KLabel. Although they were implemented, the code was rewriten such that they were no longer allowed.

Still missing: position arguments for strict and seqstrict.


andreistefanescu:
I believe this is my fault. Thank you Traian!

traiansf:
2014-03-27 5:09 GMT+02:00 Andrei Stefanescu notifications@github.com:

> I believe this is my fault. Thank you Traian!

Well, there is still work to do (strict still cannot have arguments for
klabels), but at least we're back to the original functionality.

dwightguth:
Mostly looks good. Can you either replace this test with a unit test or make the test much simpler though? The more tests we add the slower ktest runs. I don't think we need this complex of a K definition to test the feature you're dealing with here, do you?

traiansf:
I agree, in principle.  However, I was working now on adding parameters for
strict, in which case this test would be easily converted to one similar to
the one currently in the strictness directory (from which it was derived).

So, if I simplify it now, I'll add a new one as soon as I implement strict
with arguments.  What do you think would be best?

2014-03-27 19:53 GMT+02:00 dwightguth notifications@github.com:

dwightguth:
Is it possible to merge the two now? so that we have only one definition containing all the strictness tests? and then you expand upon it as needed?

traiansf:
yes, I can do that, I believe.

2014-03-28 0:48 GMT+02:00 dwightguth notifications@github.com:

traiansf:
ok, i've merged the two tests.  does it look ok now?

Issue #313, opened 2014-03-27
KLabel use in rules for function does not work well


grosu:
Consider this:

```
module TRY
  syntax KLabel ::= "#foo"
  syntax K ::= foo(K)  [function, klabel(#foo)]
  rule #foo(1) => 0
endmodule
```

If I krun the program `foo(1)` reduces as expected to 0.  However, if I change the program into `foo(foo(1))` it stays unreduced.  On the other hand, if I change the rule to `rule foo(1) => 0` then `foo(foo(1))` reduces to `foo(0)` as expected.

grosu:
And if the above was not confusing enough, it actually seems to work if you add the attribute anywhere to the original rule above.  But of course, that anywhere attribute should be redundant here, because we already said that foo is a function.

radumereuta:
I think the filter that applies 'anywhere' to rules matches on TermCons instead of KApp.
This would explain the behavior.
I don't know who is responsible for this transformation, but you might want to check this first.

grosu:
btw, the same happens with the --backend=java

traiansf:
my guess is that, due to the renaming, the function attribute does not get
propagated to the rule.  And since the java backend shares this
transformation (adding function to rules topped in a functional symbol),
there should be no surprise you notice the same behavior there.

2014-03-27 20:36 GMT+02:00 Grigore Rosu notifications@github.com:

PR #315, opened 2014-03-27
Patch and test for Issue #313.

traiansf:


src/javasources/KTool/src/org/kframework/compile/transformers/ResolveFunctions.java
```diff
@@ -33,8 +37,27 @@ public ASTNode transform(Rule node) throws TransformerException {
         if (body instanceof KApp) {
             Term l = ((KApp) body).getLabel();
             if (l instanceof KLabelConstant) {
-                if (((KLabelConstant)l).isPredicate()) {
+                KLabelConstant label = (KLabelConstant) l;
+                if (label.isPredicate()) {
                     node = addFunction(node);
+                } else {
```
dwightguth: it looks like you're trying to mimic having some kind of label.isFunction method without actually having one. Is there a reason you didn't go that route?

traiansf: No particular reason.  Would you like me to add a method isFunctional to KLabelConstant?

Then, what would be the decision in case a KLabel can be both functional and unfunctional -- would we still always report error?  Perhaps that would be fine, too.

dwightguth: Well, think about the semantics. If I have a production that has a particular label of arity N and it's a function, then any time I rewrite something to that klabel it will evaluate the function. Which means you can never actually match the non-functional production that has that label, unless it has a different arity, which is just weird and we don't want to support that. So yeah, I think it makes sense to do that. But we can discuss it on Monday briefly if you want.

traiansf:
I've added an isFunctional method.  
@dwightguth: Could you take another look?

dwightguth:
Looks good just one minor thing: can you use -- instead of - in your config.xml? It will make my life easier merging the JCommander stuff. But otherwise, go ahead and merge once you do that.

traiansf:
retest this please

traiansf:
@dwightguth: i've changed the xml as required, checkstyle passes on my machine, still it fails on Jenkins. Any ideas why?

grosu:
Dwight, you cannot use -- in an XML file, not even in a string.  This sucks, I know, but that's how it is.  We had this discussion before you returned, and I could not believe it.  Then I checked and indeed, no way to write -- in an XML file.  Maybe we should not be that concerned with all the corner cases of our K parsing of bubbles; others suck even worse.

dwightguth:
You can use it in strings, but not comments. But the arguments in favor of
ever commenting out tests are specious at best. We had bugs for a while
because people commented out tests when they broke. Also, it's not the only
language that doesn't allow arbitrary strings in multiline comments. K
doesn't either.

So I really think that -- is the right move here because it cleans up some
ugliness in ktest that makes it harder to maintain. but I'm telling you
this way per our conversation this morning.

dwightguth:
Traian: as I said via IM, you want the first line to be an xml declaration, not an element. Note the "?" in `<?xml ?>`

traiansf:
@dwightguth: of course, you were right all along.  I just thought the ? are part of the regular expression.  silly of me.
As we were just waiting for this to pass, I'm merging this/

PR #317, opened 2014-03-28
Fix regular expression. Adding the "|" operator to Token to allow for

radumereuta:
String declarations.

This comes in response to one of the emails sent over the k-user mailing list asking for a way to declare their own types of strings. I had to ask Michael to modify the BasicParser grammar to allow for "|" inside Token.

Once this is merged I can generalize strings and respond to the user.


dwightguth:
I'm having trouble figuring out precisely what this is doing. Can you add a unit test?

radumereuta:
Dwight, I've added a regression test as you requested. Can you revisit this pull request with a higher priority?
I'm trying to answer some users on the k-user mailing list.

dwightguth:
I said unit test for a reason... This is a perfect candidate for a unit test, one which takes a string of a production and runs it through the basic parser and checks the AST of the result. That way we add only a couple hundred milliseconds to the test suite instead of 10 extra seconds.

radumereuta:
Dwight, please have another look at this.

dwightguth:
lgtm. Who worked on the Basic Parser originally? Who should review this? If that person was you or isn't still around, get Michael to weigh in on this.

radumereuta:
It is Michael who wrote the javacc version of the BasicParser, and he is the one that did this change yesterday when we were working on the newparser.

dwightguth:
Okay. Then merge it.

PR #321, opened 2014-03-28
added option --pattern-matching to use pattern-matching instead of unification to drive Java Rewrite Engine

yilongli:


yilongli:
retest this please

yilongli:
I am getting this weird error message: http://fslwork.cs.illinois.edu:8080/job/k-framework-pull-requests/105/testReport/junit/_samples_java_rewrite_engine_tutorial_2_languages_1_simple_2_typed_2_dynamic_simple-typed-dynamic/k/simple_typed_dynamic_k/ 

And the failing tests seem to change in different runs. I am gonna need some help from someone who's familiar with Jenkins and ktest. Thanks.

radumereuta:
It seems that the parser (JSGLR) fails at some point, but I can't see why.
Considering the input, I think it's located near the step where I'm parsing the configuration.

You are right to check Jenkins and ktest first. This has happened in the future when kompile was run with low memory or in weird situations. But if you manage to replicate it consistently, tell me.

yilongli:
And the definition that failed looks really suspicious:

```
/samples/java_rewrite_engine/tutorial/2_languages/1_simple/2_typed/2_dynamic/simple-typed-dynamic.k.simple-typed-dynamic.k
```

Why is the name of `simple-typed-dynamic.k` duplicated?

src/javasources/KTool/src/org/kframework/backend/java/kil/KItem.java
```diff
@@ -302,81 +302,51 @@ public Term evaluateFunction(SymbolicConstraint constraint, TermContext context)
...
                     assert solutions.size() <= 1 : "function definition is not deterministic";
-                    assert solution.isMatching(leftHandSide) : "Pattern matching expected in concrete execution mode";
                 }
 
-                solution.orientSubstitution(rule.leftHandSide().variableSet());
-
-                Term rightHandSide = rule.rightHandSide();
-
-                if (rule.hasUnboundedVariables()) {
-                    /* rename rule variables in the constraints */
```
andreistefanescu, line 357: Is it a problem with the current implementation if the rule has unbounded variables in the right-hand side?

src/javasources/KTool/src/org/kframework/backend/java/symbolic/GroundRewriter.java
```diff
@@ -0,0 +1,306 @@
...
+import org.kframework.backend.java.kil.Term;
+import org.kframework.backend.java.kil.TermContext;
+import org.kframework.backend.java.kil.Variable;
+import org.kframework.backend.java.strategies.TransitionCompositeStrategy;
+import org.kframework.krun.api.SearchType;
+import org.kframework.utils.general.GlobalSettings;
+
+import com.google.common.base.Stopwatch;
+
+public class GroundRewriter {
```
andreistefanescu, line 28: I think this class has some overlap with SymbolicRewriter. Would is be possible to have an abstract class Rewriter with the common functionality. The part regarding rule selection, strategies, and search should be similar (at least for now).

src/javasources/KTool/src/org/kframework/backend/java/kil/KItem.java
yilongli, line 357: Yes, it is going to be a problem. Actually, the original code is dead and completely wrong. Thanks, Andrei.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/GroundRewriter.java
yilongli, line 28: I was intended to do so in the first place. However, `SymbolicRewriter` is full of boilerplate code which makes it a little difficult to extract the common code. I will add a TODO for now and do a larger refactoring later to separate different functionalities, i.e., test-generation, proving, simulation, into different classes.

yilongli:
I need some help here to pass the tests. Everything works fine on my machine. Anybody more familiar with Jenkins and ktest? Thanks. @owolabileg, @traiansf, @daejunpark

dwightguth:
retest this please

dwightguth:
I am going to try to debug what is happening with these tests by sshing into fslwork.

dwightguth:
So I'm still running this down but it looks like something about your change is introducing a behavior causing kompile to intermittently fail. Did you run the complete ktest suite on your local machine? because if you ran only some definitions it may just not have occurred. It's possible the behavior is specific to the machine ktest is being run on (e.g. # processors, etc) also. I will continue to investigate, but I suggest you at least try to reproduce this locally if you haven't already.

dwightguth:
I'm pretty sure I figured out what is wrong with this code. You are compiling the same definition at the same time with different options in parallel and it's breaking. Probably some kind of race condition on files in the -kompiled directory. I'm having trouble reproducing it, but I suggest you try passing a "--directory" flag to kompile and krun and see if it helps.

yilongli:
Hmm, I see. Thanks a lot! Actually, I am compiling the same definition with the same options. The only difference is the options in krun. Maybe there should be a way to compile the definition once and krun programs with different options?

dwightguth:
Hmm. Well this is kind of a hack, so I suggest you talk with @osa1 to see
if there's a way we can support this more cleanly, but you can probably do
something like:

```
<kompile-option name="--version" />
```

That would make it so that it didn't actually kompile the second time, and
used the same compiled definition for both sets of programs.

yilongli:
On the other hand, I suggest ktest also raise an error when two definitions are kompiled to the same place.

dwightguth:
Yeah, that's probably a better solution. In any event, you should be able
to fix this now, one way or another.

grosu:
The idea was that one can use the option -d if one wants to kompile two definitions at the same place, with different folder names for the two kompiled definitions.  This is particularly useful when you want to kompile the same language definition, but with different backends.  I thought that some of the entries in the global config.xml were already doing that ...

daejunpark:
@yilongli 

I remember that @osa1 already implemented the validation that gives an error when multiple `<test>`s have the same definition with the same `directory` attribute (considering its default value). So, I think your `config_java.xml` file should have raised an error at the beginning of ktest. I guess the validation is somehow broken. I'll take a look at it.

Also, I just realized that `directory` attribute does not work at all. Your problem should have been solved by using the following simple additional attribute in `<test>` node:

```
directory="pattern-matching"
```

But, somehow it does not work. I'll fix it as soon as possible. (Omer, could you help me if you have a time?)

Meanwhile, please use the following workaround. Add two options for each of your additional pattern-matching tests:

```
<kompile-option name="-d" value="pattern-matching" />
<krun-option name="-d" value="pattern-matching" />
```

It makes your pattern-matching tests to be executed at their own sub-directories.

I'm sorry for your inconvenience.

daejunpark:
I think this failure is due to the copyright check. You might already know that, though.

yilongli:
Yes, some of the copyright info disappeared after auto-rebasing. However, I also run into some other issues after merging with the latest update. This is going to take a while to resolve.

dwightguth:
What's the status of this?

yilongli:
I will let you know when it's ready. Or you can close it for now. I am afraid it's going to take a while. I have very limited time to work on code recently.

yilongli:
test this please

yilongli:
retest this please

dwightguth:
Jenkins: test this please

dwightguth:
Yilong: we retweaked the regex for the jenkins script so that you have to explicitly invoke Jenkins by name. The hope is that once we get into the habit of doing this it will confuse people looking at these threads less.

yilongli:
Thanks, Dwight. I was thinking of some kind of policy change.

yilongli:

dwightguth:
Hey so it's not a huge deal in this case, but in the future, can you please use "git merge" instead of "git rebase" when you are trying to resolve conflicts with the latest version of master, in a code review that I have already looked at? rebasing destroys the commits I saw before and recreates them, which causes me to be unable to view the diff incrementally.

samples/java_rewrite_engine/tutorial/2_languages/4_logik/tests/append-4.logik.out
```diff
@@ -2,29 +2,29 @@ Search results:
```
dwightguth, line 2: I wonder, what would happen if you used the regular expression versions of this tests from the k/tutorial directory? Would they pass?

grosu:
Please make sure this is well explained in the FAQ about git.  I think it is, but just in case.  This way the problem will be avoided by others, too.  Thanks.

src/javasources/KTool/src/org/kframework/backend/java/kil/KItem.java
```diff
@@ -309,85 +309,65 @@ public Term evaluateFunction(SymbolicConstraint constraint, TermContext context)
...
-                }
 
-                ConstrainedTerm leftHandSide = new ConstrainedTerm(
-                        ((KItem) rule.leftHandSide()).kList,
-                        rule.lookups().getSymbolicConstraint(context),
-                        leftHandSideConstraint,
-                        context);
-
-                Collection<SymbolicConstraint> solutions = constrainedTerm.unify(leftHandSide);
+                Collection<Map<Variable, Term>> solutions = PatternMatcher.patternMatch(this, rule, context);
```
dwightguth, line 317: If the decision to make function rules pattern match only and never unify was deliberate, can we get some kind of comment explaining the rationale here? If not, what exactly is the correct behavior here?

src/javasources/KTool/src/org/kframework/backend/java/kil/KItem.java
```diff
@@ -403,8 +383,10 @@ public Term evaluateFunction(SymbolicConstraint constraint, TermContext context)
 
             if (result != null) {
                 return result;
-            } else if (mayUseOwiseRule && !owiseResults.isEmpty()) {
-                assert owiseResults.size() == 1 : "function definition is not deterministic";
+            } else if (!owiseResults.isEmpty()) {
```
dwightguth, line 386: If we are supporting owise rules, do we have test cases for them? If so, where? I'd like to take a look.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/GroundRewriter.java
```diff
@@ -0,0 +1,307 @@
...
+import org.kframework.backend.java.kil.Rule;
+import org.kframework.backend.java.kil.Term;
+import org.kframework.backend.java.kil.TermContext;
+import org.kframework.backend.java.kil.Variable;
+import org.kframework.backend.java.strategies.TransitionCompositeStrategy;
+import org.kframework.krun.api.SearchType;
+
+import com.google.common.base.Stopwatch;
+
+// TODO(YilongL): extract common functionalities with SymbolicRewriter to superclass
```
dwightguth, line 27: this needs to be done now, not later. Code copying is an excellent way to introduce staleness, bugs, and out of sync implementations.

andreistefanescu, line 27: We already have another 2 classes that implement similar functionality, so I would suggest to merge this change, and then have a separate commit for refactoring all 3 classes.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/JavaSymbolicKRun.java
```diff
@@ -238,8 +244,20 @@ private ConstrainedTerm javaKILRun(org.kframework.kil.Term cfg, int bound) {
         Rule patternRule = transformer.transformRule(pattern, definition);
 
         List<SearchResult> searchResults = new ArrayList<SearchResult>();
-        List<Map<Variable,Term>> hits = symbolicRewriter.search(
-                initialTerm, targetTerm, claims, patternRule, bound, depth, searchType);
+        List<Map<Variable,Term>> hits;
+        if (K.pattern_matching) {
+            Term initialTerm = Term.of(cfg, definition);
+            Term targetTerm = null;
+            GroundRewriter rewriter = new GroundRewriter(definition, termContext);
```
dwightguth, line 251: if you make a superclass, this should have some kind of getRewriter method or something, and no if/else.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/JavaSymbolicKRun.java
```diff
@@ -301,10 +319,10 @@ private ConstrainedTerm javaKILRun(org.kframework.kil.Term cfg, int bound) {
         stateMap.put((org.kframework.kil.Term) org.kframework.kil.GenericToken.kAppOf("Id", "$1"), (org.kframework.kil.Term) org.kframework.kil.IntBuiltin.kAppOf("1"));
         ((org.kframework.kil.Cell) o)
                 .setContents(new org.kframework.kil.MapBuiltin(context
-                        .dataStructureSortOf("MyMap"),
```
dwightguth, line 304: can you please undo this change? it will cause merge conflicts with the unified builtins branch.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
```diff
@@ -762,6 +762,15 @@ private static Rule evaluateRule(Rule rule, Definition definition) {
...
         // TODO(AndreiS): some evaluation is required in the LHS as well
         //Term leftHandSide = rule.leftHandSide().evaluate(termContext);
+
+        Rule origRule = rule;
+        if (rule.isFunction()) {
+            /*
+             * rename variables in the function rule to avoid variable confusion
+             * when trying to apply this rule on its RHS
+             */
+            rule = rule.getFreshRule(termContext);
```
dwightguth, line 772: is this a variable capture thing? shouldn't the substitution handle this?

src/javasources/KTool/src/org/kframework/backend/java/symbolic/PatternMatcher.java
```diff
@@ -84,87 +104,118 @@ public static boolean matchable(Term subject, Term pattern, TermContext context)
...
-                    if (checkOrderedSortedCondition(variable, evaluatedLookup, context)) {
-                        for (Map<Variable, Term> subst2 : multiSubsts) {
-                            subst2.put(variable, evaluatedLookup);
-                        }
-                    } else {
+                assert lookupOrChoice instanceof DataStructureLookupOrChoice :
+                    "one side of the equality should be an instance of DataStructureLookup or DataStructureChoice";
+
+                List<Map<Variable, Term>> updatedSubsts = new ArrayList<>();
+                label:
```
dwightguth: I don't think this label is needed anymore.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/PatternMatcher.java
```diff
@@ -294,24 +351,26 @@ public void match(Term subject, Term pattern) {
 
             /* add substitution */
             addSubstitution(variable, subject);
+        } else if (subject.isSymbolic()) {
+            fail(subject, pattern);
         } else {
             /* match */
-            if (!subject.equals(pattern)) {
+            if (subject.hashCode() != pattern.hashCode() || !subject.equals(pattern)) {
```
dwightguth, line 357: this feels funny to me. are you doing this in order to make it faster? if so, why not insert this check directly into equals?

src/javasources/KTool/src/org/kframework/backend/java/symbolic/GroundRewriter.java
dwightguth, line 27: That's fine as long as it follows immediately after this change.

On Thu, May 22, 2014 at 1:27 PM, Andrei Stefanescu <notifications@github.com

dwightguth:
@grosu It is in fact explained in the git faq that you are not supposed to rebase changes that other people have already reviewed.

samples/java_rewrite_engine/tutorial/2_languages/4_logik/tests/append-4.logik.out
yilongli, line 2: I didn't notice that we have some new stuff for ktest. It seems pretty nice. Unfortunately, it requires quite some work to revise those regex to make it work for the Java backend.

As you can see below, the biggest problem is that the pretty printed results from the Java backend have sort information `Term` after the variable and much less parentheses. 

```
ERROR: ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/takeout-1.logik' '--search' '--pattern' '<solution> Solution:K </solution>'] output doesn't match with expected output (time: 3633 ms)
Unexpected program output:
Expected result to match regular expression:
\Q"L" |-> cons ( a (  ), cons ( b (  ), nil (  ) ) )\E

but found:
Search results:

Solution 1:
Solution:K -->
L:Term |-> cons ( a, cons (  b, nil  ) )

Solution 2:
Solution:K -->
L:Term |-> cons ( b, cons (  a, nil  ) )


ERROR: ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/reverse-fast-2.logik' '--search' '--bound' '1' '--pattern' '<solution> Solution:K </solution>'] output doesn't match with expected output (time: 3988 ms)
Unexpected program output:
Expected result to match regular expression:
\Q"L" |-> cons ( c (  ), cons ( b (  ), cons ( a (  ), nil (  ) ) ) )\E

but found:
Search results:

Solution 1:
Solution:K -->
L:Term |-> cons ( c, cons ( b, cons (  a, nil  ) ) )


Done with ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/fact-2.logik' '--search' '--pattern' '<solution> Solution:K </solution>'] (time 3557 ms)
ERROR: ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/reverse-slow-3.logik' '--search' '--pattern' '<solution> Solution:K </solution>'] output doesn't match with expected output (time: 3992 ms)
Unexpected program output:
Expected result to match regular expression:
\Q"L" |-> cons ( c (  ), cons ( X, cons ( a (  ), nil (  ) ) ) )\E\n\Q"T" |-> cons ( a (  ), cons ( X, cons ( c (  ), nil (  ) ) ) )\E\n\Q"X" |-> X\E

but found:
Search results:

Solution 1:
Solution:K -->
L:Term |-> cons ( c, cons ( X:Term, cons (  a, nil  ) ) )
T:Term |-> cons ( a, cons ( X, cons (  c, nil  ) ) )
X:Term |-> X


Done with ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/fact-4.logik' '--search' '--pattern' '<solution> Solution:K </solution>'] (time 3119 ms)
ERROR: ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/reverse-fast-1.logik' '--search' '--pattern' '<solution> Solution:K </solution>'] output doesn't match with expected output (time: 3443 ms)
Unexpected program output:
Expected result to match regular expression:
\Q"L" |-> cons ( c (  ), cons ( b (  ), cons ( a (  ), nil (  ) ) ) )\E

but found:
Search results:

Solution 1:
Solution:K -->
L:Term |-> cons ( c, cons ( b, cons (  a, nil  ) ) )


ERROR: ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/family-2.logik' '--search' '--pattern' '<solution> Solution:K </solution>'] output doesn't match with expected output (time: 4814 ms)
Unexpected program output:
Expected result to match regular expression:
"X".*david

but found:
Search results:

Solution 1:
Solution:K -->
X:Term |-> david

Solution 2:
Solution:K -->
X:Term |-> jim

Solution 3:
Solution:K -->
X:Term |-> steve


ERROR: ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/reverse-slow-palindromes-1.logik' '--search' '--bound' '5' '--pattern' '<solution> Solution:K </solution>'] output doesn't match with expected output (time: 5084 ms)
Unexpected program output:
Expected result to match regular expression:
"L" |-> cons \( (.*?), cons \( \1, nil \(  \) \) \)

but found:
Search results:

Solution 1:
Solution:K -->
L:Term |-> cons ( _113:Term, cons ( _113, nil ) )

Solution 2:
Solution:K -->
L:Term |-> cons ( _113:Term, cons ( _146:Term, cons ( _113, nil ) ) )

Solution 3:
Solution:K -->
L:Term |-> cons ( _113:Term, cons ( _146:Term, cons ( _146, cons ( _113, nil )
   ) ) )

Solution 4:
Solution:K -->
L:Term |-> cons ( _113:Term, nil )

Solution 5:
Solution:K -->
L:Term |-> nil


ERROR: ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/append-1.logik' '--search' '--pattern' '<solution> Solution:K </solution>'] output doesn't match with expected output (time: 3349 ms)
Unexpected program output:
Expected result to match regular expression:
\Q"V" |-> cons ( a (  ), cons ( b (  ), nil (  ) ) )\E

but found:
Search results:

Solution 1:
Solution:K -->
V:Term |-> cons ( a, cons (  b, nil  ) )


ERROR: ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/reverse-fast-palindromes-1.logik' '--search' '--bound' '5' '--pattern' '<solution> Solution:K </solution>'] output doesn't match with expected output (time: 3767 ms)
Unexpected program output:
Expected result to match regular expression:
"L" |-> cons \( (.*?), cons \( \1, nil \(  \) \) \)

but found:
Search results:

Solution 1:
Solution:K -->
L:Term |-> cons ( _123:Term, cons ( _123, nil ) )

Solution 2:
Solution:K -->
L:Term |-> cons ( _123:Term, cons ( _157:Term, cons ( _123, nil ) ) )

Solution 3:
Solution:K -->
L:Term |-> cons ( _123:Term, cons ( _157:Term, cons ( _157, cons ( _123, nil )
   ) ) )

Solution 4:
Solution:K -->
L:Term |-> cons ( _123:Term, nil )

Solution 5:
Solution:K -->
L:Term |-> nil


ERROR: ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/reverse-fast-3.logik' '--search' '--pattern' '<solution> Solution:K </solution>'] output doesn't match with expected output (time: 3477 ms)
Unexpected program output:
Expected result to match regular expression:
\Q"L" |-> cons ( c (  ), cons ( X, cons ( a (  ), nil (  ) ) ) )\E\n\Q"T" |-> cons ( a (  ), cons ( X, cons ( c (  ), nil (  ) ) ) )\E\n\Q"X" |-> X\E

but found:
Search results:

Solution 1:
Solution:K -->
L:Term |-> cons ( c, cons ( X:Term, cons (  a, nil  ) ) )
T:Term |-> cons ( a, cons ( X, cons (  c, nil  ) ) )
X:Term |-> X


ERROR: ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/takeout-2.logik' '--search' '--pattern' '<solution> Solution:K </solution>'] output doesn't match with expected output (time: 3443 ms)
Unexpected program output:
Expected result to match regular expression:
\Q"L" |-> cons ( a (  ), cons ( b (  ), cons ( c (  ), cons ( d (  ), nil (  ) ) ) ) )\E

but found:
Search results:

Solution 1:
Solution:K -->
L:Term |-> cons ( a, cons ( b, cons ( c, cons (  d, nil  ) ) ) )

Solution 2:
Solution:K -->
L:Term |-> cons ( b, cons ( a, cons ( c, cons (  d, nil  ) ) ) )

Solution 3:
Solution:K -->
L:Term |-> cons ( b, cons ( c, cons ( a, cons (  d, nil  ) ) ) )

Solution 4:
Solution:K -->
L:Term |-> cons ( b, cons ( c, cons ( d, cons (  a, nil  ) ) ) )


ERROR: ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/reverse-slow-1.logik' '--search' '--pattern' '<solution> Solution:K </solution>'] output doesn't match with expected output (time: 3720 ms)
Unexpected program output:
Expected result to match regular expression:
\Q"L" |-> cons ( c (  ), cons ( b (  ), cons ( a (  ), nil (  ) ) ) )\E

but found:
Search results:

Solution 1:
Solution:K -->
L:Term |-> cons ( c, cons ( b, cons (  a, nil  ) ) )


Done with ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/fact-1.logik' '--search' '--pattern' '<solution> Solution:K </solution>'] (time 3066 ms)
ERROR: ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/append-2.logik' '--search' '--pattern' '<solution> Solution:K </solution>'] output doesn't match with expected output (time: 3320 ms)
Unexpected program output:
Expected result to match regular expression:
\Q"L1" |-> cons ( a (  ), cons ( b (  ), nil (  ) ) )\E\n\Q"L2" |-> nil (  )\E

but found:
Search results:

Solution 1:
Solution:K -->
L1:Term |-> cons ( a, cons (  b, nil  ) )
L2:Term |-> nil

Solution 2:
Solution:K -->
L1:Term |-> nil
L2:Term |-> cons ( b, cons (  a, nil  ) )


ERROR: ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/append-4.logik' '--search' '--bound' '5' '--pattern' '<solution> Solution:K </solution>'] output doesn't match with expected output (time: 3607 ms)
Unexpected program output:
Expected result to match regular expression:
"L1" |-> cons \( (.*?), cons \( ((?!\1).*?), cons \( ((?!\1|\2).*?), cons \( ((?!\1|\2|\3).*?), nil \(  \) \) \) \) \)\n"L2" \|-> L2\n"L3" \|-> cons \( \1, cons \( \2, cons \( \3, cons \( \4, cons \( a \(  \), L2 \) \) \) \) \)

but found:
Search results:

Solution 1:
Solution:K -->
L1:Term |-> cons ( _99:Term, cons ( _133:Term, cons ( _170:Term, cons ( _210
  :Term, nil ) ) ) )
L2:Term |-> L2:Term
L3:Term |-> cons ( _99, cons ( _133, cons ( _170, cons ( _210, cons ( a, L2 ) 
  ) ) ) )

Solution 2:
Solution:K -->
L1:Term |-> cons ( _99:Term, cons ( _133:Term, cons ( _170:Term, nil ) ) )
L2:Term |-> L2:Term
L3:Term |-> cons ( _99, cons ( _133, cons ( _170, cons ( a, L2 ) ) ) )

Solution 3:
Solution:K -->
L1:Term |-> cons ( _99:Term, cons ( _133:Term, nil ) )
L2:Term |-> L2:Term
L3:Term |-> cons ( _99, cons ( _133, cons ( a, L2 ) ) )

Solution 4:
Solution:K -->
L1:Term |-> cons ( _99:Term, nil )
L2:Term |-> L2:Term
L3:Term |-> cons ( _99, cons ( a, L2 ) )

Solution 5:
Solution:K -->
L1:Term |-> nil
L2:Term |-> L2:Term
L3:Term |-> cons ( a, L2 )


ERROR: ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/reverse-fast-palindromes-2.logik' '--search' '--pattern' '<solution> Solution:K </solution>'] output doesn't match with expected output (time: 3269 ms)
Unexpected program output:
Expected result to match regular expression:
"A" \|-> A\n"B" \|-> B\n"C" \|-> A

but found:
Search results:

Solution 1:
Solution:K -->
A:Term |-> A:Term
B:Term |-> B:Term
C:Term |-> A


ERROR: ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/append-3.logik' '--search' '--pattern' '<solution> Solution:K </solution>'] output doesn't match with expected output (time: 3475 ms)
Unexpected program output:
Expected result to match regular expression:
"Y" \|-> (.*?)\n"Z" |-> cons \( a \(  \), \1 \)

but found:
Search results:

Solution 1:
Solution:K -->
Y:Term |-> Y:Term
Z:Term |-> cons ( a, Y )


Done with ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/fact-3.logik' '--search' '--pattern' '<solution> Solution:K </solution>'] (time 3174 ms)
ERROR: ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/family-1.logik' '--search' '--pattern' '<solution> Solution:K </solution>'] output doesn't match with expected output (time: 4740 ms)
Unexpected program output:
Expected result to match regular expression:
"A".*jim.*\n"B".*john.*\n"X".*david.*

but found:
Search results:

Solution 1:
Solution:K -->
A:Term |-> jim
B:Term |-> john
X:Term |-> david

Solution 2:
Solution:K -->
A:Term |-> nathan
B:Term |-> david
X:Term |-> jim

Solution 3:
Solution:K -->
A:Term |-> nathan
B:Term |-> david
X:Term |-> steve

Solution 4:
Solution:K -->
A:Term |-> nathan
B:Term |-> jim
X:Term |-> steve

Solution 5:
Solution:K -->
A:Term |-> nathan
B:Term |-> john
X:Term |-> david

Solution 6:
Solution:K -->
A:Term |-> nathan
B:Term |-> john
X:Term |-> jim

Solution 7:
Solution:K -->
A:Term |-> nathan
B:Term |-> john
X:Term |-> steve

Solution 8:
Solution:K -->
A:Term |-> steve
B:Term |-> david
X:Term |-> jim

Solution 9:
Solution:K -->
A:Term |-> steve
B:Term |-> john
X:Term |-> david

Solution 10:
Solution:K -->
A:Term |-> steve
B:Term |-> john
X:Term |-> jim


ERROR: ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/reverse-slow-palindromes-2.logik' '--search' '--pattern' '<solution> Solution:K </solution>'] output doesn't match with expected output (time: 3698 ms)
Unexpected program output:
Expected result to match regular expression:
"A" \|-> A\n"B" \|-> B\n"C" \|-> A

but found:
Search results:

Solution 1:
Solution:K -->
A:Term |-> C:Term
B:Term |-> B:Term
C:Term |-> C


ERROR: ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/reverse-slow-2.logik' '--search' '--bound' '1' '--pattern' '<solution> Solution:K </solution>'] output doesn't match with expected output (time: 4376 ms)
Unexpected program output:
Expected result to match regular expression:
\Q"L" |-> cons ( c (  ), cons ( b (  ), cons ( a (  ), nil (  ) ) ) )\E

but found:
Search results:

Solution 1:
Solution:K -->
L:Term |-> cons ( c, cons ( b, cons (  a, nil  ) ) )


ERROR: ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/fact-5.logik' '--search' '--pattern' '<solution> Solution:K </solution>'] output doesn't match with expected output (time: 3244 ms)
Unexpected program output:
Expected result to match regular expression:
\Q"X" |-> john (  )\E

but found:
Search results:

Solution 1:
Solution:K -->
X:Term |-> john


Done with ['/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/bin/krun' '/home/yilongli/Eclipse-Workspaces/sbtg-workspace/k/tutorial/2_languages/4_logik/basic/tests/append-5.logik' '--search' '--pattern' '<solution> Solution:K </solution>'] (time 2676 ms)
FAIL

FAIL (see details above)
----------------------------
Definitions kompiled: 1 (time: 0'8'' elapsed, 0'8'' CPU)
PDF posters kompiled: 0 (time: 0'0'' elapsed, 0'0'' CPU)
Programs krun: 24 (time: 0'44'' elapsed, 1'28'' CPU)
Total time: 0'52'' elapsed, 1'36'' CPU
----------------------------

```

src/javasources/KTool/src/org/kframework/backend/java/kil/KItem.java
yilongli, line 317: Yes, function rules should be applied by pattern match rather than unification, except in test generation. I have added a comment and will revise it to include the case of test generation when I start to work on that.

samples/java_rewrite_engine/tutorial/2_languages/4_logik/tests/append-4.logik.out
dwightguth, line 2: Alright. Leave it for Grigore to ponder when he gets to LOGIK then.

src/javasources/KTool/src/org/kframework/backend/java/kil/KItem.java
yilongli, line 386: I added support for `owise` rules when I was working on `pattern-matching` module for the Java backend. https://github.com/kframework/k/blob/master/tests/regression/java-rewrite-engine/pattern_matching/test.k

src/javasources/KTool/src/org/kframework/backend/java/symbolic/GroundRewriter.java
yilongli, line 27: I will try to refactor this in the next pull request. While I did copy code from `SymbolicRewriter` as a starter, I modified a lot of code. So I guess I wouldn't be able to extract a lot common code except for the interface.

@andreistefanescu what are the other class except `SymbolicRewriter`?

src/javasources/KTool/src/org/kframework/backend/java/symbolic/JavaSymbolicKRun.java
yilongli, line 251: Yep. Will do.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/JavaSymbolicKRun.java
yilongli, line 304: Are you referring to all the changes that simply remove trailing white-spaces?

src/javasources/KTool/src/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
yilongli, line 772: Nope, it's not. The method here tries to evaluate ground function terms on the RHS of a rule. So when that rule happens to be a function rule, you may run into the case that you are trying to match the RHS (subject) of the rule against its LHS (pattern). And they contain common variables, which is confusing.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/JavaSymbolicKRun.java
dwightguth, line 304: I mean that you have just told git that this line has been modified, but
it's a line that unified builtins would have also modified. This creates a
merge conflict. I'd rather not have to deal with it.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/PatternMatcher.java
yilongli: Yes, it is. I am still using it. For example, here:https://github.com/kframework/k/pull/321/files#diff-4efef68821310d071fda6e8800165042R167

src/javasources/KTool/src/org/kframework/backend/java/symbolic/JavaSymbolicKRun.java
yilongli, line 304: I will revert all the trivial changes on this file.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/PatternMatcher.java
yilongli, line 357: In fact, I am doing it for fun and experiment. The performance seems to be slightly better so I just leave it alone. I think it's hard to predict the effect of this check. If I move it into `equals`, then I will have to traverse the entire term the first time I compare it with other term. I tend to believe that it would slow down the system. Besides, that is a lot of changes to make.

On the other hand, consider the situation here: if `equals` returns true I lose the time spent on `hashCode`; if `equals` should return false, then I probably have skipped calling it and later calls to `hashCode` on any subterms can be done in constant time.

Anyway, I tend to keep the check here.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/PatternMatcher.java
dwightguth: Yes, I see the continue statement, but is the label needed? Unless I'm misreading the code, it's no longer continuing out of more than a single level of for statement, so you shouldn't need the label.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
dwightguth, line 772: Okay, but why doesn't that just work? If I have something like:

```
syntax Int ::= foo(Int) [function]
rule foo(I:Int) => foo(I -Int 1) when I >Int 0
```

it should still work to try to match I -Int 1 against I:Int, because of how the matching process should treat variables. Right?

yilongli, line 772: When you are doing pattern matching, yes. At first, function evaluation was driven by unification as well, which makes it a latent bug. Now that I have switched to use pattern matching, this _might_ not be a problem. But I would like to be safe.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/PatternMatcher.java
yilongli: Thank you. Nice catch.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
dwightguth, line 772: Okay so the issue is that in unification, if you try to unify I -Int 1 against I, it treats them as the same variable.

How is this handled in the non-function case, then? I should not be able to create a variable during execution that is treated as identical to one of the variables in the LHS of a rule, should I? Or am I still missing something?

yilongli, line 772: Yes.

Yes.

It is a corner case that only happens in partial evaluation of functions on the RHS of a rule and, maybe, macro expansion. Normally, the pattern and subject should have no identical variables because the fresh variable counter used to generate new variable is incrementing. That is to say, variables in rules are created before variables in terms to rewrite; and they always have smaller id number.

dwightguth, line 772: Okay. This sounds like a larger topic, so I will drop it for now, but I wonder maybe if we should treat rule variables and variables in the configuration term as belonging to separate namespaces. Is there any situation where you /do/ want unification to treat a variable in a rule as the same variable as another variable?

dwightguth:
I am satisfied with this assuming the tests pass.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
yilongli, line 772: I know that some theorem prover use variable banks to separate variables to different namespaces. But I think this separation would be an over-kill for now.

Not that I can think of right now.

PR #332, opened 2014-04-01
minor fixes to the machine integers

andreistefanescu:


yilongli:
lgtm.

dwightguth:
My only concern is that you are representing machine integers textually in a way which is not consistent with the K syntax that instantiates them. I thought that was what the value() method was supposed to represent?

andreistefanescu:
Technically speaking, the machine integers have no syntactical
representation in K. There is a function which creates them, mi(<bitwidth>,
<value>), but that's it. As it were, the output is hard to read.

dwightguth:
Hmm technically I think what is needed here is a way to specify the user-defined syntax of pre-existing builtins. So for example you provide some kind of bijection between domain values and textual representations, which the parser and pretty printer are aware of. Then you would indicate that for example the 64 bit integer 1 was represented by "1L" in Java.

Something to keep in mind for future design planning, I think, but not an immediate concern. I'm okay with delaying this task until it's higher priority, as long as you aren't using the value() method anywhere in the code base right now in a way that expects its output to be parseable (aside from the pretty printing concern). Maybe we should make a note of this somewhere though?

PR #333, opened 2014-04-01
clean up some ugliness in UnparserFilterNew

dwightguth:
@liyili2 please review.

### DIFF
--- a/src/javasources/KTool/.classpath
+++ b/src/javasources/KTool/.classpath
@@ -2,7 +2,7 @@
 <classpath>
 	<classpathentry kind="src" path="src"/>
 	<classpathentry kind="src" path="test"/>
-	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.6"/>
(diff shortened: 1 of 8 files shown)


liyili2:
What is the problem of UnparserFilterNew??

dwightguth:
So the reason I made this is because the findRightSyntax method introduces a dependency on the entire definition at unparsing time even though the only thing it truly needs is information about the bracket productions and subsorts in the definition. I am trying to break the dependency on the entire definition in krun to speed up initialization. In order to do this it was necessary to move the information you were looking for to the Context object where it belongs.

liyili2:
Does the krun parser use the UnparserFilterNew? I do not think so.

dwightguth:
I don't understand what you're asking here. Neither this pull request nor the bug report I created have anything to do with the krun parser. The point of this pull request is to break the dependency of the UnparserFilterNew on the serialized Definition object, and to bring the functionality it provides in line with the preexisting design pattern for storing metadata about a compiled definition. This pull request is a refactoring only: it should not affect functionality at all. The fact that the current functionality is broken is a separate issue, which is why I assigned it to you instead of spending days trying to fix it myself, when it's not my priority.

liyili2:
Sorry, I put the things in wrong place. What should I do for the pull request after review?

dwightguth:
If you feel like you understand what the change is doing, approve it and I
will merge it.

PR #336, opened 2014-04-02
enforce copyright notice on all files modified in any commit

dwightguth:
Once this is merged I will add the Jenkins integration functionality. This will break any pull request testing which hasn't merged this change yet, so I suggest you do so immediately following this merge.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 .metadata
 .cache
 .idea
+.checkstyle-files
 *~
 junit-reports
 samples/test/
--- a/build.xml
+++ b/build.xml
@@ -208,7 +208,9 @@
 
     <target name="checkstyle-copyright">
         <checkstyle config="${javasources}/KTool/checkstyle-copyright.xml">
-            <fileset dir="." includesfile=".checkstyle-files" />
+            <fileset dir="." includesfile=".checkstyle-files">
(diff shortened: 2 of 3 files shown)


dwightguth:
@radumereuta you reviewed the last checkstyle change...

dwightguth:
I added some comments regarding this file, please re-review. As far as 2014 goes, I did this on purpose. The idea is that we as a development team commit a change to this file every January. This forces files modified in the new year which already previously had copyright notices to be updated as part of each new commit, ensuring that the copyright notice remains up to date.

radumereuta:
Another question. What happens if someone wants to add a file with a different license?
For example someone needs to get a small library or algorithm from somewhere in source form.
We can't put our license there.

dwightguth:
Then you add an exclude to build.xml. I have added another comment to
explain this.

radumereuta:
I can live with that.
I think it's ok to merge then.

adamsmd:
In the LICENSE file, we use "(c)" instead of "(C)".  That is not just a 
local decision, but is based on the official NCSA license 
http://opensource.org/licenses/NCSA.

It is a minor thing, but I think the checkstyle should use "(c)" instead 
of "(C)" for consistency with that.  (Most licenses use lower case anyways.)

Michael D. Adams

PR #337, opened 2014-04-02
avoid creation of Z3 objects when constraint is a substitution (which is...

andreistefanescu:
... trivially satisfiable)


yilongli:
lgtm.

PR #338, opened 2014-04-07
LTL and symbolic backend: fixing the 'ltl' attribute issue for rules.

andreiarusoaie:
Implemented a feature which allows the LTL model-checker to work together with the symbolic backend. 
@dwightguth Please review.
@dlucanu Please review.


src/javasources/KTool/src/org/kframework/backend/symbolic/ResolveLtlAttributes.java
```diff
@@ -32,68 +57,100 @@ public ASTNode transform(Rule node) throws TransformerException {
...
+                        if (rewriteLHSLabel.getLabel().equals(LTL_SAT)) {
+                            // Prepare to extract the components of the '|=Ltl KApp body which
+                            // should ALWAYS have a variable representing the program state in the
+                            // left hand side and a predicate in the right hand side:
+                            //     B |=Ltl predicate
+                            KList rewriteLHSContents = (KList) rewriteLHSBody.getChild();
+
+                            Term expectedVariable = rewriteLHSContents.getContents().get(0);
+                            // extract the actual Variable instance from the AST
+                            expectedVariable = ((KApp) expectedVariable).getLabel();
```
dwightguth: this will throw a ClassCastException if the user has an error in their rule. You need to explicitly check this case and handle it with a friendly error message.

src/javasources/KTool/src/org/kframework/backend/symbolic/ResolveLtlAttributes.java
```diff
@@ -32,68 +57,100 @@ public ASTNode transform(Rule node) throws TransformerException {
...
+                            expectedVariable = ((KInjectedLabel) expectedVariable).getTerm();
```
dwightguth: same here.

src/javasources/KTool/src/org/kframework/backend/symbolic/ResolveLtlAttributes.java
```diff
@@ -32,68 +57,100 @@ public ASTNode transform(Rule node) throws TransformerException {
...
+                                        KException.KExceptionGroup.CRITICAL,
+                                        "the 'ltl' attribute can be used only for rules " +
+                                                "having only a variable in the left hand side of "
+                                                + LTL_SAT,
+                                        node.getFilename(),
+                                        node.getLocation()));
+                            } else {
+                                // process the variable: wrap it with generatedTop and add the path condition
+                                Variable phi = Variable.getFreshVar("K");
+                                Term ltlState = WrapVariableWithTopCell.wrapPCAndVarWithGeneratedTop(phi, expectedVariable);
```
dwightguth: if we are processing the LHS of the |=, the RHS of the |=, the RHS of the rewrite, the requires, and the ensures, why not just process the entire rule? Then you don't need to do all this custom stuff, you just need to wrap the LHS of the |= in the entire rule, and then call the pre-existing code to tweak the side condition.

andreiarusoaie:
@dwightguth @dlucanu: I addressed your comments, please re-review.

src/javasources/KTool/src/org/kframework/backend/symbolic/ResolveLtlAttributes.java
```diff
@@ -1,107 +1,138 @@
...
+                    Term smtInvalidCondition = (Term) requires.accept(ct);
+                    List<Term> filtered = ct.getFilteredTerms();
+                    filtered.add(BoolBuiltin.TRUE);
+                    Term smtValidCondition = AddPathCondition.andBool(filtered);
+                    Term predicates = AddPathCondition.andBool(ct.getGeneratedPredicates());
+
+                    // step 2: check the implication using the SMT solver using 'checkSat;
+                    // Note that we use smtValid instead of the rule condition because
+                    // predicates are not important when checking the implication.
+                    Term conditionNegation = KApp.of(KLabelConstant.NOTBOOL_KLABEL, smtValidCondition);
```
dwightguth: I still am not following why it's necessary to do all of this functionality constructing checkSat and so forth here. It seems like the task of taking a side condition and replacing it with the code to check that side condition's satisfiability should be both extractable and already implemented elsewhere in this backend...

dwightguth:
I think I would also like to see a JUnit unit test testing the functionality of this transformer against its inputs and outputs.

src/javasources/KTool/src/org/kframework/backend/symbolic/ResolveLtlAttributes.java
andreiarusoaie: It seemed to me too that this particular piece of code should be already present in the symbolic backend, but in fact it's not. The symbolic backend filters the rule condition and appends a call to checkSat, which verifies whether the (path condition /\ filtered rule condition) is NOT unsat. This is equivalent with detecting if the next possible program state is feasible.
On the other hand, here we construct  a different argument for checkSat which checks a logical implication (we want the argument (formula) sent to checkSat TO BE unsat) and we also add some predicates in the rule's side condition. Technically speaking, in two different places we build two different side conditions for rules. The common part is the replacement of the rule's condition, which I think it's easier to be done using rule.setRequires(newCondition) than creating a new Transformer for this.
If you think the above explanation is not ok, then we can go into deeper the details of this.

dwightguth: I'm not sure I'm following this. Can you show me the before and after of
both types of side conditions for me to compare side-by-side?

andreiarusoaie: Consider that the path condition is n > 0 and the rule condition is val(X, B) > 0 /\ n =/= 0 /\ isInt(n).
In the symbolic backend I have to build the following condition:
isInt(n) /\ val(X, B) > 0 /\ checkSat(n > 0 /\ n=/=0) =/= "unsat"

For LTL it should look something like this:
isInt(n) /\ isBool(val(X, B) > 0) /\ checkSat(n > 0 /\ notBool(n=/=0)) == "unsat"

andreiarusoaie:
Sure, I will add a junit test. I think that @dlucanu should also add some regression tests in K, because those would test the entire functionality of the LTL model-checker.

src/javasources/KTool/src/org/kframework/backend/symbolic/ResolveLtlAttributes.java
dwightguth: So in the first case you are checking that n > 0 /\ n =/= 0 is satisfiable,
and in the second case you are checking that n > 0 -> n =/= 0 is
semantically valid. I still would like to see two methods, one which takes
a term and returns a term that checks that term for satisfiability, and the
other which takes a term and returns a term that checks that term for
validity. Then it should be just a matter of calling the correct term on
the correct subterm of the boolean expression, and adding in the rest after
you perform the appropriate transformations.

The second part of this is confusing me though. Why are you doing
isBool(val(X, B) > 0) but not isBool(isInt(n)) ?

andreiarusoaie: Ok, fair point. One method already exists, I'll just have to create the other one. 

Oh, I'm afraid I have confused you with isBool(val(X, B) > 0), which is not what I generate. Sorry for that, it's a little late for me...
The actual thing I'm generating in that case is just:
isInt(n) /\ checkSat(n > 0 /\ notBool(n=/=0)) == "unsat"

However, if you have a function which has the smtlib attribute when defined in K then that function can be sent to the solver. Let us consider that function being the greatest common divisor - gcd. Then from the path condition n > 0 and the rule condition gcd(A, B) > 0 /\ n =/= 0 /\ isInt(n) I have to generate:

isInt(n) /\ isInt(A) /\ isInt(B) /\ checkSat(n > 0 /\ notBool(gcd(A, B) >0 /\ n=/=0)) == "unsat"

dwightguth: Okay, but I'm still confused. Now you don't have the val part of the side
condition at all? Why not?

andreiarusoaie: Err.. I meant:
isInt(n) /\ val(X, B) > 0 /\ checkSat(n > 0 /\ notBool(n=/=0)) == "unsat" 

val(X, B) > 0 stays in there. I wanted to delete only the predicate. The idea is that those functions that are not translatable to SMT remain untouched, while those which can be translated to the SMT go inside checkSat and I also generate predicates for them.

dwightguth: Okay, yeah, that's what I thought. In that case it should be trivially
translatable into a method to check validity very similar to the one you
already have. Go ahead and make that change if you haven't already.

andreiarusoaie:
@dwightguth  I have pushed something, but it's not done yet. I'll announce you when this is ready for review.

andreiarusoaie:
@dwightguth I have added a junit test. Please take a look.

src/javasources/KTool/test/org/kframework/backend/symbolic/LTLModelCheckerTest.java
```diff
@@ -0,0 +1,88 @@
...
+        attrs.set("anywhere", null);
+
+        // rule
+        Rule rule = new Rule(lhs, rhs, context);
+        rule.setRequires(requires);
+        rule.setEnsures(null);
+        rule.setAttributes(attrs);
+
+        // test input rule
+        Assert.assertEquals(rule.toString().trim(), inputRule.trim());
```
dwightguth: you should not be comparing strings, but actually using Term.equals.

andreiarusoaie: I was hoping not to go through the nightmare of writing the transformed rule from scratch, since I don't really have the time to do it. 
@dlucanu when do you need this merged? If you need it very soon then you can use this branch directly. The cink-semantics test suite successfully passes all the tests.

dwightguth: Why don't you view it as an opportunity to improve the constructors in the
org.kframework.kil package? Ideally you should be able to build terms
without having to use getters and setters.

dwightguth:
add to whitelist

radumereuta:
@dwightguth, you should find a way to differentiate jenkins commands from other comments.
I think 'add to whitelist' is something for jenkins. Do something like "Jenkins: add to whitelist" so  it's easier to spot which are which. But make it mandatory, so we don't forget. Also, do we have a wiki which explains all the commands?

dwightguth:
Okay, this is easy enough to do. I will write a little bit more up on the wiki also.

andreiarusoaie:
@dwightguth is this ready for merge?

PR #340, opened 2014-04-08
Newparser 2

radumereuta:
Stuff that we changed over the last week based on Dwight's comments on the original pull request.


dwightguth:
Looks like you merged wrong. I think I remember doing something very similar once, and it was because I rebased against the wrong branch. If you can't figure out how to fix this, please IM me.

include/builtins/string.k
```diff
@@ -5,6 +6,9 @@ module STRING-SYNTAX-HOOKS
   imports #STRING-INTERFACE
   syntax #String
   syntax String ::= #String
+  // TODO(Radu): replace the above with bellow
+  //   syntax String ::= Token{"\"" ((~[\"\n\r\\])|([\\]~[\n\r]))* "'"}
```
dwightguth, line 10: We will need to have a discussion about the exact regular expression we want for strings eventually, but that can wait.

radumereuta, line 10: Actually, we shouldn't wait that long. We have some problems with escaped characters in the current implementation, so we should try and discuss this after we finish with this pull request.

src/javasources/KTool/src/org/kframework/parser/concrete2/Parser.java
```diff
@@ -20,489 +23,471 @@
...
-                case  '\u000B' :
-                case  '\u000C' :
-                case  '\u0085' :
-                case  '\u2028' :
-                case  '\u2029' :
-                    l++; c = 1; break;
-                default :
-                    c++;
+                    case '\n'      :
+                    case  '\u000B' :
```
dwightguth, line 381: Hmm should we maybe use a unicode character property for this instead of enumerating every single matching character?

dwightguth:
I still need to go through this and match up changes with comments I made on the original review, but the main thing I notice here is the javadocs. There are some places where there are still fields in the Parser classes that I don't really understand what they're doing. Maybe that Javadoc of the algorithm as a whole would help here, I don't know. But it needs something. When you've done more of that type of work, please let me know, Michael, so I can try to do a more detailed vetting of the javadocs.

Also, please let me know when both of you are done with the changes that we discussed still making as part of this code review.

src/javasources/KTool/src/org/kframework/parser/concrete2/Parser.java
radumereuta, line 381: We took the codes exactly like in the links mentioned above that code block. Also, the fact that CRLF has two characters, complicates the algorithm, because you need to look ahead.

dwightguth, line 381: Hmm. Alright, it's probably fine.

adamsmd:
Okay, I think the main two things are the TreeCleaner stuff by Radu and 
(like you said) javadoc.

adamsmd:
Radu and I just finished with a fixing these things up.  We think we have it to a state where you can take a look at it and merge it if there is nothing major.  (Of course, don't forget to merge newparser-treecleaner (#358) and newparser-doc1 (#361) before doing that.)

PR #347, opened 2014-04-08
Updated agent examples to be run by ktest.

traiansf:
Had to remove output for p22 and p23 as they were non-deterministic.


dwightguth:
If we have nondeterministic programs, is it possible to write tests that use --search?

traiansf:
tehnically, yes, I think the search results are sorted.  Yet I was reluctant to use this feature until now.  Do you think I should try it?

dwightguth:
I think having tests which cover using IO with search would be good; I
don't think our test suite covers that right now. So yes, go ahead and try
it.

traiansf:
@dwightguth I've added the tests with search and everything seems fine.  Anything else?

dwightguth:
lgtm. I will bear responsibility for understanding this change, because it deals with the examples, so I am merging this.

PR #349, opened 2014-04-09
fixed hashCode for SymbolicConstraint; refactored hashCode caching

andreistefanescu:


dwightguth:
Other than the copyright thing, there are a couple notes I want to make here:

You are caching the hashCode value. Since two objects that are equal according to equals MUST have the same hash code, that means you need to invalidate your cache whenever the object's internal state changes in any way, because for all you know it was changed to be identical to another object with a different hashCode. It's possible all these objects are completely immutable (I haven't looked too closely at them), in which case you don't need to worry, but if not, that's going to be a serious problem for this code. Maybe you can work it out, maybe not, but it needs to at least be addressed somehow in this commit.

Second, if you want to reserve a specific hashCode value to mean something special, you need to make sure that hashCode never returns that value. Right now if you compute the hash of an object and it turns out to overflow back to 0 then you will be recomputing the hash of that object every time you call this method. You should make sure that if you compute the hash of an object and it returns 0 that the hashCode method actually sets the hashCode to another value, e.g. 1.

yilongli:
`SymbolicConstraint#Equality` is mutable. And it can be put into hash-based collections. So I don't think we should override its `hashCode` method. I once did that and ran into many problems. That's why the comment is there.

yilongli:
@dwightguth Most of the KIL classes are made immutable when I did the refactoring to cache their hashCode. But I am afraid not all subclasses of `JavaSymbolicObject` are immutable.

andreistefanescu:
@dwightguth and @yilongli thank you for the comments guys. I was about to make a really bad mistake. I reverted the caching for SymbolicConstraint and SymbolicConstraint.Equality.

@yilongli you are right that we should not cache the hash code for SymbolicConstraint.Equality, but we must override SymbolicConstraint.Equality#hashCode, otherwise search and prove may explore the same state multiple times.

@dwightguth I plan to make a separate pull request for changing the cache code from 0 to 1.

dwightguth:
Looks good. Merge if Yilong is happy.

yilongli:
I don't think it's correct. Equality can be put into hash-based collection. And once its hashCode is changed, you can no longer remove it from the collection correctly. 

There was one assertion that detects such problem:
https://github.com/andreistefanescu/k/blob/9fc2592b84a72dbff35cac30c3277bd9665e1953/src/javasources/KTool/src/org/kframework/backend/java/symbolic/SymbolicConstraint.java#L984

However, I removed it when I fixed the problem and added the comment in a5cb89a6a8fc0f37c87f207302e730a838a0830c.

andreistefanescu:
@yilongli please take a look

andreistefanescu:
Jenkins: test this please.

andreistefanescu:
@yilongli please take one more look.

yilongli:
lgtm.

dwightguth:
Jenkins: test this please

dwightguth:
Fyi: #384

PR #352, opened 2014-04-09
Improved TreeCleanerVisitor and how error reporting is being handled whe...

radumereuta:
...n calling the parser.

### DIFF
--- a/src/javasources/KTool/src/org/kframework/kil/visitors/BasicTransformer.java
+++ b/src/javasources/KTool/src/org/kframework/kil/visitors/BasicTransformer.java
@@ -1,3 +1,4 @@
+// Copyright (C) 2014 K Team. All Rights Reserved.
 package org.kframework.kil.visitors;
 
 import java.util.*;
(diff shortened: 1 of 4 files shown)


radumereuta:
@dwightguth, ignore this for now.
I forgot that all the JUnit tests match on the raw tree output. @mdmkolbe should I modify all the tests?

dwightguth:
This is not the change we discussed. I was fine with not calling TreeCleanerVisitor from Parser.java. What I wanted to see was that TreeCleanerVisitor would be called immediately on the output of Parser.java and then have it automatically either return null or throw a TransformerException if it discovers that the parse has failed, so that the code block in ProgramLoader.java was only a couple lines of code and contained no business logic. I am closing this because there are several pieces of this change which make the code less clean instead of more. Please sync up with me or Michael about how to proceed.

adamsmd:
Radu, I think the call to "result.accept(new TreeCleanerVisitor(null));" 
should happen in ProgramLoader.java instead of Parser.parser().

That fixes both the JUnit tests and keeps Parser.parser() orthogonal to 
how someone decides to simplify ambiguities.

PR #353, opened 2014-04-09
updated imp, lesson 1

grosu:
Updated IMP, Lesson 1, to work with the new unified builtins.

### DIFF
--- a/tutorial/1_k/2_imp/NOTES.md
+++ b/tutorial/1_k/2_imp/NOTES.md
@@ -1,0 +1,2 @@
+* Revise the change of `S1 S2` into `S1:Stmt S2:Stmt`, if needed; only `S2`
+really needs to be sorted.
--- a/tutorial/1_k/2_imp/lesson_1/README.md
+++ b/tutorial/1_k/2_imp/lesson_1/README.md
@@ -1,7 +1,5 @@
 ### Defining a More Complex Syntax
 
-[MOVIE [9'15"]](http://youtu.be/F39Ta1stiCM)
-
 Here we learn how to define a more complex language syntax than LAMBDA's,
 namely the C-like syntax of IMP.  Also, we will learn how to define languages
 using multiple modules, because we are going to separate IMP's syntax from
@@ -24,12 +22,12 @@ contents of the imported module in the one which imports it (making sure
 that everything is only kept once, even if you import it multiple times).
 In other words, there is currently nothing fancy in K tool's module system.
 
-IMP has six syntactic categories, as shown in imp.k: AExp for arithmetic
-expressions, BExp for Boolean expressions, Block for blocks, Stmt for
-statements, Pgm for programs and Ids for comma-separated lists of identifiers.
-Blocks are special statements, whose role is to syntactically constrain the
-conditional statement and the while loop statement to only take blocks as
-branches and body, respectively.
+IMP has six syntactic categories, as shown in `imp.k`: `AExp` for arithmetic
+expressions, `BExp` for Boolean expressions, `Block` for blocks, `Stmt` for
+statements, `Pgm` for programs and Ids for comma-separated lists of
+identifiers.  Blocks are special statements, whose role is to syntactically
+constrain the conditional statement and the while loop statement to only
+take blocks as branches and body, respectively.
 
 There is nothing special about arithmetic and Boolean expressions.  
 They are given the expected strictness attributes, except for `<=` and `&&`,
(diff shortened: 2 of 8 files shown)


grosu:
@traiansf please review (should be trivial, this is just my first pull request in the new branch; from here on I will include more files)

traiansf:
Looks ok to me.  However, @dwightguth what should we do about the jenkins
jobs for pull requests to this branch?

2014-04-09 23:47 GMT+03:00 Grigore Rosu notifications@github.com:

dwightguth:
Just merge even though they fail, but make sure whenever you commit changes
to one of them that it fixes the test for the definition you changed. Once
all the tests passing on master pass on the branch, it's ready to merge.

traiansf:
Ok, I'll merge this one.  But note that it fails first because of fact that
checkstyle is run on the files which are different from master, which means
a lot more than the pull request touched.

2014-04-10 0:30 GMT+03:00 dwightguth notifications@github.com:

dwightguth:
So I didn't actually notice that. It's probably okay in this case. But it's worth pointing out that it only actually checks files that are changed against the branch it's merging into, so in fact, Grigore could have fixed that part of the error with only two local changes. That's what I will suggest on future commits also.

grosu:
Besides, I'm afraid that you may have approved more than you thought :)  It looks like a clicked on git push after I finished revising the entire IMP definition in the tutorial  only a few seconds before you clicked on merge.  My bad, I thought that my first lesson was already merged.  I hope this is OK, because my changes were strictly applied to the tutorial, and nothing should depend on them.  But if not too late for you, Traian, you can take a look at the new changes, just in case.

Sorry about that, I will be more careful in the future.

Grigore

dwightguth:
In the future, try to only create pull requests once you're ready for them
to be merged, or, if not, clearly indicate that you are not ready for the
change to be merged yet.

grosu:
But shouldn't we have the people who introduced those inconsistencies fix them?  This way they learn to not do it again.

Grigore

grosu:
That's the idea, but the plan was to first make a "small" pull request in the unified-builtins2 branch, just to see how it goes.  I thought the pull request was merged when I pushed the second one, because I got the message from Traian saying that he will do the merge in spite of the error, etc.  Then I immediately did git push, thinking I was making another pull request; but it looks like I was faster than Traian, so my push was piggybacked to the previous pull request ... So it was all a matter of seconds.

Grigore

dwightguth:
If you limit yourself to one pull request per branch, this becomes less of
an issue, because you always are pushing different changes into a different
branch than earlier pull requests.

andreistefanescu:
Grigore, I'm not sure if you are aware, but you can drop the sort for Bool and Int variables (i.e. I1 instead of I1:Int) in the rules for "+", "/", "<=", and "!". Radu's pass would infer these sorts for you.

grosu:
What would be the right work flow, then?  Suppose I make a pull request P1 from branch B1 and I want to immediately continue to work on something else.  Which of the two approaches we should follow?

1) Continue to work and commit in the same branch B1.  If by the time I am done with my new edits my P1 pull request has been merged, then I can just push to B1 and make a new pull request P2.  If P1 is not merged when I'm done, then create a new branch B2, push my edits to that branch B2, and then make a pull request P2 from branch B2.

2) Start a new branch B2 immediately, do all my edits in B2 orthogonally to whether P1 has been merged or not, and when done with my edits, again orthogonally to whether P1 has been merged or not, push to B2 and make a pull request P2 from there.

None of the two cases is perfect, because I still have to keep track of when P1 has been merged, so I can remove my branch B2 to keep things clean.  But in your experience, which of the two approaches above do you think will minimize the risk to get into git conflicts?  I found those quite hard to solve so far, so I'd rather avoid them.

Thanks,
Grigore

dwightguth:
The problem with option 1 of the two you listed is that if you do this and
then have to modify the first change you cannot without mixing the two
changes together.

If you want to minimize conflicts, what I would recommend is, once you have
put the first change up for review, create the branch for the second change
by branching off the first branch. You will not be able to merge the second
code review until you have merged the first in this case, and you will have
to pull periodically the first branch into the second if you make changes
during code review, but it should prevent conflicts unless you have to
modify part of the first branch which was also modified by the second, but
you can't avoid that.

If you are struggling to resolve conflicts, maybe we should discuss
strategies for doing so? You won't necessarily ever be able to prevent them
in 100% of cases ...

PR #354, opened 2014-04-09
Fix maude profiler support.

chathhorn:
The profiler needs to be turned on before rewriting. Also, there was a missing line separator.


dwightguth:
Can you make a test for this so it doesn't happen again?

chathhorn:
The test I added would catch the missing newline, but I'm not sure how to test the profiler more thoroughly using ktest. The profiler output only shows up with the "--output raw" option, which also causes things like timing information and filesystem paths to appear in the output. Is there some way to make ktest check the output against a regex instead of just straight equality?

dwightguth:
checking against a regex actually sounds like a really useful feature. I
will make a pull request to add this functionality, then you can use it
once that pull,request is merged.

dwightguth:
Please see #356 for information on how to write your test. It is now merged.

chathhorn:
This is ready to be reviewed/merged again.

dwightguth:
I will own understanding this. Merging.

PR #355, opened 2014-04-10
Eliminated the inferrable sorts

grosu:
Eliminated the sorts that can be inferred, and modified the text accordingly.

@traiansf @andreistef: please review

### DIFF
--- a/tutorial/1_k/2_imp/README.md
+++ b/tutorial/1_k/2_imp/README.md
@@ -1,3 +1,5 @@
+<!-- Copyright (C) 2010-2014 K Team. All Rights Reserved. -->
+
 ## Part 2: Defining IMP
 
 Here you will learn how to define a very simple imperative language in K
--- a/tutorial/1_k/2_imp/lesson_1/README.md
+++ b/tutorial/1_k/2_imp/lesson_1/README.md
@@ -1,3 +1,5 @@
+<!-- Copyright (C) 2010-2014 K Team. All Rights Reserved. -->
+
 ### Defining a More Complex Syntax
 
 Here we learn how to define a more complex language syntax than LAMBDA's,
--- a/tutorial/1_k/2_imp/lesson_1/imp.k
+++ b/tutorial/1_k/2_imp/lesson_1/imp.k
@@ -1,3 +1,5 @@
+// Copyright (C) 2010-2014 K Team. All Rights Reserved.
+
 module IMP-SYNTAX
   syntax AExp  ::= Int | Id
                  | AExp "/" AExp              [left, strict]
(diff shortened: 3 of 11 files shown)


andreistefanescu:
lgtm

radumereuta:
I confirm that those are indeed the sorts which are being inferred for the java backend.
Please test if it works for the maude backend.

traiansf:
looks good to me, too

2014-04-10 5:11 GMT+03:00 Radu Mereuta notifications@github.com:

dwightguth:
Looks like you need to fix copyright.

grosu:
OK, I will do it today.  Sorry for taking so long, I had meetings almost back-to-back the last few days.  So in the end we write "Copyright (C) ..." or "Copyright (c) ...".  Can you please include this in the checkstyle script as well?  I ran checkstyle and there are no warnings/errors related to my files.

dwightguth:
Grigore: the problem with running the copyright check locally is that it
has to ask git what files have changed, and that command depends on how
your git repository is structured. If you're working on a feature branch,
the command is `git diff master`. But if you're working on master the
command is `git diff upstream/master`, and if you're working on a feature
branch and haven't rebased master since the last time you fetched, `git
diff upstream/master` will return incorrect output. And in your case you're
merging into unified-builtins2, which git won't know until you make the
pull request. So for you it's `git diff upstream/unified-builtins2`.

I can make a command like ant checkstyle-copyright -Dbase=master if you
want though, but you will still have to run it separately from `ant test`
or `ant checkstyle`, and pass the base parameter to the script. Which is
already very similar to the:

```
git diff master --name-only > .checkstyle-files
ant checkstyle-copyright
```

command pair I included on the wiki. So it's up to you based on what you
want me to do.

radumereuta:
@dwightguth, I think the best way to implement this is to do what you said in the last paragraph.
Everything should be contained in the ant scripts, no intermediate data. Something like <target name="makeversion">
I think you could set the default option to be origin/master. Probably this would be the most common case.

grosu:
@dwightguth @radumereuta 
Can't we actually make the checkstyle take an optional parameter, e.g.
    ant checkstyle -Dbase=master
and then have it check not only the copyright, but also the other style conventions, only on the modified files?  Excuse my ignorance (I am really far from claiming that I understand how git works), but would it be possible to also run exactly the same script when somebody does a pull request (with option the branch the pull request is made to), and thus reject the request if the style does not check?

grosu:
@dwightguth Please revise and merge if ready

I added the copyright notices both in the .k and in the .md files in tutorial/imp

dwightguth:
Grigore. This is blocked on #363 because I accidentally screwed up the merge of master into unified-builtins. I will merge this as soon as we get all the tests other than ktest passing.

dwightguth:
Jenkins test this please

PR #356, opened 2014-04-10
add regular expression pattern matching for ktest

dwightguth:
- Add a "regex" flag to programs that specifies that the expected output file will contain one regex per line.
- Added functionality to check regex files by asserting that all the regexes can be found in the output string.
- Demo this functionality by using it for search results in imp++ and logik


dwightguth:
@osa1 @daejunpark @grosu please review.

src/javasources/KTool/src/org/kframework/ktest/Proc.java
```diff
@@ -323,12 +325,12 @@ private void reportErr(String err) {
 
     private void reportErrMatch(Annotated<String, String> expected, String found) {
```
osa1: Arguments `expected` and `found` are unused now. Maybe remove these?

src/javasources/KTool/src/org/kframework/ktest/Proc.java
```diff
@@ -323,12 +325,12 @@ private void reportErr(String err) {
...
         assert reason == null;
-        reason = String.format("Expected program error:%n%s%n%nbut found:%n%s%n", expected, found);
+        reason = "Unexpected program error:\n" + strComparator.errorMessage();
     }
 
     private void reportOutMatch(Annotated<String, String> expected, String found) {
```
osa1: Same as `reportErrMatch`.

src/javasources/KTool/src/org/kframework/ktest/RegexStringMatcher.java
```diff
@@ -0,0 +1,32 @@
...
+                failedActual = actual;
+                break;
+            }
+        }
+        return matches;
+    }
+    
+    @Override
+    public String errorMessage() {
+        return String.format("Expected result to match regular expression:%n%s%n%nbut found:%n%s%n", failedRegex, failedActual);
```
osa1: This line is 128 chars long... this is not compliant to our coding conventions.

src/javasources/KTool/src/org/kframework/ktest/Test/ProgramProfile.java
```diff
@@ -0,0 +1,29 @@
...
+import org.kframework.ktest.PgmArg;
+
+/**
+ * A set of information used to specify the behavior for executing some subset of KRun programs.
+ * For example, a profile might represent all programs, or it might represent programs individually.
+ *
+ */
+public class ProgramProfile {
+    private List<PgmArg> args;
+    private boolean regex;
```
osa1, line 23: Can you document at least this attribute? (`regex`)

osa1:
Interesting feature..

A question: It's clear from diff file that `IgnoringStringComparator` is renamed to `IgnoringStringMatcher`, but it's very hard to see if there are other changes in this file. In cases like this(where both renaming and changes happened on same file), why not make two separate commits for easier reviewing?

dwightguth:
That's my fault, I messed up the renaming process by doing it in eclipse
instead of on the command line. Give me a sec and I'll fix it.

On Thu, Apr 10, 2014 at 2:07 PM, Ömer Sinan Ağacan <notifications@github.com

dwightguth:
Okay so github didn't really know how to deal with a rename with modifications it looks like, but if you look at the commits individually you should be able to see the changes.

dwightguth:
@osa1 please re-review

src/javasources/KTool/src/org/kframework/ktest/IgnoringStringMatcher.java
```diff
@@ -0,0 +1,55 @@
...
+    private String lastPattern;
+    private String lastActual;
+    
+    public IgnoringStringMatcher(boolean ignoreWS, boolean ignoreBalancedParens) {
+        this.ignoreWS = ignoreWS;
+        this.ignoreBalancedParens = ignoreBalancedParens;
+    }
+
+    @Override
+    public boolean matches(String pattern, String actual) {
```
osa1, line 18: Why `s1` renamed to `pattern` here? This is not a regex pattern, we're doing string comparison here. I'm not 100% sure that this is an improvement because from the name it looks like a regex pattern while it's actually not.

dwightguth, line 18: I will add a javadoc on the interface to explain this.

osa1:
Thanks! Apart from the part I mentioned as inline comment this looks good to me.

daejunpark:
Nice feature. Looks good. I have no objection to it. Thanks!

src/javasources/KTool/src/org/kframework/ktest/RegexStringMatcher.java
grosu: Dwight, would it make sense to incorporate a check for line length in the checkstyle script?  Also, would it make sense to send a "style violation warning" message when the style is violated, so people learn to not do it anymore?

Thanks,
Grigore

dwightguth: Since we already have code that runs checkstyle on the code modified by a
diff, this would be easy to set up. The only problem is it would require
users to fix every line in a file when they modify it, and not just their
own lines. That would be kind of intrusive. Making it only detect modified
/lines/ would be a lot harder to do and require more effort. How do you
want me to spend my time?

dwightguth:
@grosu can you please okay the changes I made to the tutorial's test suite so that I can merge this?

tutorial/1_k/4_imp++/lesson_3/tests/div.imp.out
```diff
@@ -1,76 +1,5 @@
...
-    <store>
-        #symNat(1) |-> 3
-        #symNat(2) |-> 3
-    </store>
-    <env>
-        x |-> #symNat(1)
-        y |-> #symNat(2)
-    </env>
-</T>
+<T>\s*<k>\s*\(3 / 0\) ~> y = HOLE ;\s*</k>\s*<store>\s*(.*) \|-> 3\s+((?!\1).*) \|-> 0\s*</store>\s*<env>\s*x \|-> \1\s+y \|-> \2\s*</env>\s*</T>
```
grosu, line 1: Guys, I understand why we are doing this, but it makes the job of writing tests in general and .out files in particular harder and thus demotivates the users.  Now they have to basically manually edit the output of krun, it is not as easy as it used to be:

```
krun pgm.imp >pgm.imp.out
```

We can keep it, provided we document it well in the manual, but I would still like to keep what we have now as an option (default one?).  I mean the whitespaces and the balanced parentheses.  We don't want users to suddenly have their tests fail only because we decided to pretty-print the output differently.  This opinion is not very strong, so in case you guys bring me solid argument for dropping it, I am listening.

src/javasources/KTool/src/org/kframework/ktest/RegexStringMatcher.java
grosu: Can you give us a number, like how many checkstyle inconsistencies are there?  If only a dozen, then maybe we can suffer a bit now and have people who modify files also fix the style in those files.  If hundreds, then we can probably just keep it as is, and maybe only add the copyright note; but in that case, are we doing anything to make sure that all code passes the checkstyle?

Grigore

tutorial/1_k/4_imp++/lesson_3/tests/div.imp.out
dwightguth, line 1: By default this code keeps the same behavior that it had before. It is only if you say something like:

```
<all-programs regex="true">
```

or:

```
<program name="foo.c" regex="true">
```

That this change takes effect. In fact, it makes our tests stronger and _less_ fragile to use this in cases where we are comparing krun output configurations and search results. By using regular expressions in the form I demonstrate in these examples, we are able to make the test not depend on the syntactic representation of fresh variables, on the ordering of search results within the resulting program, or on the precise wording of KRun's output headers. Instead it provides an extremely powerful syntax for specifying precisely what you do or don't care about in the output of krun.

If you approve of my reasoning, I will make sure to add a line to document this in the manual before I merge this. But this will probably be needed in the long term anyway in the tutorial because the Java and Maude backends cannot be guaranteed to generate identical representations of fresh variables or the ordering of search results.

src/javasources/KTool/src/org/kframework/ktest/RegexStringMatcher.java
dwightguth: In the worst case it could be anywhere from 50 to 100 lines per file.

grosu:
I agree that it is powerful, no doubt about that.  It is just that users will have a harder time to digest it.  So I approve it.  Go ahead and merge whenever you are ready.

But, regarding power, I think that power it gives us now will be overpowered by the capability to cannonicalize configurations, which we will need anyway.  For example, we will need the model checker to realize that two configurations containing "lambda x . x" and, resp., "lambda y . y" are equivalent.  Similarly, bindings in the store can be in different orders, depending on the names of the fresh variables.  The best way to deal with equality of configurations is to have a mechanism to cannonicalize them (this is hard and interesting problem in itself, based on ideas from techniques checking for graph isomorphism).  With that, we should then be able relatively easily to check whether two configurations are equivalent.  I think Liyi was supposed to work on this problem.

PR #357, opened 2014-04-10
Fixed the license issues with Tarjan.

radumereuta:


dwightguth:
lgtm. Apparently the Unlicense is a public domain license which means that as a derivative work, our copy of the file is copyrightable by us. I will merge this.

traiansf:
I, too have a Tarjan SCC class in the pdmc2 branch
I wonder whether these two could be merged.

2014-04-10 22:38 GMT+03:00 dwightguth notifications@github.com:

PR #360, opened 2014-04-10
Fixed typo where an assertion error was not being thrown.

adamsmd:
A minor fix that I noticed while looking over the code.


dwightguth:
I assume you checked everywhere else in the code for other cases of this error? If so, merge.

PR #362, opened 2014-04-11
add copyright check for markdown

dwightguth:
@radumereuta @grosu please review.

### DIFF
--- a/src/javasources/KTool/checkstyle-copyright.xml
+++ b/src/javasources/KTool/checkstyle-copyright.xml
@@ -16,6 +16,10 @@
     <property name="fileExtensions" value="xml,xsd" />
   </module>
   <module name="RegexpHeader">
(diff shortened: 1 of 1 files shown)


src/javasources/KTool/checkstyle-copyright.xml
```diff
@@ -16,6 +16,10 @@
     <property name="fileExtensions" value="xml,xsd" />
   </module>
   <module name="RegexpHeader">
+    <property name="header" value="^&lt;!-- Copyright \(C\) (\d\d\d\d-)?2014 K Team. All Rights Reserved. --&gt;$" />
```
grosu, line 19: I think this is indeed what we want, because in MD,  renders the comment as comment in the generated files (HTML, etc.).  If we wanted not to render them, then we would use , but I think we should use two dashes only.  So I am OK with your changes.

radumereuta:
Yep, looks right.

PR #363, opened 2014-04-11
fix whitespace from my mistake when merging

dwightguth:
If you don't want to read all the boring details of this review, just download the change and run "git show -w" to show you that it's just whitespace changes, and run "ant checkstyle" to show that it solves the checkstyle issues in the branch.


dwightguth:
@traiansf please review.

dwightguth:
Incidentally I plan to merge this even though it fails the copyright test, because technically all I'm doing here is fixing the merge I did before which I accidentally screwed up.

traiansf:
Ok, I verified. There are only whitespace changes.  feel free to merge.

PR #364, opened 2014-04-11
first draft of visitor refactoring

dwightguth:
this commit breaks everything ever, I know. But I want to get reviews done on the interface for visitors before I start making changes to the entire code base to support the new API.

Note that while I'm sure I'll have to do some fine-tuning, this is designed to serve as a drop-in replacement for what already existed, except that BasicHookWorker has now become
LocalTransformer in keeping with its similarity in function to LocalTransformer in org.kframework.backend.java.

@traiansf @radumereuta @andreistefanescu Please review. Ignore the fact that it doesn't compile for now, I don't intend to merge this yet.


src/javasources/KTool/src/org/kframework/kil/visitors/AbstractVisitor.java
```diff
@@ -0,0 +1,1045 @@
...
+ * return {@code null}.
+ */
+public abstract class AbstractVisitor<P, R> implements Visitor<P, R> {
+    protected org.kframework.kil.loader.Context context;
+    protected KompileOptions kompileOptions;
+    protected GlobalOptions globalOptions;
+    protected ExperimentalParserOptions experimentalParserOptions;
+    String name;
+    
+    protected java.util.Map<ASTNode, R> cache = new HashMap<>();
```
adamsmd: Note that this cache is very inefficient until hashCode() is better implemented.

Something that didn't occur to me when I talked to Dwight yesterday is that even with a fast hashCode() this HashMap might be slow if HashMap uses Object.equals instead of pointer equality.  I can't find anywhere that the JavaDoc says one way or the other.

src/javasources/KTool/src/org/kframework/kil/Term.java
```diff
@@ -74,10 +73,11 @@ public boolean contains(Object obj) {
      */
     public Set<Variable> variables() {
         final Set<Variable> result = new HashSet<Variable>();
-        this.accept(new BasicVisitor(null) {
+        this.accept(new AbstractVisitor<Void, Void>(null) {
             @Override
-            public void visit(Variable node) {
+            public Void visit(Variable node, Void _) {
```
adamsmd, line 78: What does "Void _" mean in this context?  (Is that Java 7 or something new in Java 8?)

src/javasources/KTool/src/org/kframework/kil/visitors/exceptions/TransformerException.java
```diff
@@ -4,7 +4,7 @@
 import org.kframework.utils.general.GlobalSettings;
 
 @SuppressWarnings("serial")
-public class TransformerException extends Exception {
+public class TransformerException extends RuntimeException {
```
adamsmd: I'm not sure there is anything to do about it, but I think making these exceptions be undeclared is unfortunate.

Could you do something like `public <P, R, T> R accept(...) throws T {}` or some variant of that.

src/javasources/KTool/src/org/kframework/kil/visitors/AbstractVisitor.java
dwightguth: All Java Collections Framework classes are implemented modulo Object.equals.

Our visitors are also not a performance bottleneck right now so while I
will compare performance before and after, I don't anticipate a problem at
this time, which means we can optimize as the need arises. Also it's worth
noting only BasicVisitor overrides cache to true, because only that class
did that check in the original implementation I am erasing. So this isn't
doing anything that wasn't already being done.

src/javasources/KTool/src/org/kframework/kil/Term.java
dwightguth, line 78: Void is the Java class associated with void operations just as Integer is
associated with int. Unfortunately Java doesn't autobox void, so I have to
actually declare this parameter, even though it really just means "no
parameter here". so I named the variable _ to indicate that it's an unused
field. How do you want me to proceed here? I don't have a strong
preference.

adamsmd, line 78: Ah, I was misreading it.  For some reason, I thought I was looking at an 
"actual argument" of a method _call_ instead of the "formal parameter" 
of a method _definition_.  What you've written is reasonable.

src/javasources/KTool/src/org/kframework/kil/visitors/exceptions/TransformerException.java
dwightguth: Thanks for the tip! This actually turned out really cool. Because the java compiler adds declarations that all methods can throw RuntimeException behind the scenes, I was able to make this work perfectly while at the same time handling the case where no exception is thrown. If only Java autoboxed void properly, this would probably be close to compiling...

dwightguth:
@traiansf @andreistefanescu @radumereuta please review.

dwightguth:
test this please

dwightguth:
right, I forgot. Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Okay, I'm pretty sure this change is done now (assuming tests pass like I think they will).

Since nobody other than Michael took the opportunity to complain about this diff I'm going to assume that the conceptual design is okay, which means I just need the code reviewed in general.

@andreistefanescu please review. Note that you will need to download the diff because it modifies more files than github allows you to view online. Since it has been more than two days since I asked you to review, please do not merge anything else until this review is completed.

@grosu  please also review the changes I made to the tutorial. Again, you will have to download the diff to your local machine. Let me know if you need help figuring out how to do this. I made the changes I made because I found a bug while refactoring the visitors where previously transformers did not apply underneath objects of class Freezer. Fixing this bug caused the output of some tests in imp++ to change.

src/javasources/KTool/src/org/kframework/kil/loader/JavaClassesFactory.java
```diff
@@ -194,7 +195,7 @@ public static ASTNode getTerm(ATerm atm) {
             if (appl.getName().endsWith("Ensures"))
                 return storeNode(atm.getUniqueIdentifier(), new Sentence(appl));
             if (appl.getName().endsWith("Rewrite"))
-                return storeNode(atm.getUniqueIdentifier(), new Rewrite(appl));
+                return storeNode(atm.getUniqueIdentifier(), new Rewrite(appl, context));
```
radumereuta, line 198: I don't like the fact that you have to send the context to a KIL class.
They were supposed to be independent of the context.
I'm not sure how this would behave when we will have to do modularity.
Please motivate your decision.

src/javasources/KTool/src/org/kframework/kil/visitors/AbstractTransformer.java
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
+package org.kframework.kil.visitors;
+
+import org.kframework.kil.ASTNode;
+import org.kframework.kil.loader.Context;
+import org.kframework.kil.visitors.exceptions.TransformerException;
+
+public abstract class AbstractTransformer extends AbstractVisitor<Void, ASTNode, TransformerException> {
```
radumereuta, line 18: javadoc

src/javasources/KTool/src/org/kframework/kil/visitors/AbstractTransformer.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public AbstractTransformer(String name, Context context) {
+        super(name, context);
+    }
+
+    @Override
+    public ASTNode defaultReturnValue(ASTNode node, Void _) {
+        // TODO Auto-generated method stub
```
radumereuta: remove TODO if not needed

src/javasources/KTool/src/org/kframework/kil/visitors/AbstractTransformer.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @Override
+    public boolean cache() {
+        return false;
+    }
+
+    @Override
+    public <T extends ASTNode> boolean changed(T o, T n) {
+        // TODO Auto-generated method stub
```
radumereuta: This one too.
Please look through every file you modified and remove the comments if you don't need them anymore.

andreistefanescu:
@dwightguth I would like to have a lot more javadoc either on Visitor or AbstractVisitor explaining the overall architecture. Right now I could guess how it works, but I should not have to.

src/javasources/KTool/src/org/kframework/kil/visitors/AbstractVisitor.java
```diff
@@ -0,0 +1,1113 @@
...
+        return visit((Term) node, p);
+    }
+
+    @Override
+    public R visit(Ambiguity node, P p) throws E {
+        if (cache() && cache.containsKey(node)) {
+            return cache.get(node);
+        }
+        return visit((Collection) node, p);
+    }
```
radumereuta: Are you sure this is the correct visitor for Ambiguity?
In BasicTransformer I had more code. It would use a try-catch block each time it visited a child.
If it caught an exception from that branch, it would remove the child.
I use this extensively in the disambiguation filters.

src/javasources/KTool/src/org/kframework/kil/visitors/BasicTransformer.java
```diff
@@ -1,625 +1,58 @@
...
         for (Term t : node.getContents()) {
             ASTNode result;
             try {
-                result = t.accept(this);
+                result = t.accept(this, null);
                 terms.add((Term) result);
             } catch (TransformerException e) {
                 exception = e;
             }
         }
```
radumereuta, line 42: Oh, I get it now. Ignore the last comment.

andreistefanescu:
I am also confused about how the caching is supposed to work. If I override a method in my visitor, and I forget to call at the end return visit((SuperClass) returnNode); then caching would not work...or am I missing something?

src/javasources/KTool/src/org/kframework/kil/visitors/LocalTransformer.java
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
+package org.kframework.kil.visitors;
+
+import org.kframework.kil.ASTNode;
+import org.kframework.kil.loader.Context;
+
+public class LocalTransformer extends AbstractTransformer {
```
radumereuta, line 13: javadoc

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/VariableTypeInferenceFilter.java
```diff
@@ -29,12 +30,12 @@ public VariableTypeInferenceFilter(Context context) {
...
 
     @Override
-    public ASTNode transform(Sentence r) throws TransformerException {
+    public ASTNode visit(Sentence r, Void _) throws TransformerException {
         r = (Sentence) r.accept(new RemoveDuplicateVariables(context));
 
         CollectVariablesVisitor vars = new CollectVariablesVisitor(context);
         r.accept(vars);
-
+
```
radumereuta: Did you just add 4 spaces here? :)

src/javasources/KTool/src/org/kframework/kil/loader/JavaClassesFactory.java
dwightguth, line 198: Well in the long term vision of K, the Context class is going to be deprecated. I intend to install Google Guice and use binding annotations to serve information directly to any class that asks for it. So for example the constructor for Rewrite would be:

```
@Inject
public Rewrite(@Assisted Term left, @Assisted Term right, @Named("Subsorts") Poset subsorts) {
    this.left = left;
    this.right = right;
    recomputeSort(subsorts);
}
```

Then modularity would be handled by creating a different Guice Module for each K module and binding different Posets to the named binding that we are using.

src/javasources/KTool/src/org/kframework/kil/visitors/AbstractVisitor.java
dwightguth: See BasicTransformer.java, which overrides this method. In order for that code to work it's necessary to throw the checked exception TransformerException, which BasicVisitor should not throw (and I suspect CopyOnWriteTransformer should not either, but we'll see about that later).

src/javasources/KTool/src/org/kframework/kil/visitors/BasicTransformer.java
dwightguth, line 42: ok

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/VariableTypeInferenceFilter.java
dwightguth: Good eye! I ran into a bug that I had to fix with the type inferencer, this must have been left over from some debugging attempts. I fixed it.

dwightguth:
@andreistefanescu : You are correct that is how caching works, however, that is also how caching previously worked with BasicVisitor where we were doing it before. Furthermore, I don't see a good way to avoid this. No matter which visit method I put it in, the user could override that method in such a way as to make the cache incorrect. I suppose I could put it in accept, but that seems strange and would require exposing the cache outside the visitor class also. Not to mention it's more lines of code that way. So unless you have a better suggestion, I don't really know what to do about this.

grosu:
I do not see any changes in the tutorial, but I see the following message on the page:

Diff truncated: maximum file count exceeded.

Anybody has the same problem?  How can I see the entire Diff?

Grigore

dwightguth:
Grigore: Github apparently doesn't properly display patches with this many files, so you will have to download the change yourself. To do this, follow the instructions at the following page, except replace "origin" everywhere with "upstream".

https://gist.github.com/piscisaureus/3342247

Then you can inspect the changes to the tutorial by running:

```
git diff master -- tutorial/
```

dwightguth:
@andreistefanescu @radumereuta please re-review.

dwightguth:
@grosu again, can you confirm that the changes I made to the tutorial are okay? You are the last thing I am waiting on before merging this.

@andreistefanescu : if you want to review this in more detail, I suggest you do it now.

dwightguth:
Looks like I found a better way to view this diff in its entirety: https://github.com/kframework/k/pull/364.diff

I am hoping to merge this in the next couple days, so if you want to provide further review comments, now is the time.

grosu:
Not only that I approve this, but I am VERY HAPPY with this!

```
-        "Input two numbers: " ~> ( HOLE,.List{","} ) ~> print ( HOLE ) ;
-           ~> int (x, y) ; ~> x = read ( ) ; ~> y = read ( ) ; ~> print ( (
-          "Their sum is: ", ((x + y), "\n")) ) ;
+        "Input two numbers: " ~> HOLE ~> print ( HOLE ) ; ~> int x, y ;
+           ~> x = (read ( )) ; ~> y = (read ( )) ; ~> print ( "Their sum is: "
+          , (x + y), "\n" ) ;
```

The `.List{","}` was very ugly and distracting.  The current output is just plain beautiful, finally.  Great work guys!

andreistefanescu:
lgtm

dwightguth:
If anyone still wants to review this, now is the time. I will be merging this sometime in the next couple days.

radumereuta:
Could you write a more detailed description of the changes you commited today?
It's a fairly large change and I want to better understand your motivations before I jump to conclusions.

dwightguth:
I changed the approved syntax for invoking a visitor on a term. You now call `visitor.visitNode(node)`. I moved AbstractVisitor into org.kframework.kil in order to make accept be protected instead of public, so that you can't use the wrong syntax by mistake. I wrote the machinery needed to remove a lot of the repetitive boilerplate code in AbstractVisitor.java, replacing it with some highly complex trickery with Java generics. To do this, I created the class org.kframework.kil.Interfaces which contains inner classes designed to annotate ASTNode classes with specific functionality. For example, Interfaces.Parent denotes an ASTNode which has children of class ASTNode, and Interfaces.MutableList denotes an ASTNode which has a getter and a setter for a list of related ASTNode children (e.g. Production has a mutable list of ProductionItems). A term with more than one type of child node uses an enum to distinguish between them, for example Sentence.Children has BODY, REQUIRES, ENSURES. I also removed the Visitable interface; it was serving no purpose, since only ASTNodes were visitable anyway.

Please let me know where, if anywhere, you want any of this commented.

radumereuta:
"replacing it with some highly complex"
Are you sure we want this? The code should be easy to understand for a new developer. Even for us, if we revisit the code after a long time, we should be able to tell what is going on.

Please make sure you have written all the necessary javadoc.
If you already did, please point me to it, so I can take a deeper look. At the moment I don't understand the change.

dwightguth:
I disagree with your assessment. I would say that the code should be easy
to /modify/ for a new developer, but a new developer should not need to
understand the deep details of the underlying infrastructure it is built on
top of. The Javadoc for this change is in AbstractVisitor.java, where it
always has been. I agree I am doing very complex things with generics,
however, by doing these things, I allow the KAST class hierarchy to be
modified almost trivially. That is a very good trade-off to make, in my
opinion.

traiansf:
I agree with Dwight on this.  It is much better to have a base mechanism
which is more complex but establishes good policies for the deriving
classes, than to have a simpler basic mechanism which requires us to do
type checking when overriding them.

2014-05-02 1:10 GMT+03:00 dwightguth notifications@github.com:

> I disagree with your assessment. I would say that the code should be easy
> to /modify/ for a new developer, but a new developer should not need to
> understand the deep details of the underlying infrastructure it is built
> on
> top of. The Javadoc for this change is in AbstractVisitor.java, where it
> always has been. I agree I am doing very complex things with generics,
> however, by doing these things, I allow the KAST class hierarchy to be
> modified almost trivially. That is a very good trade-off to make, in my
> opinion.
>

radumereuta:
Ok, I am convinced then.

dwightguth:
Unless anybody has more comments I will probably merge this in the next couple days.

PR #370, opened 2014-04-14
fixed bug in builtin list factory method

andreistefanescu:
@traiansf please check this bugfix


traiansf:
Looks good.  I wonder why it hasn't surfaced until now.  Could you please add a unit test for it, too?

traiansf:
looks good to me

Issue #375, opened 2014-04-14
unnecessary messages reported by ktest


grosu:
Why does ktest report the second and third messages below?

```
D:\github.grosu\k\tutorial\1_k\4_imp++\lesson_4>ktest tests\config.xml --skip=pdf
Kompile the language definitions...(1 in total)
SUCCESS
Kompiling definitions that are not yet kompiled.
Kompile the language definitions...(0 in total)
SUCCESS
```

When one kompiles only one definition, there is no need for those messages. They just make reading the output harder.

osa1:
@grosu with the recent change can you try again and close this if you're happy with the result?

grosu:
Sorry, but I don't know how to test this.  Your changes, Omer, so not seem to have been applied to the unified-builtins2 branch, which my pull request was following.  I tried to pull from kframework/k in my branch and I managed to run into several conflicts.  Sorry, but I don't know how to deal with this.  Hopefully somebody can merge the current master into the unified-builtins2 branch, so I can pull into my branch?

Thanks,
Grigore

dwightguth:
Grigore: I have merged master into unified-builtins2 for you. You should be
able to rebase to fastfoward your branch now, and have it work without
conflicts. Let me know if you run into problems.

osa1:
I'm closing this as currently no redundant messages are printed for skipped steps. (was fixed a while ago)

PR #377, opened 2014-04-14
prevent redundant ktest info printing (#375)

osa1:
#375


dwightguth:
I have been playing around with KTest a bit recently so I will own responsibility for understanding this change. @daejunpark I suggest you also read this diff.

I will merge now though.

daejunpark:
Looks good to me, too.

Issue #368, opened 2014-04-12
strange symbols in Windows displaying


grosu:
What are the symbols appearing before and after SUCCESS below?  Probably
colors.  Note that colors work just fine with krun (the displayed configurations
have colors in them), so we should do the exact same thing in ktest:

```
←[32mSUCCESS←[0m
----------------------------
Definitions kompiled: 1 (time: 0'30'' elapsed, 0'30'' CPU)
PDF posters kompiled: 1 (time: 0'35'' elapsed, 0'35'' CPU)
Programs krun: 3 (time: 0'7'' elapsed, 0'20'' CPU)
Total time: 1'12'' elapsed, 1'25'' CPU
----------------------------
```

grosu:
Similarly funny symbols for colors when FAIL is reported:

```
←[31mERROR: [D:\github.grosu\k\bin\krun.bat D:\github.grosu\k\tutorial\1_k\4_i
++\lesson_7\tests\spawn.imp --search] output doesn't match with expected outpu
(time: 6159 ms)←[0m
←[31mFAIL←[0m

←[31mFAIL (see details above)←[0m
----------------------------
Definitions kompiled: 1 (time: 0'8'' elapsed, 0'8'' CPU)
PDF posters kompiled: 0 (time: 0'0'' elapsed, 0'0'' CPU)
Programs krun: 4 (time: 0'8'' elapsed, 0'29'' CPU)
Total time: 0'16'' elapsed, 0'37'' CPU
----------------------------
```

dwightguth:
Looks like we need to detect whether the terminal supports ANSI color codes.

osa1:
Some parts of the framework uses `jansi` library for colorful printing and that library handles this cases(e.g. if terminal does not support colors, it doesn't print useless escape codes. I think it may even convert color codes for different terminals?). Maybe we can use `jansi` in ktest too. (I don't know why we didn't use it in the first place)

EDIT: I just read Grigore's first message and yeah, apparently jansi even converts color codes for terminal's expected ones.

osa1:
I just checked and places that `jansi` is used:

```
➜  k git:(master) ack --java AnsiConsole
src/javasources/KTool/src/org/kframework/krun/Main.java
32:import org.fusesource.jansi.AnsiConsole;
516:                    AnsiConsole.out.println(output);
525:                    AnsiConsole.out
622:                    AnsiConsole.out.println(debugger.printState(debugger
[...]

PR #379, opened 2014-04-14
fix ktest colorful output printing for non-ANSI terminals (issue #368)

osa1:
#368

This should fix it. We could do two things:
1. Replace `System.out.print[ln]` calls that prints colorful texts with `AnsiTerminal.out.print[ln]`
2. Just call `AnsiTerminal.systemInstall()` and use `System.out` as usual.

I choose second one. Please tell me if you see problems with that.

Note: We can also do that in krun. Currently it uses `AnsiTerminal.out` when it prints colorful text and we don't know for sure that no colorful text is printed using `System.out`.


dwightguth:
Lgtm. It looks like we first introduced jansi back in the dark days of krun and the people who did it are no longer around. So I will take responsibility for understanding what is going on here. Can you try to clean up krun so that it does the same thing here?

osa1:
Yep, I'll make required changes in `krun` and send a pull requires in a few hours.

PR #382, opened 2014-04-15
Freezing the code for path indexing

owolabileg:
The path indexing now works for most languages in the K tutorial. Stopping work on it for now, so that I can move to  a more general algorithm.


dwightguth:
hmm not really sure why this code isn't testing itself...

dwightguth:
Jenkins: test this please

dwightguth:
add to whitelist

dwightguth:
Jenkins: test this please

owolabileg:
I think this was because I forgot about the copyright information. I've added them now. Sorry about that.

dwightguth:
Jenkins: test this please.

dwightguth:
Can you please add @Deprecated attributes and @deprecated javadocs to the top-level entry points of this code path? The understanding I reached with Andrei was that this would be an unsupported change which we commit because when it works, it's a speedup. But that if we needed to make changes in order to make it work we would just revert back to the old algorithm. Hence I believe @deprecated is appropriate.

Once you do this, I will merge.

owolabileg:
Okay, I made the changes requested by @dwightguth. Do they suffice?

dwightguth:
Typically both the @Deprecated attribute and the @deprecated javadoc are applied to every instance of a deprecated object. Also, can you deprecate the enum constant in KompileOptions?

owolabileg:
Done.

dwightguth:
Looks good, but as I said, can you please deprecate IndexingAlgorithm.PATH?

owolabileg:
I already marked that constant as Deprecated. Maybe I am missing something else?

dwightguth:
I'm sorry, I missed it. Merging.

PR #383, opened 2014-04-15
Symbolic kompile

andreiarusoaie:
This pull request fixes the symbolic backend (which stopped working for some unknown reason) and appends a regression test for its functionality. Also, it it follows the same approach as the other backends, by having its own KRun implementation.

@dwightguth @dlucanu Please review.


andreiarusoaie:
It seems that Jenkins has been timed out after 30 minutes. How do I know if this pull request passes the tests? Note that the test I have added passed (I've looked into the console log), and I haven't done any modifications which can possibly affect the other tests.

src/javasources/KTool/src/org/kframework/backend/symbolic/krun/SymbolicKRun.java
```diff
@@ -0,0 +1,84 @@
...
+import org.kframework.krun.api.Transition;
+import org.kframework.utils.Stopwatch;
+
+import java.util.Set;
+
+/**
+ * Created by andrei.arusoaie on 4/15/14.
+ * This class implements the KRun interface for the symbolic backend.
+ */
+public class SymbolicKRun implements KRun {
```
dwightguth: I'd suggest based on the design you have in this class that you extend MaudeKRun instead of having an internal KRun field.

src/javasources/KTool/src/org/kframework/kompile/KompileOptions.java
```diff
@@ -196,10 +192,10 @@ public SMTSolver convert(String arg0) {
         public boolean noPrelude = false;
         
         @Parameter(names="--symbolic-rules", converter=TagListConverter.class, description="Apply symbolic transformations only to rules annotated with tags from <tags> set. This only has an effect with '--backend symbolic'.")
-        public Set<String> symbolicRules;
+        public Set<String> symbolicRules = new LinkedHashSet<>();
```
dwightguth: Is there a reason your code doesn't work if the keys of these sets aren't ordered? That seems to me like something that should be refactored out, rather than catered to, unless there's some design subtlety I'm missing here.

tests/regression/symbolic-backend/config.xml
```diff
@@ -0,0 +1,21 @@
...
+<tests>
+  <!-- tests for symbolic backend -->
+  <test
+      definition="symbolic-kompile/lang.k"
+      programs="symbolic-kompile/programs"
+      results="symbolic-kompile/tests"
+      extension="lang">
+    <kompile-option name="--backend" value="symbolic" />
+    <all-programs>
+       <krun-option name="--backend" value="symbolic" />
```
dwightguth: This should not be necessary. If it is, then your code is missing a modification somewhere. Please remove it and see what happens.

tests/regression/symbolic-backend/config.xml
```diff
@@ -0,0 +1,21 @@
...
+       <krun-option name="--color" value="off" />
+       <krun-option name="-cIN" value="ListItem(#symInt(n))"/>
```
dwightguth:
```
<krun-option name="-c" key="IN" value="ListItem(#symInt(n))" />
```

tests/regression/symbolic-backend/config.xml
```diff
@@ -0,0 +1,21 @@
...
+       <krun-option name="-cPC" value="true" />
```
dwightguth: see above

tests/regression/symbolic-backend/config.xml
```diff
@@ -0,0 +1,21 @@
...
+       <krun-option name="--search" />
+       <krun-option name="-bound" value="2" />
```
dwightguth: --bound

tests/regression/symbolic-backend/symbolic-kompile/lang.k
```diff
@@ -0,0 +1,66 @@
+// Copyright (C) 2012-2014 K Team. All Rights Reserved.
```
dwightguth: This looks a lot like IMP to me. Did you make some modifications to this definition so it would work with your backend? If not, I would like to see your backend tested on existing definitions, rather than adding a new one. Enforcing different definitions for different backends is something we are trying to move away from.

tests/regression/symbolic-backend/symbolic-kompile/tests/sum.lang.out
```diff
@@ -0,0 +1,43 @@
+Search results:
```
dwightguth: I would suggest you use regular expressions to capture what you want about this output, rather than enforcing an exact match, which is a very fragile testing mechanism.

dwightguth:
Jenkins: test this please

src/javasources/KTool/src/org/kframework/backend/symbolic/krun/SymbolicKRun.java
andreiarusoaie: You are right, but I wrote it this way because I want to extend it in the near future. For instance, the run() method will call search, then it will filter the 'feasible' configurations (the equivalent of the feasible execution paths) using the SMT solver, and then it will choose one configuration (pseudo-randomly) from the feasible ones and continue the execution. The main reason is that I want to decouple the RewriteEngine from the SMT solver.

If you are not ok with this then I can simply inherit MaudeKRun, but again, this will change in the future.

src/javasources/KTool/src/org/kframework/kompile/KompileOptions.java
andreiarusoaie: Well... This was a source of a NullPointerException and I thought (looking at the other variables) that it is ok to have a default value for symbolicRules variable. There is no reason I have chosen LinkedHashSet().. Probably I can change that into something else.

tests/regression/symbolic-backend/config.xml
andreiarusoaie: I agree with this.

tests/regression/symbolic-backend/config.xml
andreiarusoaie: Agreed.

tests/regression/symbolic-backend/symbolic-kompile/lang.k
andreiarusoaie: You are right. I needed a very simple language, which has IO; this one is IMP+IO. I can probably use IMPPP.

tests/regression/symbolic-backend/symbolic-kompile/tests/sum.lang.out
andreiarusoaie: Can you point to me an existent test which does that? I'm pretty interested in that. Thanks!

andreiarusoaie:
@dwightguth Thanks for your useful feedback. I'll make the changes you requested tomorrow.

dwightguth:
Looks like your definition can't exist side-by-side with the one for regular maude. You will need to set the -d flag on kompile and krun in order to make sure that when you compile imp++ for the symbolic backend you don't overwrite what was done with Maude.

andreiarusoaie:
@dwightguth Good catch. I think I fixed now. Is there anything else I can do before merging this?

dwightguth:
Looks good. I will merge this.

PR #384, opened 2014-04-16
Lowercase copyright

adamsmd:
Changed the top-of-file copyright to use a lower case "(c)" instead of upper case.  This makes it consistent with the LICENSE file and is the way it is most commonly seen.


dwightguth:
I will get to this tomorrow, but just a heads up: you will need to also both update the wiki and email thw k list about the change.

dwightguth:
Oh god, this is going to be a bitch to merge. Can you please fix the merge conflicts, update the remaining files this leaves out from more recent commits and get on this ASAP? In order for this to work I need to halt all merges while this is ongoing.

adamsmd:
Fixed.  (Unless more merges come through.)

PR #385, opened 2014-04-16
use AnsiConsole.systemInstall and remove AnsiConsole.out calls

osa1:
As discussed at #379 , I removed `AnsiConsole.out` calls and instead called `AnsiConsole.systemInstall` at K entry point. This should ensure that colorful printing will work on all platforms without any other changes.(e.g. `System.out....` calls should just handle colorful printing throughout the entire framework)

Note that this patch assumes that K tools are only called using `k3.jar`(`org.kframework.main.Main`) and not by using `main` methods of some other classes(like `krun.main`, `ktest.main` etc.) Calling other `main` methods bypasses `AnsiConsole.systemInstall` call and thus prints color codes even when printing to non-ANSI terminals.

After this patch, `AnsiConsole` will only be used in entry point of the framework:

```
➜  k git:(jansi_refactoring) ack --java AnsiConsole                                              
src/javasources/KTool/src/org/kframework/main/Main.java
8:import org.fusesource.jansi.AnsiConsole;
42:        AnsiConsole.systemInstall();
➜  k git:(jansi_refactoring) 
```

### DIFF
--- a/src/javasources/KTool/src/org/kframework/backend/java/ksimulation/Adjuster.java
+++ b/src/javasources/KTool/src/org/kframework/backend/java/ksimulation/Adjuster.java
@@ -24,7 +24,6 @@ import com.microsoft.z3.Expr;
 import com.microsoft.z3.Solver;
 import com.microsoft.z3.Status;
 import com.microsoft.z3.Z3Exception;
-import org.fusesource.jansi.AnsiConsole;
 
 public class Adjuster {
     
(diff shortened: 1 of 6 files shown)


osa1:
Test failed because of timeout(maybe Jenkins server was busy?). Can anyone retest this sometime later?

dwightguth:
Somehow a bunch of krun processes didn't get killed when one of the builds timed out. I killed them manually, and will start retesting these, and hopefully the problem will have been solved.

dwightguth:
Jenkins: test this please

dwightguth:
lgtm. Merging.

PR #387, opened 2014-04-16
Remove the extra start non-terminal from the parser.

adamsmd:
After reviewing the changes necessary, I am happy to report that it was much simpler than I thought and (unlike I had feared) simply removing the code didn't adverse reactions or change the interface at all.


radumereuta:
lgtm

Issue #369, opened 2014-04-13
Kast --parser binary NullPointerException


chathhorn:
This must be a fairly recent regression (within the last two weeks or so) that breaks the C semantics:

test.k:

```
module TEST
   syntax K ::= "a"
   configuration <k> a </k>
endmodule
```

Now:

```
$ kompile test.k
$ krun --output binary --output-file conf.bin
$ file conf.bin
conf.bin: Java serialization data, version 5
$ kast --parser binary conf.bin
Exception in thread "main" java.lang.NullPointerException
        at org.kframework.compile.sharing.TokenSortCollector.visit(TokenSortCollector.java:63)
        at org.kframework.kil.Production.accept(Production.java:165)
        at org.kframework.kil.visitors.BasicVisitor.visit(BasicVisitor.java:181)
        at org.kframework.kil.PriorityBlock.accept(PriorityBlock.java:69)
        at org.kframework.kil.visitors.BasicVisitor.visit(BasicVisitor.java:151)
        at org.kframework.kil.Syntax.accept(Syntax.java:101)
        at org.kframework.kil.visitors.BasicVisitor.visit(BasicVisitor.java:77)
        at org.kframework.kil.Module.accept(Module.java:97)
        at org.kframework.kil.visitors.BasicVisitor.visit(BasicVisitor.java:53)
        at org.kframework.kil.Definition.accept(Definition.java:105)
        at org.kframework.compile.sharing.TokenSortCollector.collectTokenSorts(TokenSortCollector.java:50)
        at org.kframework.kil.Definition.preprocess(Definition.java:130)
        at org.kframework.kast.KastFrontEnd.kast(KastFrontEnd.java:157)
        at org.kframework.main.Main.main(Main.java:58)
```

I seem to get this same exception no matter what file I pass to "kast --parser binary", including random binaries and the empty file.

chathhorn:
This regression was introduced at revision ff5f0347d4c211df0ed0e7961d6560399ae72f65 (@dwightguth?). I suspect the fix would be pretty easy.

dwightguth:
Sorry about that... if you want to get a head start on whatever you're working on I suggest you check out dwightguth/k@9a641b6. It will be merged into master as soon as it's through code review.

chathhorn:
Thanks, I really appreciate it!

dwightguth:
It is merged now. Please verify and close if you're happy.

PR #388, opened 2014-04-16
fix binary parser for krun

dwightguth:
Fix for #369.

@osa1 please review the ktest changes. @radumereuta please review the fix itself.


src/javasources/KTool/src/org/kframework/ktest/Config/ConfigFileParser.java
```diff
@@ -123,7 +123,7 @@ public ConfigFileParser(File configFile, CmdArg cmdArgs) throws IOException, SAX
      */
     private boolean isValidTestCase(Element testNode) {
         // posixOnly attribute is valid only on Posix-compliant OS
-        return testNode.getAttributes().getNamedItem("posixOnly") == null
+        return testNode.getAttributes().getNamedItem("posixInitScript") == null
```
osa1, line 126: Uh.. I should have fixed this many pull requests ago, how did I miss this? Thanks for the fix!

src/javasources/KTool/src/org/kframework/ktest/Test/TestSuite.java
```diff
@@ -189,6 +189,7 @@ public void dryRun() {
                 Proc<TestCase> p = new Proc<>(tc, tc.getPosixOnlyCmd(), tc.toPosixOnlyLogString(),
                         strComparator, timeout, verbose, colorSetting, updateOut, generateOut);
                 ps.add(p);
+                p.setWorkingDir(tc.getWorkingDir());
```
osa1, line 192: Ahh.. probably another bug. It might be a good idea to add working directory as a parameter to `Proc` constructor so that users would be forced to explicitly set it every time a `Proc` is created, this might help avoid this kind of bug since it would force user to think about process' working directory. I just checked usages of `Proc` class and currently  we're always setting working directory just after the `Proc` object is constructed.

osa1:
`ktest` parts look good to me :+1:

tests/regression/binaryparser/init.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+# Copyright (c) 2014 K Team. All Rights Reserved.
+$kompile test.k
+$krun -cPGM "a" --output binary --output-file 1.test
```
radumereuta, line 4: I should be able to run the tests on windows too.
I don't think sh scripts should be used in testing.

radumereuta:
lgtm

PR #389, opened 2014-04-17
fix race condition in ktest

dwightguth:
This fixes a bug I introduced in #356 where a race condition caused the error messages of failed tests to be incorrect in some cases. I'm not super happy with this fix, but I don't know what else to do. It seems to require a much more deep structural refactoring to get around, namely, some kind of provider class for StringMatchers. Probably Guice is the fix for this?

If someone has a better suggestion in the meantime I'm open to it.


dwightguth:
Never mind, this is the wrong solution. The right thing to do is to make StringMatcher stateless, which means it will be void and throw an exception with the error message. Closing.

PR #390, opened 2014-04-17
fix race condition with StringMatcher

dwightguth:
See #389 for background.

@osa1 please review

### DIFF
--- a/src/javasources/KTool/src/org/kframework/ktest/IgnoringStringMatcher.java
+++ b/src/javasources/KTool/src/org/kframework/ktest/IgnoringStringMatcher.java
@@ -5,9 +5,6 @@ public class IgnoringStringMatcher implements StringMatcher {
 
     private final boolean ignoreWS;
     private final boolean ignoreBalancedParens;
-
-    private String lastPattern;
-    private String lastActual;
     
     public IgnoringStringMatcher(boolean ignoreWS, boolean ignoreBalancedParens) {
         this.ignoreWS = ignoreWS;
@@ -15,10 +12,7 @@ public class IgnoringStringMatcher implements StringMatcher {
     }
 
     @Override
-    public boolean matches(String pattern, String actual) {
-        lastPattern = pattern;
-        lastActual = actual;
-        
+    public void matches(String pattern, String actual) throws MatchFailure {
         // algorithm copied from old ktest; I'm not convinced that they work correctly in all cases
         if (ignoreWS) {
             pattern = pattern.replaceAll("\\r|\\s|\\n","");
@@ -33,7 +27,10 @@ public class IgnoringStringMatcher implements StringMatcher {
             pattern = removeAllBalanced(pattern);
             actual = removeAllBalanced(actual);
         }
-        return pattern.trim().compareTo(actual.trim()) == 0;
+        if (pattern.trim().compareTo(actual.trim()) != 0) {
+            throw new StringMatcher.MatchFailure(
+                    String.format("Expected:%n%s%n%nbut found:%n%s%n", pattern, actual));
+        }
     }
 
     private static String removeAllBalanced(String s1) {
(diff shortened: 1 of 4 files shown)


src/javasources/KTool/src/org/kframework/ktest/IgnoringStringMatcher.java
```diff
@@ -33,7 +27,10 @@ public boolean matches(String pattern, String actual) {
             pattern = removeAllBalanced(pattern);
             actual = removeAllBalanced(actual);
         }
-        return pattern.trim().compareTo(actual.trim()) == 0;
+        if (pattern.trim().compareTo(actual.trim()) != 0) {
+            throw new StringMatcher.MatchFailure(
+                    String.format("Expected:%n%s%n%nbut found:%n%s%n", pattern, actual));
```
osa1, line 32: I just realized that in case of a comparison failure, `expected` and `found` strings here will be stripped versions of parameters(e.g. parens and whitespace will be removed). Is this really what we want to print to the user?

It was not like this before, original strings were printed to the user.(because comparison function was returning boolean value and printing was handled by the caller)

src/javasources/KTool/src/org/kframework/ktest/RegexStringMatcher.java
```diff
@@ -5,29 +5,15 @@
...
         for (String pattern : patterns.split("\n")) {
             Matcher m = Pattern.compile(pattern).matcher(actual);
-            matches &= m.find();
-            if (!matches) {
-                failedRegex = pattern;
-                failedActual = actual;
-                break;
+            if (!m.find()) {
+                throw new MatchFailure(String.format("Expected result to match regular expression:" +
+                        "%n%s%n%nbut found:%n%s%n", pattern, actual));
```
osa1, line 15: We're removing parens in string comparison function, why not do the same here? IIRC the motivation behind this was to keep tests passing when pretty-printer gets smarter and puts less number of parenthesis, but regex comparisons will fail if that happens one day.

osa1:
Apparently I need to be more careful with reviews. This patch looks good but I may have found some bugs in previous patches. I opened issue #391 for bugs related with output file generation patch(9bdc321 I'm not sure if that patch is reviewed though, it may be happened before code reviews started) and added two comments above which are not related with this patch(they were introduced before this patch)

Merge this if you want and we can discuss that in it's own issue.

src/javasources/KTool/src/org/kframework/ktest/Proc.java
```diff
@@ -257,23 +258,26 @@ private void handlePgmResult(int returnCode) throws IOException {
...
+                            "%sERROR: [%s] output doesn't match with expected output (time: %d ms)%s%n",
+                            red, logStr, timeDelta, ColorUtil.ANSI_NORMAL);
+                    reportOutMatch(e.getMessage());
+                    if (verbose)
+                        System.out.println(getReason());
+                    doGenerateOut = true;
+                }
             }
-
+
```
osa1, line 280: I'm not sure if we are _that_ strict about commits, but this line is actually just extra spaces.

dwightguth:
I merged this. This change by itself unblocks me, and I'm not the primary owner of ktest. If you want me to fix the concerns you raised myself, I can probably do that (since I introduced them), but I have so much on my plate I might not get to it right away because it's lower priority for me.

The generate-out thing predates code reviews which began in February.

osa1:
Okay, I'll send a pull request for error reporting and we're discussing `generate-out` handling at #391 .

As for parenthesis handling, we may need to discuss this on mailing list because with presence or regex this will be trickier to do right. (or maybe we can left this as is and just assume that regex writer handles parens)

dwightguth:
Personally I would side with the latter property: if I specify a regular
expression, it should match exactly what it matches and no more and no
less. Doing something more complex decreases the simplicity of the design
and therefore makes it harder to use. Java's regular expression language is
quite versatile and supports a lot of functionality. For example I made use
of backreferences and lookahead in the examples I wrote. The examples I
modified were also all search examples, and the nature of search as it's
implemented currently is that the results will change their ordering
depending on things like brackets, so a change to the pretty printer would
have broken those tests anyway. In fact that actually happened to me a
while back and I caught a regression that way. So I think it's good the way
it is. But we can discuss more if you want.

On Thu, Apr 17, 2014 at 4:45 PM, Ömer Sinan Ağacan <notifications@github.com

PR #393, opened 2014-04-17
fix another kast regression and add test

dwightguth:
:/ Added another regression test because apparently because of the nasty weird code we have in krun (my fault, sorry!), we have no kast coverage at all. Whoops!

I'll add some more tests tomorrow just to be on the safe side, but this should unblock the regression, if we're lucky.


dwightguth:
@radumereuta please review ASAP so we can merge this. I want to have this regression fixed.

radumereuta:
lgtm
If you make the tests pass, you can merge.

PR #395, opened 2014-04-21
refactor the way the application exits and exceptions are caught

dwightguth:
I am still in the process of testing this, so it may not pass all tests yet. When it is ready for review, I will let you know.

To summarize the change:
- The new and only way to terminate the application prematurely is to call GlobalSettings.kem.register. This throws a KEMException which is caught by org.kframework.main.Main.
- org.kframework.main.Main is the only class allowed to call System.exit
- If you spawn threads, you are responsible for joining on them so that the application only reaches the termination of the main thread when the application is ready to exit.
- In order for this to work, I removed all instances in the codebase where RuntimeException, Exception, Error, or Throwable are caught (except org.kframework.backend.kil.KItem.java). I also removed all instances in the codebase where Exception or Throwable were explicitly thrown.
  *\* I added a checkstyle check to ensure that this property is not violated.
  *\* It is now an antipattern to catch an exception, print a stack trace, and do nothing. The correct course of action is to print a stack trace if GlobalOptions.debug is set in the current option set, and register a KException if the error is unrecoverable.


dwightguth:
Okay, this is ready for review.

@yilongli please review.

@andreistefanescu @radumereuta @andreiarusoaie @osa1 @traiansf please provide sanity checks on any refactorings I made to the way exceptions are handled in your code.

@dlucanu please test the graphical debugger to see if it works now. I didn't see any problems but you would know better what to test for. Please also test to see whether the issue you had with java.awt.headless is still present. If it is, I will probably need to take further action on this diff to take care of that.

src/javasources/KTool/src/org/kframework/krun/Main.java
```diff
@@ -576,20 +593,40 @@ public static void normalExecution(Term KAST, String lang, RunProcess rp,
...
     // isSwitch variable is true if we enter in debug execution from normal
     // execution (we use the search command with --graph option)
+    /**
+     * 
+     * @param kast
+     * @param lang
+     * @param state
+     * @param context
+     * @return true if the application completed normally; false otherwise
+     */
```
radumereuta, line 612: Finish this Javadoc?

src/javasources/KTool/src/org/kframework/parser/DefinitionLoader.java
```diff
@@ -102,7 +102,7 @@ public static Definition parseDefinition(File mainFile, String mainModule, boole
 
             // transfer information from the BasicParser object, to the Definition object
             org.kframework.kil.Definition def = new org.kframework.kil.Definition();
-            def.setMainFile(mainFile.getCanonicalPath());
+            def.setMainFile(mainFile.getAbsolutePath());
```
radumereuta: Why?
I was asking for the CanonicalPath because I wanted the system to resolve the ".." from the filename. This is used many times in error reporting. The path to the files must be shown to the user in the most simple way.

src/javasources/KTool/src/org/kframework/parser/DefinitionLoader.java
```diff
@@ -124,9 +124,9 @@ public static Definition parseDefinition(File mainFile, String mainModule, boole
             Stopwatch.instance().printIntermediate("Basic Parsing");
 
             //This following line was commented out to make the latex backend 
-            //parse files importing from other files
+            //parse files importing from other files
             def = (Definition) def.accept(new RemoveUnusedModules(context, autoinclude));
-
+
```
radumereuta: Extra whitespace.

src/javasources/KTool/src/org/kframework/parser/generator/ParseRulesFilter.java
```diff
@@ -175,10 +175,6 @@ else if (ss.getType().equals(Constants.RULE))
                 return config;
             } catch (TransformerException te) {
                 te.printStackTrace();
-            } catch (Exception | Error e) {
-                e.printStackTrace();
-                String msg = "Cannot parse sentence: " + e.getLocalizedMessage();
-                GlobalSettings.kem.register(new KException(ExceptionType.ERROR, KExceptionGroup.CRITICAL, msg, ss.getFilename(), ss.getLocation()));
```
radumereuta, line 181: I'm not sure about removing this block. It provides useful information sometimes.
If SDF crashed, it would be caught here and an error message shown.
Will your modifications handle such cases?

src/javasources/KTool/src/org/kframework/utils/errorsystem/KExceptionManager.java
```diff
@@ -20,20 +20,23 @@ public KExceptionManager(GlobalOptions options) {
...
-            if (e.type == ExceptionType.ERROR)
-                errors = true;
             System.err.println(StringUtil.splitLines(e.toString()));
         }
-        if (errors)
-            System.exit(1);
+    }
+    
+    /**
+     * Thrown to indicate that the K Exception manager has terminated the applciation due to an error.
```
radumereuta: Typo?

radumereuta:
Also, go into org.kframework.parser.ProgramLoader
and change the lines with if (context.globalOptions.verbose)
to use the new debug option.

src/javasources/KTool/src/org/kframework/parser/DefinitionLoader.java
dwightguth: .. is a relative path. getAbsolutePath resolves relative paths. The reason
getCanonicalPath throws IOException and getAbsolutePath does not is because
getCanonicalPath resolves symlinks, which we don't want to do.

src/javasources/KTool/src/org/kframework/parser/generator/ParseRulesFilter.java
dwightguth, line 181: If sdf crashes, you will get a stack trace, which is what you want when a
bug happens. Your code here even catches Error which is clearly wrong
because Errors are not catchable: the jvm throws them when it's no longer
safe to execute /anything/.

src/javasources/KTool/src/org/kframework/parser/DefinitionLoader.java
radumereuta: File f = new File("..");
System.out.println(f.getAbsolutePath());
System.out.println(f.getCanonicalPath());

output:
D:\work\k\samples\test..
D:\work\k\samples

So absolute path doesn't resolve relative paths, but canonical does:
"A canonical pathname is both absolute and unique.... This typically involves removing redundant names such as "." and ".." from the pathname, resolving symbolic links (on UNIX platforms), and converting drive letters to a standard case (on Microsoft Windows platforms)."

I want the ".." and "." to be eliminated. those are hard to follow or easy to overlook when viewing error messages.
Regarding symlinks, well... I agree... you may not want those to be resolved. Suggestions?

src/javasources/KTool/src/org/kframework/parser/generator/ParseRulesFilter.java
radumereuta, line 181: Ok, I'll see what happens. I can't think of a way to test it right now.

src/javasources/KTool/src/org/kframework/parser/DefinitionLoader.java
dwightguth: It looks like File.getAbsolutePath doesn't automatically resolve based on
the current working directory. I did not know that. Using getAbsolutePath
is still capable of being correct, you just apparently have to specify what
the path is relative to first. I will take action on this when I resolve
the other issues, and write a unit test for this. We'll see what I have to
do to make it work.

radumereuta: Ok, tell me to review it again when you finish.

andreistefanescu:
The java backend code looks good.

yilongli:
The general mechanism looks good to me. 
@owolabileg The BTrace script used to profile krun may need some slight change though. We will modify it later.

src/javasources/KTool/src/org/kframework/parser/DefinitionLoader.java
dwightguth: Turns out I misunderstood the difference between these two methods. I am just going to revert this change and add a catch block.

dwightguth:
@radumereuta I took care of your suggestions. Please re-review.

radumereuta:
lgtm

src/javasources/KTool/src/org/kframework/krun/ioserver/main/IOServer.java
```diff
@@ -170,17 +174,17 @@ public Command createCommand(String[] args, Socket socket, Logger logger) {
...
             } catch (InvocationTargetException e) {
-                return new CommandUnknown(args, socket, logger, fs);
+                Throwable t = e.getTargetException();
+                if (t instanceof Error) {
+                    throw (Error)t;
+                } else if (t instanceof RuntimeException) {
+                    throw (RuntimeException)t;
+                } else {
+                    throw new AssertionError("should not throw checked exceptions from unreachable code", e);
+                }
```
andreiarusoaie: Can you explain how this message is useful for an user?

src/javasources/KTool/src/org/kframework/parser/basic/Basic.jj
```diff
@@ -396,7 +397,7 @@ public class Basic {
     SimpleCharStream s = new SimpleCharStream(new StringReader(str), line, column);
     try {
       for (int i = 0; i <= index; i++) { s.readChar(); }
-    } catch (IOException e) { /* Impossible */ e.printStackTrace(); System.exit(1); }
+    } catch (IOException e) { throw new AssertionError("did not expect IOException", e); }
```
andreiarusoaie, line 400: The message confuses me: "did not expect IOException"... What if it happens?

src/javasources/KTool/src/org/kframework/krun/ioserver/main/IOServer.java
dwightguth: I just remembered this whole code block existed because we wanted to support Java 6, which we don't actually support any longer. So I am removing it, and the need to address this goes away.

src/javasources/KTool/src/org/kframework/parser/basic/Basic.jj
dwightguth, line 400: What will happen is the same thing that happened before, a stack trace followed by the application exiting. If you think that's a bad solution, you can discuss changing it, but that's not my responsibility as part of this refactoring.

andreiarusoaie, line 400: You are right. This is not part of this refactoring.

andreiarusoaie:
I don't have any other comments.

dwightguth:
@osa1 @traiansf @dlucanu please review per my request for comment above.

traiansf:
Looks good. Thanks Dwight for doing this.  It was well needed.

dwightguth:
Jenkins: retest this please

traiansf:
looks good to me

2014-04-30 20:33 GMT+03:00 dwightguth notifications@github.com:

dwightguth:
@osa1 since the changes to ktest were minimal and straightforward, I'm not going to wait on your review for them, so if you want to comment, you should do that in the next couple days before I merged this.

@dlucanu Since the graphical debugger was broken before and is now at the very least mostly not broken, I am not going to wait on your review for that either. When you have time, please check it out and if you run into issues let me know and I will address them if they are related to this change.

If nobody else has feedback I will probably merge this tomorrow afternoon.

osa1:
@dwightguth ktest part looks good but I was reading other parts too and since this patch is huge it was taking time. Don't wait for me, I can open an issue if something gets caught to my eye.

src/javasources/KTool/src/org/kframework/kcheck/KCheckFrontEnd.java
```diff
@@ -121,7 +128,6 @@ public static void verbose(CommandLine cmd, Context context) {
         //if (GlobalSettings.verbose) {
             context.printStatistics();
         //}
```
radumereuta, line 130: why still have these commented lines?

dwightguth, line 130: Why still have kcheck at all? It doesn't work anymore. I just didn't delete
the whole thing because I wasn't sure if Andrei A wanted any of the code.
But the code path is deprecated, so when my change broke this code, I just
commented it out rather than fixing it.

andreiarusoaie, line 130: I will soon transform kcheck into an option for krun, namely --prove for the symbolic backend.
However, I still need some code from it (thank you Dwight for not deleting it).  So, I'm proposing you guys to just ignore the code that belongs to kcheck when reviewing the code.

PR #397, opened 2014-04-22
Added the definition of #String directly into K. I also had to fix some ...

radumereuta:
...stuff in order for it to work:
- renamed escapeK into enquuoteString for better naming
- fixed a bug in KSyntax2Grammar generation
- added ktests for the new parser.


src/javasources/KTool/src/org/kframework/kil/Production.java
```diff
@@ -87,9 +87,16 @@ public Lexical getLexical() {
     }
 
     public boolean isConstant() {
+        // TODO(Radu): properly determine if a production is a constant or not, just like below
```
dwightguth, line 90: Where all is this called from? Can we remove calls to this method wherever possible? Where can't we remove calls to it from?

src/javasources/KTool/src/org/kframework/kil/Token.java
```diff
@@ -50,7 +50,7 @@ public static KApp kAppOf(String sort, String value) {
         } else if (sort.equals(IntBuiltin.SORT_NAME)) {
             return IntBuiltin.kAppOf(value);
         } else if (sort.equals(StringBuiltin.SORT_NAME)) {
-            return StringBuiltin.kAppOf(value);
+            return StringBuiltin.kAppOf(StringUtil.unquoteString(value));
```
dwightguth, line 54: good catch. @andreistefanescu , this was a bug before, correct?

src/javasources/KTool/src/org/kframework/parser/generator/ProgramSDF.java
```diff
@@ -19,12 +19,13 @@
 import org.kframework.parser.concrete2.KSyntax2GrammarStatesFilter;
 import org.kframework.utils.BinaryLoader;
 import org.kframework.utils.StringUtil;
+import org.kframework.utils.general.GlobalSettings;
 
 /**
  * Collect the syntax module, call the syntax collector and print SDF for programs.
- *
+ *
```
dwightguth: whitespace!

src/javasources/KTool/src/org/kframework/parser/generator/ProgramSDF.java
```diff
@@ -147,7 +148,7 @@ public static StringBuilder getSdfForPrograms(Definition def, Context context) {
 
         for (String t : ctv.terminals) {
             if (t.matches("[a-zA-Z\\_][a-zA-Z0-9\\_]*")) {
-                sdf.append("    \"" + StringUtil.escape(t) + "\" -> DzDzID {reject}\n");
+                sdf.append("    \"" + StringUtil.escape(t) + "\" -> IdDz {reject}\n");
```
dwightguth, line 148: What happened here?

dwightguth:
looks decent, but you need to fix copyright before I know if it's breaking any of the tests.

src/javasources/KTool/src/org/kframework/kil/Production.java
radumereuta, line 90: This is related to some other idea I have about the Context object. We should discuss it sometime.
Basically I would like to have ASTNodes which are independent from everything else, and the Context gives them meaning.
This would help modularity in the future.
For example: a production of the form `X ::= "terminal"`
is considered a constant, only if there is another declaration of the form `X ::= Token{...}`

The same for sort: term.getSort(context).

Anyway. Probably I can make more sense face to face.

src/javasources/KTool/src/org/kframework/kil/Token.java
radumereuta, line 54: Actually, I consider this to be a hack.
I think we shouldn't unquote strings when we represent them in the memory.
I'm saying this because Daejun had some problems in the javascript semantics when we didn't handle properly some special characters.

For example, if you write in a program "a\bc" after it goes through kompile and krun, you will get the result "abc" (notice the backslash disappeared). Now that's a bug, we need to handle it somehow. But not in this commit. I just followed the convention we currently have in this case.

src/javasources/KTool/src/org/kframework/parser/generator/ProgramSDF.java
radumereuta, line 148: I found a bug. Here I'm testing to see what keywords do I have to reject from the Id category.
Id was moved from K into SDF a long time ago, and I forgot to relink the sorts.

Now, this is also limitation of how we dealt with SDF, because it is hard coded. When I move everything into K, I could test every single token definition against every terminal and optimize the work of the parser. Right now I'm generating all possible combinations of "terminal" -> LexicalSort {reject}.

src/javasources/KTool/src/org/kframework/kil/Token.java
andreistefanescu, line 54: The escaping of strings is a mess. We need to fix it.

dwightguth, line 54: It's not a bug though. It's part of the current formal semantics of strings
in K. K does not have the \b escape. If you want to represent the string
"a\bc" in K, you should write "a\x08c". They correspond to the same exact
sequence of unicode code points, they just obey different conventions for
expressing and escaping strings.

If you want the \b escape in your language, you should manually construct a
MyString token sort with exactly the regular expression you want. If on the
other hand you want to have "strings" be completely unprocessed in K, you
can also do that, but then things like substring and find will not work at
all how you expect them to.

The String sort is designed to represent the mathematical domain of unicode
character strings: it is not designed to represent the syntax of Strings in
any particular language other than K.

src/javasources/KTool/src/org/kframework/kil/Production.java
dwightguth, line 90: I still want to see this method removed if it's not functionally correct.

dwightguth:
looks good, just that one minor point I just commented on.

src/javasources/KTool/src/org/kframework/kil/Token.java
radumereuta, line 54: Then I should restrict the String lexicals to only parse things that are supported by the backends.
This way they get a nice error message from the parser, instead getting bugs at runtime like Daejun had a few weeks ago.
I was thinking of the following regex:
Token{"\"" ((~[\"\n\r\])|([\][nrtfxuU\"]))\* "\""}

Anyone against this?

src/javasources/KTool/src/org/kframework/kil/Production.java
radumereuta, line 90: I agree, but I don't think this is the right pull request to remove it.
It might break other things because it is used in many places.
And before that, I think we should discuss about doing the same thing to `Term.getSort`

dwightguth, line 90: Okay, that rationale for making a separate change makes sense to me. Can
you make an issue?

Is it actually necessary for Term.getSort though? All concrete syntax is
associated with a production, and therefore with a sort.

radumereuta, line 90: There is the corner case where Rewrite is actually a computed sort.
Plus, working on modularity would be easier, since I would have to transform only one object, the Context.

src/javasources/KTool/src/org/kframework/kil/Token.java
dwightguth, line 54: The regex is actually more complex than that. \x must be followed by 2 hex
digits, \u by 4 and \U by 8. But I agree in principle that restricting the
syntax of strings is the right thing to do.

src/javasources/KTool/src/org/kframework/kil/Production.java
dwightguth, line 90: The idea would be that Terms would not have context but operations on terms
would? Okay, I think I agree. That is something similar to what javac does.
The idea would be that the only information stored in the Term itself would
be information which is directly computable from the term in question. Does
that make sense?

radumereuta, line 90: Yes, that's the main idea.
I'm trying to make Terms as independent as possible, and only the Context you provide to the transformation gives them meaning.
I remember a horrible experience with DOM, where each element has access to the containing Document object. Because of that, every insertion or adoption of nodes is two times as hard, because you need to handle this pointer. It would be best if we avoid such high maintenance systems in K, because at the moment, they don't bring anything useful.

src/javasources/KTool/src/org/kframework/kil/Token.java
radumereuta, line 54: Should I do this right now? I think I should.

radumereuta:
I think this is ready to merge.
Did I miss any comment?

dwightguth:
Looks good. You double checked to make sure that everywhere where we are unquoting strings from user input for the first time that we are catching IllegalArgumentException and throwing a better error message? If so, go ahead and merge.

dwightguth:
Actually, I take it back, can you do one more thing? Can you add a set of unit tests for unquoteString and quoteString?

radumereuta:
I think now it's ready to be merged.

src/javasources/KTool/test/org/kframework/utils/UtilsTest.java
```diff
@@ -0,0 +1,33 @@
...
+package org.kframework.utils;
+
+import junit.framework.Assert;
+import org.junit.Test;
+
+public class UtilsTest {
+
+    @Test
+    public void StringUtilUnquote() throws Exception {
+        Assert.assertTrue(StringUtil.unquoteString("\"\n\"").equals("\n"));
```
dwightguth: This isn't quite the test you wanted, I think. Shouldn't it be "\"\n\"" ?

src/javasources/KTool/test/org/kframework/utils/UtilsTest.java
```diff
@@ -0,0 +1,33 @@
...
+package org.kframework.utils;
+
+import junit.framework.Assert;
+import org.junit.Test;
+
+public class UtilsTest {
+
+    @Test
+    public void StringUtilUnquote() throws Exception {
+        Assert.assertTrue(StringUtil.unquoteString("\"\\n\"").equals("\n"));
```
radumereuta, line 11: Right, this is what I actually meant.
Parsing the string in the right, would generate into memory the content of the strings on the left.
The internal representation that I'm aiming for is the new line character alone.

dwightguth:
lgtm. You can merge when the tests pass.

Issue #396, opened 2014-04-22
ktest can't run kast instead of krun


radumereuta:
I want to add tests for the new parser, and I need to run `kast` for specific programs, and krun for others.
At the moment I can only specify `krun` but I need to compare the output of the parsers directly.

daejunpark:
This issue will be further discussed the wiki page of ktest design:
https://github.com/kframework/k/wiki/KTest-Design

Issue #392, opened 2014-04-17
Allow ktest users to specify a phase topology


dwightguth:
One of the examples that came out of #369 was to run krun, and then run kast or krun on the output of running krun. Our test framework does not currently support this, or indeed any more complex workflows than:
1. kompile all definitions
2. run all programs for each definition one at a time.

It would be good to be able to have some kind of generalization of the phase mechanism which would allow you to specify building blocks like "kompile a definition", "test this program", "test all programs matching <blah>" etc and compose them together into larger test cases.

I am creating this issue to track the possibility of adding this feature. Note that once we have something that does this it will be necessary to delete the test created in #388 and replace it with a phase descriptor.

daejunpark:
This issue will be further discussed the wiki page of ktest design:
https://github.com/kframework/k/wiki/KTest-Design

PR #399, opened 2014-04-25
KTEST: Support alternative krun executable. Issue #396

daejunpark:
An alternative krun executable can be provided by using the additional `<krun-exec>` node. This option is useful especially for running `kast` instead of `krun` for certain programs. (See Issue #396.)

@dwightguth @osa1 Please review this code.
@radumereuta Please check if this change is sufficient for you.

CHANGES:
- Create additional `<krun-exec>` node (optional) in both `<all-programs>` and `<program>` nodes.
- Allow multiple programs in the `name` attributes of `<program>` node.

USAGE:

```
  <test
    definition="lang.k"
    programs="programs"
    results="results"
    extension="lang"
  >
    <all-programs>
      <krun-option name="--color" value="off" />
      <krun-option name="--output" value="none" />
    </all-programs>
    <program name="pgm1.lang pgm2.lang">
      <krun-exec name="kast" />
    </program>
  </test>
```

LIMITATIONS and CONCERNS:
- The name of the additional node `krun-exec` might be inappropriate.
- Currently, an alternative krun executable should reside in `KHOME/bin` directory as `kompile`, `krun`, and `kast`.
- It would be better to come up with a more general solution that also solves the issue #392.

### DIFF
--- a/src/javasources/KTool/src/org/kframework/ktest/Config/ConfigFileParser.java
+++ b/src/javasources/KTool/src/org/kframework/ktest/Config/ConfigFileParser.java
@@ -430,10 +430,13 @@ public class ConfigFileParser {
             if (childNode.getNodeType() == Node.ELEMENT_NODE
                     && childNode.getNodeName().equals("program")) {
                 Element elem = (Element) childNode;
-                ret.put(elem.getAttribute("name"), new ProgramProfile(parseKrunOpts(
-                        elem.getChildNodes()),
-                        Boolean.parseBoolean(elem.getAttribute("regex")),
-                        parseKrunExec(elem.getChildNodes())));
+                String[] names = splitNodeValue(elem.getAttributeNode("name"));
+                for (String name : names) {
+                    ret.put(name,
+                            new ProgramProfile(parseKrunOpts(elem.getChildNodes()),
+                                    Boolean.parseBoolean(elem.getAttribute("regex")),
+                                    parseKrunExec(elem.getChildNodes())));
+                }
             }
         }
         return ret;
--- a/src/javasources/KTool/src/org/kframework/ktest/Config/ConfigFileParser.java
+++ b/src/javasources/KTool/src/org/kframework/ktest/Config/ConfigFileParser.java
@@ -363,15 +363,17 @@ public class ConfigFileParser {
     private ProgramProfile parseAllPgmsKrunOpts(NodeList nodes) {
         List<PgmArg> ret = new LinkedList<>();
         boolean regex = false;
+        String krunExec = "krun";
         for (int childNodeIdx = 0; childNodeIdx < nodes.getLength(); childNodeIdx++) {
             Node childNode = nodes.item(childNodeIdx);
             if (childNode.getNodeType() == Node.ELEMENT_NODE
                     && childNode.getNodeName().equals("all-programs")) {
                 ret.addAll(parseKrunOpts(childNode.getChildNodes()));
                 regex = Boolean.parseBoolean(((Element)childNode).getAttribute("regex"));
+                krunExec = parseKrunExec(childNode.getChildNodes());
             }
         }
-        return new ProgramProfile(ret, regex);
+        return new ProgramProfile(ret, regex, krunExec);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/javasources/KTool/src/org/kframework/ktest/Config/ConfigFileParser.java
```diff
@@ -409,9 +430,13 @@ private ProgramProfile parseAllPgmsKrunOpts(NodeList nodes) {
             if (childNode.getNodeType() == Node.ELEMENT_NODE
                     && childNode.getNodeName().equals("program")) {
                 Element elem = (Element) childNode;
-                ret.put(elem.getAttribute("name"), new ProgramProfile(parseKrunOpts(
-                        elem.getChildNodes()),
-                        Boolean.parseBoolean(elem.getAttribute("regex"))));
+                String[] names = splitNodeValue(elem.getAttributeNode("name"));
+                for (String name : names) {
```
osa1, line 434: I don't understand this part. Is this related with this patch or is this a bugfix unrelated with this patch?

(also, instead of parsing elements for every `name` in `names`, we can parse them one time then add names to `ret` using that parsed values)

dwightguth:
I don't think this is the right design. I think you should create a "kast" phase that works as follows:
- Executes after "pdf" but before "krun"
- <all-programs> and <program> take a "parser" attribute. This attribute, if present, is automatically passed to krun's --parser flag.
- By default, the kast phase calls kast on all programs and compares output on stdout to "<pgm>.kast". If the parser field is set, that parser is used instead.
- the <test> node takes

dwightguth:
:/ hit the button too soon
- the <test> node takes a "exclude-krun" and "exclude-kast" field. The semantics of the two are that a program is excluded from each phase respectively. The existing exclude excludes both.

I am hesitant to merge this with this design, so I suggest we agree on a replacement and then implement it in place of this. But when the time comes, you are free to implement this a piece at a time in order to unblock Radu more quickly, and to use whatever tricks are needed to make the tests pass in the meantime.

If you like, we can discuss my proposed design on Monday in the meeting. Thoughts?

radumereuta:
I'm ok with waiting for a better solution.
Frankly, I would like more control from ktest. I would like to create a task and say exactly what to run over what programs. I would like to say something <krun programsDir="programs/**/*.imp" output="tests/output/*" />
For a specific definition, I want to run these programs with krun with the output being somewhere else.
The way I see it would work best would be something like this:

```
<task>
   <kompile options="-v imp.k" failTaskOnError="true" />
   <kompile options="-pdf imp.k" />
   <krun failTaskOnError="false">
       <fileset name="programs/**/*.imp" />
   </krun>
   <kast options="-parser newprogram">
       <fileset name="programs/**/*.imp" />
   </kast>
</task>
```

andreiarusoaie:
I  like Radu's proposal. It allows us to test different backends too, which would be quite useful.

osa1:
I'm not an active ktest user so there's high probability of me missing/forgetting the point but have you guys considered writing unit tests instead? Maybe we can implement an API for writing unit tests, which provides functions for calling different tools in the framework, comparing output, printing statistics etc.

I think this provides maximum amount of flexibility to the ktest user. For example in this case, Radu could fall-back to Java API to run whatever executable he wants, and use our API to compare outputs like current ktest does and again using our API register the results so that statistics would be printed as before. This speeds up the development of new features. Once we make sure that a new addition to the API is necessary(e.g. some functionality is used again and again by test writers) we can implement that but nobody would have to wait for someone to extend ktest.

Some K users probably don't want to write Java for tests, and for that cases we can always keep current ktest implementation, but as a front-end(i.e. config file parser) of ktest API. So ktest executable still works as before, but for maximum flexibility you could use ktest API instead of ktest executable.

What do you guys think about this?

dwightguth:
Closing per our discussion on Monday. @denis-bogdanas will investigate the possibility of integration with ant once he has time, and then we will rediscuss how to redesign ktest.

osa1:
What did you discuss?

dwightguth:
We decided to investigate the possibility of making a ktest ant task which
configures this information.

PR #406, opened 2014-04-30
Work on priorities for the new parser:

radumereuta:
```
- uses the offline visitor to filter out illegal parse trees.
- uses the old inefficient way to calculate the reachability between two productions
           (TODO: implement an efficient graph reachability algorithm)
- updated the examples and added them to ktest
- TODO: make the same for left and right associativity
```


adamsmd:
The changes to PriorityFilter and PriorityFilter2 are there because we are using KApp instead of TermCons, right?

radumereuta:
Yes. It's still kind of ugly, but a lot of code from there is because we don't have unique labels, and the old parser doesn't guarantee that under a KApp I will always find a KList object.
Hopefully we can generate more uniform terms.

dwightguth:
I'm not satisfied with the way this code treats the distinction between TermCons and KApp. I think we need to do one of two things:
1. Make the parser kernel return TermCons, and perform flattening of terms only after all disambiguation is complete. This would have the advantage that it would require minimal changes to the disambiguation filters in order to make them work, however, those filters are eventually being set aside, and it would also serve to reinforce the non-one-to-one mapping between productions and KLabels. But it would also give us the flexibility to write the grammar exactly how we want it to be, while separating out the concern of how to turn a term into an AST.
2. Rigorously enforce a more rigid structure with respect to KLabels. All disambiguation would be done on KApps, which means that it would be impossible to express a disambiguation rule which applies to one production with a particular KLabel but not the others. We would be able to demand that all KApps of a particular label had the arity of that production, and that all KItems have as a child a KList. Then you would never ask for the Production while performing disambiguation, because all the information needed to perform disambiguation would be associated with the KApp itself.

I will have more detailed comments regarding the code as it is currently written to follow.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PriorityFilter.java
```diff
@@ -18,16 +26,18 @@ public PriorityFilter(Context context) {
...
         if (tc.getProduction().isListDecl()) {
-            tc.getContents().set(0, (Term) tc.getContents().get(0).accept(new PriorityFilter2(tc, context)));
-            tc.getContents().set(1, (Term) tc.getContents().get(1).accept(new PriorityFilter2(tc, context)));
+            tc.getContents().set(0, (Term) tc.getContents().get(0).accept(new PriorityFilter2(label, Side.LEFT, context)));
+            tc.getContents().set(1, (Term) tc.getContents().get(1).accept(new PriorityFilter2(label, Side.RIGHT, context)));
         } else if (!tc.getProduction().isConstant() && !tc.getProduction().isSubsort()) {
             for (int i = 0, j = 0; i < tc.getProduction().getItems().size(); i++) {
                 if (tc.getProduction().getItems().get(i) instanceof Sort) {
                     // look for the outermost element
                     if ((i == 0 || i == tc.getProduction().getItems().size() - 1)
```
dwightguth, line 37: you are not safe to only look at the outermost nonterminals in the production. Consider the following production:

syntax Foo ::= Exps Int "+" Int

In the case where the Exps is epsilon, "5 + 6" parses as a Foo. Then 5 is on the outside of the parse and needs to be subject to priorities.

If you're interested in the formal theory of precisely when it is necessary to perform this filter, I suggest you look at Maude's gather attribute, and when a nonterminal in a production is by default set to gather "&", which signifies that priority is irrelevant. Maude is not perfect in the regard of parsing and pretty printing, but they do get priority and associativity /exactly/ right when most parser implementations do not, in the full generality.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PriorityFilter.java
```diff
@@ -36,4 +46,37 @@ public ASTNode transform(TermCons tc) throws TransformerException {
...
+
+    @Override
+    public ASTNode transform(KApp kapp) throws TransformerException {
+        // get the production
+        // look only for the outermost elements
+        if (kapp.getLabel() instanceof KLabelConstant) {
+            KLabelConstant kls = (KLabelConstant) kapp.getLabel();
+            String label = kls.getLabel();
+            List<Production> prods = context.productionsOf(label);
+            if (prods.size() > 0) {
```
dwightguth, line 58: see my more general comments on the thread itself.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PriorityFilter.java
```diff
@@ -36,4 +46,37 @@ public ASTNode transform(TermCons tc) throws TransformerException {
...
+                // there are terms declared using only the klabel form
+                // in that case don't do any priority disambiguation
+                // TODO: check to see if the returned productions are overloaded (if more than one)
+                Production prod = prods.get(0);
```
dwightguth, line 62: This is not a safe solution at all, nor can I countenance a TODO here. Any implementation that we choose to accept here must satisfy the following property: that the behavior of this filter is well defined according only to the information the user specifies about priorities, and that if the user specifies conflicting information about priorioties that the implementation cannot handle gracefully, an error is reported.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PriorityFilter.java
```diff
@@ -36,4 +46,37 @@ public ASTNode transform(TermCons tc) throws TransformerException {
...
+                if (kapp.getChild() instanceof  KList) {
```
dwightguth, line 63: Can you provide me an example term where this check returns false?

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PriorityFilter.java
```diff
@@ -36,4 +46,37 @@ public ASTNode transform(TermCons tc) throws TransformerException {
...
+                    KList klist = (KList) kapp.getChild();
+                    assert klist.getContents().size() == prod.getArity() : "Production arity and KApp don't match";
```
dwightguth, line 65: If we choose to go with KLabels the way you do here, this can't be an assert. A user can do this mistakenly, and it needs to have a friendly error message. Of course, if we go with concrete syntax for this filter, then it no longer becomes relevant.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PriorityFilter2.java
```diff
@@ -16,20 +19,27 @@
 
 public class PriorityFilter2 extends BasicHookWorker {
 
-    private TermCons parent;
+    /**
+     * Specifies whether the current node is the left most or the right most child of the parent.
+     * This is useful because associativity can be checked at the same time with priorities.
+     */
+    public static enum Side {LEFT, RIGHT}
```
dwightguth, line 26: Because it is necessary to check more than just the outermost nonterminals, this is not sufficient information. You need an integer index within the arity of the production.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PriorityFilter2.java
```diff
@@ -16,20 +19,27 @@
...
 
-    public PriorityFilter2(TermCons parent, Context context) {
+    public PriorityFilter2(String parentLabel, Side side, Context context) {
         super("Type system", context);
-        this.parent = parent;
+        this.parentLabel = parentLabel;
+        this.side = side;
     }
-
+/*
```
dwightguth, line 35: if this constructor is not used, please delete it.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PriorityFilter2.java
```diff
@@ -40,6 +50,21 @@ public ASTNode transform(TermCons tc) throws TransformerException {
...
+                KException kex = new KException(ExceptionType.ERROR, KExceptionGroup.CRITICAL, msg, kapp.getFilename(), kapp.getLocation());
+                throw new PriorityException(kex);
+            }
+        }
+
+        return kapp;
+    }
+
     @Override
     public ASTNode transform(Ambiguity node) throws TransformerException {
```
dwightguth, line 69: Should this method still exist, since I believe the same thing is being done in BasicTransformer?

tests/newparser/config.xml
```diff
@@ -2,32 +2,39 @@
...
+          programs="03-arithmetic-expressions-amb/"
+          extension="test"
+          results="03-arithmetic-expressions-amb/">
+        <all-programs>
+            <krun-option name="-parser" value="kast -parser newprogram" />
+        </all-programs>
+    </test>
 
-	<test definition="04-ambiguous-constructor/test.k" />
+    <!-- TODO(Radu): once ktest can call kast, add programs for tests. See issue
```
dwightguth, line 35: I believe I suggested what you can do to fix this, without having to wait for this code change? That should be part of this commit, to test everything.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PriorityFilter.java
radumereuta, line 63: When you have a variable of type KList directly under the KApp.
Note that this filter is currently being run over the rule AST also.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PriorityFilter2.java
radumereuta, line 69: This class extends BasicHookWorker - diffrent from BasicTransformer.
BasicTransformer visits all the nodes in an AST, while BasicHookWorker visits only what you extend.

In this case I want to apply the priority filter to the direct descendants of the parent node if it happens to be a TermCons or a KApp. If it is an Ambiguity node, I want to go over it. But if it is anything else, then stop.

tests/newparser/config.xml
radumereuta, line 35: Yes, I remember why I couldn't do that in the first place.
Basically, I'm printing terms which have ambiguities in them. Krun and maude don't know how to handle those, and it crashes with "unexpected token amb <---" (something like this).

So in the end, to be able to properly handle those tests, I think I need the functionality I was asking at the Monday meeting.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PriorityFilter.java
radumereuta, line 37: At the moment we want to just replace SDF. Have the exact same thing so we can also test.
Also, I don't think this is something that we want to support. We would like to have a simple and easy to explain mechanism. Maude fails in this regard. It's hard to explain what gather does, and you always  have to check to manual to make sure you did it right.

The more complex productions, like your example, could be handled by the rewrite engine which takes fully general disambiguation rules.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PriorityFilter2.java
radumereuta, line 26: I disagree on this one, check my previous comment.

tests/newparser/config.xml
dwightguth, line 35: Well no, not exactly. If that's what's going on, then you are stepping outside the use case of kast itself. Kast is intended to return completely disambiguated terms. What you should do is run the code through AmbFilter, like we would reasonably expect kast to do to arbitrarily resolve the ambiguities, and then match on the warnings that this generates on the command line when this happens.

radumereuta, line 35: Fair point.

radumereuta:
Also, I'm considering your comment about TermCons. My only concern with this is that it is horrible to explain, although it solves so many problems.
My other solution would have been to remove TermCons and always have KApps, but with two different sets of klabels. One specific to parsing which assigns unique labels to everything (user syntax, rewrites, cells, etc...) and another level which transforms the unique labels into the klabels or Rewirte, Cell... which can be used in the rest of the tool.

Pros: 1. finally get rid of TermCons which is confuzing. 2. keep the two levels distinct which made parsing and applying filters easier.
Cons: have to modify the current code

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PriorityFilter.java
dwightguth, line 37: Well you don't have to do what Maude did, that wasn't what I was claiming. I was claiming that Maude has a functional correctness that this lacks. I am willing to accept the idea that we don't want to change this for now because we just want to get this working, but it's /wrong/ now. Code that priorities should handle will not be handled, and the behavior will not be what users expected. The correct behavior is that priority applies on all nonterminals which are not surrounded on both sides by terminals.

This is starting to persuade me that we should just make the output of the parser kernel be concrete syntax trees. There are advantages to that approach regardless with respect to the documentation backend, and it would make it no longer necessary to modify every single parser filter (which you will have to do the way you are doing now).

As far as I can tell the only thing that would be needed to make that happen would be to replace the WrapLabelRule that exists currently with a WrapTermCons or something.

Thoughts on this?

dwightguth:
Well, I agree that TermCons is confusing as it exists right now. But I've been wanting to refactor that code for a while now, and eventually you or I could get to what I have in mind if we went that route. The problem with TermCons right now is the idea of the cons itself. A cons is a string which introduces a dependency on SDF. But suppose we renamed TermCons "ProductionTerm" and replaced the cons field with a field of type Production. Then it would be much more logical how ProductionTerm works, and you would still get the nice benefits of having access to information about the Production in the parser without having to guess which production we're referring to.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PriorityFilter.java
radumereuta, line 37: I don't know what makes you claim that the current semantics of priorities is 'wrong'.
We chose to implement priorities this way because it is a simple system that is easy to explain and easy to implement. The fact that it doesn't handle your example doesn't convince me. It's not something that is widely used in programming languages. It could be potentially be handled by the full general rewrite engine.
Even so. Now it's not the time to 'fix' this. First of all, we need a replacement for SDF, and after that we can think of ways to extend it.

Regarding the concrete syntax trees... I don't know where you got this from. All the parsing systems that I've worked with produce abstract syntax trees. And for good reasons, it makes everything more complicated. The current infrastructure of K has problems handling this format. The latex and html backends need to be modified. Also if you want the split that Grigore wants between front-end and backend, you will need two versions of latex and html...
I think the better solution here is to solve the problems with labels referencing multiple productions. That's what TermCons does (in a weird way). I think the best solution for this case would be to have two levels of klabels:
1. for parsing - maybe derived from the `cons` field? At this level I will solve all the ambiguities (priorities, associativity...)
2. for the rest of the compilation steps.

You mentioned the parser returning TermCons. That's kind of hard to do. The parser actually creates `KList`s, and WrapLabelRule only puts a KApp with a label on top of that. TermCons has the list of children builtin. It would require heavy modifications either for the parser, or for everything that uses TermCons.

KApp is more flexible. This is also why I added it to the filters. I want to remove TermCons at the same time with the SDF parser.

I hope I'm making sense.

dwightguth, line 37: The notion of a priority is that it represents a case where you use the fact that node A is an immediate child of node B to resolve an ambiguity by rejecting that particular parse. In particular, consider the expression "1+2*3". Here the ambiguity surrounds the number "2". It can be part of a parse beginning at position 0 and ending at position 2, or a parse beginning at position 2 and ending at position 4. The goal of priorities is to disambiguate this case.

In order to implement the solution to this problem, it is necessary to have a formal specification for under precisely what circumstances node A cannot nest inside node B. Your formal specification is that node A cannot nest inside node B if it is the first or last child of node B. This makes perfect sense if you approach the situation only from the perspective of binary infix operators. Since the problem of disambiguation using priorities typically arises in expression grammars, and the expression grammars of most modern programming languages use infix binary operators for expressions, it is not terribly surprising that this approach would be mistaken for the correct one.

However, when you generalize to arbitrary context free grammars such that terminals and nonterminals can be intermixed any way you choose, this starts to break down as you propose it. Consider for example the following grammar:

```
syntax Exp ::= Int | "^" Exp Exp | Exp Exp "$"
```

In this grammar, the string "^ 5 6 7 $" is ambiguous. If I declared priorities one way, it would parse as (^ 5 6) 7 $, and if I declared them the other way it would parse as ^ 5 (6 7 $). However, if I declare the former, then it becomes the case that the ^ production cannot nest directly inside the $ production. This causes the "5 ^ 6 7 $" term to fail to parse, even though the string is actually unambiguous. And if I declare the latter, then $ cannot nest inside ^, and therefore I cannot parse the string "^ 5 6 $ 7".

In order to resolve this, it is necessary that we obey the following property:
- If the priorities say that A cannot nest inside B, then we reject any parse in which A is an immediate child of B _and A is inside a nonterminal that begins or ends the production of B_)

Thus this works because in the production `syntax Exp ::= "^" Exp Exp`, the first nonterminal is not adjacent to the end of the production, while the second nonterminal is.

What the point of this illustrates, I think, is that the proper behavior of a parser that supports priorities is _dependent on the production itself_. That is to say, the behavior of the parser's priority rule cannot be determined from the AST alone: depending on which production the ASTNode was generated from, the correct behavior of the filter changes. In order to make the decision, it is necessary to know where in the production the terminals are.

Now that that's explained, let's approach the second point here. You would like to make a one-to-one mapping between productions and terms inside the parser? But that is _precisely_ what I am proposing. I am simply proposing the following design:

```
class ProductionTerm extends Term {
   Production prod;
   List<Term> children;

  KApp toKApp() {
    return new KApp(prod.getKLabel(), new KList(children));
  }
}

class Production extends ASTNode {
  KLabelConstant klabel;
  List<ProductionItem> items;
}

class KApp extends Term {
  Term label;
  Term child;
}

class KLabelConstant extends KLabel {
  String label;
}

class Context {
Map<String, Set<Production>> productions; //takes a label as a key
}
```

For comparison's sake, if I understand you correctly, here's what you are proposing, if I am generous:

```
class KApp extends Term {
  Term label; // may or may not be a KLabelConstant, so you have to check every single time
  Term child; // may or may not be a KList, and may or may not have the same arity as the label
}

class KLabelConstant {
  String label; //may or may not be an actual KLabel, might be a cons instead
}

class Context {
  Map<String, Set<Production>> productions; //takes a label as a key
  Map<String, Production> productions; //also takes a label as a key, but does something completely different
}
```

Do you see the problems with the second design? If not, I can't help you, except to say that it is /not/ a design I will allow to be merged.

dwightguth, line 37: One more thing: right now WrapLabelRule looks roughly like this (I am
simplifying a bit):

```
class WrapLabelRule {
  KLabel label;
  KList apply(KList klist) {
    return new KList(new KApp(label, klist));
  }
}
```

All I propose is the following change to make it return TermCons, until we
complete the refactoring I spoke of:

```
class WrapLabelRule {
  String cons;
  KList apply(KList klist) {
    return new KList(new TermCons(cons, klist.getContents()));
  }
}
```

If there is some deeper complexity required here, I don't see it, so you'll
have to explain for me. This is the only place in the parser where the KApp
class is used, so I don't see how anything could be dependent on the fact
that that class exists in beween the outer KList and the inner List<Term>.
Am I missing some nuance of the domain?

andreistefanescu:
I think we should focus on making the transition to the Java parser, and use the priority filters we already have. Fixing the filters would take time and we may end up not using them anymore. So I agree with Radu.

dwightguth:
After talking with Andrei it became clear I was not communicating myself well. To summarize:
- We need information about the production in order to properly apply priority filters
- When we support the full generality of rule-based disambiguation, this can change back, but for now, we move faster if the parser kernel supports TermConses
- I believe using TermCons in the parser kernel should be a simple change, and should save us time spent rewriting all 20-odd disambiguation filters.

Radu: does this make sense? Please help me understand if I am wrong about the complexity of the change I am proposing.

andreistefanescu:
I think this is a reasonable suggestion... @radumereuta what do you think?

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PriorityFilter.java
dwightguth, line 37: Also: I misunderstood the theory and misread the code. I looked at it again and it's correct already. It's only the change being discussed now that I have issue with.

radumereuta:
I have a few problems with what you proposed regarding TermCons:
1. From what I understand how the parser works, swapping KApp and KList with TermCons is not such a trivial thing to do. At least I know for sure that I will have to significantly modify the KSyntax2Grammar filter.
1. It increases the coupling between the java classes. This is something that would bring us troubles in the future if we want to make the parser a stand alone product for other people to use.

The solution that I proposed, is somewhere in between. We're trying to avoid the problem where we have more than one possible productions for a single `klabel`. The way I suggest we solve this, is to create KApps with unique labels in the parsing stage. Apply the disambiguation filters, and after everything is resolved, collapse the labels to something that we decide are the labels for the backend. 

I think this solution also fits better with what Grigore is planing with regard to using kast everywhere.

We should discuss this on Monday, although, I'm concerned about the rate of my progress. More precisely the fact that I am being contested so hard at every single step, no matter how trivial or small it is. Yes, there are many issues with the parser. I'm trying fix them, but I'm not going to be able to do it from the first try. There are to many dependencies.

dwightguth:
I am closing this since it is an out of date pull request. See #455.

Issue #366, opened 2014-04-12
Windows 8 installation error


grosu:
After fixing the JAVA_HOME undocumented problem, I got another build error:

```
D:\github.grosu\k>ant
Buildfile: D:\github.grosu\k\build.xml

build:

concrete.signature:

concrete.signature.helper:

all.signature:

concrete.stratego2java:

concrete.stratego2java.helper:

all.stratego2java:

build-subprojects:

init:

build-project-java:
     [echo] KTool all: D:\github.grosu\k\src\javasources\KTool\build.xml

build:

jar:
   [delete] Deleting: D:\github.grosu\k\lib\java\k3.jar
      [jar] Building jar: D:\github.grosu\k\lib\java\k3.jar

alllocal:

makeversion:

BUILD FAILED
D:\github.grosu\k\build.xml:19: The following error occurred while executing thi
s line:
D:\github.grosu\k\build.xml:68: Execute failed: java.io.IOException: Cannot run
program "git": CreateProcess error=2, The system cannot find the file specified
        at java.lang.ProcessBuilder.start(ProcessBuilder.java:1041)
        at java.lang.Runtime.exec(Runtime.java:617)
        at org.apache.tools.ant.taskdefs.launcher.Java13CommandLauncher.exec(Jav
a13CommandLauncher.java:41)
        at org.apache.tools.ant.taskdefs.Execute.launch(Execute.java:428)
        at org.apache.tools.ant.taskdefs.Execute.execute(Execute.java:442)
        at org.apache.tools.ant.taskdefs.ExecTask.runExecute(ExecTask.java:628)
        at org.apache.tools.ant.taskdefs.ExecTask.runExec(ExecTask.java:669)
        at org.apache.tools.ant.taskdefs.ExecTask.execute(ExecTask.java:495)
        at org.apache.tools.ant.UnknownElement.execute(UnknownElement.java:292)
        at sun.reflect.GeneratedMethodAccessor4.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAcces
sorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:606)
        at org.apache.tools.ant.dispatch.DispatchUtils.execute(DispatchUtils.jav
a:106)
        at org.apache.tools.ant.Task.perform(Task.java:348)
        at org.apache.tools.ant.Target.execute(Target.java:435)
[...]

Issue #365, opened 2014-04-12
installation problem under windows


grosu:
I followed blindly the steps under src/INSTALL.  Both javac and ant work when I type their name in the terminal.  Yet, when I type ant in the main k folder i get the following failure.  I do not have time to look into it now as I have to board my airplane to Japan, but this is frustrating, as I was hoping I can work on the airplane.  Isn't Java just supposed to be platform independent and work?  What are we doing wrong that after all these years, we still run into such low level issues?  I mean, nobody else in the team tried to follow the src/INSTALL instructions on windows?

```
D:\github.grosu\k>ant
Unable to locate tools.jar. Expected to find it in C:\Program Files\Java\jre7\li
b\tools.jar
Buildfile: D:\github.grosu\k\build.xml

build:

concrete.signature:

concrete.signature.helper:
    [mkdir] Created dir: D:\github.grosu\k\src\javasources\parsers\Concrete\incl
ude
     [java]   including ../parsers/Concrete/syntax/Concrete.sdf
     [java]   including ../parsers/Concrete/syntax/Common.sdf
     [java]   including ../parsers/Concrete/syntax/KTechnique.sdf
     [java]   including ../parsers/Concrete/syntax/KBuiltinsBasic.sdf
     [java]   including ../parsers/Concrete/syntax/Integration.sdf
     [java]   including ../parsers/Concrete/syntax/Variables.sdf
     [echo] Packed to .def
     [java] [ org.strategoxt.tools.main-sdf2rtg | info ] Calling external tool D
:\github.grosu\k\src\javasources\KTool/../../../lib/native/cygwin/sdf2table -i D
:\github.grosu\k\src\javasources\KTool/../parsers/Concrete/include/Concrete.def
-o C:\Users\Grigore\AppData\Local\Temp\StrategoXT8237994697871264757.tmp -n -m C
oncrete
     [java] cygwin warning:
     [java]   MS-DOS style path detected: D:\github.grosu\k\src\javasources\KToo
l/../parsers/Concrete/include/Concrete.def
     [java]   Preferred POSIX equivalent is: /cygdrive/d/github.grosu/k/src/java
sources/parsers/Concrete/include/Concrete.def
     [java]   CYGWIN environment variable option "nodosfilewarning" turns off th
is warning.
     [java]   Consult the user's guide for more details about POSIX paths:
     [java]     http://cygwin.com/cygwin-ug-net/using.html#using-pathnames
[...]

PR #410, opened 2014-05-02
Updating the README for developers to better explain that JAVA_HOME 

radumereuta:
environment variable is required to point to the JDK installation directory.
Also specify that git needs to run in the command line. The GUI is not enough.

This is in response to issues #365 and #366

### DIFF
--- a/src/README
+++ b/src/README
@@ -3,7 +3,7 @@
 
 2. Prerequisites
 
-2.1 Java Development Kit (tested with JDK7 or higher)
+2.1 Java Development Kit (required JDK7 or higher)
 You can follow the instructions here: 
 http://docs.oracle.com/javase/7/docs/webnotes/install/index.html depending on 
 the type of your machine.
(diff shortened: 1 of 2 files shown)


traiansf:
looks good to me

src/README
```diff
@@ -17,7 +17,14 @@ Linux and Mac:
...
+    installation instructions from the 'INSTALL' file.
+
+Ant usually requires setting an environment variable 'JAVA_HOME' pointing
+to the installation directory of the JDK (not to be mistaken with JRE).
+	
+2.3 Git - command line
+Having a GUI client is not enough. Most distributions have an installation
+option to make git accessible in the command line too.
 
 You can test if it works by calling 'ant' in a Terminal.
```
traiansf, line 29: This line should have been above section 2.3 and maybe you need a similar line for git here

radumereuta, line 29: Actually no. That was the problem reported in issue #366. If you don't have git installed as a command line tool, ant will still fail.

dwightguth:
Grigore is right, we shouldn't depend on git like this. It's not good practice. If you want to close out issue #366, you need to actually do a check on whether git is in the path from within build.xml.

radumereuta:
@dwightguth is it ready for merge now?
If yes, please go ahead.

PR #419, opened 2014-05-05
Apparently you can write how many times you want a pattern to be repeated

radumereuta:
This makes the regular expression for strings (java version at least) more readable.


dwightguth:
Lgtm. Merge if tests pass.

PR #421, opened 2014-05-06
Added the reject for keywords mechanism in the new parser.

radumereuta:
- the grammar generator takes as arguments a set of all tokens found in the definition
- when finding a Token declaration it compares all tokens to see which one overlap and add them as an exception to the matcher
- added a regression test which compares the output and err of krun
- added the AmbFilter at the end of the new parser.

@dwightguth  please review.
It's a bit different from what we discussed yesterday, but I think this is more efficient.
I'm letting the regular expression match on the string, and at the end, compare with a list of keywords that I know can overlap (computed during grammar generation time).

Just like before, keywords are automatically rejected, but now, if you don't want them to be, you can just declare them of the sort you want them to match. Example:
syntax Id ::= Token{[a-z]+}
syntax Stmt ::= "return" Exp ";"
syntax Id ::= "return"

The common case is that you want to reject those as being Id, but now you have a way to get around that.


src/javasources/KTool/src/org/kframework/parser/concrete2/Grammar.java
```diff
@@ -463,6 +463,7 @@ public RuleState(String name, NonTerminal nt, Rule rule) {
     public abstract static class PrimitiveState extends NextableState {
         /** The sort of the KApp */
         public final String sort;
+        public final Set<String> rejects;
```
dwightguth: I don't think this should be part of PrimitiveState. It should be part of the implementation in RegexState. The idea of a PrimitiveState is that it represents an opaque function for determining whether a particular input string matches this state. Your code change makes it non-opaque here, which is an undesirable design change.

tests/newparser/05-reject-keywords/a.test
```diff
@@ -0,0 +1 @@
+let abc
```
dwightguth: Can I see a few more examples testing some of the different corner cases you tested yesterday when we were in your office?

dwightguth:
Overall looks good. I agree this is a better design than what I proposed, with the minor caveat I mention.

dwightguth:
lgtm. Merge if tests pass.

PR #423, opened 2014-05-07
Moving the Int and Float declaration from SDF into K.

radumereuta:
Let's see if it passes the tests.

### DIFF
--- a/include/builtins/int32.k
+++ b/include/builtins/int32.k
@@ -1,7 +1,7 @@
 // Copyright (c) 2014 K Team. All Rights Reserved.
 module INT32
 
-syntax Int32 ::= Token{"[0-9]+"i32""}
+syntax Int32 ::= Token{"[0-9]+i32"}
                | Int32 "+Int32" Int32 [function, left, smtlib(+), latex({#1}\mathrel{+_{\scriptstyle\it Int32}}{#2}), hook(#INT32:_+Int32_)]
 
 endmodule
--- a/include/builtins/int32.k
+++ b/include/builtins/int32.k
@@ -1,6 +1,6 @@
 module INT32
 
-syntax Int32 ::= Token{"[0-9]+i32"}
+syntax Int32 ::= Token{"[0-9]+"i32""}
                | Int32 "+Int32" Int32 [function, left, smtlib(+), latex({#1}\mathrel{+_{\scriptstyle\it Int32}}{#2}), hook(#INT32:_+Int32_)]
 
 endmodule
--- a/include/builtins/int32.k
+++ b/include/builtins/int32.k
@@ -1,3 +1,4 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
 module INT32
 
 syntax Int32 ::= Token{"[0-9]+"i32""}
--- a/src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferAvoidFilter.java
+++ b/src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferAvoidFilter.java
@@ -1,3 +1,4 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
 package org.kframework.parser.concrete.disambiguate;
 
 import java.util.ArrayList;
--- a/src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferAvoidFilter.java
+++ b/src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferAvoidFilter.java
@@ -4,6 +4,8 @@ import java.util.ArrayList;
 
 import org.kframework.kil.ASTNode;
 import org.kframework.kil.Ambiguity;
+import org.kframework.kil.IntBuiltin;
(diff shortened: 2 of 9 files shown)


radumereuta:
@dwightguth please review.

include/builtins/int32.k
```diff
@@ -1,6 +1,7 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
 module INT32
 
-syntax Int32 ::= Token{"[0-9]+i32"}
+syntax Int32 ::= Token{"[0-9]+"i32""}
```
dwightguth: this code path is deprecated. If you want to support machine integers, you should use the mechanism in include/builtin/mint.k.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferAvoidFilter.java
```diff
@@ -26,6 +29,13 @@ public ASTNode transform(Ambiguity amb) throws TransformerException {
                     prefer.add(tc);
                 if (tc.getProduction().getAttributes().containsKey("avoid"))
                     avoid.add(tc);
+            } else if (variant instanceof KApp) {
```
dwightguth, line 32: this feels funky to me. A user who defines unary minus in their language can specify the prefer themselves one way or the other, can't they? Should we really be doing this in Java code?

include/builtins/int32.k
radumereuta: I actually don't know what is this. The last time it was modified by Pat. I just found an error in the regular expression and I fixed it.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferAvoidFilter.java
radumereuta, line 32: Indeed is funky. It is hard coded, because at the moment GenericToken doesn't have a reference to the production where I can find the prefer/avoid attributes. I will do it properly in the priorities pull request.

include/builtins/int32.k
dwightguth: Okay. Well, please discard the change then if it wasn't intentional. No
point in making changes to a code path that we aren't using anymore.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferAvoidFilter.java
dwightguth, line 32: That's not quite what I meant. I meant, maybe we should do something like this?

```
(in int.k)
syntax Int ::= Token{...} [integers]
```

And then the user can just say `syntax prefer '-_ > integers` if they want to add unary minus to their language.

dwightguth:
Is this really that simple of a change? I'm kind of impressed. You're sure that the tokens are getting instantiated properly as IntBuiltin down the line?

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferAvoidFilter.java
radumereuta, line 32: That could be a solution...
But at the moment we shouldn't change the semantics of the disambiguation filters.
Prefer and avoid are now global (not relative).

radumereuta:
They are actually immediately instantiated into IntBuiltin.
This is something that I've been complaining, because I have to do cases each time I want to process them. When I will ad the priorities, I plan to modify the new parser to instantiate TermCons and Constants. Both being instances of ProdTerm. A class which references the Production from which they come from (like we discussed on Monday - hopefully).

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferAvoidFilter.java
dwightguth, line 32: Okay. That makes sense to me. Can you make an issue for this?

dwightguth:
Can you make regression tests for this functionality? If you do that, and make the issue I requested, I am happy with this.

dwightguth:
Actually wait, hang on. This code path is actually being exercised in the SDF case, correct?

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferAvoidFilter.java
radumereuta, line 32: About what?
This is a small part of disambiguation with rewrite rules.
The semantics of priorities, as well as prefer and avoid has to be discussed at one point in the full meeting to agree on something that is best for K.

So I don't see the point of creating an issue about a subtopic. (If I understood correctly your request)

dwightguth, line 32: Okay, yeah, I guess that makes sense. I meant to track the fact that we intended to change this behavior, but I guess that's pretty much tracked regardless.

radumereuta:
Yes, it is tested with all the examples that have unary minus in them (simple, kool...).

PR #426, opened 2014-05-08
Refactor visitor exceptions

dwightguth:
- BasicTransformer is renamed ParseForestTransformer, and is to be used only for parse forests
- TransformerException is renamed ParseFailedException, and is to be used only to signify that the current parse tree in a parse forest should be pruned. In cases where an empty parse is not acceptable, the user should always call TransformerException#report() to terminate the application if the parse failed all the way to the top.
- CopyOnWriteTransformer no longer throws any checked exceptions.
- Some minor changes were made to code which violated the above mentioned principles.


radumereuta:
I think ParseForestTransformer is not the best name for this. It should be named based on what it does, not on where it is used. Just like CopyOnWriteTransformer.

I think a better name would be InPlaceTransformer. This way TransformerException is still a good name.

dwightguth:
TransformerException is /not/ a good name, because the only place TransformerExceptions are ever thrown is to indicate that a particular parse tree is not valid, or that a parse failed, and the only place TransformerExceptions are ever /caught/ which is not undesirable boilerplate code arising from poor use of checked exceptions, are when a parser is invoked and the exception is thrown to indicate that no valid parse was found. All other instances of "throws TransformerException" were frivolous and decreased code quality by forcing us to catch them in very fragile, meaningless ways.

Once you accept the premise that the entire purpose of TransformerException is to deal with ambiguities, it becomes clear that that is the entire purpose of BasicTransformer as well. This is the true distinction between the two transformer patterns, not whether or not the tree is modified in place (which is a distinction I leave for legacy reasons, but in fact it's almost never desirable to use a transformer in the way BasicTransformer does).

Thus the renaming operations. I do this because it makes it very clear the purpose of this exception and this transformer where before the purpose behind both classes was hidden behind the facade of a meaningless assumption about how they would be used, which never bore out in reality.

radumereuta:
Ok, I'm convinced.
Do you have any other things to add to this?
If not, you can merge.

dwightguth:
@radumereuta can you please confirm that everything looks good after the merge from master, and then I will merge it back in?

radumereuta:
lgtm. ok to merge

PR #427, opened 2014-05-08
fix bug in #parse code path

dwightguth:
So apparently using a custom parser for #parse was broken after all? This whole time? Gah.


dwightguth:
Actually, don't review this yet, I want to add a test.

tests/regression/kast/customparser
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+if [ "$1" = "foo" ]
+then
+  echo ".K"
+else
+  kast --parser ground "$@"
+fi
```
radumereuta: I don't understand what you are trying to do here.
Again, why do you need a .sh script which works only on unix systems?

dwightguth: You're right, it's not necessary for what I'm trying to do. Give me a sec and I'll rewrite the test without it.

radumereuta:
This looks better. If it's done, you can merge it.

PR #430, opened 2014-05-09
Updated the definition for the stream equivalence in samples/prog-equiv/stream.

dlucanu:
Corrected the definition for the the stream equivalence and added some more examples. The definition is still incomplete, e.g., the function checking a substitution must be rewritten, but it is fine now for the demo purposes.

PS This is my first pull request, I hope it works.


dwightguth:
Looks like you need to fix the copyright.

dlucanu:
Oops. Thank you. I will fix it asap.

dlucanu:
I did it. 
@dwightguth Should I do something else? Thanks.

samples/prog-equiv/imp/peq.k
```diff
@@ -1,3 +1,5 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
```
dwightguth: This should be copyright 2012-2014 because parts of this file are as old as 2012. I suggest when you decide what copyright header to put on a file you use `git blame` to determine how old the file is.

samples/prog-equiv/streams/seq.k
```diff
@@ -1,3 +1,5 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
```
dwightguth: see above

dwightguth:
Once you fix the issue with the years of the copyright range, you can merge this.

dwightguth:
Unfortunately this seems to have gotten sidestruck by the work I did with Andrei S. You will need to resolve the merge conflicts by preferring whatever version of the copyright notice is currently in master over your own code.

Issue #431, opened 2014-05-10
Unable to run graphical debugger using symbolic backend


daniel-ionut-vicol:
Steps:
1. run kompile on a definition using symbolic backend 
(ex: kompile seq.k --backend symbolic)
2. launch the graphical debugger for a program
(ex:krun programs/morse.seq -cPC true --debug-gui)

Result : 
When the graphical debugger is launched it reports :
[Error] Critical: Currently supported backends are 'maude' and 'java'

dwightguth:
Fixed in #432. @dlucanu can you please verify and then close this if you're happy?

PR #432, opened 2014-05-10
Fixed : Unable to run graphical debugger using symbolic backend #431

daniel-ionut-vicol:
fixed debugger symbolic backend issue by getting the KRun object from main class
- added KRun as parameter to gui controller
- get KRun from main class when continuing execution from a state

### DIFF
--- a/src/javasources/KTool/src/org/kframework/krun/gui/Controller/RunKRunCommand.java
+++ b/src/javasources/KTool/src/org/kframework/krun/gui/Controller/RunKRunCommand.java
@@ -38,13 +38,13 @@ public class RunKRunCommand {
         debugger = krun.debug(cfg);
     }
 
-    public RunKRunCommand(KRunState state, String lang, Context context) {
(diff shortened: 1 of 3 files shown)


dwightguth:
Jenkins: ok to test

dwightguth:
Jenkins: ok to test

src/javasources/KTool/src/org/kframework/krun/gui/Controller/RunKRunCommand.java
```diff
@@ -53,7 +44,7 @@ public RunKRunCommand(KRunState state, String lang, Context context) {
         this.KAST = state.getRawResult();
         this.lang = lang;
         rp = new RunProcess();
-        this.krun = createKrun(context);
+        this.krun = Main.obtainKRun(context);
```
dwightguth: this should also take a KRun as a constructor arg. That way the obtainKRun method can stay private, and you can pass in the old krun object a second time at the call sites for this constructor.

Issue #418, opened 2014-05-05
Latex: Preserving variable sorts


denis-bogdanas:
Right now latex output don't display the sorts of variables. This is confusing even in bubbled output, and completely inappropriate for math format intended to be used in the paper. So we probably have to display all the sorts that were specified by the user, and hide the sorts that were inferred automatically.

grosu:
I think the idea was to include all the user-declared sorts of all the variables in the generated Latex, and only dynamically display the inferred sorts when one hoover the mouse over the variable.  Actually, I think the latter  was something we wanted to generalize to whole terms, not only to variables, that is, to display the inferred sort of the term rooted in a given symbol (operation or variable) whenever one hoovers the mouse over that symbol.

traiansf:
Ok, I've pushed something on the latex-fix branch.  I'm using the isUserTyped() method to check whether a variable was user-types. However, there is a problem still to be solved.  Apparently all occurrences of of a user-sorted variable identify themselves as user-typed so the latex shows the variable for all occurrences of a typed var.
@radumereuta - can you suggest another mechanism to detect user-typedness?

radumereuta:
I've just pushed in the same branch a fix for variables being typed everywhere.
Please test if everything works now. I'm not sure if it is a safe change or not.

PR #433, opened 2014-05-10
Latex fix

traiansf:
By our powers combined, here is a proposed fix for Issue #418 
@grosu especially, but also everybody else: please check pdf output and suggest improvements.

@dwightguth please review


traiansf:
Closing this as it is replaced by #435

Issue #425, opened 2014-05-08
Latex: <br/> inside rule looks horrible


denis-bogdanas:
Expected:
![expected](https://cloud.githubusercontent.com/assets/5214914/2919288/a14917b8-d6d9-11e3-857e-655b36375b90.png)

Actual:
![actual](https://cloud.githubusercontent.com/assets/5214914/2919290/a5cdd724-d6d9-11e3-8805-6ddad8b1f548.png)

Source code:

```
rule [Invoke-compute-methodInfo-DecC]:
<k>
  Qual:KResult
    . methodInfo(Sig:Signature, QualC:ClassType, .K => MetaT, .K => DecC, _,_)
    ([Args:KList])
  ...
</k>
<classType> QualC </classType>
<classMetaType> MetaT:ClassMetaType </classMetaType>
<methods>... Sig |-> DecC:ClassType ...</methods>
```

radumereuta:
@denis-bogdanas 'horrible' is an opinion, not a description of the bug.
Not only it's not helpful, but it's a mockery of other people work. Please try to be at least helpful in the future, and describe what is your problem.

traiansf:
Although I don't feel as passionate as @radumereuta about the matter, I would really like to understand what you mean, i.e.,  where 'horrible'  comes from and what (technically) is the proposed solution.

denis-bogdanas:
I attached two images, one is the expected result, the other one is the actual one. All in accordance to the standards of bug reporting. Also the source code. Please let me know if you need any more info.

traiansf:
Well, I am not convinced that the "actual" result is bad (its purpose was to "wrap" the rule so that its identity is not confused) and that the expected result is better.   I would like some rationale for why is that better and maybe some group discussion to find the best solution/compromise.

denis-bogdanas:
A group discussion sounds fair, especially because we are discussing about subjective tastes.

I tried to render one module of the semantics using <br/>, and the parentheses did not add any extra clarity. Where a rule ends and another one begins is pretty clear because of the keyword RULE. And the parentheses are definitely not desired when we are rendering for a paper - we want to make things as compact as possible. 

At least I would be great to have the option to render with and without parentheses. Preferably by supplying an argument to \usepackage[...]{k}

grosu:
First, Denis and Radu, please calm down and stick to the facts.  We don't have time for endless subjective debates, and this is not the right place for them anyway.  Meet and discuss this directly, your offices are next to each other.

[...]

PR #434, opened 2014-05-12
Patch for Issue #425

traiansf:
Removing brackets for `<br/>` as normal brackets can be used for disambiguation.

@grosu please check tutorial/1_k/5_types/lesson_6 — it is the only example from the tutorial using `<br>` in rules. I think it does not need additional parens so I did not add them.

@denis-bogdanas please check it works for you now. It is true that in the style you're using the rules are easier to delimit, but in the poster style the rule keyword gets aligned to the center of the rule (if  `<br/> is at the top), which made rules using`<br/>` harder to delimit.

### DIFF
--- a/src/javasources/KTool/src/org/kframework/backend/latex/LatexFilter.java
+++ b/src/javasources/KTool/src/org/kframework/backend/latex/LatexFilter.java
@@ -197,14 +197,12 @@ public class LatexFilter extends BackendFilter {
             return null;
         }
         if (hasBR) {
-            if (parens) result.append("\\left(");
             result.append("\\begin{array}{@{}c@{}}");
         }
         List<Term> contents = col.getContents();
         printList(contents, "\\mathrel{}");
         if (hasBR) {
             result.append("\\end{array}");
-            if (parens) result.append("\\right)");
         }
         return null;
     }


dwightguth:
I don't really have an opinion about whether or not this change is a good idea, it looks like that's a separate discussion you will have to have. But I am fine with this from a coding perspective, so decide amongst yourselves whether this is what you want to do, and if it is, you can merge it.

denis-bogdanas:
I see this page displays the message "This pull request can be automatically merged."
So if I pull the latest K now, will it contain this patch or I have to do something else to get it?

dwightguth:
No, you would have to click the "merge pull request" button first in order for it to be visible in master. But you can check out the latex-br-fix branch that this pull request is on and use it if you wish.

denis-bogdanas:
The central alignment of cells leads to awkward results. Take a look below.
Rules manually formatted by me:
manual

present patch:
patch-434

Source code for the first rule. Manual:

```
\krule[Invoke-compute-methodInfo-ContextType]{
\kprefix[Green]{k}{{\AnyVar[K]}\terminal{.}{{}\terminal{methodInfo}({\variable[Signature]{Sig}}, {\AnyVar[K]},{\AnyVar[K]},{\variable[ClassType]{DecC}},{\reduce{\dotCt{K}}{\variable[ContextType]{ContextT}}},{\reduce{\dotCt{K}}{\variable[AccessMode]{Acc}}})}({\AnyVar[K]})}
\mathrel{}\\\mathrel{}\kall[Red]{classType}{\variable[ClassType]{DecC}}
\mathrel{}\kall[Blue]{methodSignature}{\variable[Signature]{Sig}}
\mathrel{}\kall[Blue]{methodContextType}{\variable[ContextType]{ContextT}}
\mathrel{}\kall[Blue]{methodAccessMode}{\variable[AccessMode]{Acc}}
}{}{}{}{}
```

Present patch:

```
\krule[Invoke-compute-methodInfo-ContextType]{
\begin{array}{@{}c@{}}
\kprefix[Green]{k}{{\AnyVar[K]}\terminal{.}{{}\terminal{methodInfo}({\variable[Signature]{Sig}},{\AnyVar[K]},{\AnyVar[K]},{\variable[ClassType]{DecC}},{\reduce{\dotCt{K}}{\variable[ContextType]{ContextT}}},{\reduce{\dotCt{K}}{\variable[AccessMode]{Acc}}})}({\AnyVar[K]})}
\mathrel{}\\\mathrel{}\kall[Red]{classType}{\variable[ClassType]{DecC}}
\mathrel{}\kall[Blue]{methodSignature}{\variable[Signature]{Sig}}
\mathrel{}\kall[Blue]{methodContextType}{\variable[ContextType]{ContextT}}
\mathrel{}\kall[Blue]{methodAccessMode}{\variable[AccessMode]{Acc}}
\end{array}}{}{}{}{}   
```

denis-bogdanas:
Things are even worse for long rules. The rule below did not overflow over the right page margin prior to the patch:
big rules

traiansf:
I've only removed the parentheses,  The centering was there all along.  I can remove that, too, but some people might argue that centering is better than aligning to the left.

@grosu What do you think?

denis-bogdanas:
Please take a look at the last screenshot as well. If we don't add any `<br>`, then the rule is rendered nicely on 3 rows. If we put a single `<br>` because we want the rule to do line breaking a little bit different than default way, then it don't do any automatic line breaking anymore and get overflowed. This is definitely a bug.

traiansf:
I don't subscribe to this point of view.  Since line breaks can happen both at the top level and inside of cells, I need to use an environment to allow them.  I'm now using array.  
If you don't break the cells at all, Latex automatically breaks the lines (only at the top level) and of course it does it nicely.

However, if you decide to break them yourself, I believe it should be your task to make sure they fit nicely afterwards.

The only thing I can think of doing would be to treat specially `<br/>`s when they occur at the top level, but then the effect of `<br/>` would be different (and different looking) between top level and inside of cell.

denis-bogdanas:
Hmm, I've never seen line breaks inside a cell, and supposed they are not supported there. Can you give us an example?

dwightguth:
What is the next step here?

traiansf:
I think Grigore was supposed to issue an opinion about this.

2014-05-23 21:03 GMT+03:00 dwightguth notifications@github.com:

dwightguth:
@grosu can you weigh in then? I'd like to decrease my backlog of unmerged
pull requests.

grosu:
Traian, I really like it that I can put parentheses if I want to and they appear as large brackets!  Excellent, we should definitely keep this.  However, for some reason, the colors of the cells disappear when I put parentheses.  Could it be because of my acroread?

grosu:
Regarding the alignment of the cells, left vs. center, I don't know what to say.  We kind of got used with the center alignment.  Can we add another HTML tag, <div>, for that?  Here is how it works:

  http://www.echoecho.com/htmltext06.htm

This way, people are given more freedom in how to align their cells.

traiansf:
Not sure why colors would disappear. I'm only putting everything inside of \left( \right)

W.r.t. defining alignment in cells, we could add another attribute if @radumereuta allows us;  however this would only solve the issue that  @denis-bogdanas pointed out if we align everything not specifically mentioned by the user to the left, or if we special case the rules.  BTW, in this case, how would one specify alignment, as a tag align(center) in the rules tags?

dwightguth:
So can I merge this or not, please?

dwightguth:
Jenkins: test this please

denis-bogdanas:
I'm fine with the merge, it is a new useful feature after all, even if it have some issues. Those could be addressed in a subsequent pull request. I'm using an older version of K anyway.

PR #435, opened 2014-05-12
Latex fix2

traiansf:
By our powers combined, here is a proposed fix for Issue #418 
@grosu especially, but also everybody else: please check pdf output and suggest improvements.

@dwightguth please review

Note: this is the same pull request as #433  only rebased from master.


dwightguth:
Can  you help me understand why VariableTypeFilter had to be changed?

radumereuta:
VariableTypeFilter would be applied after type inferencer for variables. It would look for every appearance of a variable and change it's sort into the inferred sort. After that I would apply the type filter and eliminate branches with unwanted ambiguities.

What I changed, was the way I update the isUserTyped() parameter. This was used only at the beginning of this step, where I had to collect all variable declarations. After that it is used only by the latex filter. So the change is safe.

dwightguth:
Okay. I am merging this, then.

PR #437, opened 2014-05-12
Copyright

dwightguth:
@andreistefanescu was fed up with fixing copyright every time, and when he complained, I realized I knew a way to script this (I had thought about it previously and not figured it out). So I am saving us all a bunch of time by adding copyright to all the .java, .k, .jj, and .xml files in the repo.

@andreistefanescu can you do a spot check? Don't look over everything, if the tests pass that will be good enough.


radumereuta:
that's actually a good thing.
Hope we won't have too many troubles with merging with other branches, like unified-builtins

andreistefanescu:
lgtm

PR #438, opened 2014-05-13
Copyright on unified builtins

dwightguth:
I decided to do the same commit on unified-builtins2 separately and then merge afterwards, rather than trying to merge before Andrei's change is merged.

@andreistefanescu please review again.


dwightguth:
Closing per discussion with @andreistefanescu

PR #440, opened 2014-05-13
ensure that the IOServer thread terminates

dwightguth:
Previously the application only exited if system.exit was called, when the IO server had been initialized. This causes ktest to time out instead of failing if an uncaught exception is thrown by the pretty printer. This code fixes that by using java.nio in order to allow the thread to be interrupted while waiting on a socket, causing the application to shutdown in an orderly fashion.

Does anyone happen to know anything about threading who might be able to review this?


radumereuta:
Maybe @denis-bogdanas ?
He's our java expert.

pdaian:
I looked over the change briefly from a Java perspective at @dwightguth's request, and it looks good.  Certainly shouldn't introduce any behaviors that the original code didn't exhibit.

dwightguth:
I am going to merge this now, since Philip thinks it looks good, and there isn't really anyone around to review the IO server side of things.

PR #444, opened 2014-05-14
change kompile to use unique temp files

dwightguth:
There was a race condition in the pdf backend if you compiled the same definition twice at the same time. I believe this to be because they share a .k directory. I am hoping that this fix, which finally implements the separation between temp files and outputs that -kompiled and .k were supposed to exhibit, will fix this problem.

I still need to test this more thoroughly. I will let you know when it's ready for review.


dwightguth:
This is ready to be reviewed. @andreistefanescu  @traiansf  @radumereuta  please review the modifications to your sections of the code.

andreistefanescu:
lgtm

radumereuta:
Since I will have to revise most of the table storing and loading for the new parser, I won't be too picky right now, so lgtm.

PR #445, opened 2014-05-14
fix ktest log string escaping

dwightguth:
@osa1 @daejunpark please review.


src/javasources/KTool/src/org/kframework/utils/StringUtil.java
```diff
@@ -519,4 +521,27 @@ private static String splitLine(String str, int col) {
...
+            for (String arg : args) {
+                sb.append('"');
+                sb.append(StringUtils.replace(StringUtils.replace(arg, "\\", "\\\\"), "\"", "\\\""));
+                sb.append("\" ");
+            }
+            sb.deleteCharAt(sb.length() - 1);
+            return sb.toString();
+        }
+        throw new IllegalArgumentException("unsupported OS");
+    }
```
osa1, line 546: I'm pretty sure this was handled before. This is what `org.kframework.ktest.QuoteHandling` class for, but that class looks unused right now. I just searched in commit history to see why quoting is removing and when is it removed.
- https://github.com/kframework/k/commit/ca29b751d823c16f86ef59aeb8a4b15b6f9e9c96#diff-39b4d4e21743e2de4c36b28c757e87beL59
- https://github.com/kframework/k/commit/f7b1e896c92794ff3a0aee47eaede2ff1e41a214#diff-a5cf3aaf006fff3d93af758c62573efdL162

This two commits(relevant lines are annotated in the links) removes uses of `quoteArgument`. There should be at least one more removal of this method(for PDF command printing), but I couldn't find it.

I have no idea why it was removed before but why not put these lines back instead of implementing same thing again? (we now `quoteArgument` was working fine without problems)  And why it was removed before? (commit messages do not say anything useful)

dwightguth, line 546: Oh, that's probably my fault. I tweaked some stuff pertaining to PgmArg and
must have done it wrong because I didn't know what that did. I will do as
you suggest tomorrow. Feel free to close this diff unmerged.

osa1:
I'm closing this as @dwightguth will put old working code back as can be seen in relevant commit in my comment above.

dwightguth:
I am reopening this because I think we need to discuss this further. I am looking at QuoteHandling now and it seems to have an algorithm which is less correct than the one I proposed. It does not correctly handle strings with `, $, |, ~, etc. It also will crash ktest if the string contains both ' and ", and it will not correctly print out something executable on windows.

dwightguth:
@osa1 can we please have this discussion? Andrei is waiting for this change and I don't like the idea of doing what you propose if the code we had before was incorrect, which it was.

osa1:
Sorry, I didn't know you were waiting for my comments. My implementation is not handling as many cases as yours(although it has slightly different use case -- for example it'n not always wrapping arguments with quotes and instead it sometimes escapes characters etc.) so I say let's merge this and remove QuoteHandling class.

dwightguth:
Alright. I will test this change one more time and then merge it if the tests pass.

PR #446, opened 2014-05-14
make error message better for case when user specifies too many files fo...

dwightguth:
...r kompile


andreistefanescu:
So what happens if there are multiple tags parsed as files. They should not just be ignored.

andreistefanescu:
Never mind, if you are taking care of that in other issues. I did not see that.

andreistefanescu:
lgtm

PR #449, opened 2014-05-14
fix all samples except prog-equiv

dwightguth:
- symbolic backend test doesn't work still
- agent is broken because it relies on search and substitution
- prog-equiv is broken because it declares new constructors of sort Map

@andreiarusoaie we will not delay the merge trying to fix the symbolic backend, so if you want it to continue to work moving forward I suggest you fix it within unified-builtins2 while we are still waiting to merge.
@dlucanu your code in peq.k uses features of the framework that are being removed by this branch. You will no longer be able to declare new constructors of sorts Map, List, and Set. We will not delay the merge trying to fix this definition, so if you want it to continue to be regression tested moving forward, you will need to figure out how to fix the definition.

### DIFF
--- a/src/javasources/KTool/src/org/kframework/compile/transformers/DesugarStreams.java
+++ b/src/javasources/KTool/src/org/kframework/compile/transformers/DesugarStreams.java
@@ -14,7 +14,6 @@ import java.util.ArrayList;
 public class DesugarStreams extends CopyOnWriteTransformer {
     
     ArrayList<String> channels = new ArrayList<String>();
-    boolean newList;
 
     public DesugarStreams(org.kframework.kil.loader.Context context) {
         super("Desugar streams", context);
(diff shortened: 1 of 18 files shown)


dwightguth:

src/javasources/KTool/src/org/kframework/compile/transformers/DesugarStreams.java
```diff
@@ -16,12 +16,11 @@
     ArrayList<String> channels = new ArrayList<String>();
     boolean newList;
```
andreistefanescu: Is this field still used?

dwightguth: Oh, no, thanks. I will remove it tomorrow.

andreiarusoaie:
@dwightguth go ahead and merge this as it is and I'll fix it later. I was not able to fix it today and I will not be able to fix it soon, so do not wait for me.

andreistefanescu:
lgtm

dlucanu:
If there is not allowed to declare constructors for the sort Map, List, Set, how can a user to work with symbolic stuff? I need a mean to say that this is a symbolic Map (just a symbolic variable of sort Map). Does it work "#symMap", "symList", "symSet"? Or do we have better solutions now? (maybe I missed some discussions on this subject).

dwightguth:
@andreistefanescu can correct me if I'm wrong, but I believe the answer is
that you will need to add support for symbolic maps, lists, sets directly
to your kompile and krun backends, rather than trying to define them in K
itself.

dlucanu:
Andrei A told me that he tried and it seems that the branch devoted to unifying builtins includes support for "#symMap", "#symSet", ....
Is it correct? If yes, when is it expected to merge that branch?

dwightguth:
The unified-builtins2 branch will hopefully be merged sometime in the next
several weeks, but this change is actually specific to that branch. You can
check it out now and fix the definition which is broken. If you do so, we
will keep it inside the K test suite when the branch is merged, otherwise
it will no longer be regression tested after we merge the branch. That is
what I meant to say.

Issue #231, opened 2014-01-04
Precompiled PDF of ref manual would be very helpful


jcamins:
The reference manual is provided only as a K file (that contains only LaTeX, but needs to be preprocessed by K somehow before it can be used), but there are no instructions for compiling the reference manual. The barrier for entry would be greatly lowered if there were a PDF of the manual available for printing, etc.

denis-bogdanas:
An attempt to compile the manual produces an exception:

d:\k-framework\documentation>kompile --backend pdf ref-manual.k
Exception in thread "main" java.lang.NullPointerException

version: bcbfff7
Platform: windows

denis-bogdanas:
d:\k-framework\documentation>kompile.bat --backend pdf ref-manual.k
[Warning] Compiler: pdflatex returned a non-zero exit code.  The pdf might be
generated, but with bugs. please inspect the latex logs.

d:\k-framework\documentation>kompile --version
K-framework nightly build.
Git Revision: 2d8f90d
Build date: Thu May 15 16:46:23 CDT 2014

dwightguth:
Oncall needs to fix the compilation errors.

dwightguth:
There are now instructions to build the pdf: however, it still needs to be made to build automatically. Additionally, the pdf needs at some point to be updated with the latest version of K and synced with the to-be-processed.

WalkerCodeRanger:
Tried to follow instructions to build ref manual for v3.5.2 on windows.  It failed.  Really need the prebuilt pdf.  Seems like k project is exactly what I need, but the documentation is so inaccessible and lacking I am having a lot of trouble getting started.  If I can't get past these issues I may have to use a different tool.

DustinWehr:
So does nobody have a build of the ref manual that they can upload somewhere???
Ditto for pdfs built from all the latex in the tutorial examples.
I am trying to evaluate whether my company should use K, and having to successfully install everything (I too have failed to build the reference manual) before I can even read the formatted documentation is a BIG red flag.

grosu:
@DustinWehr Unfortunately, as an academic team we didn't have the resources to maintain the reference manual.  A demotivating factor for that was also the fact that K changed a lot lately.  In fact, we are working on a new backend of K implemented in Haskell, targeted for proofs and the generation of proof objects to serve as correctness certificates for smart contracts or any other programs. 

[...]

PR #450, opened 2014-05-15
fix crash in issue #231

dwightguth:
@radumereuta please review


dwightguth:
Closing. I investigated the code further and it turns out you're supposed to be able to use the pdf backend on files without modules in them. I will rework this.

PR #453, opened 2014-05-15
fix latex backend for files without modules

### LINKED ISSUE #231: Precompiled PDF of ref manual would be very helpful

dwightguth:
@traiansf please review.

Fixes #231.


traiansf:
I'm fine with it.  We could consider printing a warning if the file contains no modules, though.  Feek free to merge it.

PR #466, opened 2014-05-16
Flattenterms

liyili2:


dwightguth:
Jenkins: add to whitelist

src/javasources/KTool/src/org/kframework/compile/transformers/ResolveSyntaxPredicates.java
```diff
@@ -38,7 +44,8 @@ public ASTNode visit(Sentence node, Void _)  {
         for (Variable var : vars) {
 //            if (!var.isUserTyped()) continue;
             if (var.isSyntactic()) continue;
-            if (MetaK.isKSort(var.getSort())) continue;
+            if ((var.getExpectedSort()==null && MetaK.isKSort(var.getSort()))
```
dwightguth: I think you need to speak to @radumereuta about this. It seems very odd to me to be expecting a variable like this to be null. It also seems strange to me to need to change the actual logic of this step beyond what is actually modified by FlattenTerms.

What exactly was the problem you ran into when you tried to solve things the way I suggested? Why was that route abandoned?

liyili2: The problem is like:

When I move the flattenTerms before resolveSyntaxPredicate, For some terms, the resolveSyntaxPredicate step cannot detect the isInt relation among the term's variables. For example:

  rule <k> '`,(.KList) => .K ...</k>
       <array>... P |-> (_ => V) ...</array>
       <ptr> P:Int </ptr>
       <in> ListItem(V:Int) => .List ...</in>

The resolveSyntaxPredicate step should collect the variables P and V as #andBool(isInt(P),isInt(V)). 

However, if I move the flattenTerms before resolveSyntaxPredicate, the term will become 
...
       <ptr> P:KItem </ptr>
       <in> ListItem(V:KItem) => .List ...</in>

Hence, the resolveSyntaxPredicate cannot detect the isInt relation among variables P and V.

liyili2: I tried two solutions. 

One is what I do now. I add to check the getExpectedSort instead of getSort in resolveSyntaxPredicate. This solution cause less problem. There are some test cases which was failed. However, all definitions can be compiled.

The other solution is to modify the variable class and set default getExpectedSort to be the same as getSort instead of null. Then I change here to check getExpectedSort only. However, this solution will cause some definitions to be failed. Some definitions cannot be parsed correctly (there will introduces ambiguity). It means that somehow the concrete parser relies on the fact that getExpectedSort is null in the beginning.

liyili2: I mean I test the fixing in resolveSyntaxPredicate or variable class individually. Then I test to move flattenTerms before resolveSyntaxPredicate. For the above two solutions, The first cannot pass every test cases if I did not move flattenTerms before resolveSyntaxPredicate. The second one cannot pass all test cases even if I did not move  flattenTerms before resolveSyntaxPredicate.

dwightguth: So, the two parts of the change I proposed are dependent on each other. You
have to change both ResolveSyntaxPredicates and FlattenTerms. Since the
idea is for KAST to have the semantic sorts of variables, it is not correct
for FlattenTerms to set all the variable sorts to KItem. That is why I
proposed you change that class, and why I do not like the current solution.
But then, in order to make everything work, it's necessary that the sort be
changed later, hence the change to ResolveSyntaxPredicates.

You are explaining a lot of different changes here, and I am having trouble
following which problems you encountered under which version of the change.
Can you please explain specifically just the problem you encountered under
the solution I just named, and nothing else?

liyili2: I do not think I need to explain to you every detail, right. I mean I did not finish it yet, there is still problem when I move flattenTerms before ResolveSyntaxPredicates. Wait me to finish it.

dwightguth: If I ask a question, it's because I think I need to know the answer. It's
important that I am able to track progress on a change like this otherwise
time will be wasted going in the wrong direction. For example, you have
reverted the change I suggested you make regarding the sorts of variables
in FlattenTerms, and I don't know why yet. I would have to ask that
question when you are done and ready for code review, and then if I didn't
like the answer we would have to start the entire diff over on a different
path. I am just trying to establish the answer now so I know whether or not
I need to have you go back to the path you abandoned, or if there is some
obstacle here I genuinely did not anticipate.

liyili2: Because there solution is wrong. I cannot even run some program. The program will loop forever.
for example:
If I have this definition:
// Copyright (c) 2013-2014 K Team. All Rights Reserved.
/*
Expected:
'ListWrap(1,,1,,1)
*/
module LABEL-AND-CONTEXT
  configuration <k> $PGM:K </k>

  syntax KLabel  ::= "'ListWrap" | "'A"
  syntax KResult ::= Int
  context 'ListWrap(_,,HOLE,,_)

  rule 'A(_) => 1
endmodule

and run this program:
'ListWrap('A(.KList),,'A(.KList),,'A(.KList))

dwightguth: Unfortunately this is caused by a nontermination bug in the maude backend
of krun. The program does not actually loop forever, it is merely that the
compiled definition in maude is loaded and has errors, and the bug causes
krun to not crash with an error message. I have a pull request now that
should address this issue, which will be merged into the unified builtins
branch soon. In the meantime, I suggest you do the following:
1. Reload the change that you made before with the removed code from
   FlattenTerms and the Variable.setSort call
2. Try compiling the example again, then run, from the main k directory:

```
$ lib/maude/binaries/maude.linux
Maude > load
tests/regression/label-and-context/label-and-context-kompiled/main.maude

This should show you the errors that you will need to resolve.


On Tue, May 20, 2014 at 10:36 AM, Liyi Li <notifications@github.com> wrote:

> In
> src/javasources/KTool/src/org/kframework/compile/transformers/ResolveSyntaxPredicates.java:
>
> > @@ -38,7 +44,8 @@ public ASTNode visit(Sentence node, Void _)  {
> >          for (Variable var : vars) {
> >  //            if (!var.isUserTyped()) continue;
> >              if (var.isSyntactic()) continue;
> > -            if (MetaK.isKSort(var.getSort())) continue;
> > +            if ((var.getExpectedSort()==null && MetaK.isKSort(var.getSort()))
>
> Because there solution is wrong. I cannot even run some program. The
> program will loop forever.
> for example:
> If I have this definition:
> // Copyright (c) 2013-2014 K Team. All Rights Reserved.
> /*
> Expected:
> 'ListWrap(1,,1,,1)
> */
> module LABEL-AND-CONTEXT
> configuration $PGM:K
>
> syntax KLabel ::= "'ListWrap" | "'A"
> syntax KResult ::= Int
> context 'ListWrap(*,,HOLE,,*)
>
> rule 'A(_) => 1
> endmodule
>
> and run this program:
> 'ListWrap('A(.KList),,'A(.KList),,'A(.KList))
>
> —
> Reply to this email directly or view it on GitHub<https://github.com/kframework/k/pull/466/files#r12850577>
> .
>
```

liyili2: Thank you very much.

I think I had other solution. It is a combination with your solution and my solution.

I will add the flattenTerms step with setSort of KItem before FlattenSyntax steps.

Then I will create another flattenTerms step without setSort before ResolveSyntaxPredicate step. I think this will work.

liyili2: Another piont I want Radu to know also is like I think the concrete parser (I am not sure if it is the new one or not) has a fundamental problem. The concrete parser relies on the fact that the extended sort to be null initially. I think this is wrong, the extended sort should always be a more concrete sort than the original sort of a term.

liyili2: good. The new implementation works.

liyili2: Ok. I found that even if I remove the getExtendedSort in resolveSyntaxPredicate, it will work too. I wonder if you think it is a good idea to put getExtendedSort in resolveSyntaxPredicate or it is good to put getSort in resolveSyntaxPredicate step?

liyili2:
What does it mean by  Merged build triggered

dwightguth:
It means it's waiting on another Jenkins build to begin testing.

build.xml
```diff
@@ -2,9 +2,10 @@
 <!-- Copyright (c) 2012-2014 K Team. All Rights Reserved. -->
 <project basedir="." default="build" name="KTool Build file">
```
dwightguth, line 4: Changes to this file should be reverted.

src/javasources/KTool/src/org/kframework/backend/BasicBackend.java
```diff
@@ -88,6 +89,7 @@ public boolean autoinclude() {
         steps.add(new ResolveSyntaxPredicates(context));
         steps.add(new ResolveBuiltins(context));
         steps.add(new ResolveListOfK(context));
+        steps.add(new FlattenTermsTwo(context));
```
dwightguth, line 92: ...no. Not happening.

You need to be aware of the fact that duplicating code when you need its functionality to change is very bad code practice and leads to unmaintainable and buggy code. While I agree with the sentiment that perhaps variables should have their sort set to KItem here, that code belongs in FlattenSyntax and is expressible in a few lines. I have already had to clean up several places in the code base where you have done this previously, and more still remain. If you want to develop code in thus code base, you have to learn not to do this.

src/javasources/KTool/src/org/kframework/compile/transformers/FlattenTermsTwo.java
```diff
@@ -0,0 +1,236 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
```
dwightguth, line 1: See above about this class.

src/javasources/KTool/src/org/kframework/compile/transformers/ResolveSyntaxPredicates.java
```diff
@@ -26,6 +26,12 @@ public ASTNode visit(Syntax node, Void _)  {
         return node;
     }
     
+    /**
```
dwightguth: These comments are out of date.

src/javasources/KTool/src/org/kframework/compile/utils/MetaK.java
```diff
@@ -88,6 +88,7 @@ public static String cellUnit(String cellName) {
         public static final String coolingTag = "cool";
         public static final String hole = "[]";
         public static final String freshCons = "Bool1FreshSyn";
+        public static final String freshLabel = "'fresh(_)";
         public static final String plusIntCons = "Int1PlusSyn";
```
dwightguth, line 92: You should delete freshCons and replace all references to it.

src/javasources/KTool/src/org/kframework/compile/transformers/FlattenTerms.java
```diff
@@ -222,11 +223,6 @@ public ASTNode visit(Variable node, Void _)  {
             }
 
             node = node.shallowCopy();
-            if (kompileOptions.backend.java() || K.backend.equals("java")) {
```
dwightguth, line 225: This method can be cleaned up by removing every line following the point where something other than node is returned, and just returning node.

build.xml
liyili2, line 4: I do not understand this.

liyili2, line 4: I guess I will copy it from unified-building class.

src/javasources/KTool/src/org/kframework/compile/utils/MetaK.java
liyili2, line 92: Please do not. I want to finish something first. I do not want to do other people's job.

Why I need to worry about different backends? I think the job I have now is to divide the basicBackend into two stages instead of caring about all backends. There are a lot of backends who is doing this.

src/javasources/KTool/src/org/kframework/compile/transformers/FlattenTerms.java
liyili2, line 225: I do not know how to do it.

I want to do other things instead of cleaning.

src/javasources/KTool/src/org/kframework/backend/BasicBackend.java
liyili2, line 92: I agree with this.

src/javasources/KTool/src/org/kframework/compile/transformers/FlattenTermsTwo.java
liyili2, line 1: What is the meaning of reverted?

I do not understand this.

In addition, FlattenTermsTwo is different from FlattenTerms step

src/javasources/KTool/src/org/kframework/backend/BasicBackend.java
liyili2, line 92: The below tests show that it is not the cases that you just need to move few codes into FlattenSyntax. I guess We should merge FlattenTerms with FlattenSyntax completely. I cannot see the advantage to do that instead of putting FlattenTermsTwo before FlattenSyntax.

liyili2, line 92: After I move resolveListOfK beyond FlattenTerms step. Every works magically.  Based on your suggestion of removing the FlattenTermsTwo step and moving the stuff about variable into FlattenSyntax step.

liyili2:
This does not make sense. The version f07892b and 385d22f are exact the same

Why 385d22f is succeeded but f07982b is out of time?

liyili2:
It does not make sense. f08982b are the same as version 385d22f. Why is the case 385d22f succeeded but f08982b fails.

dwightguth:
closing these in favor of #484

PR #467, opened 2014-05-16
fix #parse thing, oh god for real finally geez

dwightguth:
@radumereuta please review

Also, please try to review diffs more carefully in the future. Once you look at a diff for the first time, review changes incrementally instead of trying to review the entire diff again. In this case, last time I tried to fix this bug, you missed the fact that I had accidentally reverted the bug fix itself, and since my test that I thought covered it actually didn't, I failed to realize it and ended up not fixing the bug. You would have noticed if you had seen the incremental revision of the diff the second time because it would have contained a line modification that should not have been there, but I suspect you missed it because instead you viewed the diff as a whole, where there was nothing to see except the absence of something, which is much harder to detect.


radumereuta:
lgtm

PR #470, opened 2014-05-19
fixed bug in I/O builtins when returning K

andreistefanescu:


dwightguth:
Test for this?

yilongli:
Minor thing: is it better to use something like `KSequence.EMPTY` and `KList.EMPTY` instead of constructors?

andreistefanescu:
@dwightguth I don't think a test makes sense for this. It's a minor change.
@yilongli Yes, I agree. We should do that in a next commit.

Issue #475, opened 2014-05-19
Warnings about undeclared variables


chathhorn:
kompile seems to be warning about all undeclared variables, even if there has been a previous declaration:

module BUG
rule X:Int => X
endmodule

kompile -w all bug.k

Produces:
[Warning] Compiler: Variable 'X' was not declared. Assuming sort Int and expected sort Int.
      File: /home/chathhorn/scratch/undef/bug.k
      Location: (3,20,3,21)  

(in addition to 414 more warnings due to the undeclared variables in the builtins)

PR #478, opened 2014-05-20
Fixing issue #475. I've made a change a while back to fix latex,

radumereuta:
but that broke warning reporting for typed variables.

Also, unfortunately I can't add a regression test, because ktest doesn't allow me to compare the output of kompile.


dwightguth:
Make a unit test then. Something minimal that is expected to throw a
KEMException.

dwightguth:
No sorry I guess it would assert something about a KException, since it's a
warning and not an error.

radumereuta:
I'm not sure how to do that.

radumereuta:
Also. A test would not fit in the JUnit philosophy of being fast.
I have to run the entire kompile in order to properly test this (well... the expensive part at least).

dwightguth:
I don't understand. If the warning is generated by VariableTypeInferenceFilter when you have a certain type of rule, just construct such a rule as input, then call VariableTypeInferenceFilter on it. You can mock any dependencies necessary. What's the complexity here?

radumereuta:
@dwightguth  please review. I think it's ready for merge.

Issue #481, opened 2014-05-22
LTL model checking don't work for negative results


denis-bogdanas:
Steps to reproduce:
- Download php semantics from http://www.phpsemantics.org/
- cd kphp-sources\
- kompile php.k --transition step
- krun --debug-info --parser="java -jar parser/parser.jar" --ltlmc="<>Ltl(eqTo(gv(variable(\"x\")),val(\"world\")))" examples/hello-world-ltl.php

Result:

[Error] Critical: Cannot parse result xml from maude. If you believe this to be
in error, please file a bug and attach
D:\kphp-sources.k\krun8698076417810780425\maudeoutput.xml

The following, positive formula works instead:
krun.bat --debug-info --parser="java -jar parser/parser.jar" --ltlmc="<>Ltl(eqTo(gv(variable(\"x\")),val(\"hello\")))" examples/hello-world-ltl.php

platform: windows

dwightguth:
I will fix this tomorrow. Long past time we had a regression test for this.

denis-bogdanas:
Looks like Danielle had the same problem. Here he presents a workaround (section LTL model checking ):
http://www.doc.ic.ac.uk/~maffeis/phpsemantics/documentation.html

dwightguth:
This is fixed. @denis-bogdanas , can you verify and close?

PR #483, opened 2014-05-22
fix issue #481

dwightguth:


dwightguth:
@denis-bogdanas took a look at this and it makes sense to him, so I am going to merge this if the tests pass.

Issue #428, opened 2014-05-08
Exception in thread "pool-1-thread-1" org.kframework.utils.errorsystem.KExceptionManager$KEMException


dwightguth:
$ cat test.k
module TEST

syntax Exp ::= Int | Exp "+" Exp | foo(Int)

rule 5 => #parse("A:Exp + foo(A)", "Exp")

endmodule
$ cat a.test
5
$ krun --parser "kast --parser rule" a.test
Exception in thread "pool-1-thread-1" org.kframework.utils.errorsystem.KExceptionManager$KEMException
    at org.kframework.utils.errorsystem.KExceptionManager.register(KExceptionManager.java:23)
    at org.kframework.kil.visitors.exceptions.TransformerException.report(TransformerException.java:24)
    at org.kframework.parser.concrete.disambiguate.VariableTypeInferenceFilter.visit(VariableTypeInferenceFilter.java:70)
    at org.kframework.parser.concrete.disambiguate.VariableTypeInferenceFilter.visit(VariableTypeInferenceFilter.java:1)
    at org.kframework.kil.Sentence.accept(Sentence.java:141)
    at org.kframework.kil.AbstractVisitor.visitNode(AbstractVisitor.java:95)
    at org.kframework.parser.DefinitionLoader.parsePattern(DefinitionLoader.java:359)
    at org.kframework.parser.ProgramLoader.processPgm(ProgramLoader.java:116)
    at org.kframework.krun.RunProcess.runParser(RunProcess.java:139)
    at org.kframework.krun.ioserver.commands.CommandParse.run(CommandParse.java:32)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
    at java.lang.Thread.run(Thread.java:744)

This is because GlobalSettings.kem.register is being called from inside the IO server when parsePattern is called and TypeSystemFilter throws a TransformerException. Make sure you create a regression test for this when you fix the bug.

PR #486, opened 2014-05-22
Fixing issue #428.

radumereuta:
- errors found in VariableTypeInference were reported immediately instead being thrown further.
- fixed some typos.


radumereuta:
@dwightguth I've tested everything about the parser, and it works as expected. The only things that I don't understand, are left in Krun. I don't feel comfortable enough to work on krun. Can you have a look, and fix the test?

I found that are various inconsistencies when calling krun with --parser.

radumereuta:
This is actually really strange. The tests pass on jenkins. It must be something wrong with my windows machine then.

dwightguth:
If I had to hazard a guess I'd say it's probably because the successful execution of this test depends on environment variables, and Windows is messing it up somehow. I suggest you try to create a minimal reproducible example and then create an issue. I may not have time to get to it right away however.

This looks good to me right now, so I am merging this.

radumereuta:
You can test it with the example I just added. Ktest fails for it.

dwightguth:
If you have an example that breaks the test, can you re-create the pull
request containing that example? As you can see, since it worked when you
ran ktest, I assumed it was a windows-specific thing and went ahead and
merged.

radumereuta:
I wanted to say, that if you run ktest on windows for this example, it fails. Although it shouldn't.
So the negative test should be for you to run it on windows.

dwightguth:
Per my discussion with Radu, fixing this would require a significant
overhaul of the internal vs. external parsing framework. it's on my todo
list, but way low, after a bunch of other stuff. You can make an issue and
assign it to me though if you want.

PR #490, opened 2014-05-22
Added description for automatic keyword rejection in to-be-processed.txt.

radumereuta:
Anyone who speaks better English than me, please review.


dwightguth:
Please also make a note that "syntax Blah ::= "blah" [reject]" is just syntactic sugar for hooking into this process, and does not restrict rejection to only the Blah sort.

radumereuta:
Ok, I've added a few more lines, but do we want that?
I think [reject] should reject only the sort you specified.
I think right now it's a limitation on how I implemented it.

dwightguth:
Okay. Then make an issue for it, and leave it out of the manual. I agree
with you, I just thought you didn't.

radumereuta:
@dwightguth please review. I've decided to just fix it right now. It would save us a lot of problems in the future.

Also, can you have a look and see why checkstyle-copyright fails? I've put the proper comments.

dwightguth:
You have a space after the period. Otherwise looks good to me, and you can merge if tests pass.

PR #492, opened 2014-05-22
Made it so timing.log is opened each time I want to write something in it

radumereuta:
Since this is called only when --debug is provided to kompile, it shouldn't be a time penalty for normal usages.

@dwightguth please review.

### DIFF
--- a/src/javasources/KTool/src/org/kframework/parser/generator/ParseConfigsFilter.java
+++ b/src/javasources/KTool/src/org/kframework/parser/generator/ParseConfigsFilter.java
@@ -107,10 +107,8 @@ public class ParseConfigsFilter extends ParseForestTransformer {
                 config = new AmbFilter(context).visitNode(config);
 
                 if (globalOptions.debug) {
-                    try {
-                        Formatter f = new Formatter(new FileWriter(context.dotk.getAbsolutePath() + "/timing.log", true));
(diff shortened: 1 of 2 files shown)


src/javasources/KTool/src/org/kframework/parser/generator/ParseConfigsFilter.java
```diff
@@ -101,6 +106,15 @@ public ASTNode visit(StringSentence ss, Void _) throws ParseFailedException {
                 // last resort disambiguation
                 config = new AmbFilter(context).visitNode(config);
 
+                if (globalOptions.debug) {
+                    try {
```
dwightguth: this needs to be a try with resources:

```
try (Formatter f = ....) {
...
} catch (IOException e) {
...
}
```

src/javasources/KTool/src/org/kframework/parser/generator/ParseRulesFilter.java
```diff
@@ -134,9 +121,14 @@ else if (ss.getType().equals(Constants.RULE))
                 // last resort disambiguation
                 config = new AmbFilter(context).visitNode(config);
 
-                if (globalOptions.verbose) {
-                    f.format("Parsing rule: Time: %6d Location: %s:%s\n", (System.currentTimeMillis() - startTime), ss.getFilename(), ss.getLocation());
-                    f.flush();
+                if (globalOptions.debug) {
+                    try {
```
dwightguth: see above

PR #493, opened 2014-05-22
Unified builtins fresh

andreistefanescu:


src/javasources/KTool/src/org/kframework/backend/java/kil/Token.java
```diff
@@ -19,6 +22,17 @@
  */
 public abstract class Token extends Term {
 
+    public static Token of(String sort, String value) {
+        switch (sort) {
+            case BoolToken.SORT_NAME:
```
dwightguth, line 27: why only Bool and Int here?

src/javasources/KTool/src/org/kframework/backend/java/kil/Variable.java
```diff
@@ -160,13 +160,13 @@ public ASTNode accept(Transformer transformer) {
      * Renames serialized anonymous variables to avoid name clashes with existing anonymous
      * variables.
      */
-    private Object readResolve() {
+    Object readResolve() {
```
dwightguth, line 163: this should stay private if you're not calling it somewhere else.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
```diff
@@ -669,6 +650,18 @@ public ASTNode visit(org.kframework.kil.Rule node, Void _)  {
         return rule;
     }
 
+    private void transformConjunction(Collection<Term> requires, Term term) {
```
dwightguth, line 653: since you call this with ensures as well as requires, can you rename this parameter?

src/javasources/KTool/src/org/kframework/compile/utils/MetaK.java
```diff
@@ -87,6 +87,7 @@ public static String cellUnit(String cellName) {
         public static final String heatingTag = "heat";
         public static final String coolingTag = "cool";
         public static final String hole = "[]";
+        @Deprecated
         public static final String freshCons = "Bool1FreshSyn";
```
dwightguth, line 91: is this still used anywhere?

src/javasources/KTool/src/org/kframework/kil/Variable.java
```diff
@@ -16,7 +15,8 @@
     private String name;
     /** True if the variable was written with an explicit type annotation */
     private boolean userTyped = false;
-    private boolean fresh = false;
+    private boolean freshVariable = false;
```
dwightguth: I mislike doing it this way. There are constructors you didn't modify which will ignore this information. We already have problems because of things like this. Is there a reason you couldn't do something like:

```
private Boolean freshVariable;
public boolean isFreshVariable() {
    if (freshVariable == null) {
        freshVariable = name.startsWith("?");
    }
    return freshVariable;
}
```

?

tests/regression/config.xml
```diff
@@ -66,23 +66,33 @@
...
+    </all-programs>
+  </test>
+
   <include file="java-rewrite-engine/config.xml" directory="java-rewrite-engine" />
 
   <!-- Regression Tests -->
   <test definition="issue552/issue.k" />
   <test definition="issue529/issue.k" />
   <test definition="defaultTerms1/issue.k" />
+<!-- TODO(Traian) uncomment this test once !X works in Maude
```
dwightguth, line 88: what is broken about this test? I would feel better if we didn't comment out tests with the assumption that they would be uncommented later; history demonstrates that in practice this doesn't happen, and then we have regressions.

src/javasources/KTool/src/org/kframework/kil/Variable.java
radumereuta: What constructor are you referring to?
I've asked Andrei to strip the "?" and "!" from the beginning of the variable so I can still run the CheckVariablesFilter correctly. This means that "?" and "!" is a semantic modifier to the variable, and is not part of the name.

The result of this is that a rule like:
rule A => A +Int !A
will produce an error message. I think this is safer and less likely for users to make a mistake.
But if more people disagree with this, we could change it (I know that there are tools which consider the annotations as part of the name).

grosu: !A and ?A are the actual names of the variables.  So if you want to reuse the same variable several times in a rule, each of them should use the ! or the ?.

I agree with you that it is a good idea to report an error message if somebody tries to use both A and !A (or ?A) in the same rule.  However, we do not need to report an error if they use both !A and ?A, but no A.

Grigore

radumereuta: Glad to report that it doesn't fail in the CheckVariablesFilter, but I also tried this:

```
module TEST
  rule .K => ?A ~> !A
endmodule
```

 and I got the following error message:

```
[Warning] Compiler: Module K-FRESH undefined.
        File: D:\work\k\include\modules\substitution.k
        Location: (8,3,8,17)
        Compilation Phase: class
org.kframework.compile.FlattenModules$FlattenModulesVisitor
```

@andreistefanescu any idea why?

dwightguth: My concern here is something different than Grigore's, although I agree with Grigore.

The problem with this solution is it creates a fragmented authority for determining whether a variable in a rule is ! or ?. I look at the freshVariable flag and I don't know if it tells me the correct information, because I could be looking at a variable whose flag was set incorrectly by one of the other constructors.

In my opinion, the variable class, outside the frontend, should have the following fields only:

```
String name; //name of variable, including ? or !
String declaredSort; //EXPLICIT sort of variable. set only if the variable should be runtime checked to be that sort
String sort; //inherited from Term, set by type inferencer to concrete syntactic sort of variable.
```

radumereuta: I'm not sure exactly what's the best practice here. My concern is with past experiences related to # in front of sort names. It wasn't a very reliable way to query if a sort is lexical or not.
Plus, if we are to change it now, CheckVariablesFilter needs to be modified, and maybe other steps too.

dwightguth: How about a compromise? The fields freshVariable and freshConstant and name become blank final variables, and so each constructor has to initialize them.

src/javasources/KTool/src/org/kframework/backend/java/kil/Token.java
andreistefanescu, line 27: Because we do not have float yet (you are working on it), and the machine integers do not have a string representation. The remaining Tokens are uninterpreted.

dwightguth, line 27: String?

On Fri, May 23, 2014 at 1:14 PM, Andrei Stefanescu <notifications@github.com

src/javasources/KTool/src/org/kframework/kil/Variable.java
radumereuta: I think this is actually a good solution.

src/javasources/KTool/src/org/kframework/backend/java/kil/Variable.java
andreistefanescu, line 163: This is actually a very tricky bug: readResolve is not called anywhere, it is used during java deserialization, and if marked public, it is not inherited by the subclasses of Variables.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
andreistefanescu, line 653: Sure.

src/javasources/KTool/src/org/kframework/compile/utils/MetaK.java
andreistefanescu, line 91: It is used in some dead code. I will eliminate it next week, as part of a separate commit.

tests/regression/config.xml
andreistefanescu, line 88: This test is not broken, you can find it above for the java backend; @traiansf is going to implement this functionality for Maude, and then uncomment the test.

src/javasources/KTool/src/org/kframework/backend/java/kil/Variable.java
dwightguth, line 163: okay, fine. Make it final then, and make a comment.

On Fri, May 23, 2014 at 1:16 PM, Andrei Stefanescu <notifications@github.com

tests/regression/config.xml
dwightguth, line 88: Not my point. Fresh worked before in both backends. It should continue to
work in both backends indefinitely. If that means you need to wait until
Traian implements it in Maude to commit this, so be it.

On Fri, May 23, 2014 at 1:20 PM, Andrei Stefanescu <notifications@github.com

src/javasources/KTool/src/org/kframework/kil/Variable.java
grosu: I am assuming your are talking about the KAST classes here, which are only used by the kompiler.  That's because we do not want to spend any runtime overhead to read the first character in a string to see whether the variable is fresh or universally quantified.  The Java rewrite engine should encode the rules the possible way it can, so that it can quickly find this information without deciphering strings.  I'm sure that's the way it is, but I just re-iterate it to be safe.

Grigore

PR #497, opened 2014-05-23
Floats

dwightguth:
Recreating this pull request against the new branch.


src/javasources/KTool/src/org/kframework/backend/java/builtins/BuiltinFloatOperations.java
```diff
@@ -16,198 +14,196 @@
...
+        return term1.exponent();
     }
-
-     public static Term mul(Term term1, Term term2, TermContext context) {
-        if (term1.equals(UninterpretedToken.of("#Float","0.0")) ||
-                term2.equals(UninterpretedToken.of("#Float","0.0")))
-            return UninterpretedToken.of("#Float","0.0");
-        return null;
+    
+    public static IntToken precision(FloatToken term, TermContext context) {
```
radumereuta, line 56: I'm sorry for making you do a lot of work, but I would like javadoc on all of these functions. Although they are pretty standard, I am more worried about the differences.

tests/regression/config.xml
```diff
@@ -344,9 +333,9 @@
   </test>
 
 <test
-      definition="tests/regression/issue#428/test.k"
-      programs="tests/regression/issue#428"
-      results="tests/regression/issue#428"
+      definition="issue#428/test.k"
```
radumereuta, line 348: Why is this part of this change?

dwightguth, line 348: Oh, it's not. I merged master into unified builtins2 and accidentally broke
the branch. I had to commit an unreviewed fix amounting to this exact patch
hunk. You can ignore this, it will get taken care of by the time I fix the
broken tests in this pull request.

dwightguth:
Jenkins: retest this please

dwightguth:
Jenkins: retest this please

include/builtins/float.k
```diff
@@ -3,8 +3,15 @@ require "bool.k"
 require "int.k"
 module FLOAT-SYNTAX-HOOKS
   imports #FLOAT-INTERFACE
-  syntax #Float ::= Token{[\+\-]?[0-9]+"."[0-9]+([eE][\+\-]?[0-9]+)?[fFdD]?}
-                 [regex("([\\+-]?\\d(\\.\\d*)?|\\.\\d+)([eE][\\+-]?(\\d(\\.\\d*)?|\\.\\d+))?[fFdD]?")]
+  syntax #Float ::= Token{[\+\-]?[0-9]+"."[0-9]+([eE][\+\-]?[0-9]+)?([fFdD]|([pP][0-9]+[xX][0-9]+))?}
```
andreistefanescu, line 6: I think there is no point in having #Float anymore. Float should suffice.

dwightguth:
Jenkins: test this please

include/builtins/float.k
radumereuta, line 6: I agree. I tried to remove #Float, but it's too deeply embedded in the backends, so I gave up. If you can do it. please remove it.

dwightguth, line 6: This should be a separate commit. I did a few cleanup things regarding this, but I don't want this change to make a bunch of modifications to String/Id/Bool/Int, which the change you are proposing would require.

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

src/javasources/KTool/src/org/kframework/kil/FloatBuiltin.java
```diff
@@ -0,0 +1,182 @@
...
+    private final BigFloat value;
+    private final int exponent;
+    
+    private FloatBuiltin(BigFloat value, int exponent) {
+        this.value = value;
+        this.exponent = exponent;
+    }
+
+    private static Pattern precisionAndExponent = Pattern.compile("(.*)[pP](\\d+)[xX](\\d+)");
+    public static Pair<BigFloat, Integer> parseKFloat(String s) {
```
andreistefanescu, line 74: why does this not return a FloatBuiltin?

dwightguth, line 74: I am calling this function also from FloatToken.of. Would you rather
FloatToken took a constructor of a FloatBuiltin? I could do it that way too.

andreistefanescu:
Looks good. You can merge.

src/javasources/KTool/src/org/kframework/kil/FloatBuiltin.java
andreistefanescu, line 74: Yes, I would prefer that.

dwightguth, line 74: I am keeping this as is per our discussion because it needs to be called from a FloatBuiltin constructor.

Issue #402, opened 2014-04-25
incorrect handling of streams cells in java rewrite engine


dwightguth:
```
$ cd tests/regression/java-rewrite-engine/io/test2
$ kompile test.k --backend java
[Warning] Compiler: Expecting a list item but got class
org.kframework.kil.KApp.
Won't transform.
    File: /home/dwightguth/k/tests/regression/java-rewrite-engine/io/test2/test.k
    Location: (21,11,21,31)
    Compilation Phase: Resolve Blocking Input
[Warning] Compiler: Expecting a list item but got class
org.kframework.kil.KApp.
Won't transform.
    File: /home/dwightguth/k/tests/regression/java-rewrite-engine/io/test2/test.k
    Location: (23,11,23,28)
    Compilation Phase: Resolve Blocking Input
[Warning] Compiler: Expecting a list item but got class
org.kframework.kil.KApp.
Won't transform.
    File: /home/dwightguth/k/tests/regression/java-rewrite-engine/io/test2/test.k
    Location: (25,11,25,30)
    Compilation Phase: Resolve Blocking Input
$ krun programs/readInt.test
    [0, 25.48 ms]
    <T>
        <k>
            read ( int )
        </k>
        <in>
            ListItem(#buffer ( "" ))
            ListItem(#istream ( 0 ))
        </in>
        <out>
            ListItem(#ostream ( 1 ))
            ListItem(#buffer ( "" ))
        </out>
    </T>
```

This should work without warnings and the programs should accept input on stdin. But it doesn't.

PR #501, opened 2014-05-24
fixing #402 in branch Unified builtins2

yilongli:
fixing https://github.com/kframework/k/issues/402


yilongli:

include/io/io.k
```diff
@@ -22,12 +22,16 @@ module IO
...
        (ListItem(S:String) => .List)
        _:List [stdout, stderr]
+
   rule ListItem(#buffer(Buffer:String => Buffer +String Float2String(F)))
        (ListItem(F:Float) => .List)
        _:List [stdout, stderr]
   rule ListItem(#buffer(Buffer:String => Buffer +String Int2String(I)))
        (ListItem(I:Int) => .List)
        _:List [stdout, stderr]
+  rule ListItem(#buffer(Buffer:String => Buffer +String Char2String(C)))
```
dwightguth, line 32: Do we even _have_ a Char sort?

dwightguth:
This diff is really hard to read because of the way it mixes together whitespace changes, refactorings, and the bugfix itself. I have no way of knowing whether or not I have adequately reviewed it in this state because it's too hard to be sure.

Can you please remake this diff from scratch with more care? To start with, disable all autoreformatting of code, because that causes most of the noise in this diff and we don't really gain anything by doing it right now. We can discuss making this a hard code style guideline if you want, but there are correct ways of doing so and this isn't it.

Then, please separate out the bugfix and its tests from the refactorings you did, and put them in conceptually distinct diffs. For example, one which refactors ResolveBlockingInput, one which changes new KSequence to KSequence.EMPTY, etc.

Not only is this a very good practice to get into, but it also improves code quality by making it easier for me to review code effectively and not miss things in all the jumbled noise of the diff.

include/io/io.k
traiansf, line 32: We used to have something like this and it was used by the C semantics for
doing char-level I/O access.

The idea was that if one tries to read a string from a stream cell, they
would get a token;  if they ask for a Char, they would get the next
character.

  Not sure whether it still works, though.

2014-05-27 18:22 GMT+03:00 dwightguth notifications@github.com:

yilongli, line 32: I don't know. But I saw one above and figured that it might be a missing case. https://github.com/yilongli/k/blob/unified-builtins2/include/io/io.k#L17

andreistefanescu:
lgtm

PR #507, opened 2014-05-28
Improve sdf

radumereuta:


dwightguth:
Just to satisfy my need to be sure that this is the right solution, can you please provide me with a quick example of SDF code which contains the broken follow restriction, and an input and output to a parse for which JSGLR returns incorrect results?

radumereuta:
I managed to reduce the problem to only this:
context-free syntax
    K "==" K -> K {cons("EqEq")}
    K "==Set" K -> K {cons("EqSet")}
    K "==String" K -> K {cons("EqString")}
    K K -> K {cons("KK")}

lexical restrictions
    "==" -/- [S].[t].[r].[i].[n].[g]
    "==" -/- [S].[e].[t]

The problem is that SDF only allows one follow restriction, and it takes the first one.
Tried piping, and it works in the same way.
I don't see a way around this.

dwightguth:
I have no idea how to solve this either. This is as good a solution as any.

PR #509, opened 2014-05-28
Trying to improve the amount of PDFs that are generated in ktest.

radumereuta:
Don't review yet, I'm trying to see if the tests pass, or if the right amount of PDFs are skipped.

### DIFF
--- a/tests/config.xml
+++ b/tests/config.xml
@@ -41,7 +41,6 @@
       definition="samples/quine/explicit/quine-explicit.k"
       programs="tests/examples/quine-explicit"
       results="tests/examples/quine-explicit"
-	  skip="pdf"
       extension="quine" >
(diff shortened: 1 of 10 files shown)


radumereuta:
@dwightguth please review.

tests/config.xml
```diff
@@ -41,6 +41,7 @@
       definition="samples/quine/explicit/quine-explicit.k"
       programs="tests/examples/quine-explicit"
       results="tests/examples/quine-explicit"
+	  skip="pdf"
```
dwightguth: shouldn't you keep this one?

radumereuta: Not really sure. Anyway, you can't reproduce the example from the PDF, since every single space matters in this definition.

PR #512, opened 2014-05-29
Java builtin (user-level) substitution

traiansf:
This push request contains an implementation of an user-level substitution mechanism for the Java backend.

Note however that it cannot be used directly instead of the one used in substitution.k because the java engine cannot process that file (and those depending on it).  

I have something written to address that in a separate branch (see  aeff8a8afa9ebf3dc4316d972b3a55a9308f9101 );  however, I would like to push this first as the substitution works and I have a provided a regression test for it.


tests/configjavarw.xml
```diff
@@ -499,5 +499,16 @@
       <krun-option name="-backend" value="java" />
     </all-programs>
   </test>
+
+   <test
```
andreistefanescu: import user_substitution/tests/config.xml

andreistefanescu:
Looks good.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/UserSubstitutionTransformer.java
```diff
@@ -0,0 +1,117 @@
...
+            assert kItem.kList() instanceof KList : "KList must be concrete";
+
+            KLabel kLabel = (KLabel) kItem.kLabel();
+            KList kList = (KList) kItem.kList();
+            if (kLabel instanceof KLabelConstant) {
+                KLabelConstant kLabelConstant = (KLabelConstant) kLabel;
+                if (kLabelConstant.isBinder()) {
+                    assert kList.getContents().size()==2 && !kList.hasFrame() :
+                            "Only supporting binders of the form lambda x. e for now";
+                    Term boundVar = kList.get(0);
```
dwightguth: this makes me a little nervous... is there a reason why using the preexisting binder attributes would be too complex to implement now?

tests/configjavarw.xml
```diff
@@ -1,503 +0,0 @@
-<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
dwightguth, line 1: can you confirm that all of these tests are captured elsewhere? I'm pretty sure they're not...

tests/regression/java-rewrite-engine/user_substitution/lambda.k
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2012-2014 K Team. All Rights Reserved.
+
+module LAMBDA
+  syntax K ::= K "[" K "/" K "]" [hook(Substitution:userSubstitution)]
```
dwightguth, line 4: How much work would be involved in order to make the interface for substitution be the same for Maude and Java? In some sense, that is one of the goals of the unified builtins branch, that the same definition be compilable with both backends. I am willing to let this slip if we have to as long as everything that works now in master continues to work and is tested, but I want to at least know what work is entailed before I make that decision.

tests/configjavarw.xml
traiansf, line 1: Ok, I'm going to revert this file.  The reason was that these tests are not used for testing (Jenkins) and so their presence there is confusing (at least for me it was)

traiansf:
I've added code to make substitution work with generalized binders, but I'm afraid there is no current test for them for either Maude or Java.

tests/regression/java-rewrite-engine/user_substitution/lambda.k
traiansf, line 4: I would like to merge this first and then complete the work on the unified-hooks branch. It should not take more than a couple of days. I would say hours, but I said days to be on the safe side.

dwightguth:
Looks like you have a nontermination bug.

traiansf:
Yep, apparently the binder mappings are numbered starting from 1 (as they
represent positions in under the binder).  I wonder whether it would be a
good idea to rebase them at 0 at parsing time, so that they are consistent
with Java's way of indexing collections.

2014-06-03 18:08 GMT+03:00 dwightguth notifications@github.com:

src/javasources/KTool/src/org/kframework/backend/java/symbolic/UserSubstitutionTransformer.java
```diff
@@ -0,0 +1,121 @@
...
+            KLabel kLabel = (KLabel) kItem.kLabel();
+            KList kList = (KList) kItem.kList();
+            if (kLabel instanceof KLabelConstant) {
+                KLabelConstant kLabelConstant = (KLabelConstant) kLabel;
+                if (kLabelConstant.isBinder()) {
+                    List<Term> termList = new ArrayList<>();
+                    termList.addAll(kList.getContents());
+                    Multimap<Integer, Integer> binderMap =  kLabelConstant.getBinderMap();
+                    for (Integer key : binderMap.keySet()) {
+                        Term boundVar = termList.get(key - 1);
```
dwightguth: comment here about why -1?

dwightguth: same with value-1

dwightguth:
just the minor things here about documenting and then you can merge this.

traiansf:
I tried to add a test for generalized binders and found pretty severe problems with the substitution, both from a stand-alone point of view, but also from interaction with the meta-substitution.
I've changed the code computing binderMap as all code using it was assuming it starts at 0.

@andreistefanescu @dwightguth  please re-review.

src/javasources/KTool/src/org/kframework/backend/java/kil/KLabelConstant.java
```diff
@@ -181,9 +181,17 @@ public TermContext termContext() {
...
 
+    public boolean isMetaBinder() {
+        return hasAttribute("metabinder");
+    }
+
     public boolean isBinder() {
+        return hasAttribute("binder");
+    }
+
+    private boolean hasAttribute(String attribute) {
```
dwightguth, line 183: this feels a bit strange to me. I'd feel better if a check were put in place to ensure that all productions with the  same klabel had the same binder attribute. But then, it's Andrei's code, not mine. And I don't feel that strongly. So it's up to him.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/UnboundedTermsCollector.java
```diff
@@ -0,0 +1,78 @@
...
+import com.google.common.collect.HashMultiset;
+import com.google.common.collect.Multimap;
+import com.google.common.collect.Multiset;
+import org.kframework.backend.java.kil.*;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+* Created by Traian on 04.06.2014.
```
dwightguth: okay sure but what's it do?

src/javasources/KTool/src/org/kframework/backend/java/symbolic/UnboundedTermsCollector.java
```diff
@@ -0,0 +1,78 @@
...
+            }
+        });
+     }
+
+    private void handleBinderVariables(KItem kItem, boolean add) {
+        // TODO(AndreiS): fix binder when dealing with KLabel variables and non-concrete KLists
+        if (!(kItem.kLabel() instanceof KLabel) || !(kItem.kList() instanceof KList)) {
+            return;
+        }
+        assert kItem.kLabel() instanceof KLabel : "KLabel variables are not supported";
```
dwightguth: are these two lines still needed? they seem to be unreachable.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/UserSubstitutionTransformer.java
```diff
@@ -0,0 +1,157 @@
...
+            super(context);
+        }
+
+        @Override
+        public ASTNode transform(KItem kItem) {
+            // TODO(AndreiS): fix binder when dealing with KLabel variables and non-concrete KLists
+            if (!(kItem.kLabel() instanceof KLabel) || !(kItem.kList() instanceof KList)) {
+                return super.transform(kItem);
+            }
+            assert kItem.kLabel() instanceof KLabel : "KLabel variables are not supported";
```
dwightguth: same here

dwightguth:
Did nothing need to be changed on the Maude code path when you changed binders to start at 0?

src/javasources/KTool/src/org/kframework/backend/java/kil/KLabelConstant.java
andreistefanescu, line 183: This goes for a few other attributes, like strict and function. We should have a check in kompile that ensures productions have consistent attributes. @dwightguth can you take care of it?

dwightguth, line 183: Okay, that makes sense. I have made an issue and assigned it to me, I will
try to get to it sometime this week. This is fine for now. If you fix the
minor issues and answer my question, you can merge, Traian.

traiansf:
Nobody was using those functions in the Maude code path, apparently...

2014-06-09 17:06 GMT+03:00 dwightguth notifications@github.com:

src/javasources/KTool/src/org/kframework/backend/java/kil/KLabelConstant.java
traiansf, line 183: I think here there was a case when people were redeclaring lambda to say
it's a result.  Maybe this is what Andrei tried to address here?

2014-06-09 17:02 GMT+03:00 dwightguth notifications@github.com:

src/javasources/KTool/src/org/kframework/backend/java/symbolic/UnboundedTermsCollector.java
traiansf: I'm not totally convinced they are unreachable... they used to be allowed to be variables, in which case we didn't want to handle them. @andreistefanescu  what do you think about this?

dwightguth: but look at the lines above. if either of these conditions is false, the function returns.

traiansf: Yes.  this means we simply ignore Kitems where either the label or the list
is a variable.

2014-06-10 18:06 GMT+03:00 dwightguth notifications@github.com:

PR #517, opened 2014-05-29
enhanced builtin equality for symbolic terms; fix for immutable SymbolicConstraint#Equality

andreistefanescu:


src/javasources/KTool/src/org/kframework/backend/java/builtins/TermEquality.java
```diff
@@ -10,22 +14,57 @@
 public class TermEquality {
 
     public static BoolToken eq(Term term1, Term term2, TermContext context) {
```
yilongli, line 16: some Javadoc regarding the return value?

src/javasources/KTool/src/org/kframework/backend/java/builtins/TermEquality.java
```diff
@@ -10,22 +14,57 @@
...
+                return BoolToken.TRUE;
+            case FALSE:
+                return BoolToken.FALSE;
+            default:
+                return null;
+        }
     }
 
     public static BoolToken ne(Term term1, Term term2, TermContext context) {
-        checkArguments(term1, term2);
```
yilongli, line 18: same as above

yilongli:
lgtm.

dwightguth:
I admit I'm a little bit lost here... it looks like you are moving some functionality surrounding solving constraints into BuiltinBoolOperations? Is this really the design we want? I thought that the BuiltinFooOperations were just supposed to be concrete execution and then symbolic stuff happened elsewhere? Maybe I'm just misunderstanding, but I'd like to know what the design is here.

andreistefanescu:
I am not moving the functionality, I am simply making it available in the hook. Yes, I am currently extending the functionality of this hook to handle also symbolic terms, but I don't think this is a problem, at least not now.

dwightguth:
Okay, but either all hooks should handle symbolic stuff, or none of them should, right? Are you prepared to make that commitment?

andreistefanescu:
For now, this is a best effort approach. I don't know how to write this from K in a reasonable fashion, and I need this functionality. Later, we can review this policy.

andreistefanescu:
After my discussing with Dwight, I am merging this, and we may revise it later,

Issue #519, opened 2014-05-30
ktest uncaught exception


grosu:
Arguments need -- instead of -, but better error message needed:

```
fslwork$ ktest tests/config.xml -programs=. -results=.
org.apache.commons.cli.UnrecognizedOptionException: Unrecognized option: -programs=.
    at org.apache.commons.cli.Parser.processOption(Parser.java:363)
    at org.apache.commons.cli.Parser.parse(Parser.java:199)
    at org.apache.commons.cli.Parser.parse(Parser.java:85)
    at org.kframework.ktest.CmdArgs.CmdArgParser.<init>(CmdArgParser.java:81)
    at org.kframework.ktest.KTest.<init>(KTest.java:36)
    at org.kframework.ktest.KTest.main(KTest.java:107)
    at org.kframework.main.Main.main(Main.java:58)
Exception in thread "main" java.lang.NullPointerException
    at org.kframework.ktest.KTest.main(KTest.java:112)
    at org.kframework.main.Main.main(Main.java:58)
```

traiansf:
What I did for RVPredict was to catch the exception and simply print the
message, which would be Unrecognized option: -programs=.
then to exit.

2014-05-30 5:25 GMT+03:00 Grigore Rosu notifications@github.com:

daejunpark:
It seems to be same with the issue #485, and Omer gave a pull request #522.
Will it be also fixed with the pull request?

osa1:
Yes.

osa1:
Fixed with #522.

PR #522, opened 2014-05-30
fix exception reporting in ktest

### LINKED ISSUE #485: ktest throws exception for wrong typed option
```
$ ktest config.xml --update out
org.apache.commons.cli.UnrecognizedOptionException: Unrecognized option: --update
    at org.apache.commons.cli.Parser.processOption(Parser.java:363)
    at org.apache.commons.cli.Parser.parse(Parser.java:199)
    at org.apache.commons.cli.Parser.parse(Parser.java:85)
    at org.kframework.ktest.CmdArgs.CmdArgParser.<init>(CmdArgParser.java:81)
    at org.kframework.ktest.KTest.<init>(KTest.java:36)
    at org.kframework.ktest.KTest.main(KTest.java:107)
    at org.kframework.main.Main.main(Main.java:58)
Exception in thread "main" java.lang.NullPointerException
    at org.kframework.ktest.KTest.main(KTest.java:112)
    at org.kframework.main.Main.main(Main.java:58)
```

osa1:
Fixes #485

### DIFF
--- a/src/javasources/KTool/src/org/kframework/ktest/KTest.java
+++ b/src/javasources/KTool/src/org/kframework/ktest/KTest.java
@@ -111,8 +111,7 @@ public class KTest {
             e.printStackTrace();
             GlobalSettings.kem.register(
                     new KException(KException.ExceptionType.ERROR,
-                            KException.KExceptionGroup.CRITICAL,
-                            e.getMessage(), "command line", "System file."));
+                            KException.KExceptionGroup.CRITICAL, e.getMessage()));
         } catch (InvalidConfigError e) {
             LocationData location = e.getLocation();
             GlobalSettings.kem.register(
@@ -122,8 +121,7 @@ public class KTest {
         } catch (ParseException | InvalidArgumentException e) {
             GlobalSettings.kem.register(
                     new KException(KException.ExceptionType.ERROR,
-                            KException.KExceptionGroup.CRITICAL,
-                            e.getMessage(), "command line", "System file."));
+                            KException.KExceptionGroup.CRITICAL, e.getMessage()));
         }
         return false;
     }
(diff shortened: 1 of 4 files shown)


osa1:
Also fixes #519 .

osa1:
Also fixes #523 .

src/javasources/KTool/src/org/kframework/ktest/KTest.java
```diff
@@ -119,6 +119,11 @@ public static boolean main(String[] args) {
                     new KException(KException.ExceptionType.ERROR,
                             KException.KExceptionGroup.CRITICAL,
                             e.getMessage(), location.getSystemId(), location.getPosStr()));
+        } catch (ParseException | InvalidArgumentException e) {
+            GlobalSettings.kem.register(
+                    new KException(KException.ExceptionType.ERROR,
+                            KException.KExceptionGroup.CRITICAL,
+                            e.getMessage(), "command line", "System file."));
```
dwightguth: haha, no. if there's no file or location, omit those parameters, don't put useless junk.

dwightguth:
The correct pattern here is to have a command line option --debug which is a boolean and, when specified, causes exceptions which exit the application to print a stack trace. So, for example:

```
} catch (SAXException e) {
    if (debug) {
        e.printStackTrace();
    }
    GlobalSettings.kem.register(...);
}
```

osa1:
I fixed some irregularities on the way. I suggest reviewing this commit by commit.

dwightguth:
I don't like the way you are handling the case when options parsing fails, but it's not worth fighting over because this is still using the old options framework that I will be getting rid of eventually. I will merge this.

osa1:
> I don't like the way you are handling the case when options parsing fails, but it's not worth fighting over because this is still using the old options framework that I will be getting rid of eventually.

Do you have any suggestions? I have to search through the argument list manually.

dwightguth:
No, as I said, it's fine the way it is now, because we don't really have a
better solution. But when I redo this with JCommander, it will get dealt
with. Don't worry about it.

On Mon, Jun 2, 2014 at 11:04 AM, Ömer Sinan Ağacan <notifications@github.com

PR #525, opened 2014-05-30
Rename TermContext to GlobalContext

cos:
The TermContext is actually the evolving (non-revertible) global context.


cos:
Jenkins: test this please

cos:
realized it may be easier to leave the TermContext name, and add a GlobalContext class for the cases where its global nature is relevant. We need to discuss this. (relevant for @andreistefanescu @dwightguth)

PR #526, opened 2014-05-30
rename TermContext to State

cos:
just an automated refactoring -- intrusive but should be clean


dwightguth:
Jenkins: add to whitelist

yilongli:
I am afraid the documentation is going to stay the same.

cos:
ok. hold the merge for now. IntelliJ put in some junk.

cos:
gr... lots of talk about "term context" even though it is not "term" and is not "context"... I'll close this, and try again cleanly. It seems like I'll have to slowly move though hundreds of files to check that the comment/string replacements are correct.... uff

PR #528, opened 2014-05-30
Rename TermContext to State

cos:
ok. so I took a lighter approach. I only changed the obvious places in the documentation, i.e., those mentioning "term context".

regarding method renames, I only changed those that were talking about `termContext`, not other variants.

there are many places where the State object is still referred by `context`. While not ideal, I think it is not necessarily incorrect -- the State can be considered the context in which something happens. And it is at least better than `TermContext`.

Let's see what Jenkins says, then it may be good to have another pair of eyes check the replacements.

### DIFF
--- a/src/javasources/KTool/src/org/kframework/backend/java/kil/State.java
+++ b/src/javasources/KTool/src/org/kframework/backend/java/kil/State.java
@@ -13,7 +13,7 @@ import com.google.common.collect.HashBasedTable;
 import com.google.common.collect.Table;
 
 /**
- * An object containing context specific to a particular configuration.
+ * This class represents a state in the transition system.
  */
 public class State extends JavaSymbolicObject {
 
--- a/src/javasources/KTool/src/org/kframework/backend/java/kil/State.java
+++ b/src/javasources/KTool/src/org/kframework/backend/java/kil/State.java
@@ -15,15 +15,15 @@ import com.google.common.collect.Table;
 /**
  * An object containing context specific to a particular configuration.
  */
-public class TermContext extends JavaSymbolicObject {
+public class State extends JavaSymbolicObject {
 
-    private static Map<Definition, TermContext> cache1 = new HashMap<Definition, TermContext>();
-    private static Table<Definition, FileSystem, TermContext> cache2 = HashBasedTable.create();
+    private static Map<Definition, State> cache1 = new HashMap<Definition, State>();
+    private static Table<Definition, FileSystem, State> cache2 = HashBasedTable.create();
 
     private final Definition def;
     private final FileSystem fs;
     
-    private TermContext(Definition def, FileSystem fs) {
+    private State(Definition def, FileSystem fs) {
         this.def = def;
         this.fs = fs;
     }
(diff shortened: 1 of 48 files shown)


yilongli:
I guess searching for `State context` and renaming `context` to `state` should solve the problem?

cos:
@yilongli , yes :) Still, on giving it more thought, I think the name `context` is better. Saying "State state" is redundant. The meaning here is that we are moving the context around, which in this case happens to be the entire State. 

Long term, from the discussion with Dwight, we will remove all this state piggybacking for the cases where it is not needed. 

On a related note, there are many places in the code base with this kind of redundancy, e.g. https://github.com/kframework/k/blob/master/src/javasources/KTool/src/org/kframework/backend/java/builtins/MetaK.java#L56-L71 . Personally, I could do without the javadoc `@param`s in this case. They are redundant considering the names of the parameters, but make yet another piece to maintain. @dwightguth , what is the project's policy here?

dwightguth:
I see no reason to have javadoc which does not add value. You can remove them if you want.

src/javasources/KTool/src/org/kframework/backend/java/kil/State.java
```diff
@@ -15,15 +15,15 @@
 /**
  * An object containing context specific to a particular configuration.
```
dwightguth: If the idea for this class is that it represents a state in the transition system, can this reflect that?

src/javasources/KTool/src/org/kframework/backend/java/ksimulation/Waitor.java
```diff
@@ -62,13 +61,13 @@ public Waitor(Context implRules,Context specRules,org.kframework.kil.Term implTe
         
         
         Term term = Term.of(implTerm, impl.getDefinition());
-        TermContext termContext = TermContext.of(impl.getDefinition(), new PortableFileSystem());
-        ConstrainedTerm implConstraint = new ConstrainedTerm(term, termContext);
+        org.kframework.backend.java.kil.State state = org.kframework.backend.java.kil.State.of(impl.getDefinition(), new PortableFileSystem());
```
dwightguth, line 64: import

dwightguth, line 64: or is there a naming conflict here?

src/javasources/KTool/src/org/kframework/backend/java/ksimulation/Waitor.java
```diff
@@ -62,13 +61,13 @@ public Waitor(Context implRules,Context specRules,org.kframework.kil.Term implTe
...
+        ConstrainedTerm implConstraint = new ConstrainedTerm(term, state);
         pair[0] = implConstraint;
         
         term = Term.of(specTerm, spec.getDefinition());
-        termContext = TermContext.of(spec.getDefinition(), new PortableFileSystem());
-        ConstrainedTerm specConstraint = new ConstrainedTerm(term, termContext);
+        state = org.kframework.backend.java.kil.State.of(spec.getDefinition(), new PortableFileSystem());
```
dwightguth, line 69: same as above

dwightguth:
You did a text search for all the appropriate references and determined there isn't anything lingering? What I can see looks good.

src/javasources/KTool/src/org/kframework/backend/java/ksimulation/Waitor.java
cos, line 64: Yes, it's a name conflict.

cos:
I did go through all the files, but keep in mind there are almost three hundred references to `State`, so another pair of eyes may catch more things. 

Still, considering the tests pass, I think the worst that could happen is having inappropriate variable names, which shouldn't be that bad -- we have much greater problems with State that we need to address, so we can fix names along the way.

dwightguth:
I agree. Are you sure about the name conflict in Waitor.java? I don't see any other class named State being imported or in that package...

cos:
`Waitor` extends `Thread` which has a `State` inner class.

dwightguth:
Oh okay. I am good with this then.

dwightguth:
@cos if you still believe that this is the correct approach that you want to take, you can merge this.

cos:
I gave up -- targeting unified-builtins2 instead.

PR #535, opened 2014-06-01
State knows about top configuration [don't review]

cos:
abusing the pull request jenkins integration to get my stuff tested

### DIFF
--- a/src/javasources/KTool/src/org/kframework/backend/java/builtins/BuiltinBitVectorOperations.java
+++ b/src/javasources/KTool/src/org/kframework/backend/java/builtins/BuiltinBitVectorOperations.java
@@ -3,7 +3,7 @@ package org.kframework.backend.java.builtins;
 
 import org.kframework.backend.java.kil.BuiltinList;
 import org.kframework.backend.java.kil.Term;
-import org.kframework.backend.java.kil.TermContext;
+import org.kframework.backend.java.kil.State;
 
 import java.util.List;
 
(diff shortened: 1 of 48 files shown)


cos:
Ok. So it failed. The only change here is removing the caching for `State` (former `TermContext`). @yilongli, you added the caching. I tried to figure out where you needed it but it is not straightforward to search for `==` on a particular object type. Any ideas why it fails?

yilongli:
Here is the commit which cached the TermContext: https://github.com/yilongli/k/commit/7b8cde3c5525ab1ac8068a0ec94e92520fb19c8a
After second thought, I don't think I have intentionally used `==` to check equality between `TermContext` objects, because such operation probably doesn't exist in our code. I guess I am just trying to save some space and time.

cos:
Yes, I've looked at this commit but it doesn't do anything else aside from adding the builder pattern and caching. The problem is that the build failed after I (manually) reverted it. Jenkins failed with timeout, while my local failed with errors.

Creating one object with a few fields once in a while (after a rule application, and before some visitors) should not have a significant impact on performance. Do you remember why you made this change?

yilongli:
At that point, `TermContext` is basically only used as a global object, so I decided to make a factory method for it. That's probably the only reason I made the change.

yilongli:
Can you first remove the caching mechanism in the factory method to see if it is indeed the root cause? Maybe the renaming refactoring is not that robust at all...

cos:
Tested the rename idependently (see the other pull request) and it works.

yilongli:
If you can give me the simplest example that fails on your laptop, I can try to figure out what is the problem.

dwightguth:
Jenkins: test this please

cos:
It may be related to the current broken master -- I've forked from the HEAD. Let's see what's the problem there first and get back to this.

cos:
the change has been done in another branch -- see #578 for the big picture.

PR #537, opened 2014-06-01
Cleanup builtin map api

yilongli:
@andreistefanescu @dwightguth please review.

1) Use builder pattern to avoid unnecessary defensive-copying in both the `transform` methods and `BuiltinMap` constructors;
2) Make `BuiltinMap` immutable.

P.S.
I made a stupid mistake in profiling the rewrite engine when I tried to use persistent map for `BuiltinMap` in https://github.com/kframework/k/pull/531. The performance is actually worse (5~10% slower) since there are usually a lots of lookups and very few updates. The speedup of update by persistent data structure doesn't out-weigh the slowdown of lookup (even with reasonably large `store` cell). Therefore, the use of persistent data structure only makes sense for `KSequence` and, maybe, `KList`, whose structures change very frequently.

### DIFF
--- a/src/javasources/KTool/src/org/kframework/backend/java/builtins/BuiltinMapOperations.java
+++ b/src/javasources/KTool/src/org/kframework/backend/java/builtins/BuiltinMapOperations.java
@@ -9,9 +9,7 @@ import org.kframework.backend.java.kil.Variable;
 
 import com.google.common.base.Preconditions;
 
-import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Map;
 import java.util.Set;
 
 
@@ -30,11 +28,11 @@ public class BuiltinMapOperations {
     }
 
     public static BuiltinMap construct(BuiltinMap term1, BuiltinMap term2, TermContext context) {
+        Preconditions.checkArgument(!term1.hasFrame() || !term2.hasFrame(), 
+                "both map arguments have frames, but the combined map cannot have two frames");
+        
         Variable frame = null;
-        if (term1.hasFrame() && term2.hasFrame()) {
-            throw new IllegalArgumentException(
-                    "both map arguments have frames, but the combined map cannot have two frames");
-        } else if (term1.hasFrame()) {
+        if (term1.hasFrame()) {
             frame = term1.frame();
         } else if (term2.hasFrame()) {
             frame = term2.frame();
@@ -48,14 +46,12 @@ public class BuiltinMapOperations {
     }
 
     public static BuiltinMap update(BuiltinMap term, Term key, Term value, TermContext context) {
-        if (!term.hasFrame()) {
-            BuiltinMap.Builder builder = BuiltinMap.builder();
-            builder.putAll(term.getEntries());
-            builder.put(key, value);
-            return builder.build();
-        } else {
-            throw new IllegalArgumentException("argument " + term + " has frame");
-        }
+        Preconditions.checkArgument(!term.hasFrame(), "argument " + term + " has frame");
+        
+        BuiltinMap.Builder builder = BuiltinMap.builder();
+        builder.putAll(term.getEntries());
+        builder.put(key, value);
+        return builder.build();
     }
 
 }
--- a/src/javasources/KTool/src/org/kframework/backend/java/builtins/BuiltinMapOperations.java
+++ b/src/javasources/KTool/src/org/kframework/backend/java/builtins/BuiltinMapOperations.java
@@ -7,6 +7,8 @@ import org.kframework.backend.java.kil.Term;
 import org.kframework.backend.java.kil.TermContext;
 import org.kframework.backend.java.kil.Variable;
 
+import com.google.common.base.Preconditions;
+
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
@@ -21,12 +23,10 @@ import java.util.Set;
 public class BuiltinMapOperations {
 
     public static BuiltinSet keys(BuiltinMap term, TermContext context) {
-        if (!term.hasFrame()) {
-            Set<Term> elements = new HashSet<Term>(term.getEntries().keySet());
-            return new BuiltinSet(elements);
-        } else {
-            throw new IllegalArgumentException("argument " + term + " has frame");
-        }
+        Preconditions.checkArgument(!term.hasFrame(), "argument " + term + " has frame");
+        
+        Set<Term> elements = new HashSet<Term>(term.getEntries().keySet());
+        return new BuiltinSet(elements);
     }
 
     public static BuiltinMap construct(BuiltinMap term1, BuiltinMap term2, TermContext context) {
@@ -40,17 +40,19 @@ public class BuiltinMapOperations {
             frame = term2.frame();
         }
 
-        Map<Term, Term> entries = new HashMap<>(term1.getEntries());
-        entries.putAll(term2.getEntries());
-        return new BuiltinMap(entries, frame);
-
+        BuiltinMap.Builder builder = BuiltinMap.builder();
+        builder.putAll(term1.getEntries());
+        builder.putAll(term2.getEntries());
+        builder.setFrame(frame);
+        return builder.build();
     }
 
     public static BuiltinMap update(BuiltinMap term, Term key, Term value, TermContext context) {
         if (!term.hasFrame()) {
-            Map<Term, Term> entries = new HashMap<>(term.getEntries());
-            entries.put(key, value);
-            return new BuiltinMap(entries);
+            BuiltinMap.Builder builder = BuiltinMap.builder();
+            builder.putAll(term.getEntries());
+            builder.put(key, value);
+            return builder.build();
         } else {
             throw new IllegalArgumentException("argument " + term + " has frame");
         }
--- a/src/javasources/KTool/src/org/kframework/backend/java/kil/BuiltinMap.java
+++ b/src/javasources/KTool/src/org/kframework/backend/java/kil/BuiltinMap.java
@@ -148,14 +148,11 @@ public class BuiltinMap extends Collection implements Iterable<Map.Entry<Term, T
     public static class Builder {
         
         private boolean done = false;
-        private boolean setEntries = false;
-        private boolean setFrame = false;
         
         private Map<Term, Term> entries = new HashMap<>();
         private Variable frame = null;
 
         public void put(Term key, Term value) {
-            Preconditions.checkArgument(!setEntries, "Cannot put in new key-value pair once the entries are set.");
             entries.put(key, value);
         }
         
@@ -166,10 +163,7 @@ public class BuiltinMap extends Collection implements Iterable<Map.Entry<Term, T
          * @param map
          */
         public void putAll(Map<Term, Term> map) {
-            Preconditions.checkArgument(!setEntries, "Cannot put in new key-value pair once the entries are set.");
-            for (Map.Entry<Term, Term> e : map.entrySet()) {
-                this.put(e.getKey(), e.getValue());
-            }
+            entries.putAll(map);
         }
         
         public Term remove(Term key) {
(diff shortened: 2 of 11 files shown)


src/javasources/KTool/src/org/kframework/backend/java/builtins/BuiltinMapOperations.java
```diff
@@ -21,12 +23,10 @@
 public class BuiltinMapOperations {
 
     public static BuiltinSet keys(BuiltinMap term, TermContext context) {
-        if (!term.hasFrame()) {
-            Set<Term> elements = new HashSet<Term>(term.getEntries().keySet());
-            return new BuiltinSet(elements);
-        } else {
-            throw new IllegalArgumentException("argument " + term + " has frame");
-        }
+        Preconditions.checkArgument(!term.hasFrame(), "argument " + term + " has frame");
```
dwightguth, line 24: why use preconditions for only one of these three methods?

src/javasources/KTool/src/org/kframework/backend/java/kil/BuiltinMap.java
```diff
@@ -53,19 +43,12 @@ public Term get(Term key) {
     }
 
     public Map<Term, Term> getEntries() {
-        return Collections.unmodifiableMap(entries);
-    }
-
-    public Term put(Term key, Term value) {
-        return entries.put(key, value);
+        return UnmodifiableMap.decorate(entries);
```
dwightguth, line 46: Collections.unmodifiableMap?

dwightguth, line 46: actually, hmm. Can you explain the reasoning behind using an unmodifiable map instead of an immutable map?

src/javasources/KTool/src/org/kframework/backend/java/kil/BuiltinMap.java
```diff
@@ -147,20 +130,112 @@ public void accept(Visitor visitor) {
...
+        
+        /**
+         * Copies all key-value pairs of the given map into the BuiltinMap being
+         * built.
+         * 
+         * @param map
+         */
+        public void putAll(Map<Term, Term> map) {
+            Preconditions.checkArgument(!setEntries, "Cannot put in new key-value pair once the entries are set.");
+            for (Map.Entry<Term, Term> e : map.entrySet()) {
```
dwightguth: entires.putAll(map)?

src/javasources/KTool/src/org/kframework/backend/java/kil/BuiltinMap.java
```diff
@@ -147,20 +130,112 @@ public void accept(Visitor visitor) {
...
+            return UnmodifiableMap.decorate(entries);
+        }
+        
+        /**
+         * Sets the entries without copying the contents. Once the entries are
+         * set, no more modification is allowed.
+         * 
+         * @param map
+         */
+        public void setEntries(Map<Term, Term> map) {
```
dwightguth: this feels like a really odd method. how is it being used?

src/javasources/KTool/src/org/kframework/backend/java/kil/BuiltinMap.java
```diff
@@ -147,20 +130,112 @@ public void accept(Visitor visitor) {
...
+            entries = map;
+        }
+        
+        /**
+         * Sets the frame of the BuiltinMap being built. Once the frame is set,
+         * it cannot be changed.
+         * 
+         * @param frame
+         */
+        public void setFrame(Variable frame) {
```
dwightguth, line 194: same here

src/javasources/KTool/src/org/kframework/backend/java/kil/BuiltinMap.java
```diff
@@ -147,20 +130,112 @@ public void accept(Visitor visitor) {
...
-            return builtinMap;
+        
+        public BuiltinMap build() {
+            Preconditions.checkArgument(!done, "Only one BuiltinMap can be built from a builder.");
+            done = true;
+            // YilongL: Guava's ImmutableMap.copyOf(entries) is not smart enough
+            // to avoid actually copying the entries, because entries is not an
+            // ImmutableMap yet; using Apache's decorate method because it would
+            // avoid creating nesting wrappers
+            return new BuiltinMap((UnmodifiableMap<Term, Term>) UnmodifiableMap.decorate(entries), frame);
```
dwightguth, line 229: Are you sure?

http://docs.guava-libraries.googlecode.com/git/javadoc/src-html/com/google/common/collect/ImmutableMap.html#line.271

dwightguth, line 229: oh wait I see what you're saying. Never mind. But I still wonder why we're not using ImmutableMap.Builder.

dwightguth:
Most of my concern here surrounds the structure of the new BuiltinMap. Why is it using an unmodifiableMap internally and not an ImmutableMap? Why aren't we using ImmutableMap.Builder? Etc.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
```diff
@@ -426,12 +426,13 @@ public ASTNode visit(org.kframework.kil.SetBuiltin node, Void _)  {
...
     @Override
     public ASTNode visit(org.kframework.kil.MapBuiltin node, Void _)  {
-        HashMap<Term, Term> entries = new HashMap<Term, Term>(node.elements().size());
+        BuiltinMap.Builder builder = BuiltinMap.builder();
         for (Map.Entry<org.kframework.kil.Term, org.kframework.kil.Term> entry :
                 node.elements().entrySet()) {
             Term key = (Term) this.visitNode(entry.getKey());
             Term value = (Term) this.visitNode(entry.getValue());
-            entries.put(key, value);
+//            entries.(key, value);
```
andreistefanescu: delete this line

src/javasources/KTool/src/org/kframework/backend/java/kil/BuiltinMap.java
andreistefanescu, line 229: I think you should be able to use ImmutableMap.Builder.

src/javasources/KTool/src/org/kframework/backend/java/builtins/BuiltinMapOperations.java
yilongli, line 24: You are right. I forgot to make the change to all three of them.

src/javasources/KTool/src/org/kframework/backend/java/kil/BuiltinMap.java
yilongli: Thank you, Dwight. I actually wanted to say `this.put(e.getKey(), e.getValue())`.

src/javasources/KTool/src/org/kframework/backend/java/kil/BuiltinMap.java
yilongli: In `CopyOnWriteTransformer#transform(BuiltinMap)`, when the `entries` of the `BuiltinMap` instance being transformed remains intact, it can be reused without being copied.

src/javasources/KTool/src/org/kframework/backend/java/kil/BuiltinMap.java
yilongli, line 194: same as above

src/javasources/KTool/src/org/kframework/backend/java/kil/BuiltinMap.java
yilongli, line 229: OK, I will use `ImmutableMap.Builder` instead:)

yilongli, line 229: Well, I actually remember that `Immutable.Builder` doesn't have a `remove` method. That's probably why I change to use Apache `decorate` method instead.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
yilongli: done

src/javasources/KTool/src/org/kframework/backend/java/kil/BuiltinMap.java
dwightguth: okay, but why the restriction that a builder can only be modified once? seems arbitrary.

yilongli: If you use this method to set the `entries`, it is an `ImmutableMap`. It can not be modified without copying. Besides, there is no need to set and then modify the entries. I am trying to restrict the api so that they can be used correctly.

dwightguth: I guess I just don't see the benefits of the added complexity of a variable that forces you to not be able to call certain methods twice, even though not being able to make certain types of modifications to ImmutableMaps is the entire reason we created this class in the first place. Sure, if you call this method with an ImmutableMap, then put will throw an exception, but how is that any different from the exception you are currently doing funky logic in order to throw manually?

src/javasources/KTool/src/org/kframework/backend/java/kil/BuiltinMap.java
dwightguth: okay but why? it's the exact same effect to call entries.putAll and it's less complex code.

yilongli: You are right. There was some additional check in the `put` method. I guess I removed them at some point. I will use `entries.putAll` instead.

src/javasources/KTool/src/org/kframework/backend/java/kil/BuiltinMap.java
yilongli: Because calling this method twice is not the right way to build a BuiltinMap. I have already answered why I couldn't use a ImmutableMap.Builder in the other reply.

dwightguth: Okay but why isn't it the right way to build a BuiltinMap? You're trying to
enforce a contract here, but I just don't see what we actually gain from
enforcing that contract. What's the bad thing that happens if someone
decides they happen to want to build a map this way?

yilongli: Because the way to build a BuiltinMap is driven by the use cases in the `CopyOnWriteTransformer`. We are not building a library for users; I don't think there is a point in arguing over this minor issue. If you really really need such functionality, I can do it for you.

dwightguth: it's not that I need this functionality; it's that your restrictions on the
behavior of the code introduce complexity into the class that I don't
understand the need for. The idea behind a builder is that it provides an
API for constructing an unchangeable object. How you intend to use it in
CopyOnWriteTransformer should not come into it, except insofar as it
provides the design constraints on the implementation. In this case, you
want to avoid unnecessary copying.

Simple code is better than complex code. This class as it currently stands
is quite complex and has a number of subtleties that you need to understand
about it, despite the fact that I don't perceive the advantage those
subtleties actually provide in achieving the design constraint, which is an
immutable BuiltinMap class that can be constructed with as little copying
as possible.

As far as minor issues go, there are most certainly minor issues that are
not worth arguing over in a code review. The architectural structure of a
class and unnecessary complexity in it are simply not one of them, though.
Ensuring that the code that gets committed has the right design is the
entire _purpose_ of a code review.

src/javasources/KTool/src/org/kframework/backend/java/kil/BuiltinMap.java
dwightguth, line 229: So I just noticed something. You can call setEntries with a modifiable map, build the map, and then modify the map with your preexisting reference to the modifiable view. If you want to use unmodifiable for this instead of immutable, you will need to ensure somehow via scoping and access rules that no references to the modifiable map escape after calling build().

yilongli, line 229: Yes, I will do that.

src/javasources/KTool/src/org/kframework/backend/java/kil/BuiltinMap.java
yilongli: You have got a point. I have tried to revise the code.

PR #538, opened 2014-06-02
update to latest version of jansi

dwightguth:
This bug is so unbelievably strange I have no idea whatsoever if this will fix it, because I still have no idea what's causing it. But it's worth a try.


osa1:
What's the bug?

grosu:
It didn't fix it, Dwight.  I think we should ignore the color checking in ktest.  For uniformity, we should do it the same way we ignore the white spaces and the balanced parentheses: with an option --ignore-colors <on|off> with default on.  Does this sound reasonable?

dwightguth:
That seems reasonable, yes.

@osa1 do you want to undertake this one? it would involve modifying IgnoringStringMatcher and RegexStringMatcher.

The bug is that in some cases because of a bug in JANSI, krun outputs the color codes even though output has been piped to a file. I have no idea how to fix it because I can't figure out how to reproduce it reliably, so I don't see a way to fix it other than what @grosu proposes. You're welcome to look into it further, but in the meantime, we have a very fragile test case in master. So it would be good to be taken care of quickly.

dwightguth:
Update: just talked to Grigore and he says he actually tested this wrong, so it's possible it actually does work.

@osa1 my description of the bug is above. If you are satisfied, can I merge this?

osa1:
So did updating the jansi solve it? I just tried to reproduce it but failed. Piping to files or other programs(`cat` etc.) works fine here. (tried on several different definitions and programs)

Also, I just realized that if we have only one cell in the output(and no nested cells) it's not printed in colors even while printing to the colorful terminal. Is this intended?

osa1:
Is this problem happening while running krun or ktest? If this is happening only while running krun under ktest then maybe because of a Java threads related thing jansi may be failing to detect output stream's color support.

dwightguth:
No, it affects krun by itself. It's just that we can't fix the bug because
it's strange and seems to be a bug in JANSI itself, and I can't figure out
how to reproduce it.

On Tue, Jun 3, 2014 at 10:36 AM, Ömer Sinan Ağacan <notifications@github.com

osa1:
OK. I can't do anything without reproducing the bug so I say let's merge this and try to find a way to reproduce the bug if it happens again after this patch.

Btw, in that case it'd be very helpful if reporter could also write down exact command and definition/program files etc.

dwightguth:
Oh no I can reproduce the bug. You reproduce it by checking out a completely fresh copy of the k framework and then testing imp++ lesson 3. The problem is that if you take a completely identical directory to what you get from doing that, and then run the exact same command, under some set of circumstances I can't identify, it works instead. But yeah, I will merge this.

osa1:
Well, that imp++ lesson 3 tests are not working on clean copy at all:

```
➜  ~  git clone https://github.com/kframework/k k_clean   
Cloning into 'k_clean'...
remote: Reusing existing pack: 117999, done.
remote: Counting objects: 424, done.
remote: Compressing objects: 100% (322/322), done.
remote: Total 118423 (delta 199), reused 92 (delta 51)
Receiving objects: 100% (118423/118423), 209.11 MiB | 808.00 KiB/s, done.
Resolving deltas: 100% (72660/72660), done.
Checking connectivity... done
Checking out files: 100% (3491/3491), done.
➜  ~  cd k_clean 
➜  k_clean git:(master) ant build
Buildfile: /home/omer/k_clean/build.xml
... snip ...
BUILD SUCCESSFUL
Total time: 41 seconds
➜  k_clean git:(master) export PATH=/home/omer/k_clean/bin:$PATH
➜  k_clean git:(master) which ktest
/home/omer/k_clean/bin/ktest
➜  k_clean git:(master) which krun
/home/omer/k_clean/bin/krun
➜  k_clean git:(master) cd tutorial/1_k/4_imp++/lesson_3/      
➜  lesson_3 git:(master) ktest tests/config.xml -v
Running initial scripts...
SUCCESS
Kompile the language definitions...(1 in total)
ERROR: ['/home/omer/k_clean/bin/kompile' '/home/omer/k_clean/tutorial/1_k/4_imp++/lesson_3/imp.k' '--superheat' 'division' '--supercool' 'lookup increment'] failed with error (time: 373 ms)
error was: [Error] Critical: Creating multiple kompiled definition in the same directory
is not allowed.
        File: command line
        Location: /home/omer/k_clean/tutorial/1_k/4_imp++/lesson_3/./config-kompiled

FAIL
Generate PDF files...(0 in total)
SUCCESS
SUCCESS

FAIL (see details above)
----------------------------
Definitions kompiled: 1 (time: 0'0'' elapsed, 0'0'' CPU)
PDF posters kompiled: 0 (time: 0'0'' elapsed, 0'0'' CPU)
Programs krun: 0 (time: 0'0'' elapsed, 0'0'' CPU)
Total time: 0'0'' elapsed, 0'0'' CPU
----------------------------
```

Am I doing something wrong?

dwightguth:
If you delete that config-kompiled directory does it work?

On Tue, Jun 3, 2014 at 11:24 AM, Ömer Sinan Ağacan <notifications@github.com

osa1:
:jack_o_lantern: my bad. yeah now it works and tests pass.

grosu:
A big thank you guys, it works now!!!

PR #551, opened 2014-06-05
Use cons list in ksequence

yilongli:
Use cons list as the underlying data structure of KSequnce to enable efficient reuse of the tail.

Achevied 3~6x speedup on related operations of KSequence, but only slight overall performance improvement (< 10%) because this is not yet the bottleneck.


yilongli:
@andreistefanescu @dwightguth please review.

src/javasources/KTool/src/org/kframework/backend/java/kil/KSequence.java
```diff
@@ -123,4 +196,45 @@ public void accept(Visitor visitor) {
...
+            IllegalAccessException, NoSuchFieldException, SecurityException {
+        in.defaultReadObject();
+
+        Object[] termArray = (Object[]) in.readObject();
+        PStack<Term> stack = ConsPStack.empty();
+        for (int i = termArray.length - 1; i >= 0; i--) {
+            stack = stack.plus((Term) termArray[i]);
+        }
+        Field f = KSequence.class.getDeclaredField("contents");
+        f.setAccessible(true);
```
dwightguth: I know you can totally do this, but it makes my skin crawl. Any chance you could use the serialization proxy pattern instead?

PR #559, opened 2014-06-07
Caching rules before disambiguation.

radumereuta:
Unfortunately the diff also contains windows to linux line termination conversion which makes the diff a bit harder to read.


dwightguth:
lgtm. You'll need to find out why tests failed after we get Jenkins set up again in the new office, but I see no problems with it as written.

radumereuta:
@dwightguth can you tell me why is this failing?

dwightguth:
Test Result (3 failures / +1)

```
/samples/quine/explicit/quine-explicit.k.quine-explicit.k
/tests/regression/kast/test.k.3
/tests/regression/kast/test.k.4
```

Is that enough information for you to reproduce the errors?

radumereuta:
I managed to fix the quine example, but not the regression one. I don't even know how to start this, because I get different behaviors from krun when I run it on windows and linux:
Linux:

```
radum@fslwork:~/work/k/tests/regression/kast$ krun --parser "kast --parser ground" 3.test
<k>
    #parse ( "foo" , "bar" )
</k>
```

Windows:

```
d:\work\k\tests\regression\kast>krun 3.test --parser "kast.bat --parser ground"
<k>
    .K
</k>
```

dwightguth:
This is because of the nasty hideous way that external parsers are turned
into internal parsers right now, which uses string matching. I wrote it
myself, but it's still terrible and I shouldn't have. You can fix it if you
want, but I don't have time to right now.

The answer you're looking for is that something is broken in the code path
which calls the parser without creating a kast process.

dwightguth:
Jenkins: test this please

dwightguth:
The linux bug is because the entire prelude is broken and none of its rules are being included in the compiled maude definition... yeah.

No goddamn clue why that might be happening or how it wasn't frighteningly obvious. Sorry.

radumereuta:
@dwightguth please review.
The problem was that == instead of .equals() when comparing Strings.

dwightguth:
Ah, that explains the OS-specific nature, because String interning is JVM-dependent, and linux usually uses OpenJDK.

radumereuta:
Please merge it if you agree with it.
I'll try to merge master into unified-builtins2 afterwards.

dwightguth:
Merged. Please merge into unified-builtins2

yilongli:
BTW, should we also introduce static-analysis tools like FindBugs into our developing process. It could have saved Radu a lot of time in revealing this bug.

radumereuta:
That is actually a really good idea.
We may need a week to enforce it, but I think it would be worth it.

cos:
I agree, FindBugs would help.

cos:
On a second look, Eclipse gives me 768 compilation warnings. We could start with fixing those before going into FindBugs.

radumereuta:
Do you know how to make them tested automatically?

yilongli:
No, but I am sure @dwightguth can take care of this.

cos:
I think they are warnings from javac. So they should be available in any development/testing environment, including jenkins. Still, as with fixing line endings, fixing warnings at once would disrupt the entire code base. It can only work if everybody does it at the same time. Alternatively, we would add a rule that any file that is changed should have no compilation warnings. But we cannot enforce this (easily) though Jenkins afaik.

grosu:
I also agree that it is a good idea to use as many code analysis tools as possible/feasible.  LEt us not forget those developed by a company called Runtime Verification, Inc. They have two famous RV tools, rvmonitor and rvpredict.  Now seriously, Owolabi ran K through rvmonitor with about 200 properties (API protocols) for the most common 4 packages and guess what: he found lots of bugs, which can lead to non-deterministic behaviors/errors.  But we will send another message about it when the time comes.  Anyway my point is that we should start using our own tools ourselves, once we exhaust the easily available and fast ones.

Issue #555, opened 2014-06-06
ktest warnings


grosu:
Guys, can you please give priority to the bugs that keep us from merging the unified-builtins2 branch?  For example the one below.  Also, when you fix a bug and make a pull request for it, it is best to wait until the bug fix is merged before you close an issue.

```
fslwork$ pwd
/home/grosu/github.grosu/k/tutorial/1_k/4_imp++/lesson_2
fslwork$ kompile imp.k
[Warning] Compiler: unable to transform the KApp: '#freeVars(# Ks:KList ())
because of multiple productions associated:
["#freeVars" "(" KList ")" , "#freeVars" "(" KList "," Set "," Set ")" ]
    File: /home/grosu/github.grosu/k/include/modules/substitution.k
    Location: (110,3,110,47)
    Compilation Phase: Compile collections to internal K representation
[Warning] Compiler: unable to transform the KApp: '#freeVars(# Ks:KList ())
because of multiple productions associated:
["#freeVars" "(" KList ")" , "#freeVars" "(" KList "," Set "," Set ")" ]
    File: /home/grosu/github.grosu/k/include/modules/substitution.k
    Location: (115,3,115,61)
    Compilation Phase: Compile collections to internal K representation
[Warning] Compiler: unable to transform the KApp: '#freeVars(# Ks:KList (),,
'.Set(),, '.Set())
because of multiple productions associated:
["#freeVars" "(" KList ")" , "#freeVars" "(" KList "," Set "," Set ")" ]
    File: /home/grosu/github.grosu/k/include/modules/substitution.k
    Location: (115,3,115,61)
    Compilation Phase: Compile collections to internal K representation
[Warning] Compiler: unable to transform the KApp: '#freeVars(# (),,
GeneratedFreshVar523:Set ,, Free:Set )
because of multiple productions associated:
["#freeVars" "(" KList ")" , "#freeVars" "(" KList "," Set "," Set ")" ]
    File: /home/grosu/github.grosu/k/include/modules/substitution.k
    Location: (118,3,118,45)
    Compilation Phase: Compile collections to internal K representation
[Warning] Compiler: unable to transform the KApp: '#freeVars(# X:KItem ,,
Ks:KList (),, Bound:Set ,, Free:Set )
because of multiple productions associated:
["#freeVars" "(" KList ")" , "#freeVars" "(" KList "," Set "," Set ")" ]
    File: /home/grosu/github.grosu/k/include/modules/substitution.k
[...]

PR #560, opened 2014-06-07
fix issue #555

yilongli:
Only throw the warning if the `kLabelConstant` associated with multiple productions is actually being transformed to built-in data structure.


yilongli:

PR #566, opened 2014-06-08
Term context is not a JavaSymbolicObject

cos:


cos:
@andreistefanescu, why is TermContext a JavaSymbolicObject?

cos:
I'll close this for now. Bigger fish.

Issue #533, opened 2014-05-31
ktest --- running ktest in ktest


daejunpark:
We need to have `ktest` be able to run `ktest` as well as `krun`.

Recently, the issues of #518, #523, and #485 reported that previously-working-features did not work any longer. As @osa1 said [https://github.com/kframework/k/issues/518#issuecomment-44646856], this is partially because of a failure of code review. However, I think this is mainly due to the _absence_ of regression test.

We do not have a regression test for `ktest` at all. We cannot have the regression test for `ktest` by using `ktest` because `ktest` cannot be executed by `ktest`.

Along with the same line of the earlier issue #396, we need to have `ktest` execute arbitrary processes other than `krun`.

osa1:
Absence of regression tests for ktest is a problem. We already repeated lots of bugs fixed in the past.

However, while I'm not against testing ktest using ktest iteself, I don't understand how can we do that(using current implementation). Because ktest is a complex tool and testing it may need complex features. On the other hand, unit tests may give us more flexibility without any extra effort for modifying ktest(e.g. without adding that feature to ktest).

Are we going to have a "test language" that provides us all the required primitives for testing and then write "regression-533.ktest" like files for tests?(this is not testing ktest with ktest though) Or are we going to write ktest/K definition files that use something like `<ktest> ... </ktest>`?(then use ktest to read that fields and somehow run tests) (I'm not sure if that makes sense though). This sounds seriously cool but I'm not sure if that's worth the effort.

I can see four alternative implementations here, first one being the current approach but it needs adding lots of new features to current ktest:
1. Specify ktest tests in config.xml files.
2. Write a ktest language using K, and specify tests in that language.
3. Modify K to have tags like `<ktest> ... </ktest>` in definition files and specify tests using that.
4. We may skip all these and use ordinary unit tests for that.

I think it might be good to discuss how to add regression tests for some recent bugs in these alternative implementations and see advantages/disadvantages. But for now, even though I'm not an experienced K user, from an implementors point of view I can see this problems with each approach:
1. We need to modify ktest itself for lots of new features and make ktest even more complex beast. Not flexible at all.
[...]

PR #574, opened 2014-06-10
Move to-be-processed.txt to github wiki

daejunpark:
See https://github.com/kframework/k/issues/533#issuecomment-45579871


dwightguth:
Did you diff the master and unified-builtins2 versions of this file to make sure that any differences were captured in the wiki page? if so, you can merge both of these.

daejunpark:
Yes, the master version is a superset of the unified-builtin2 version, and I've created the wiki page based on the master version. So, nothing was missed.

PR #575, opened 2014-06-10
Move to-be-processed.txt to github wiki -- unified-builtin2

daejunpark:
See https://github.com/kframework/k/issues/533#issuecomment-45579871


daejunpark:
Same with the pull request #574.

PR #577, opened 2014-06-10
add checkstyle check to assert that line endings are unix-style

dwightguth:
@radumereuta please review.

You can test this easily by viewing a couple files to make sure they look correct, then downloading the diff and running `git show --ignore-space-at-eol`


radumereuta:
I've just talked with @andreistefanescu @cos and @yilongli, and they all agree that we should delay this merge until we switch master with unified-builtins2.

Also, I think you are missing the trailing spaces check.

So I suggest you make a pull request with only the checks, and no modifications to the code. This way we will still have the important part of the code (and merge it when it's more convenient), but we don't complicate the pull requests for the rest of us.

Thoughts?

dwightguth:
I will close this out. We can reopen it once the merge happens, and I will take care of trailing spaces.

radumereuta:
You could also leave it opened so we don't forget about it.

cos:
@dwightguth I am still  worried that we'll have issues merging branches across the line endings fix. See this discussion: http://stackoverflow.com/questions/9776527/merging-without-whitespace-conflicts
It refers to white spaces, but I think it is similar for line endings. It seems the issue is that, although they will not show in the diff, the line endings would be added back. The solution would be to have a hook that fixes the endings each time, but it is already getting involved.

we could approach this "by process" instead. we could use the merge of unified-builtins2 (or some other point in time) as a milestone for pending pull requests. at the milestone, we fix the line endings. if your pull request missed it, it is your job to cherry pick.

(I may be wrong about this being a problem -- if you know that it isn't than we should apply the patch as soon as possible)

dwightguth:
So, this diff creates a checkstyle check. As long as all merges into
unified-builtins2 are tested via jenkins (which they should be), it will
fail the build if there are windows style line endings, and then the user
can fix them.

I can hold off on this for now if that's what you think is best, but I'm
not really sure I follow the thread of the argument you're making. Can you
try to explain more precisely what the use case is that you are concerned
will not behave reasonably?

PR #581, opened 2014-06-10
updated simple-untyped.k

grosu:
Updated simple-untyped.k to work in the new unified-builtins2 branch.
@andreisf @traianf please review (it is almost identical with the one under samples/java-rewrite-engine)

### DIFF
--- a/tutorial/2_languages/1_simple/1_untyped/simple-untyped.k
+++ b/tutorial/2_languages/1_simple/1_untyped/simple-untyped.k
@@ -182,7 +182,7 @@ arguments. */
                 | "if" "(" Exp ")" Block "else" Block   [avoid, strict(1)]
                 | "if" "(" Exp ")" Block
                 | "while" "(" Exp ")" Block
-                | "for" "(" Stmts Exp ";" Exp ")" Block // Stmts and not Stmt, because macro expansion yields too general sort info, which fails unification
+                | "for" "(" Stmts Exp ";" Exp ")" Block
                 | "return" Exp ";"                      [strict]
                 | "return" ";"
                 | "print" "(" Exps ")" ";"              [strict]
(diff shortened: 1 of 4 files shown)


tutorial/2_languages/1_simple/1_untyped/simple-untyped.k
```diff
@@ -79,7 +79,7 @@ are only included in the parser generated to parse programs (and used by the
 concrete identifiers that play a special role in the semantics, like
 \texttt{main} below. */
 
-  syntax Id ::= "main"
+  syntax Id ::= Token{"main"}
```
andreistefanescu: "main" should be just syntactic sugar for Token{"main"}. @radumereuta I thought you fixed this issue some time ago.

tutorial/2_languages/1_simple/1_untyped/simple-untyped.k
```diff
@@ -416,14 +416,13 @@ inside the loop body with fewer dimensions.  The second variable,
 \texttt{\$2}, iterates through and initializes each element of the current
 dimension: */
 
-  syntax Id ::= "$1" | "$2"
+  syntax Id ::= Token{"$1"} | Token{"$2"}
```
andreistefanescu: see above

grosu: My fault that I assumed it didn't work only because the simple language under java-rewrite-engine was using Token{...}.  I just removed the Token{...} wrappers and everything worked fine.  Everything pushed again.

Grigore

tutorial/2_languages/1_simple/1_untyped/simple-untyped.k
```diff
@@ -182,7 +182,7 @@ arguments. */
                 | "if" "(" Exp ")" Block "else" Block   [avoid, strict(1)]
                 | "if" "(" Exp ")" Block
                 | "while" "(" Exp ")" Block
-                | "for" "(" Stmt Exp ";" Exp ")" Block
+                | "for" "(" Stmts Exp ";" Exp ")" Block // Stmts and not Stmt, because macro expansion yields too general sort info, which fails unification
```
traiansf: line too long here.

traiansf:
Looks good.  But  there are a lot of changes from the original one. 
I think we need to 
(1) tag  a new version of K right before merging back the unified-builtins branch
(2) write a migration document documenting all the various API changes (including map updates, KItem, dots)

tutorial/2_languages/1_simple/1_untyped/simple-untyped.k
yilongli: In fact, this is a comment I added in the Java backend version of `simple-untyped.k`. I think it's better not to put it in the tutorial. The problem is that the macro used to expand for-loop is written in a way that expects the first argument of `for` construct to have sort `Stmts` instead of `Stmt`. This works in the maude backend but not Java backend.

dwightguth:
If you fixed simple-untyped, can you uncomment it in the regression test suite?

radumereuta:
I remember fixing Token{"main"} for both parsers. So you should be fine using it.

grosu:
@dwightguth I added simple-untyped to the test suite.
@yilongli I removed that comment about Stmts, but added a paragraph in the Latex description about it, because it is important.
@radumereuta ok, good to know.
@traiansf I think we all agreed to release a new stable version of K's master before we make the unified-builtins2 branch master.  Regarding your 2, we need a change_log.txt file, don't we?  I hope we can put that together based on the names of the various commits ...

dwightguth:
I am happy with this. As far as I'm concerned, you can merge it if everyone else is satisfied.

grosu:
Since nobody complained anymore, I assume it is OK to merge simple-untyped.

PR #586, opened 2014-06-11
Unified builtins2 master merge

radumereuta:
Adding the latest modifications committed to master into unified-builtins.


radumereuta:
@dwightguth please review.
There were some conflicts in the config.xml files, but I kept the unified-builtins tests, and added the skip="pdf" to some tasks.

dwightguth:
Please merge again now that another of your diffs has gone into master, and when you've done that,  I will review this.

radumereuta:
@dwightguth please review again.

dwightguth:
I don't think you merged tests/configjavarw.xml correctly. That file should probably be deleted actually. Can you delete it in unified-builtins2, and then make sure that every difference between the tests in that file and the tests in the appropriate sections of the other xml files are in sync?

radumereuta:
I'm not sure I get it.
configjavarw.xml is in unified-builtins2 branch. All I did, was to add skip="pdf" to the tasks.

dwightguth:
Yes, but the point is it's not actually tested. The tests that are in that file should actually go in other files. It was a bad design decision leaving that file around because it makes it easy to mistakenly create a bad merge like this one. You need to actually modify the lines of code which are actually tested in the other config.xml files in other directories in order to ensure that they are in sync with master.

grosu:
Guys, not sure whether this is strictly related, but I think that the config-backend.xml files should include and extend/modify the default ones.  We should avoid boilerplate code as much as possible, even in config.xml files.  If ktest does not support that, then we need to extend it.  The last thing we want to do is to copy-and-paste the existing config.xml and only slightly modify it.  That is bad, bad.

radumereuta:
I agree that the config.xml files should be clean, but I'm not sure I am the best person to clean them up.
I don't know what is the status of the unified-bultins branch, or which tests are working or not.
This pull request just makes the differences between master and unified-builtins more uniform. So any such cleaning should be done in a separate pull request.

dwightguth:
Fine, I will do it. Merging.

PR #588, opened 2014-06-12
Maude binder patch

traiansf:
As Dwight mentioned, Maude indexes were starting at 1, so my commit 6759f713aa9c1020a3bef5f4073aefd0a23dbcb9 broke substitution for Maude.

This fixes it by increasing the indexes when generating Maude-specific equations.

@dwightguth please review


dwightguth:
Okay. If you tested this diff against my own agent change and it worked, and if the tests pass, you can merge this.

PR #593, opened 2014-06-12
Fix i549

radumereuta:


radumereuta:
@dwightguth please review.

dwightguth:
So, this is good, but not quite good enough. I ran into other cases where a KApp had arity 1 and its representation in Java was not a KList. For example:

```
syntax KItem ::= "foo"

rule isKResult(foo) => true
```

If this parses as a KApp, it should contain a child which is a KList of length 1 also.

radumereuta:
So, is it ok if I verify only if the child node is not of sort KList?
I think it solves the problem in general.

dwightguth:
That should work, I think, yes.

Issue #596, opened 2014-06-13
too restricted sort inference for Bag variable


yilongli:
Here is the regression test I added for testing bag support in the Java backend:
https://github.com/yilongli/k/commit/dd576f149d897152511b2f7579d4218de66ab0f1#diff-74220b6ce2fead025ed939dcfc09303dR14

As you can see, I had to explicitly sort the variable `B` to `Bag`; otherwise, it will be inferred as a `BagItem` and cause a failure later during unification in the Java backend.

radumereuta:
Well... this is a tricky one, and can be reproduced by just two sentences:

```
module TEST
  configuration <bag> <b multiplicity="*"> . </b> </bag>
  rule <bag> B => .Bag </bag>
endmodule
```

The problem is when looking for the type of cell 'bag'. Since in the configuration I have only one cell, I'm assuming the contents of the cell must be BagItem. I do similar things for cells like <next> 0 </next> where I'm considering sort Int.
But I guess in this case multiplicity plays a role in the sort of the cell...

dwightguth:
Oncall should test whether this is still an issue in the new parser.

radumereuta:
I remember fixing this for the current implementation, but in the new parser I'm not doing type checking based on the configuration yet, so when we get to this problem again, we should decide what to do.

dwightguth:
Essentially I think what we want here is to have a "." production for every sort (List, Map, Set, Bag, K), and then disambiguate cells parsed using the RULE-CELLS module based on the declared sort of the cell. I am thinking the simplest way to do this would be to just make leaf cells in that module retain the sort of their child. So for example instead of K, the <state> cell in imp would become the following production:

```
syntaix Cell ::= "<state>" #OptionalDots Map #OptionalDots "</state>"
```

radumereuta:
What would be the sort of `.K`? Because I had issues before when you could put K anywere and I coulnd't distinguish anymore between the different dots?

PR #595, opened 2014-06-13
Add explicit support for bag in java backend

yilongli:
Still testing. Not ready for review.

PR #597, opened 2014-06-13
Issue#596

radumereuta:
Fixing issue #596 
Also #595 should be merged at the same time with this one.

### DIFF
--- a/samples/java_rewrite_engine/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
+++ b/samples/java_rewrite_engine/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
@@ -720,7 +720,7 @@ shortly).  The location where the object will be stored is also made
 available in the \textsf{crntObj} cell, so that method closures can
 refer to it (see rule above). */
 
-  syntax KItem ::= (Id,BagItem)
+  syntax KItem ::= (Id,Bag)
 
   rule <k> new Class:Id(Vs:Vals) ~> K
            => create(Class) ~> storeObj ~> Class(Vs); return this; </k>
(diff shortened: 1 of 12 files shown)


yilongli:
I made a mistake when creating https://github.com/kframework/k/pull/595. I closed it and created a new one at https://github.com/kframework/k/pull/600.

dwightguth:
Per our discussion, the correct solution is to remove sort BagItem.

yilongli:
What is the sort of a cell, if not BagItem? Bag? I am not sure I understand why removing sort BagItem is the correct solution.

radumereuta:
Apparently it isn't.
I'm not sure what is the correct solution.

yilongli:
In case it's indeed the correct solution, which I doubt, I can take care of the failing tests for the Java backend.

grosu:
I don't remember when removing the sort BagItem was discussed ... Anyway, we do not have bandwidth for innovations now; all we want is to merge unified-builtins2 asap, without losing anything we had before.  Any other design decisions can be applied afterwards.

dwightguth:
We removed MapItem, SetItem, ListItem. Why not BagItem?

grosu:
Oh, the sorts, sorry, I thought the wrappers.  So we still write ListItem(X) => Y, but we are not allowed to write X:ListItem => Y.  Yes, I knew that and it makes sense.

But now getting back to Yilong's original question: "What is the sort of a cell, if not BagItem? Bag?".  Yilong, do you need to match generically on one cell?  We need to re-design the entire configuration syntax and internal representation soon anyway, so maybe this can wait until then?  But if you really need the BagItem sort right now, then please let us know.

dwightguth:
You do not. in the worst case you can always say `B:Bag when size(B) ==Int
1`

yilongli:
I think in terms of how the Java backend works. The Java backend might still use BagItem in several places. I don't think I need BagItem. But I will have to look at the source code to understand what goes wrong.

dwightguth:
I no longer want to insist on a large scale refactoring of our use of Bag at this time, but if there's still a bug here, can it get fixed?

radumereuta:
I plan to have another look at it after I'm done with the type inferencer.
I'm still thinking how to handle the sort for rewrites, and it should be ready to be merge.

radumereuta:
Let's delay solving the BagItem issue for the new kast.
I would have to go too much into SDF to fix it, and it won't bring us too much benefit.

PR #598, opened 2014-06-14
added lambda

grosu:
Since we decided that we allow definitions in the unified-builtins2 branch to pass if they pass with the Maude backend only, here is lambda.

### DIFF
--- a/tutorial/1_k/1_lambda/lesson_1/NOTES
+++ b/tutorial/1_k/1_lambda/lesson_1/NOTES
@@ -1,0 +1,5 @@
+We now support the following line to the syntax module:
+
+  syntax priorities 'lambda_._ > '__  // the priority may be beackwards ...
+
+This will allow for more programs with no parenthesis. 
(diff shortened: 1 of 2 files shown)


grosu:
@andreistefanescu @traiansf @dwightguth There isn't much to check here, but just in case, please make sure that this is what we agreed upon.  This is after several beers and 5 bottles of hot sake with our Japanese visitors :)

radumereuta:
Since you are working on this example, please add the following line to the syntax module:

```
syntax priorities 'lambda_._ > '__
```

This will allow for programs with no parenthesis.

traiansf:
@radumereuta Well, actually, the priority needs to go the other way around
:-)

2014-06-14 18:44 GMT+03:00 Radu Mereuta notifications@github.com:

grosu:
Actually, this has no priority at all :)  Now seriously, the idea is to put all our energy in switching to the unified-builtins2 branch.  And for that, we decided to only make it work with the Maude backend, the same way it works in the Master now.  We should refrain from trying/doing anything new right now before the merge.  So I'll just put a note about it and move on.  Thanks for reporting it, Radu!

Grigore

PR #612, opened 2014-06-16
Kool update

grosu:
All three variants of KOOL updated to work with the unified-builtins2 branch.

### DIFF
--- a/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
+++ b/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
@@ -209,7 +209,7 @@ for the KOOL object-oriented constructs.
                 | "if" "(" Exp ")" Block "else" Block   [avoid, strict(1)]
                 | "if" "(" Exp ")" Block
                 | "while" "(" Exp ")" Block
-        	| "for" "(" Stmt Exp ";" Exp ")" Block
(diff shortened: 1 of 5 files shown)


grosu:
@andreistefanescu @traiansf @yilongli Please review asap.  We'd like to be done with the merge by tomorrow afternoon (the 1:35 meeting).

traiansf:
Looks ok, generally.  There are two things I feel uneasy with, though.
1. I know it looks better and is easier to write, but the fact that we don't put sorts when rewriting into builtins makes me feel uneasy.  The reason is that the (semantics) sorting used to act as explicit guards on the transformation while now we have to rely on sort inference doing its job right.
2. I don't particularly like (and I bet @grosu doesn't either) the fact that what used to be easily written with a single rule (albeit matching in the middle of a associative list) now takes 2 rules iterating over that list.  This patterns occurs at least 3 times per definition (if not more) and appears in all definitions.

dwightguth:
Per my discussion with Traian, we are merging this now. The sort inference and associative matching issues are larger issues.

grosu:
Just for the record: I agree with Traian fully on the second issue.  On the first one, I think we can rely on the inferred types provided that we have a super-immediate way to know what sorts have been inferred for variables.  This needs to be discussed.

Issue #602, opened 2014-06-15
error message needed for sorts which do not allow for fresh variables


grosu:
By mistake, I wrote the following rule with `!N:Nat` instead of `!N:Int`:

```
  rule <k> closure(Rho,X,E) V:Val => E ~> env(Rho') ...</k>
       <env> Rho' => Rho[X <- !N] </env>
       <store>... .Map => (!N:Nat |-> V) ...</store>
```

The `krun` showed configurations like the following one:

```
<T>
    <k>
        closure ( x |-> fresh("Nat") , y , (x y) )
    </k>
    <store>
        fresh("Nat") |-> closure ( .Map , x , (lambda y . (x y)) )
    </store>
    <env>
        .Map
    </env>
</T>
```

containing `fresh("Nat")` inside.  An error message saying that the sort `Nat` does not admit fresh variables is needed.

PR #615, opened 2014-06-16
Fixing issue #602

yilongli:
Fixing https://github.com/kframework/k/issues/602.

Now if you use `!N:Nat` in the rule, the kompiler will catch this error and give the following error message:

```
[Error] Compiler: Unsupported sort of fresh variable: Nat
Only sorts [Id, Int] admit fresh variables.
    File: ...
    Location: (24,20,24,26)
    Compilation Phase: class org.kframework.compile.checks.CheckVariables
```


andreistefanescu:
lgtm

yilongli:
The build failed because of the following error.

```
[Error] Compiler: Unsupported sort of fresh variable: K
Only sorts [Id, Int] admit fresh variables.
    File:
/var/lib/jenkins/jobs/k-framework-pull-requests/workspace/include/modules/substitution.k
    Location: (289,41,289,45)
    Compilation Phase: class org.kframework.compile.checks.CheckVariables
```

It turns out that there is one rule in the `substitution` module which is not supported by the current fresh variable generator:

```
  syntax K ::= freshVariables(Set,K)
  rule freshVariables(.Set, T:K) => T
  rule freshVariables(SetItem(Tv:K) S:Set, T:K)
    => freshVariables(S, T[#changeLabel(!Tv',Tv)/Tv])

  syntax K ::= #changeLabel(K,K)  [function]
  rule #changeLabel(_:KLabel(Ks:KList),L':KLabel(_)) => L'(Ks)
```

I am blocked by this. How should I proceed? Comment out the rule? Change `Tv'` to some other sort? @andreistefanescu @grosu

andreistefanescu:
@traiansf how does the substitution work right now?

traiansf:
Well, I think it works because your check is a static check while the code
there relies on the fact that although the variable there is K, it will
only be used on sorts which admit freshness.

2014-06-17 7:39 GMT+03:00 Andrei Stefanescu notifications@github.com:

grosu:
Traian, I do not understand your reply.  You say that "it works" but it does not.  Or are you saying that the check that Yilong implemented is too restrictive?  I personally like this static check.

Another solution could be to include the sort K in the list of those with allowed freshness, but I am not sure how that works with subsorting of other sorts under K.

traiansf:
2014-06-17 14:18 GMT+03:00 Grigore Rosu notifications@github.com:

> Traian, I do not understand your reply. You say that "it works" but it
> does not. Or are you saying that the check that Yilong implemented is too
> restrictive? I personally like this static check.

I'm saying that the check is too restrictive, if we want such behavior.  I
also like this check, but freshVariables is not caring  about sorts.  The
mere usage of the #changeLabel is a witness to this thing.

I'm quoting from your message before those definitions:
// Grigore: I added the operation below here, because I want it to be //
generic. I don't know how to generate globablly fresh variables // of
desired type any other way ... sorry. This is used in type inference

Substitution works because it's not using freshVariables.  The only
definition using it seems to be

   tutorial/1_k/5_types/lesson_9/lambda.k

So I guess we just need to find another (less hacky) way to do
freshVariables, and keep the check in-place.

yilongli:
I will wait for a solution to deal with `freshVariables` so that I can finish this pull request.

dwightguth:
@traiansf is it possible for you to look into a solution for this? This pull request has been sitting around longer than I'd like.

traiansf:
I think the check for fresh Generators should only be done for fresh constants (a.k.a. !-variables).  

Therefore it should be changed as per next diff.  @yilongli can you do it, since it is in your request? @andreistefanescu @dwightguth Does this seem reasonable for you?

```
diff --git a/src/javasources/KTool/src/org/kframework/compile/checks/CheckVariab
index b1026db..aa4eff9 100644
--- a/src/javasources/KTool/src/org/kframework/compile/checks/CheckVariables.jav
+++ b/src/javasources/KTool/src/org/kframework/compile/checks/CheckVariables.jav
@@ -61,8 +61,9 @@ public class CheckVariables extends BasicVisitor {

     @Override
     public Void visit(Variable node, Void _) {
-        if (node.isFreshVariable() || node.isFreshConstant()) {
-            if (!context.freshFunctionNames.containsKey(node.getSort())) {
+        boolean freshConstant = node.isFreshConstant();
+        if (node.isFreshVariable() || freshConstant) {
+            if (freshConstant && !context.freshFunctionNames.containsKey(node.g
                 GlobalSettings.kem.register(new KException(
                         KException.ExceptionType.ERROR,
                         KException.KExceptionGroup.COMPILER,
```

yilongli:
Yeah, I can do that.

dwightguth:
If you fix conflicts and it still passes tests, you can merge this.

PR #621, opened 2014-06-17
Fun updated

grosu:
FUN also works with unified-builtins2.  The substitution-based definition only needed to change one . into .Map.  That was nice.


dwightguth:
Looks like you just need to fix output on this one to make the tests pass.

grosu:
Wow, things move fast!  It worked with yesterday's version of unified-builtins2, but now the pretty printer is better.  Cool.  I'm going to fix the output file immediately.

Grigore

grosu:
Done.

@andreistefanescu @traiansf @dwightguth Please revise (should be really easy)

andreistefanescu:
lgtm

PR #625, opened 2014-06-17
Introduce State object

cos:
**do not review yet**


cos:
will create another pull request applying after we have isNormalized

PR #629, opened 2014-06-19
fix kast --parser binary on java backend

dwightguth:
Nasty hack, I know. I still want to wait for the new kast to fix this properly though.

@radumereuta please review.


radumereuta:
Should do the job for now.

Issue #633, opened 2014-06-19
Error when parsing with null pointer


liyili2:
This error show up this morning when I update my unified-builtins branch to the newest one (I did update yesterday). Before today, this set or rules can be parsed.

When I have this set of rules:

```
module C-SYMBOLIC-VALUE
     syntax SymLoc ::= "NullPointer"
     syntax Bits ::= Int
     syntax CSize ::= Int
     syntax CSize ::= SymVal
     syntax SymLoc ::= SymVal
     syntax Bits ::= SymVal
     syntax SymVal ::= "symVal"
     syntax Int ::= offset(SymLoc) [function]
     rule offset(symVal) => symVal

endmodule
```

It will show this error:

```
Exception in thread "main" java.lang.NullPointerException
    at org.kframework.parser.concrete.disambiguate.CorrectRewritePriorityFilter.visit(CorrectRewritePriorityFilter.java:54)
    at org.kframework.parser.concrete.disambiguate.CorrectRewritePriorityFilter.visit(CorrectRewritePriorityFilter.java:28)
    at org.kframework.kil.Ambiguity.accept(Ambiguity.java:63)
    at org.kframework.kil.AbstractVisitor.visitNode(AbstractVisitor.java:84)
    at org.kframework.kil.AbstractVisitor.genericVisitChild(AbstractVisitor.java:113)
    at org.kframework.kil.AbstractVisitor.visit(AbstractVisitor.java:326)
    at org.kframework.kil.AbstractVisitor.visit(AbstractVisitor.java:344)
    at org.kframework.kil.Rule.accept(Rule.java:72)
    at org.kframework.kil.AbstractVisitor.visitNode(AbstractVisitor.java:92)
    at org.kframework.parser.generator.DisambiguateRulesFilter.visit(DisambiguateRulesFilter.java:49)
    at org.kframework.parser.generator.DisambiguateRulesFilter.visit(DisambiguateRulesFilter.java:26)
    at org.kframework.kil.AbstractVisitor.visit(AbstractVisitor.java:344)
    at org.kframework.kil.Rule.accept(Rule.java:72)
    at org.kframework.kil.AbstractVisitor.visitNode(AbstractVisitor.java:84)
    at org.kframework.kil.AbstractVisitor.genericVisitList(AbstractVisitor.java:135)
    at org.kframework.kil.AbstractVisitor.visit(AbstractVisitor.java:300)
[...]

PR #635, opened 2014-06-19
fix code for computing least sort of KItems

dwightguth:
I fully expect removing this nasty hack from Context.java will break things. So don't review until you see the tests pass.

### DIFF
--- a/src/javasources/KTool/src/org/kframework/kil/loader/Context.java
+++ b/src/javasources/KTool/src/org/kframework/kil/loader/Context.java
@@ -190,7 +190,6 @@ public class Context implements Serializable {
 //            labels.put(label, s = new HashSet<String>());
 //        }
 //        s.add(cons);
-        putLabel(p.getLabel(), cons);
         putLabel(p.getKLabel(), cons);
     }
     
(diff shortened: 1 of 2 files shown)


dwightguth:
@yilongli please review.

radumereuta:
@dwightguth @yilongli  you should have a look at #633 first. I think there is a similar problem there.

src/javasources/KTool/src/org/kframework/kil/loader/Context.java
```diff
@@ -190,7 +190,6 @@ public void putLabel(Production p, String cons) {
 //            labels.put(label, s = new HashSet<String>());
 //        }
 //        s.add(cons);
-        putLabel(p.getLabel(), cons);
```
yilongli, line 193: Is this change related to this fix?

dwightguth, line 193: Yes. Previously if you had two productions whose klabel differed only in
the initial apostrophe, both productions would be considered as being
associated with the klabel without the apostrophe. That is because of this
line of code, which is a bug.

yilongli:
lgtm

PR #638, opened 2014-06-19
add ==K for Bags

dwightguth:

### DIFF
--- a/include/builtins/k-equal.k
+++ b/include/builtins/k-equal.k
@@ -17,9 +17,6 @@ module K-EQUAL-HOOKS
   syntax Bool ::= left:
   				  K "==K" K           [function, hook(#K-EQUAL:_==K_), latex({#1}\mathrel{=_K}{#2}), equalEqualK]
 				| K "=/=K" K          [function, hook(#K-EQUAL:_=/=K_), latex({#1}\mathrel{\neq_K}{#2}), notEqualEqualK]
-				// TODO(dwightguth): fix once bags are K
-  				| Bag "==K" Bag           [function, hook(#K-EQUAL:_==K_), latex({#1}\mathrel{=_K}{#2}), equalEqualK]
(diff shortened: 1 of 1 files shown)


include/builtins/k-equal.k
```diff
@@ -17,6 +17,9 @@ module K-EQUAL-HOOKS
   syntax Bool ::= left:
   				  K "==K" K           [function, hook(#K-EQUAL:_==K_), latex({#1}\mathrel{=_K}{#2}), equalEqualK]
 				| K "=/=K" K          [function, hook(#K-EQUAL:_=/=K_), latex({#1}\mathrel{\neq_K}{#2}), notEqualEqualK]
+				// TODO(dwightguth): fix once bags are K
+  				| Bag "==K" Bag           [function, hook(#K-EQUAL:_==K_), latex({#1}\mathrel{=_K}{#2}), equalEqualK]
+				| Bag "=/=K" Bag          [function, hook(#K-EQUAL:_=/=K_), latex({#1}\mathrel{\neq_K}{#2}), notEqualEqualK]
```
radumereuta: Shouldn't this be ==Bag?

dwightguth: Should it? I thought the intuition was that ==K works on all terms...

I don't feel that strongly about it though. Andrei, what do you think?

radumereuta: I don't feel strongly about this either. I'm just making a suggestions. I think we had ==Bag in the past.
Although this way looks more uniform.

andreistefanescu: This should be ==Bag, because the convention is to use the sort of the arguments, not of the result (to be consistent with >=Int, not >=Bool).

dwightguth: Okay. I will change it.

andreistefanescu:
lgtm

PR #650, opened 2014-06-23
changed machine integers to be more amenable for symbolic computation

andreistefanescu:


andreistefanescu:
@dwightguth @yilongli please review. I cannot test it with the CIL semantics because I get the NPE from the parser.

dwightguth:
lgtm. Merge if tests pass. I will take it from here if there are more issues with cil once we have fixed the sort diamond issue.

yilongli:
lgtm

PR #653, opened 2014-06-23
Issue 607

traiansf:
Apparently there were some issues with the substitution when using bounded symbolic variables.  Not sure why these were not caught by the lambda tests.  To address them, I had to force syntactic equality in the process of computing the free variables.  The main thing I'm not proud of is basically changing the definition of -Set to a syntactic one (by abusing ==K true).  @andreistefanescu says that is probably ok.  If anybody objects to that, I can create a -SetSyn or something similar for syntactic difference.

@andreistefanescu @grosu @dwightguth please review.

### DIFF
--- a/include/builtins/set-impl.k
+++ b/include/builtins/set-impl.k
@@ -7,16 +7,15 @@ module SET-IMPL
 
   /* set difference */
 //  syntax Set ::= Set "-Set" Set   [function, hook(Set:difference), latex({#1}-_{\it Set}{#2})]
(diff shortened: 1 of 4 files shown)


grosu:
I cannot see all the implications of what you've done, but I like the changes in lesson_4.  The fact that works is good enough for me, so lgtm.

Grigore

dwightguth:
lgtm2

Andrei, you can merge if you are happy.

andreistefanescu:
lgtm

PR #656, opened 2014-06-24
Tool fixes to handle spaces in file paths

### LINKED ISSUE #655: robustly handle spaces in directory names
Several K tools fail if they encounter spaces in path names. Some of the problems arise when these spaces are in paths leading to the K installation directory. Others come up if there are spaces in paths to user files. In either case, the tools should be quoting or escaping spaces properly to robustly handle these situations.

liblit:
These changes collectively fix #655. Most of the changes add proper double-quoting to shell scripts so that they properly handle spaces in the path leading to the K installation directory. The `KRunner` tweaks handle the case of spaces in the path leading to the user's `*-kompiled` directory. For that we use backslash-escaping, a feature supported by Maude's `load` command.

Unfortunately, running `ant test` currently fails even for a pristine snapshot of the K master branch in directories with no spaces. Therefore, I can't definitively say that this fixes all spaces-in-paths bugs while introducing no new regressions. I can say that my changes are both necessary and sufficient to make the first several tutorial steps work when there are spaces in paths. Those steps involve using `kompile`, `krun`, and `kast`. I have not tried other K tools suck as `kagreg`, `kpp`, or `ktest`, even though I have made the obvious fixes to those shell scripts too.

### DIFF
--- a/bin/kagreg
+++ b/bin/kagreg
@@ -4,8 +4,8 @@ ulimit -s `ulimit -H -s`
 if [ -z "$K_OPTS" ];
   then export K_OPTS="-Xms64m -Xmx1024m -Xss32m"
 fi
-if "$(dirname $0)/../lib/scripts/checkJava"; then
-  java -Djava.awt.headless=true $K_OPTS -ea -jar "$(dirname $0)/../lib/java/k3.jar" -kagreg "$@"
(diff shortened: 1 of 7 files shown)


liblit:
Drat, I noticed a small glitch. I'll submit a new pull request once that glitch is fixed. Sorry for the noise!

PR #657, opened 2014-06-24
Tool fixes to handle spaces in file paths

liblit:
These changes collectively fix issue #655. Most of the changes add proper double-quoting to shell scripts so that they properly handle spaces in the path leading to the K installation directory. The `KRunner` tweaks handle the case of spaces in the path leading to the user's `*-kompiled` directory. For that we use backslash-escaping, a feature supported by Maude's `load` command.

Unfortunately, running `ant test` currently fails even for a pristine snapshot of the K master branch in directories with no spaces. Therefore, I can't definitively say that this fixes all spaces-in-paths bugs while introducing no new regressions. I can say that my changes are both necessary and sufficient to make the first several tutorial steps work when there are spaces in paths. Those steps involve using `kompile`, `krun`, and `kast`. I have not tried other K tools suck as `kagreg`, `kpp`, or `ktest`, even though I have made the obvious fixes to those shell scripts too.

### DIFF
--- a/bin/kagreg
+++ b/bin/kagreg
@@ -5,7 +5,7 @@ if [ -z "$K_OPTS" ];
   then export K_OPTS="-Xms64m -Xmx1024m -Xss32m"
 fi
 if "$(dirname "$0")/../lib/scripts/checkJava"; then
-  java -Djava.awt.headless=true $K_OPTS -ea -jar "$(dirname "$0)/""../lib/java/k3.jar" -kagreg "$@"
+  java -Djava.awt.headless=true $K_OPTS -ea -jar "$(dirname "$0")/../lib/java/k3.jar" -kagreg "$@"
 else
   exit 1
 fi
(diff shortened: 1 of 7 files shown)


grosu:
Thanks, Ben.  We will take care of it and merge it asap.  You said that `ant test` does not work even with a fresh installation of K.  That should not happen, unless it is a non-critical Latex issue (packages missing on your platform).  Please let us know if it fails on anything else except on generating the PDF files.

Grigore

liblit:
Thanks, Grigore. Since you've clarified that `ant test` failures should not be occurring, I have filed a new issue #659 with more details about that. I might not have time to track down the root cause right now, but I can at least document the regression fully.

dwightguth:
Jenkins: ok to test

liblit:
Jenkins rejected these changes: checkstyle flagged a tab character that I introduced. OK, no problem, I can fix that and resubmit. I am a bit new to github: is the proper procedure for me to close this pull request, correct that glitch, and then create a whole new pull request?

dwightguth:
You don't have to. Just commit a new commit to the same branch and push it,
and it will be added to the current pull request.

liblit:
Aah, OK. Good to know. I wish I'd known that before I closed #656 just to recreate it as #657. Ah well, I'll know better next time. Thanks, and sorry for the extra noise from that!

PR #661, opened 2014-06-24
fix some bugs in ListBuiltin

dwightguth:
- make CollectionBuiltin abstract
- make ListBuiltin constructors take List
- make ListBuiltin constructor private
- fix bugs in ListBuiltin.of
- Make #noIO not require a list variable

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -178,7 +178,7 @@
 	</target>
 
     <target name="junit" depends="test-build">
-       <junit haltonfailure="true" fork="true" dir="${javasources}/KTool">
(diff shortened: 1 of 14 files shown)


dwightguth:

andreistefanescu:
Do you need to make any modifications to the transformers?

dwightguth:
If I understand correctly, the one line change to AbstractVisitor which causes it to call ListBuiltin.of when it copies objects should be sufficient to take advantage of this diff.

andreistefanescu:
lgtm

PR #667, opened 2014-06-25
If I can't find a LUB for the rewrite, then use K. This is a temporary fix.

radumereuta:


radumereuta:
@dwightguth please review.

PR #668, opened 2014-06-25
Maude stuff

dwightguth:
@andreistefanescu @traiansf please review.

### DIFF
--- a/include/builtins/mint.k
+++ b/include/builtins/mint.k
@@ -16,6 +16,10 @@ module MINT
   /* Machine integer of bit width and value. */
   syntax MInt ::= mi(Int, Int)    [function, hook(#MINT:constructor)]
 
+  // needed because maude doesn't generate predicates for functions, but in
+  // the maude backend mi isn't really a function
+  rule isMInt(mi(_, _)) => true [predicate]
+
   /* Function returning the bit width of this machine integer. */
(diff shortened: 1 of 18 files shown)


include/builtins/mint.k
```diff
@@ -16,6 +16,10 @@ module MINT
   /* Machine integer of bit width and value. */
   syntax MInt ::= mi(Int, Int)    [function, hook(#MINT:constructor)]
 
+  // needed because maude doesn't generate predicates for functions, but in
+  // the maude backend mi isn't really a function
+  rule isMInt(mi(_, _)) => true [predicate]
```
andreistefanescu: I'm not sure I want this rule to be imported in the java backend. @traiansf was there a way to say that this rule is a maude implementation rule only?

traiansf: This was supposed to be part of the unified-hooks branch which I did not requested a pull for, yet.

traiansf: However, I don't really understand why matching on mi is allowed since mi is a function.  maybe it should be called something else?

dwightguth:
@andreistefanescu does this look better?

andreistefanescu:
I'm not sure if adding a production to MInt does not break anything. But if the tests pass, you can merge.

dwightguth:
Jenkins: test this please

PR #671, opened 2014-06-25
Added a check to test for duplicate module names.

radumereuta:
Got a weird nasty NPE because I didn't have this check.

### DIFF
--- a/include/autoinclude-java.k
+++ b/include/autoinclude-java.k
@@ -15,6 +15,6 @@ module AUTO-INCLUDED-MODULE
   imports LIST
   imports MAP
   imports SET
-  imports IO
+  imports K-IO
 endmodule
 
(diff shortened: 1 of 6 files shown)


radumereuta:
@dwightguth please review.

dwightguth:
:/ looks like we don't currently obey this restriction. You need to fix agent.

radumereuta:
@traiansf I think agent is under your keep.
Do you need that module name to be called 'IO' ?
I found a bug in the parser which I can avoid by having unique module names.

If you disagree with having unique names for modules, then I need a semantics on what to do if that happens (merge the modules, replace...)

chathhorn:
Shouldn't you change the standard module names if you're going to enforce this? I don't see why the standard modules should get all the good names.

dwightguth:
We could make the builtin modules begin with "K-"... e.g. K-IO.

radumereuta:
@dwightguth, @traiansf please review.

traiansf:
lgtm :-)

Issue #659, opened 2014-06-24
`ant test` produces hundreds of checkstyle 5.7 failures


liblit:
I have a completely pristine checkout of the latest K master branch. I run `ant` followed by `ant test`. The latter produces 656 failures: one for each ofthe 656 files checked. This looks like a failure of the testing system rather than actual style violations.

I won't give the full output here, as it is quite long and fairly repetitive. Here are some representative samples, starting with the first few lines of the `checkstyle` phase of `ant test`'s execution:

```
checkstyle:
[checkstyle] Running Checkstyle 5.7 on 656 files
[checkstyle] Can't find/access AST Node typecom.puppycrawl.tools.checkstyle.api.DetailAST
[checkstyle] Can't find/access AST Node typecom.puppycrawl.tools.checkstyle.api.DetailAST
[checkstyle] Can't find/access AST Node typecom.puppycrawl.tools.checkstyle.api.DetailAST
...
[checkstyle] /home/liblit/Downloads/k/src/javasources/KTool/src/org/kframework/backend/Backend.java:0: Got an exception - java.lang.ClassCastException: antlr.CommonAST cannot be cast to com.puppycrawl.tools.checkstyle.api.DetailAST
[checkstyle] /home/liblit/Downloads/k/src/javasources/KTool/src/org/kframework/backend/BackendFilter.java:0: Got an exception - java.lang.ClassCastException: antlr.CommonAST cannot be cast to com.puppycrawl.tools.checkstyle.api.DetailAST
[checkstyle] /home/liblit/Downloads/k/src/javasources/KTool/src/org/kframework/backend/BasicBackend.java:0: Got an exception - java.lang.ClassCastException: antlr.CommonAST cannot be cast to com.puppycrawl.tools.checkstyle.api.DetailAST
...
BUILD FAILED
/home/liblit/Downloads/k/build.xml:204: Got 656 errors and 0 warnings.
```

[...]

PR #673, opened 2014-06-25
add known issue to readme

dwightguth:
FYI @liblit

If this seems clear to you let me know and I'll merge it.

### DIFF
--- a/src/README
+++ b/src/README
@@ -67,6 +67,10 @@ Common error messages:
 +  Git is not accessible in the command line. Please reinstall git and make
    make sure to check to option to make it available in the command line.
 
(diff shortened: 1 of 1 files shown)


liblit:
In the discussion of issue #659, I concluded that having `ant-antlr` installed on my system triggered the problem. But that package does not actually include `antlr` per se. So the proposed advice does not seem applicable. When I have `ant-antlr` installed, it is indeed true that "no version of antlr is installed in the ant system classpath other than antlr 2.7.7", but the problem still arises.

dwightguth:
@liblit is this better?

liblit:
Yes, that seems just right. Thanks!

PR #674, opened 2014-06-26
Krun options

dwightguth:
**please review**
- use JCommander for krun
- clean up krun frontend to be more generic
- fix some hacky behavior in some of the options


dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
@radumereuta @andreistefanescu @traiansf please review. @osa1 please review ktest change

samples/quine/explicit/quine-explicit.k
```diff
@@ -5,11 +5,11 @@
         (\texttt{radu.mereuta@info.uaic.ro})}
 \organization{University of Al. I. Cuza, Ia\c{s}i}
 */
-//@Usage: kompile and then run with krun --output none
+//@Usage: kompile and then run with krun -cPGM='""' --output none
 module QUINE-EXPLICIT
 
   configuration
-    <k> "" ~> "" </k>
+    <k> $PGM ~> $PGM </k>
```
radumereuta: That's not going to be ok.
A quine program must take no input. Please revert this.
I should be able to just run a definition and not give it any input. If krun can't do that, then it has a problem.

samples/quine/short/quine-short.k
```diff
@@ -2,8 +2,9 @@
 module QUINE-SHORT
 
   configuration
+    <k> $PGM:K </k>
     <out stream="stdout"> .List </out>
-    <prefix> "// Copyright (c) 2013-2014 K Team. All Rights Reserved.\nmodule QUINE-SHORT\n\n  configuration\n    <out stream=\"stdout\"> .List </out>\n" </prefix>
+    <prefix> "// Copyright (c) 2013-2014 K Team. All Rights Reserved.\nmodule QUINE-SHORT\n\n  configuration\n    <k> $PGM:K </k>\n    <out stream=\"stdout\"> .List </out>\n" </prefix>
```
radumereuta: Same as above. $PGM not acceptable in Quine.

src/javasources/KTool/src/org/kframework/utils/ColorUtil.java
```diff
@@ -30,17 +29,19 @@
      * A cache to avoid computing the closest terminal color for a given color each time it is needed.
      */
     private static Map<Map<Color, String>, Map<Color, String>> colorToCodeConvertCache;
-
+
```
radumereuta: You need to add the check for trailing whitespace and windows line ending.

radumereuta:
I managed to skimp over all of it, but I don't trust myself enough. That's a big diff, so wait until you have more opinions.
If you fix the quine examples, then lgtm.

samples/quine/explicit/quine-explicit.k
dwightguth: krun can do that. it's ktest that can't. If you want I can revert this, but then we have to take quine out of the regression test suite. I really don't want to keep the broken ability to specify configuration parameters that don't even exist.

traiansf: I don't understand what's the issue here... I think Radu is right to require no program for the definition. For ktest to work all one needs is a program which parses, even if the program is not used in the configuration.  We already have other tests which behave similarly...

traiansf:
It looks good to me in general, although I don't subscribe to the fact that you want all tests to actually use their program.  If you are really determined to do that, then I would suggest altering ktest to allow krunning definitions without programs so such definitions can be tested, too.  This should make everybody happy.  I'm working on some examples now (model checking pushdown systems) where it would not make much sense to have programs testing them, but rather using different LTL formulas for the same spec.

As Radu said, it's a huge diff, so is is something in particular you would want me to look at?
If not, and all tests pass, then it lgtm.

dwightguth:
I believe I have addressed all review comments. If nobody else has suggestions, I will merge this Monday.

osa1:
Sorry for the delay. I'll review ktest related parts today.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/JavaExecutionOptions.java
```diff
@@ -0,0 +1,30 @@
...
+import org.kframework.krun.K;
+
+import com.beust.jcommander.Parameter;
+
+
+public final class JavaExecutionOptions {
+    @Parameter(names="--generate-tests", description="Test programs will be generated along with normal search.")
+    public boolean generateTests = false;
+    
+    @Parameter(names="--deterministic-functions", description="Throw assertion failure during execution in the java backend if function definitions are not deterministic.")
```
osa1: is it possible to make this line and similarly long ones follow style guidelines?

src/javasources/KTool/src/org/kframework/backend/latex/PdfBackend.java
```diff
@@ -64,7 +64,7 @@ public void run(Definition definition) throws IOException {
         File latexFile = latexBackend.getLatexFile();
         File pdfFile = generatePdf(latexFile);
         if (pdfFile.exists()) {
-            copyFile(pdfFile, new File(options.directory, FilenameUtils.removeExtension(new File(definition.getMainFile()).getName()) + ".pdf"));
+            copyFile(pdfFile, options.output());
```
osa1: I think this is wrong. We need to replace output file extension file `.pdf`.

osa1:
@dwightguth This branch is missing some commits I've done that have some changes in PdfBackend, LatexBackend and Kompile. Those fix some bugs so we need them. Are we going to continue developing on `unified-builtins2` branch? If that's the case, who's supposed to pick/merge those commits from `master`?

src/javasources/KTool/src/org/kframework/kompile/KompileFrontEnd.java
```diff
@@ -201,22 +205,23 @@ private static void genericCompile(KompileOptions options, Backend backend, Stri
...
-            }
+                        KExceptionGroup.CRITICAL, msg, null,
+                        kompiled.getCanonicalPath()));
+        }
+    }
+    
+    private static void checkIsFile(File kompiled) throws IOException {
+        if (kompiled.exists() && kompiled.isDirectory()) {
+                String msg = "Cannot write output to " + kompiled.getCanonicalPath()
+                        + "; file already exists and is a directory.";
```
osa1: What's wrong with having a directory with same name?

src/javasources/KTool/src/org/kframework/kompile/KompileFrontEnd.java
```diff
@@ -102,53 +98,61 @@ public void run() {
         Backend backend = null;
         switch (options.backend) {
             case PDF:
+                checkIsFile(options.output());
```
osa1: ~~I don't think those `checkIsFile` calls are correct. Not all backends generate a folder with `definitionName + "-kompiled"` name. (for example, latex and pdf backends).~~ (I just saw KompileOptions.java and I'm adding my concerns there)

Another confusing thing is that `checkIsFile` is actually looking for a directory, not a file. So why it's name is`checkIsFile` and not `checkIsDir` ?

src/javasources/KTool/src/org/kframework/kompile/KompileOptions.java
```diff
@@ -65,8 +65,35 @@ public File mainDefinitionFile() {
     /**
      * Directory in which the compiled definition should be put.
      */
-    @Parameter(names={"--directory", "-d"}, description="Path to the directory in which the output resides. An output can be either a kompiled K definition or a document which depends on the type of backend. The default is the current directory.")
-    public File directory = new File(".");
+    @Parameter(names={"--output", "-o"}, description="Path to the file in which to place the output. An output can be either a kompiled K definition or a document which depends on the type of backend. The default is <file>-kompiled for executable backends, where <file> is the name of the file being compiled, without file extension. The defaults for the pdf, latex, and html backends are <file>.pdf, <file>.tex, and <file>.html respectively.")
+    private File output;
+    
+    public File output() {
```
osa1: I don't like that part for this reason: If we want to change some output files in the future we need to figure that this code decides on what files we're generating and we need to modify this part too. Also, it's not possible to specify multiple files and we can't specify whether we're generating a file of directory. We'll get errors if we're generating a file but we have a folder with same name.(and the other way around)

Instead why not put generated files to backend interface and make frontend ask to backend for "tell me what files are you generating so that I can check them for you" ?

So main problem is that we're keeping related parts in completely unrelated parts. Some info related with specific backends are defined in kompile frontend.

src/javasources/KTool/src/org/kframework/krun/K.java
```diff
@@ -25,124 +13,22 @@
...
-    }
-
-    public static Color getTerminalColor() {
-        if (terminalColor == null) {
-            terminalColor = ColorUtil.getColorByName("black");
+    
+    public static void setTool(Tool tool) {
+        if (K.tool != null) {
+            assert false : "Can only call setTool once";
         }
```
osa1: Why not use

``` java
assert (K.tool == null) : "Can only call setTool once";
```

instead?

src/javasources/KTool/src/org/kframework/ktest/Proc.java
```diff
@@ -249,7 +249,7 @@ public void run() {
      * @param returnCode return code of the process
      */
     private void handlePgmResult(int returnCode) throws IOException {
-        String red = ColorUtil.RgbToAnsi(Color.RED, colorSetting);
+        String red = ColorUtil.RgbToAnsi("red", colorSetting, Color.BLACK);
```
osa1: When is this changed? Why terminal color is assumed to be black?

src/javasources/KTool/src/org/kframework/ktest/Test/TestSuite.java
```diff
@@ -117,7 +117,7 @@ public boolean run() throws InterruptedException, TransformerException,
         if (!skips.contains(KTestStep.KRUN))
             ret &= runKRunSteps(filterSkips(successfulTests, KTestStep.KRUN));
 
-        String colorCode = ColorUtil.RgbToAnsi(ret ? Color.green : Color.red, colorSetting);
+        String colorCode = ColorUtil.RgbToAnsi(ret ? "green" : "red", colorSetting, Color.BLACK);
```
osa1: Again, I don't understand why terminal color is assumed to be black.

src/javasources/KTool/src/org/kframework/ktest/Test/TestCase.java
```diff
@@ -74,6 +79,16 @@ public TestCase(Annotated<String, LocationData> definition,
                     ProgramProfile krunOpts,
                     Map<String, ProgramProfile> pgmSpecificKRunOpts,
                     Set<KTestStep> skips) throws InvalidConfigError {
+        File definitionFile = new File(definition.getObj());
+        File output = new File(definitionFile.getParentFile(), 
+                FilenameUtils.removeExtension(definitionFile.getName()) + "-kompiled");
+        for (PgmArg arg : kompileOpts) {
+            if (arg.arg.equals("--output")) {
```
osa1: We may need to use `--output` argument in some other places in ktest too. I'll review this part in details later.

osa1:
We don't have unit or regression tests for frontend and we're relying completely on code reviews. But this patch is huge, with lots of unrelated changes. How are changes in colors related with option parsing/injecting? There are probably other unrelated changes too, like moving/renaming files(I didn't read the whole patch line by line, it's impossible without spending at least a whole day). So here we're trying to review lots of different patches, combined together. I think patches like this, with the fact that we don't have tests at all, will cause bugs like we experienced before.

Why not spend time on building unit/regressions tests for front-end parts instead of this kinds of things? I believe tests should have top priority right now, all others are relatively less important.

EDIT: Also, we're still violating style guidelines, even with all the automated tools that warn us against doing that.(checkstyle, jenkins etc.)

src/javasources/KTool/src/org/kframework/kompile/KompileFrontEnd.java
dwightguth: you can't write a file to a location where a file with the same name is a directory without deleting the directory first. Doing that didn't seem like the right solution.

src/javasources/KTool/src/org/kframework/kompile/KompileFrontEnd.java
dwightguth: checkIsFile is checking that if the output already exists that it's a file. I will add comments on the functions though.

src/javasources/KTool/src/org/kframework/ktest/Proc.java
dwightguth: previously it was already assuming the terminal color is black, it was just depending on implicit static state. By removing the static state, it was necessary to specify this additional information. It doesn't affect anything to assume the terminal is black, except that if the background of the terminal is actually red, it won't do anything special. But again that's already the case.

dwightguth:
@osa1 in response to your question, as I requested earlier, it is the responsibility of the developer (i.e. you) to merge changes made in the master branch into unified-builtins.

As regards line length, I will work on a checkstyle check for this next, but it should be a separate change. As regards the output files, I changed them to work this way to be consistent with other compiler tools, like gcc, where previously the way the code worked was hacky at best. As regards testing, in honesty the code base as it currently exists is not testable. Removing static state like this and other changes I have planned for the immediate future are a needed first step before that is possible.

I will address the rest of your concerns as much as possible in the commit, but I think we need to discuss some of this in more detail in a higher bandwidth format than github. Do you have a google account that I can use to IM you?

src/javasources/KTool/src/org/kframework/ktest/Proc.java
osa1: We can definitely live with it but I don't think this API is very good for several reasons:
1. How do we expect user of this API to decide what's the current terminal's background color?
2. If we want to use black as default, how do we expect user to know what default we're using? Instead I think we should provide a function that uses the default value for color.
3. Let's say in the future we decided to use `$COLORFGBG` to detect background color. We'll have to modify every single call site.

But like I said it's not a huge problem. (we have more important problems IMO, like tests)

osa1:
Sure, you can use my email address(as seen in k-list) for GTalk/IM. I'm also online at irc.freenode.org with nickname osa1.

src/javasources/KTool/src/org/kframework/kompile/KompileFrontEnd.java
osa1: Interesting, I didn't know that.

src/javasources/KTool/src/org/kframework/ktest/Proc.java
dwightguth: I will put a todo at each of these call sites telling me to fix this code when I introduce dependency injection. Dependency injection is the tool that would allow us to implement this functionality correctly in a clean way.

dwightguth:
@osa1 please re-review

src/javasources/KTool/src/org/kframework/backend/java/kil/ConstrainedTerm.java
```diff
@@ -256,7 +256,8 @@ public Term term() {
             solutions.add(candidate);
         }
 
-        if (K.do_testgen && !solutions.isEmpty()) {
+        if (context.definition().context().javaExecutionOptions.generateTests
```
osa1, line 259: Isn't execution option related with arguments that are passed to krun? (or in other words, it's related with krun and not related with the definition in any sense) Why is this info part of the definition context?

dwightguth, line 259: It's a migration pattern. We need to get this info into any class in
kompile or krun which asks for it. Using context which is already passed
everywhere is how I chose to do it. The change I am working on next adds a
dependency on Google Guice, a dependency injection framework. Eventually,
the constructor to ConstrainedTerm will take this parameter as a Boolean.
That value will be initialized by Guice's injector after having been
initialized at the beginning of the process execution from an inner class
of JavaExecutionOptions, which will no longer be passed anywhere following
the initialization of the application.

Does that answer satisfy you? In order to do this, I must first eliminate
static state, which is the primary purpose of this change.

src/javasources/KTool/src/org/kframework/backend/maude/krun/MaudeKRun.java
```diff
@@ -45,17 +44,46 @@
...
+        xmlOutFile = new File(krunTempDir, "maudeoutput.xml");
+        processedXmlOutFile = new File(krunTempDir, "maudeoutput_simplified.xml");
+        
+        if (!context.krunOptions.global.debug) {
+            Runtime.getRuntime().addShutdownHook(new Thread() {
+                public void run() {
+                    try {
+                        FileUtils.deleteDirectory(krunTempDir);
+                    } catch (IOException e) {
+                        e.printStackTrace();
```
osa1, line 77: Is this really how we want to report a failure? (I'm not against it but we had problems in the past, about uncaught exceptions with unhelpful messages to the user)

Maybe we need something like "please file a bug report this with info" message.

Another thing that may be a problem is that we don't re-throw the exception. (although that may not be a problem in this case since this will be run on shutdown)

Since this is probably not the only place where we just print stack trace and ignore the exception, we can probably ignore this for now and fix them all later.

dwightguth, line 77: Well, this is actually kind of a funny case. The standard code pattern here
is to check the debug flag and print the stack trace only if it's true. But
this shutdown hook is only created if debug is false. So printing a stack
trace actually is appropriate here. Otherwise you couldn't ever get this
stack trace. But I agree that some kind of "please file this as a bug"
message is appropriate for uncaught exceptions. I will fix this in a
separate pull request though.

src/javasources/KTool/src/org/kframework/backend/maude/krun/MaudeKRun.java
```diff
@@ -426,7 +472,7 @@ private static String getSearchType(SearchType s) {
         if (s == SearchType.PLUS) return "+";
         if (s == SearchType.STAR) return "*";
         if (s == SearchType.FINAL) return "!";
-        return null;
+        throw new NullPointerException("null SearchType");
```
osa1, line 475: This is an unreachable code. Maybe we should use same pattern for every unreachable cases in the code base. I just searched in the code base for "unreachable" and saw these patterns used:

``` java
throw e; //unreachable
return null; // unreachable code
throw new AssertionError("unreachable");
```

At least adding a comment or changing the string argument so that a reader can know this is supposed to be unreachable would be nice.

Best option would be using exhaustive pattern matching like in typed functional languages but I don't know how to do that in Java or if it's even possible.

dwightguth, line 475: But it's not unreachable, it can be reached by passing null to this
function. That's just never done. But throwing NullPointerException when a
method disallows a null argument is standard practice: the JDK does this.

src/javasources/KTool/src/org/kframework/krun/KRunFrontEnd.java
```diff
@@ -0,0 +1,535 @@
...
+        } else {
+            throw new AssertionError("currently only three execution backends are supported: "
+                    + "MAUDE, SYMBOLIC, and JAVA");
+        }
+    }
+
+    /**
+     * Execute krun in normal mode (i.e. not in debug mode).
+     * @param initialConfiguration
+     * @param context
```
osa1: param parts are wrong.

src/javasources/KTool/src/org/kframework/utils/ColorUtil.java
```diff
@@ -376,25 +377,25 @@ private static String getEightBitTerminalCode(int code) {
         return "\u001b[38;5;" + code + "m";
     }
 
-    public static String RgbToAnsi(Color rgb, ColorSetting colorSetting) {
-        colors(); //init static maps if needed
+    public static String RgbToAnsi(String rgb, ColorSetting colorSetting, Color terminalColor) {
```
osa1: Why use strings instead of enums for color? In other places you did the opposite. (for example in K.Tool, that was represented as a string before)

src/javasources/KTool/src/org/kframework/backend/maude/krun/MaudeKRun.java
osa1, line 475: Right, joys of working with observable bottom values.

src/javasources/KTool/src/org/kframework/backend/java/kil/ConstrainedTerm.java
osa1, line 259: OK.

src/javasources/KTool/src/org/kframework/utils/ColorUtil.java
dwightguth: This is to prevent the classloader from loading java.awt.color prematurely,
which would interfere with our ability to set whether the jvm is headless.
Do you want me to make a comment of it somewhere?

dwightguth: wait a minute, I think I did this wrong. Let me try to revert this part of the change and see what happens.

dwightguth:
@osa1 if you have more comments, go ahead. I think I addressed everything you've given so far though. If there's nothing else, can I merge?

osa1:
Yes, thanks. Sorry if I asked for too many things. I hope it's in a better shape right now.

dwightguth:
No, your feedback was good, detailed, caught some good stuff. Thank you.

I will merge this tomorrow afternoon unless anyone has more feedback. Last call @andreistefanescu @traiansf @radumereuta

andreistefanescu:
lgtm

Issue #684, opened 2014-06-27
Better error message needed when forgetting ? in front of fresh variables


grosu:
This rule

```
  rule <k> lambda X:Id . E:Exp => T:Type -> (E)::K ~> env(TEnv) ...</k>
       <tenv> TEnv:Map => TEnv[X <- T] </tenv>
```

yields the following error message:

```
[Error] Compiler: Unbounded Variable T:Type .
    File: /home/grosu/github.grosu/k/tutorial/1_k/5_types/lesson_5.5/lambda.k
    Location: (44,37,44,38)
    Compilation Phase: class org.kframework.compile.checks.CheckVariables
```

Everything works fine if I replace `T` with `?T`.

radumereuta:
What would you have expected to happen, and why?

grosu:
An error message saying that unbounded variables should start with either ? or !.

Because that's what we decided, to make sure people do not mistype a variable and suddenly get symbolicity without even realizing it.  When you type ? or ! you know what you are doing.

radumereuta:
But there are other cases when you just simply miss typed the name of a variable on the right hand side.
I think the current error message is correct, and is more general. There are other cases to consider other than fresh.

grosu:
I don't understand how to make it any clearer than I just did, without saying that 1+1=2.  Let me know what is not clear to you, or otherwise explain clearer what you mean.

PR #685, opened 2014-06-27
Updating the error message for unbound variables. Fixing issue #684.

radumereuta:
@dwightguth @grosu please review.
Now the error message will say: 

```
[Error] Compiler: Unbounded variable X:KList should start with ? or ! to make
it fresh.
    File: D:\work\k\samples\test\test.k
    Location: (6,17,6,18)
    Compilation Phase: class org.kframework.compile.checks.CheckVariables
```

### DIFF
--- a/src/javasources/KTool/src/org/kframework/compile/checks/CheckVariables.java
+++ b/src/javasources/KTool/src/org/kframework/compile/checks/CheckVariables.java
@@ -7,6 +7,7 @@ import org.kframework.kil.loader.Context;
 import org.kframework.kil.visitors.BasicVisitor;
 import org.kframework.kompile.KompileOptions;
 import org.kframework.utils.errorsystem.KException;
+import org.kframework.utils.errorsystem.KException.ExceptionType;
 import org.kframework.utils.general.GlobalSettings;
 
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


src/javasources/KTool/src/org/kframework/compile/checks/CheckVariables.java
```diff
@@ -121,19 +122,16 @@ public Void visit(Sentence node, Void _) {
...
-                            "Unbounded Variable " + v.toString() + ".",
-                            getName(), v.getFilename(), v.getLocation()));
-                }
+                if (!options.backend.java())
+                    exType = KException.ExceptionType.ERROR;
+                else
+                    exType = KException.ExceptionType.WARNING;
+                GlobalSettings.kem.register(new KException(exType,
+                        KException.KExceptionGroup.COMPILER,
+                        "Unbounded variable " + v.toString() + "should start with ? or ! to make it fresh.",
```
grosu: I would get rid of "to make it fresh".  Other than that, it is exactly what we wanted here.  Thanks, Radu.

dwightguth:
Okay making the behavior change based on the backend here makes no sense.

dwightguth:
Jenkins: test this please

grosu:
I don't know how the implementation is, but this is a general check that needs to be done, not a backend-specific one.  So we should have this check before any backend.

traiansf:
Isn't it possible for the unbounded variable on the right to be universally
quantified as it sometimes happens in logic programming?

2014-06-27 21:32 GMT+03:00 Grigore Rosu notifications@github.com:

grosu:
I don't understand your question, Traian.  Can you be more explicit?  I mean, what does this have to do with the backend?

traiansf:
The error message is basically saying that an unbounded variable on the
right hand side must be existentially quantified (as both ? and ! are).
I was just asking whether it is unconceivable to have unbounded variables
which are universally quantified.

For example, in the Prolog rule
relStar(X,Z) :- rel(X,Y), relStar(Y, Z).
Y is universally quentified, and is also unbounded.  I was asking whether
something like this cannot happen in a K definition.

2014-06-28 16:23 GMT+03:00 Grigore Rosu notifications@github.com:

grosu:
1) In reachability logic (RL), formulae of the form Phi => forall X . Phi' are not that meaningful (in terms of logic programming (LP), the entire RL rule correspons to the conclusion of the LR clause).  What are the transitions that these specify?  There is likely no gamma' that satisfies forall X . Phi'.

2) We want to eventually support the entire RL in K.  When that happens, we will allow qunatifiers in LHS and RHS terms.  For example:

  rule Phi[Y,Z] /\ (forall X . X > Y implies X > Z) => Phi'[Y,Z,?W] /\ (forall X . X > ?W)

So we will have a general solution then.

Grigore

radumereuta:
@andreistefanescu  I removed the check for the java backend just like you suggested. Now I'm always throwing an error message.
@grosu I've also updated the message like you suggested.

dwightguth:
Looks like you need to fix a Java backend test which violates this rule.

radumereuta:
I'm not sure how. @andreiarusoaie @yilongli can you have a look?

andreistefanescu:
Adding ? in from of the variable name in the failing test should do it.

PR #687, opened 2014-06-27
Fixing set membership

traiansf:
...which seems to be the cause for current problems for the substitution.  I'm not satisfied with the solution, though.


andreistefanescu:
For now, this is ok. We should have a discussion at some point about Maps and Sets and the operations on them, but let's merge unified-builtins2 first.

grosu:
LGTM for now, so we can merge (and aim at merging the unified-builtins2 branch asap).  I agree with Andrei that we want to rediscuss the collections and the operations on them.  The discussion is probably more ample, as it seems that we want to have a generic way to deal with syntactic (==) vs. semantic (=) equality, both at the level of matching and in RHS of rules.

PR #692, opened 2014-07-01
Fixing some of the hidden warnings found in /include directory 

radumereuta:
(typing all variables, singleton variables become anonymous, . is .K)

@dwightguth please review.


dwightguth:
Looks good. Can you please investigate why the ktest test we have for this didn't catch these?

radumereuta:
It's because we can't test the outputs from kompile. These warning messages are shown only when you run:
`kompile -w all <file>`
Last time when we discussed about this, we didn't have this functinality.

PR #696, opened 2014-07-03
Fix var type inference2

radumereuta:
Taking a second approach where I'm trying to infer the most general sort that fits in a particular context.

Let's see if tests pass first.


radumereuta:
Jenkins: test this please

radumereuta:
Closing so it doesn't test all the intermediate results. I will restore the pull request shortly.

PR #698, opened 2014-07-04
Fix var type inference2

radumereuta:
Giving it another try.


radumereuta:
To move ahead faster, I will call this ready to be reviewed.
@dwightguth please review the java code.
@grosu I had to change some of the definitions since I added new checks. One of them is looking at function rewrite and checks to see if I have a decreasing sort in the right hand side. That's why I had to modify some of the simple and kool examples.

Also, another problem that @traiansf helped me with, is the #if ...#fi overloaded operators. I would like @andreistefanescu to have a look at that.

At the moment the CIL semantics fails because of the checks that I've added, but we can still code review this until that is being fixed. I've asked @chathhorn to have a look. I tried to fix it myself, but I had to change too much, and I didn't feel comfortable with my results.

samples/java_rewrite_engine/kernel-cil/kernel-cil.k
```diff
@@ -1033,8 +1033,8 @@ corresponding store lookup operation. */
 // Array element: evaluate the array and its index;
 // then the array lookup rule above applies.
 
-  context lvalue(_:Exp[HOLE::Exp])
-  context lvalue(HOLE::Exp[_:Exp])
+  context lvalue(_:Exp[HOLE:>Exp])
```
grosu: don't you want `_:>Exp` instead of `_:Exp` to save a runtime check?

samples/java_rewrite_engine/kernel-cil/kernel-cil.k
```diff
@@ -1033,8 +1033,8 @@ corresponding store lookup operation. */
...
+  context lvalue(HOLE:>Exp[_:Exp])
```
grosu: don't you want `_:>Exp` instead of `_:Exp` to save a runtime check?

samples/java_rewrite_engine/simple/untyped_with_floats/simple-untyped.k
```diff
@@ -1076,8 +1076,8 @@ corresponding store lookup operation. */
 // Array element: evaluate the array and its index;
 // then the array lookup rule above applies.
 
-  context lvalue(_:Exp[HOLE::Exp])
-  context lvalue(HOLE::Exp[_:Exp])
+  context lvalue(_:Exp[HOLE:>Exp])
```
grosu: don't you want `_:>Exp` instead of `_:Exp` to save a runtime check?

samples/java_rewrite_engine/simple/untyped_with_floats/simple-untyped.k
```diff
@@ -1076,8 +1076,8 @@ corresponding store lookup operation. */
...
+  context lvalue(HOLE:>Exp[_:Exp])
```
grosu: don't you want `_:>Exp` instead of `_:Exp` to save a runtime check?

samples/java_rewrite_engine/tutorial/2_languages/1_simple/2_typed/1_static/simple-typed-static.k
```diff
@@ -583,7 +583,7 @@ exceptions which are not caught. */
 /*@ The function \texttt{mkDecls} turns a list of parameters into a
 list of variable declarations. */
 
-  syntax Decl ::= mkDecls(Params)  [function]
+  syntax Stmts ::= mkDecls(Params)  [function]
```
grosu, line 586: Nice catch!  Kudos to the new parser.

samples/java_rewrite_engine/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -466,7 +466,7 @@ interestingly, the semantics of return stays unchanged. */
 
   rule while (E) S => if (E) {S while(E)S}  [structural]
 
-  rule <k> print(V:Val, Es => Es); ...</k> <out>... .List => ListItem(V) </out>
+  rule <k> print(V:Val, Es:Vals => Es); ...</k> <out>... .List => ListItem(V) </out>
```
grosu: Why did you have to change this rule?  What you want here is `Es:Exps`, and the sort inferencer should figure it out.  Doesn't it work yet?  If not, then at least you should put some comment that this is temporary.

samples/java_rewrite_engine/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -977,7 +977,7 @@ member's location based on the current class of the object. */
   rule <k> lvalue(X:Id => this . X) ...</k>  <env> Env </env>
     when notBool(X in keys(Env))  [structural]
 
-  context lvalue(HOLE . _)
+  context lvalue((HOLE . _)<:Exp)
```
grosu, line 981: Nice!

tests/regression/java-rewrite-engine/bag/test.k
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) 2014 K Team. All Rights Reserved.
 module TEST
 
-  configuration <T> <k> $PGM:K </k> <bag> <b multiplicity="*"> . </b> </bag> </T>
+  configuration <T> <k> $PGM:K </k> <bag> <b multiplicity="*"> .K </b> </bag> </T>
```
grosu, line 4: Why do you have to do this?  Haven't we agreed that for each term you should pick the largest sort that it can have in its context?

grosu:
Other than the above, it lgtm.

samples/java_rewrite_engine/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
traiansf: Actually inferring that as Exps makes the code non-deterministic, as the engine might choose to print the first argument before evaluating the others.  This actually broke a test with the Java rewrite engine (`new-in-field.kool`) which has a line like this:

```
print("(new C2()).c is:",(new C2()).c,"\n");
```

where C2's constructor prints two other lines.  The current semmantics is non-deterministic, in the sense that it allows printing `"new C2().c is"` before executing the constructor, which injects its own prints before returning and evaluating c to be printed.

grosu: Wow, this is quite interesting.  Nice!  OK then, let's require Es to be Vals, but please then use Vs instead.  Radu, can you please also add a note that we want to evaluate all the arguments of print before we print them, because of the side effects?  I will revise it later, but to make sure we don't forget.

Thanks,
Grigore

src/javasources/KTool/src/org/kframework/kil/Hole.java
```diff
@@ -13,16 +13,16 @@
 
     public static final Hole KITEM_HOLE = new Hole(KSorts.KITEM);
 
-    private Hole(Element element) {
```
dwightguth, line 16: I'm not going to fight about this, but only because when I have time to work on the new kast, location information will be completely redone.

src/javasources/KTool/src/org/kframework/kil/KApp.java
```diff
@@ -70,7 +70,7 @@ public KApp(String location, String filename, Term label, Term child) {
      * @param child the KList which the given KLabel is applied to. A non-null instance of {@link KList}, {@link Variable} of sort KList, or {@link Ambiguity}.
      */
     public KApp(Term label, Term child) {
-        super(KSorts.KITEM);
+        super(label.getLocation(), label.getFilename(), KSorts.KITEM);
```
dwightguth: I'm not going to fight about this, but only because when I have time to work on the new kast, location information will be completely redone.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/CellTypesFilter.java
```diff
@@ -39,7 +44,10 @@ public ASTNode visit(Syntax cell, Void _) {
 
     @Override
     public ASTNode visit(Cell cell, Void _) throws ParseFailedException {
-        String sort = context.cellKinds.get(cell.getLabel());
+        String sort = context.cellSorts.get(cell.getLabel());
+        // if the k cell is opened, then the sort should be K because of ... desugaring
+        if (cell.getLabel().equals("k") && cell.getEllipses() != Ellipses.NONE)
```
dwightguth: this is not a general solution. What if a different cell has a ksequence in it? what if the K cell has Map in it? You need to come up with a solution that does not special case one specific cell.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/CellTypesFilter.java
```diff
@@ -76,14 +84,16 @@ else if (cell.getLabel().equals(MetaK.Constants.pathCondition))
...
+            super(CellTypesFilter2.class.getName(), context);
             this.expectedSort = expectedSort;
             this.cellLabel = cellLabel;
         }
 
         @Override
         public ASTNode visit(Term trm, Void _) throws ParseFailedException {
-            if (!context.isSubsortedEq(expectedSort, trm.getSort())) {
+            // accept only terms that have sort <= to the cell type, or terms of sort K
+            if (!context.isSubsortedEq(expectedSort, trm.getSort()) &&
```
dwightguth, line 75: I think you need to check the kind here as well. What if the user puts K in a cell that contains a bag? What error will they get?

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/CellTypesFilter.java
```diff
@@ -100,6 +110,7 @@ public ASTNode visit(Bracket node, Void _) throws ParseFailedException {
         @Override
         public ASTNode visit(Rewrite node, Void _) throws ParseFailedException {
             Rewrite result = new Rewrite(node);
+            result.setSort(expectedSort);
```
dwightguth, line 96: this should not be necessary. ReplaceChildren sets the sort

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/CollectExpectedVariablesVisitor.java
```diff
@@ -19,18 +19,18 @@ public CollectExpectedVariablesVisitor(Context context) {
...
-        Set<VarHashMap> newVars = new HashSet<VarHashMap>();
+        Set<VarHashMap> newVars = new HashSet<>();
         for (Term t : node.getContents()) {
             CollectExpectedVariablesVisitor viz = new CollectExpectedVariablesVisitor(context);
             viz.visitNode(t);
             // create the split
             for (VarHashMap elem : vars) { // for every local type restrictions
                 for (VarHashMap elem2 : viz.vars) { // create a combination with every ambiguity detected
-                    newVars.add(combine(elem, elem2));
+                    newVars.add(elem.deepClone().addAll(elem2));
```
dwightguth: can you add a comment here explaining why you believe cloning is necessary here?

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/CorrectCastPriorityFilter.java
```diff
@@ -36,9 +37,14 @@ public ASTNode visit(Cast cst, Void _) throws ParseFailedException {
         // if I find a Cast near a variable, then I remove the cast and translate the sort restrictions to the variable
```
dwightguth, line 37: is this comment now out of date?

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferDotsFilter.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    @Override
+    public ASTNode visit(Syntax cell, Void _) {
+        return cell;
+    }
+
+    @Override
+    public ASTNode visit(Rule rl, Void _) throws ParseFailedException {
+        if (rl.getBody() instanceof Ambiguity) {
```
dwightguth, line 51: what if the rewrite isn't at the top? Is it correct to do nothing in that case?

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferDotsFilter.java
```diff
@@ -0,0 +1,172 @@
...
+
+    private void processRW(Rewrite node) throws ParseFailedException {
+        // if the left hand side is a function, check to see if the right hand side has the same sort
+        if (node.getLeft() instanceof TermCons && ((TermCons) node.getLeft()).getProduction().containsAttribute("function")) {
+            node.setRight(preferWeak(node.getLeft().getSort(), node.getRight()), context);
+            node.setSort(node.getLeft().getSort());
+        }
+    }
+
+    private Term preferWeak(String expectedSort, Term t) throws ParseFailedException {
```
dwightguth: javadoc

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferDotsFilter.java
```diff
@@ -0,0 +1,172 @@
...
+            return node;
+        }
+        return t;
+    }
+
+    @Override
+    public ASTNode visit(Cell cell, Void _) throws ParseFailedException {
+        String sort = context.cellSorts.get(cell.getLabel());
+        // if the k cell is opened, then the sort should be K because of ... desugaring
+        if (cell.getLabel().equals("k") && cell.getEllipses() != Ellipses.NONE)
```
dwightguth: see previous comment about this case

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferDotsFilter.java
```diff
@@ -0,0 +1,172 @@
...
+            sort = KSorts.K;
+
+        if (sort == null) {
+            if (cell.getLabel().equals("k"))
```
dwightguth: okay, this is scary. please explain why you're doing this instead of trying to ensure the correct sort information is inferred in the context for these cells.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferDotsFilter.java
```diff
@@ -0,0 +1,172 @@
...
+            String msg = "Cell '" + cell.getLabel() + "' was not declared in a configuration.";
+            throw new ParseFailedException(new KException(ExceptionType.ERROR, KExceptionGroup.COMPILER, msg, getName(), cell.getFilename(), cell.getLocation()));
+        }
+        return super.visit(cell, _);
+    }
+
+    @Override
+    public ASTNode visit(TermCons tc, Void _) throws ParseFailedException {
+        // choose only the allowed subsorts for a TermCons
+        if (!tc.getProduction().getItems().isEmpty() && tc.getProduction().getItems().get(0) instanceof UserList) {
```
dwightguth: this is the third time I've had to tell you to use the helper functions we declared for this. Please don't let there be a fourth time.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferDotsFilter.java
```diff
@@ -0,0 +1,172 @@
...
+            UserList ulist = (UserList) tc.getProduction().getItems().get(0);
+            tc.getContents().set(0, (Term) preferStrict(ulist.getSort(), tc.getContents().get(0)));
+            tc.getContents().set(1, (Term) preferStrict(tc.getProduction().getSort(), tc.getContents().get(1)));
+        } else {
+            int j = 0;
+            Production prd = tc.getProduction();
+            for (int i = 0; i < prd.getItems().size(); i++) {
```
dwightguth:
```
for (int i = 0; i < tc.getContents.size(); i++) {
    tc.getContents().set(i, (Term) new TypeSystemFilter2(tc.getProduction().getChildSort(i), context).visitNode(tc.getContents().get(i));
}
```

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/TypeSystemFilter2.java
```diff
@@ -14,31 +14,47 @@
...
-                        ExceptionType.ERROR,
-                        KExceptionGroup.CRITICAL,
-                        "type error: unexpected term '" + trm + "' of sort '" + trm.getSort()
-                                + "', expected sort '" + maxSort + "'.",
-                        trm.getFilename(),
-                        trm.getLocation());
-                throw new ParseFailedException(kex);
-            }
+        boolean error = false;
+        if (strict && !trm.getSort().equals(maxSort))
```
dwightguth, line 41: something's confusing me here. If strict is true, this filter rejects well-typed terms? Why?

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/TypeSystemFilter2.java
```diff
@@ -78,6 +94,14 @@ public ASTNode visit(Rewrite node, Void _) throws ParseFailedException {
                 (Term) this.visitNode(node.getLeft()),
                 (Term) this.visitNode(node.getRight()),
                 context);
+        result.setSort(maxSort);
```
dwightguth, line 97: this feels skeevy to me. Eventually we will want to compute all sort information immutably instead of setting it manually everywhere. Can you explain why Rewrite.computeSort isn't good enough for you, and why, if it does not behave correctly, it can't simply be changed?

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/TypeSystemFilter2.java
```diff
@@ -78,6 +94,14 @@ public ASTNode visit(Rewrite node, Void _) throws ParseFailedException {
...
+        // if the left hand side is a function, check to see if the right hand side has the same sort
+        if (node.getLeft() instanceof TermCons && ((TermCons) node.getLeft()).getProduction().containsAttribute("function")) {
+            TypeSystemFilter2 tsf = new TypeSystemFilter2(node.getLeft().getSort(), context);
+            node.setRight((Term) tsf.visitNode(node.getRight()), context);
+            result.setSort(node.getLeft().getSort());
```
dwightguth, line 102: same here

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/VarHashMap.java
```diff
@@ -2,6 +2,7 @@
 package org.kframework.parser.concrete.disambiguate;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
```
dwightguth: this looks like you're trying to implement a multimap to me. Can't you just use the multimap class in one of the collections libraries we have (either guava or apache collections 4.0)?

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/CellTypesFilter.java
radumereuta, line 75: That's caught by the next line. You will get the expected error message.
Anyway, I'm inverting the boolean expression. I think it's more readable that way.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/CellTypesFilter.java
radumereuta, line 96: I have another pull request for computing the sort of rewrite.
We're no longer looking for LUB. We want the maximum sort for overloaded operators. For rewrite this would be the context in which it is found.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/CollectExpectedVariablesVisitor.java
radumereuta: You will have to look at the comments in the entire file.
Basically, I'm trying to move all the ambiguities at the top, and get the list of all possible constraints for variables.
I'm not actually removing sharing, I'm collecting only the constraints for each variable.
That clone is done each time I need to create a new set of constraints.
Please tell me if I should be more descriptive.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/CorrectCastPriorityFilter.java
radumereuta, line 37: No, it's just that I had to take one more case into consideration. Brackets.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferDotsFilter.java
radumereuta, line 51: rule myFunction(X => .)
When moving the rewrite at the top, then the type decreasing property still holds.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferDotsFilter.java
radumereuta: These are standard cells, and most of them are generated during kompilation.That's why, if I don't find a declared sort ( if sort == null) I'm skipping to the default.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferDotsFilter.java
radumereuta: Sorry, this was code ported from some older files.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferDotsFilter.java
radumereuta: Cool... didn't know about getChildSort.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/TypeSystemFilter2.java
radumereuta, line 41: This is only for Cast.
After talking with Grigore, we kind of agreed that cast should specify the exact sort of a production.
Otherwise something like `.::K` would still be ambiguous.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/TypeSystemFilter2.java
radumereuta, line 97: Well... the sort of the rewrite is a bit tricky.
As I said before, we don't want to choose minimum sort for operators, but instead maximum allowed by the context.
This is a simple way to look at overloaded operators, and if you want something else, then you have casting.
What I did here is a bit of a hack. I'm trying to bypass the `computeSort` function. Probably I will have to revisit the entire thing, although I'm not entirely convinced how to proceed.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/TypeSystemFilter2.java
radumereuta, line 102: See above.

radumereuta:
@dwightguth please re-review.
I've addressed some of your comments, and posted some questions.
I still have to fix CIL and fix the merge conflicts, but I will do that after I get the review done so It's easier to read.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/CellTypesFilter.java
dwightguth, line 96: Okay if you're making changes to this pull request based on planned changes in another pull request, I either need to see that pull request or you need to explain to me in detail the plan you have made for this part of the code base that has you making these steps.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferDotsFilter.java
dwightguth, line 51: I have no idea what you just explained or how it supposedly answered this question. This transformer does nothing to rewrites that aren't at the top of the rule. If that's correct, I need you to be more detailed in your explanation of why, so I can understand.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferDotsFilter.java
dwightguth: yes okay so it turns out I didn't actually want you to explain your choice, I wanted you to handle this the right way. Sorry for the confusion. context.cellSorts should be the authoritative source for all info about cell sorts. If a cell isn't in it, that's a bug.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/TypeSystemFilter2.java
dwightguth, line 41: ok

dwightguth:
You did not address all of my comments. Please go back through them again and make sure you have addressed everything.

radumereuta:
JENKINS: retest this please.

radumereuta:
Jenkins: test this please

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/PreferDotsFilter.java
radumereuta, line 51: Probably @andreistefanescu would be able to explain better how this goes, but here is how I view it.
In regard to defining functions, there are a few rules:
 a) the function must be at the top
 b) the rewrite must have a decreasing sort (right hand side of rewrite must have sort <= left hand side sort)
From what Andrei told me, if the rewrite is not at the top, then the compilation process will bring it at the top at one point, so I will end up rewriting myFunction to myFunction. So, yes, this would be safe.
Does this explanation seem reasonable?

tests/regression/java-rewrite-engine/bag/test.k
radumereuta, line 4: This is a configuration. The sort of the `<b>` cell is not yet defined, so I will have an ambiguity between sort K and sort Bag. The maximal sort is ambiguous in this case since K and Bag are disjoint.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/CellTypesFilter.java
radumereuta, line 96: @dwightguth please re-review.
Ok... I think I've addressed all of your comments, except for this one.
Previously, we wanted to set the sort of the rewrite as LUB of left and right, but we decided to change the way we disambiguate overloaded operators, by selecting the biggest sort possible in that context. Also, if the rewrite involves a function, then the sort is given by the left hand side (of course, if possible in the context).

I think I've done that. The funny thing is, all the examples still pass, so it seems it doesn't matter what the sort of the rewrite is, as long as it passes the type checking.
My proposal is to delay solving this inconsistency until we run into problems, and we have a better understanding of what we want.

src/javasources/KTool/src/org/kframework/parser/concrete/disambiguate/CellTypesFilter.java
dwightguth: You still have not addressed this comment. I don't have time right now to go through this whole diff checking everything I've commented to  make sure you really addressed it. The fact that I've found at least one comment you have not addressed is demonstration that this is still not ready for me to re-review.

_Please_, next time go through _all_ my comments and actually check that they have been addressed. You are wasting my time and slowing both of us down by asking me to review code that is not ready for re-review.

radumereuta:
@dwightguth I've reworked the getCellSort function. Please review.

dwightguth:
Jenkins: test this please

dwightguth:
lgtm. Merge if tests pass.

dwightguth:
Jenkins: test this please

radumereuta:
@dwightguth what is going on here? Why isn't CIL passing? I thought you fixed it when I asked you today. There is something seriously wrong, because the reported error does not correspond to any of the versions of CIL currently found on github (master or java_rewrite_engine).

Please have a look into it. I want to be done already with this pull request.

dwightguth:
There is still another bug with the code Andrei wrote. I am going to fix the C semantics repo to a version that contains your diff but not his. Then we can merge this without being blocked on Andrei's bug.

dwightguth:
Jenkins: test this please

PR #703, opened 2014-07-06
Prototype k rewrite machine

yilongli:
I call it K rewrite machine but it's really just a simple interpreter so far. Eventually, this would become a a fast abstract rewrite machine and the starting point of native code generation.


yilongli:

For newly added classes, it's easier to only review the final version. For other changes, it's better to review each commit in order.

This is quite a big change. Thanks a lot!

src/javasources/KTool/src/org/kframework/backend/java/indexing/IndexingPair.java
```diff
@@ -63,6 +66,56 @@ else if (kSequence.size() == 1) {
...
+     * Retrieves {@code IndexingPair} from an input stream pattern based on the
+     * first and last elements.
+     * 
+     * @param instream
+     *            the input stream pattern
+     * @param definition
+     *            the definition
+     * @return the indexing pair
+     */
+    public static IndexingPair getInstreamIndexingPair(BuiltinList instream, Definition definition) {
```
dwightguth, line 80: This feels really weird what you are doing with the input and output stream cells... Can you explain to me why you can't solve the problem more generically without introducing two new types of indexing information?

src/javasources/KTool/src/org/kframework/backend/java/kil/Cell.java
```diff
@@ -56,6 +56,22 @@ public Kind contentKind() {
...
+     * violates the immutability contract on {@code Term} and its subclasses.
+     * This method is only used in the {@link KAbstractRewriteMachine} to allow
+     * efficient in-place rewriting; as a result, it is the caller's
+     * responsibility to make defensive copy before calling this operation to
+     * avoid modifying shared {@code Cell}s. Besides, this operation should be
+     * used together with some mechanism to properly invalidates the cached
+     * hashCode of all the affected {@code Term}s.
+     * 
+     * @param content the new content
+     */
```
dwightguth, line 73: Can you add a TODO here with my name on it telling me to fix this with dependency injection? I believe the correct long-term solution here is to bind an abstract Cell class to either a mutable or immutable variant of the class.

src/javasources/KTool/src/org/kframework/backend/java/kil/KItem.java
```diff
@@ -335,7 +332,14 @@ public Term evaluateFunction(SymbolicConstraint constraint, TermContext context)
                     /* rename rule variables in the rule RHS */
                     rightHandSide = rightHandSide.substituteWithBinders(freshSubstitution, context);
                 }
-                rightHandSide = rightHandSide.substituteAndEvaluate(solution, context);
+                if (K.do_fast_exec) {
```
dwightguth: so, we're hoping to eventually break this piece off into a new and separate component. What you do here interferes with that. In order for that to happen, it's necessary that the code for the efficient rewriter be able to live in a separate place (i.e. no ifs on a global state variable inside code shared between the two engines). Can you restructure this so it has only one code path which calls out to some function implemented separately for the two engines?

src/javasources/KTool/src/org/kframework/backend/java/kil/Rule.java
```diff
@@ -42,6 +52,13 @@
     private final boolean containsKCell;
     private final boolean hasUnboundVars;
     
+    private boolean compiledForFastRewriting;
```
dwightguth, line 60: these need javadocing

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/ExtendedSubstitution.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.List;
+import java.util.Map;
+
+import org.kframework.backend.java.kil.Cell;
+import org.kframework.backend.java.kil.Term;
+import org.kframework.backend.java.kil.Variable;
+
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+
```
dwightguth, line 13: javadoc

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/Instruction.java
```diff
@@ -0,0 +1,87 @@
...
+
+import java.io.ObjectStreamException;
+import java.io.Serializable;
+import java.util.Map;
+
+import org.kframework.backend.java.util.Utils;
+
+import com.google.common.collect.Maps;
+
+public final class Instruction implements Serializable {
```
dwightguth, line 18: javadoc

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/Instruction.java
```diff
@@ -0,0 +1,87 @@
...
+    
+    private static final long serialVersionUID = -4395918031714809591L;
```
dwightguth: do you plan to increment this version id every time you want to break compatibility of this class's serialized form? If you don't, you'd better not specify this.

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
```diff
@@ -0,0 +1,283 @@
...
+import org.kframework.backend.java.symbolic.DeepCloner;
+import org.kframework.backend.java.symbolic.PatternMatcher;
+import org.kframework.backend.java.symbolic.Transformer;
+import org.kframework.backend.java.util.Profiler;
+
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+
+/**
+ *
```
dwightguth, line 32: javadoc

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
```diff
@@ -0,0 +1,283 @@
...
+        this.instructions = rule.instructions();
+        this.context = context;
+    }
+    
+    public static boolean rewrite(Rule rule, Term subject, TermContext context) {
+        KAbstractRewriteMachine machine = new KAbstractRewriteMachine(rule, (Cell<?>) subject, context);
+        return machine.rewrite();
+    }
+    
+    @SuppressWarnings({ "rawtypes", "unchecked" })
```
dwightguth: I don't like it. Better to fix the cause of warnings wherever possible instead of just suppressing them. If you can justify why that's not possible, though, I'll listen.

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/ExtendedSubstitution.java
```diff
@@ -0,0 +1,44 @@
...
+class ExtendedSubstitution {
+
+    private Map<Variable, Term> subst;
+    private List<Cell<?>> writeCells;
```
dwightguth, line 34: what's a write cell?

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
```diff
@@ -0,0 +1,283 @@
...
+                
+                /* perform local rewrites under write cells */
+                for (Cell cell : solution.writeCells()) {
+                    String cellLabel = cell.getLabel();
+                    Term rightHandSide = getWriteCellRHS(cellLabel);
+                    if (rule.cellsToCopy().contains(cellLabel)) {
+                        rightHandSide = DeepCloner.clone(rightHandSide);
+                    }
+
+                    cell.unsafeSetContent((Term) rightHandSide.accept(substAndEvalTransformer));
```
dwightguth, line 102: do you need to reset the computed hash code when you do this?

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
```diff
@@ -0,0 +1,283 @@
...
+        
+        while (true) {
+            nextInstr = nextInstruction();
+           
+            if (nextInstr == Instruction.UP) {
+                return;
+            }
+
+            if (nextInstr == Instruction.CHOICE) {
+                assert !isStarNested : "nested cells with multiplicity='*' not supported";
```
dwightguth, line 153: should this be a proper error message?

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
```diff
@@ -0,0 +1,283 @@
...
+    private Term getWriteCellRHS(String cellLabel) {
+        return rule.rhsOfWriteCell().get(cellLabel);
+    }
+    
+    private Collection<Cell> getSubCellsByLabel(Cell<?> cell, String label) {
+        return ((CellCollection) cell.getContent()).cellMap().get(label);
+    }
+
+    /**
+     * Returns the CNF of the formula which consists of extanded substitutions.
```
dwightguth: yeah this doesn't tell me what this function does in terms I can understand

src/javasources/KTool/src/org/kframework/compile/transformers/AddLocalRewritesUnderCells.java
```diff
@@ -0,0 +1,205 @@
...
+import org.kframework.kil.Term;
+import org.kframework.kil.loader.Context;
+import org.kframework.kil.visitors.BasicVisitor;
+import org.kframework.kil.visitors.CopyOnWriteTransformer;
+
+import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
+
+/**
+ * Adds local rewrites under the cells of interest computed in
```
dwightguth, line 25: this doesn't tell me what adding local rewrites under a cell is, it just copies the wording I already inferred from the name of the class.

src/javasources/KTool/src/org/kframework/compile/transformers/GenerateRewritingSchedule.java
```diff
@@ -0,0 +1,156 @@
...
+import org.kframework.kil.Module;
+import org.kframework.kil.ModuleItem;
+import org.kframework.kil.Rewrite;
+import org.kframework.kil.Rule;
+import org.kframework.kil.Syntax;
+import org.kframework.kil.loader.Context;
+import org.kframework.kil.visitors.CopyOnWriteTransformer;
+
+/**
+ *
```
dwightguth: javadoc

src/javasources/KTool/src/org/kframework/kil/Rule.java
```diff
@@ -20,6 +26,14 @@
 
     private List<BuiltinLookup> lookups = Collections.emptyList();
     private Map<Variable, Integer> concreteDataStructureSize = Collections.emptyMap();
+    
+    private boolean compiledForFastRewriting = false;
```
dwightguth, line 30: this feels wrong to me. what happens if a rule is created at some point and this stuff isn't set? If a rule is transformed, will this info get out of date? why not compute this information in the kompile backend, after we've settled these concerns?

src/javasources/KTool/src/org/kframework/backend/java/indexing/IndexingPair.java
yilongli, line 80: In the Javadoc I explained that I had to retrieve indices from input/output cells different than k cell. It's just a workaround for now. I will implement the generic pattern match automata anyway in the future. And all these specific methods for handling different cells should all go away.

src/javasources/KTool/src/org/kframework/backend/java/kil/Cell.java
yilongli, line 73: OK. 

I am not sure about the correct long-term solution, though the current implementation is definitely not very good. I need to come back at some point to work out an appropriate class hierarchy for this.

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/Instruction.java
yilongli, line 18: Done.

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/ExtendedSubstitution.java
yilongli, line 13: Done.

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/Instruction.java
yilongli: Done.

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
yilongli: The problem is the following statements.

```
for (Cell cell : solution.writeCells()) { // writeCells() returns List<Cell<?>>
    ...
    cell.unsafeSetContent((Term) rightHandSide.accept(substAndEvalTransformer));
}
```

I couldn't find a way to make the compiler happy. I would be more than happy to revise this code if you know a solution. I will just remove the annotations for now.

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
yilongli, line 102: No. Because I guarantee that I would never put a `Term` into hash-based container during the whole rewriting process. Once I have different hierarchies for mutable and immutable terms, this would look safer since the mutable terms will not cache their hashcodes.

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
yilongli, line 32: Done.

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/ExtendedSubstitution.java
yilongli, line 34: Javadoc added.

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
yilongli: javadoc referring to the original version of the functionality added

src/javasources/KTool/src/org/kframework/compile/transformers/GenerateRewritingSchedule.java
yilongli: done

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
yilongli, line 153: This is a limitation of the current implementation which is shared by PatternMatcher and SymbolicUnifier as well. They are all implemented as assertions. I prefer to keep it like this for now because it's easier for me to debug. However, we should definitely change all assertions in the Java backend to emit proper error message before we decide to make it the default backend.

src/javasources/KTool/src/org/kframework/backend/java/indexing/IndexingPair.java
dwightguth, line 80: alright, that's fair I guess

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
dwightguth, line 102: ok

src/javasources/KTool/src/org/kframework/backend/java/kil/Rule.java
yilongli, line 60: done

src/javasources/KTool/src/org/kframework/compile/transformers/AddLocalRewritesUnderCells.java
yilongli, line 25: done

src/javasources/KTool/src/org/kframework/kil/Rule.java
yilongli, line 30: This `compiledForFastRewriting` must be computed after all auto-generated rules are included; this is explained in the Javadoc of `ComputeCellsOfInterest`.

If a rule is transformed, this `compiledForFastRewriting` might change from true to false. But it is the responsibility of the kompilation pass to set it to false.

This information and many others are indeed computed and set by the kompile backend.

In short, there are two key time points for me to compute and set these fields. One is after all rules are properly created but the rewrite has not been pushed to the top. The other is when all transformations on rules are done.

yilongli:
@dwightguth please review again

dwightguth:
okay, if you want to add some more comments, you can do that, otherwise you can merge this.

PR #711, opened 2014-07-11
make the maude backend output more deterministic

dwightguth:
@traiansf please review

The Maude backend had some issues that were making the output to maude JVM dependent. I added some code whose purpose is to ensure that the ordering and numbering of rules is the same accross different Java versions.

### DIFF
--- a/include/builtins/set-impl.k
+++ b/include/builtins/set-impl.k
@@ -30,8 +30,8 @@ module SET-IMPL
 
   /* element membership */
 //  syntax Bool ::= K "in" Set    [function, hook(Set:in)]
-  rule K:K in SetItem(K) _:Set => true
-  rule K1:K in SetItem(K2:K) S:Set => K1 =K K2 orBool K1 in S
+  rule K1:K in SetItem(K2:K) _:Set => true when K1 =K K2
+  rule K1:K in SetItem(K2:K) S:Set => K1 in S when notBool(K1 =K K2)
   rule _ in .Set => false
 
   /* set cardinality */
--- a/include/builtins/set-impl.k
+++ b/include/builtins/set-impl.k
@@ -30,8 +30,8 @@ module SET-IMPL
 
   /* element membership */
 //  syntax Bool ::= K "in" Set    [function, hook(Set:in)]
-  rule K1:K in SetItem(K2:K) _:Set => true when K1 =K K2
-  rule K1:K in SetItem(K2:K) S:Set => K1 in S when notBool(K1 =K K2)
+  rule K1:K in SetItem(K2:K) _:Set => true requires K1 =K K2
+  rule K1:K in SetItem(K2:K) S:Set => K1 in S requires notBool(K1 =K K2)
   rule _ in .Set => false
 
   /* set cardinality */
--- a/tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.k
+++ b/tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.k
@@ -589,7 +589,7 @@ checks for cycles. */
             ...</classes> _ => .)
            <out>... .List => ListItem("Class \"" +String Id2String(C)
                                   +String "\" is in a cycle!\n") </out>
-       </T>  [transition]
+       </T>  [inheritance-cycle]
 
 
 /*@ \subsubsection{New}
--- a/tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.k
+++ b/tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.k
@@ -589,7 +589,7 @@ checks for cycles. */
             ...</classes> _ => .)
            <out>... .List => ListItem("Class \"" +String Id2String(C)
                                   +String "\" is in a cycle!\n") </out>
-       </T>  [structural]
+       </T>  [transition]
 
 
 /*@ \subsubsection{New}
--- a/tutorial/2_languages/2_kool/2_typed/2_static/tests/cycle.kool.out
+++ b/tutorial/2_languages/2_kool/2_typed/2_static/tests/cycle.kool.out
@@ -1,2 +1,15 @@
-Class \\"C2\\" is in a cycle!
-Class \\"C1\\" is in a cycle!
+Search results:
+
+Solution 1:
+<T>
+    <out>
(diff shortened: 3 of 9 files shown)


dwightguth:
@traiansf please review

tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.k
```diff
@@ -589,7 +589,7 @@ checks for cycles. */
             ...</classes> _ => .)
            <out>... .List => ListItem("Class \"" +String Id2String(C)
                                   +String "\" is in a cycle!\n") </out>
-       </T>  [structural]
+       </T>  [transition]
```
traiansf: Why transition?

dwightguth: Because if more than one class is in a cycle, this rule will match on any
of them, making it cause non confluence if it is an equation.

tutorial/2_languages/2_kool/2_typed/2_static/tests/cycle.kool.out
```diff
@@ -1 +1,2 @@
-Class "C2" is in a cycle!
+Class \\"C2\\" is in a cycle!
+Class \\"C1\\" is in a cycle!
```
traiansf: So basically that's why you made the rule above a transition, to detect all cycles?

traiansf: Maybe you should ask @grosu if he approves of this change to the code.  Although you detect all errors, it might be reasonable to stop at the first error.

traiansf:
Ok, I am convinced.  But Grigore should have a look, too.

tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.k
grosu: Dwight,

I think you should actually give the rule some name/tag, and them kompile the definition with the --transition option include that name/tag.

Traian, that is needed because the definition of the type checker is in fact non-deterministic :-)  If you have two classes in a cycle, then any of them can be picked first by that rule and reported as being in a cycle.  Making it a transition allows us to see and test  both behaviors.

Grigore

tutorial/2_languages/2_kool/2_typed/2_static/tests/cycle.kool.out
grosu: But the problem is that it is not clear which one is first.  We got the other one when we switched to Java 8, for whatever reason which I do not even remotely want to understand :)

Grigore

traiansf: Weren't we sorting the search results at some point?  I thought we did.

dwightguth:
@grosu please re-review

include/builtins/set-impl.k
```diff
@@ -30,8 +30,8 @@ module SET-IMPL
 
   /* element membership */
 //  syntax Bool ::= K "in" Set    [function, hook(Set:in)]
-  rule K:K in SetItem(K) _:Set => true
-  rule K1:K in SetItem(K2:K) S:Set => K1 =K K2 orBool K1 in S
+  rule K1:K in SetItem(K2:K) _:Set => true when K1 =K K2
```
grosu: Please use `requires` instead of `when`, because the later is deprecated.  Also, don't we want to write these two rules as one rule

```
rule K1:K in SetItem(K2:K) S:Set => #if (K1 =K K2) then true else K1 in S fi
```

I think this should be faster, if it works.

tutorial/2_languages/2_kool/2_typed/2_static/tests/cycle.kool.out
grosu: Same comment as Traian: don't we sort the solutions anymore?

grosu: Also, how does it know when one output of one execution ends and the other one starts?  I mean, the output of one execution may spread over more than one line.  Is it that the `regex=true` option in `config.xml` assumes that in that case you would write `\n` for new lines in patterns?

I know I said it in the past, but I feel uneasy about this regex thing.  I thank you for solving the problem at the time, Dwight, but it just feels that this is not the long-term solution.  In this particular case, I would _not_ use `regex=true` and I would have the cycle.out file have two solutions in it, the old-fashioned way.

include/builtins/set-impl.k
traiansf: The problem might be that =K (unlike ==K) is not guaranteed to evaluate, so you might end up with a stuck term there.

grosu: I see, didn't think about that.  But don't you want that anyway when you do symbolic execution for reasoning, combined with case analysis?  This is not important, so we can forget it for now.

Grigore

tutorial/2_languages/2_kool/2_typed/2_static/tests/cycle.kool.out
dwightguth: okay. I will do it that way, then.

dwightguth:
@grosu review again

tutorial/2_languages/2_kool/2_typed/2_static/tests/cycle.kool.out
```diff
@@ -1 +1,15 @@
-Class "C2" is in a cycle!
+Search results:
+
+Solution 1:
+<T>
```
grosu: To make the .out file look nicer, I would search for a pattern `<T> <out> ListItem(#buffer(S:String)) </out> </T>`.
This would also make it more modular to changes later on; if we decide to drop ListItem or #buffer, then we would only need to change the config.xml file in one place.

dwightguth:
Blocked on #720

Grigore, are you okay with the temporary solution I just added being merged? I want to get us moving on being able to use Java 8... I can add a comment to the related issue asking for it to be changed once the bug is fixed though.

grosu:
OK.  I wanted to propose exactly that after I saw your previous message to Traian.

dwightguth:
Okay. So I can merge this if all the tests pass?

grosu:
yes.

Grigore

PR #713, opened 2014-07-11
fixing-issue#709

yilongli:
@dwightguth please review


andreistefanescu:
==List and =/=List are deprecated. Use ==K and =/=K instead.

yilongli:
Right, I forgot this. I guess I should also remove the declaration of ==List as well?

andreistefanescu:
Yes. Plese remove ==Map and == Set as well.

yilongli:
done please review again @andreistefanescu @dwightguth

andreistefanescu:
Looks good. @traiansf please take a look and see if this is correct for the Maude backend.

traiansf:
Looks ok, although I don't understand why samples/imp-hoare/imp.k had to be completely reformatted.

Issue #708, opened 2014-07-09
detecting error message for declare syntax K


liyili2:
When someone declare syntax K ::= something in their definitions, we should have an error message to suggest that this is wrong in java backend.

yilongli:
@radumereuta I am not sure what is the right place to put the check. Basically, all I need to do is to throw an error when the user tries to extend sort K.

radumereuta:
It should be here:
org.kframework.compile.checks.CheckSyntaxDecl

yilongli:
fixed via https://github.com/kframework/k/pull/714

PR #714, opened 2014-07-11
forbid extending sort K in the Java backend

yilongli:
Fixed https://github.com/kframework/k/issues/708

@dwightguth @radumereuta please review


yilongli:
test failed because of this production: https://github.com/yilongli/k/blob/fixing-issue%23708/tests/regression/java-rewrite-engine/user_substitution/lambda.k#L4

What I did here is to throw an error when user tries to extend sort K in their definition. So can we move this production somewhere inside `include`? @traiansf

traiansf:
Yes, we could.  Actually one thing to do would be to separate substitution.k into an interface module and an implementation module, have the interface module contain that line and the implementation the remainder of what is now in substitution.k (similarly to what happened with set-imp and the other collections.  Then this lambda could include only the interface and be happy.

This should change when I get to merge my unified-hooks code anyway.

yilongli:
@traiansf So should I go ahead and make the change or wait for your unified-hooks branch? I think you know best how to separate the module. Thanks.

src/javasources/KTool/src/org/kframework/compile/checks/CheckSyntaxDecl.java
```diff
@@ -63,10 +66,18 @@ else if (!(pi instanceof Terminal)) {
         if (node.isSubsort()) {
             String sort = ((Sort) node.getItems().get(0)).getName();
             if (Sort.isBasesort(sort) && !context.isSubsorted(node.getSort(), sort)) {
-                String msg = "Extending  built-in sorts is forbidden: K, KResult, KList, Map,\n\t MapItem, List, ListItem, Set, SetItem, Bag, BagItem, KLabel, CellLabel";
+                String msg = "Subsorted to built-in sorts is forbidden: K, KResult, KList, Map,\n\t MapItem, List, ListItem, Set, SetItem, Bag, BagItem, KLabel, CellLabel";
```
radumereuta: Shouldn't this be 'Subsorting' ?

yilongli: Yes...

src/javasources/KTool/src/org/kframework/compile/checks/CheckSyntaxDecl.java
```diff
@@ -63,10 +66,18 @@ else if (!(pi instanceof Terminal)) {
...
                 GlobalSettings.kem.register(new KException(KException.ExceptionType.ERROR, KException.KExceptionGroup.COMPILER, msg, getName(), node.getFilename(), node.getLocation()));
             }
         }
+        
+        if (context.kompileOptions.backend.java()
+                && !node.getFilename().startsWith(KPaths.getKBase(false) + File.separator + "include") 
+                && node.getSort().equals(KSorts.K)) {
+            String msg = "Sort K is not user-extendable in the Java backend:\n\t" + node + "\n\tConsider extending KItem instead.";
```
radumereuta: Maybe we should do the same thing in both java and maude backends.
I always assumed that K is the base sort for all syntax, but since the java backend was introduced, it seems KItem is the preferred sort.
Maybe we should look over all the definitions, and I could go through the SDF stuff and fix all these inconsistencies.

traiansf:
Ok, I'll try to do that.

src/javasources/KTool/src/org/kframework/compile/checks/CheckSyntaxDecl.java
yilongli: There are many places in the tutorial that extend K. I will add a TODO comment here for now. Maybe we can get back to it later?

yilongli:
@dwightguth please review

@traiansf I commented out the test for user-substitution in order to merge.

src/javasources/KTool/src/org/kframework/compile/checks/CheckSyntaxDecl.java
```diff
@@ -63,10 +66,19 @@ else if (!(pi instanceof Terminal)) {
         if (node.isSubsort()) {
             String sort = ((Sort) node.getItems().get(0)).getName();
```
radumereuta: You can replace this with getSubsort()

yilongli:
@andreistefanescu I need some help here. Right now, the build failed because there are three places extending sort K in https://github.com/kframework/c-semantics/blob/java_rewrite_engine/cil-semantics/cil-typing.k  I am not sure if they can be simply replaced by KItem.

yilongli:
BTW, another question is: should we allow function to return sort K? If so, I should only forbid extending sort K with constructor.

src/javasources/KTool/src/org/kframework/compile/checks/CheckSyntaxDecl.java
grosu: Oh, I could not even imagine that we do different things in the Java and Maude backends.  We should definitely have the same K language in ALL backends.  There should be _nothing_ special done in a particular backend.  The only differences between backends should be in the set of builtins that they implement/provide, and remember that also means the pm (pattern-match) function, which tells which patterns can be matched by the backend.  Thus, the only backend-specific error messages should be of the forms:

1) Backend XYZ does not support builtin operation FGH.
2) Backend XYZ does not support matching pattern PQR.

If we do anything else specific to a backend, then something is very wrong there and will be eventually changed.

Grigore

grosu: This is bad!  They probably sneaked because we decided to claim victory in unified-builtins2 if we pass everything with the Maude backend.  I volunteer to change all those places as soon as needed.

Grigore

andreistefanescu:
Functions may return K.

tests/regression/java-rewrite-engine/config.xml
```diff
@@ -278,9 +278,9 @@
       <krun-option name="--smt" value="none" />
     </all-programs>
   </test>
-
+<!--
```
dwightguth: what? no, why are we commenting out this test? If you intend to immediately fix it, fix it first and /then/ merge. Commenting it out now runs the risk of introducing regressions from another change during the time it's commented out.

traiansf: It's commented per our discussion yesterday. I'm working on fixing it.

2014-07-16 18:09 GMT+03:00 dwightguth notifications@github.com:

yilongli:
I have replaced all non-function productions of sort K with sort KItem. This affects a lot of files, so I categorized them into different commits.
@grosu please review all the changes in directories `tutorial` and `include`
@dlucanu please review changes related to your definitions
@dwightguth please review everything

Besides, there are lots of functions returning sort K. I believe some of them will only return sort KItem. However, I don't have the time to distinguish them one by one, so I just leave them unchanged.

include/builtins/array.k
```diff
@@ -1,47 +1,47 @@
...
-
-  // array size
-  rule size-of-array(store(A:Array, _:K, _:K)) => size-of-array(A) 
-  rule size-of-array(const-array(Size:Int, _:K)) => Size
-
-endmodule
+require "int.k"
+
+module ARRAY-SYNTAX
+  syntax Array ::= "store"       "(" Array "," Int "," K ")"  [smtlib(store)]
```
grosu, line 5: You know you can write

```
syntax Array ::= store(Array,Int,K)  [smtlib(store)]
```

right?

include/builtins/k-equal.k
```diff
@@ -53,7 +53,7 @@ module K-EQUAL-HOOKS
   // equality over KList
   syntax Bool ::= KList "==KList" KList    [function, hook(KEqual:_==KList_)]
 
-  syntax K ::= dummy(KList)
+  syntax KItem ::= dummy(KList)
```
grosu, line 35: Wouldn't it be easier to just use KAST here, which is simpler because it avoids any encodings, and write just

```
syntax KLabel ::= "dummy"
```

Maybe even `#dummy`.

include/builtins/array.k
yilongli, line 5: Yes, of course. But my change is to mechanically replace `syntax K ::=` with `syntax KItem ::=` for non-function productions. This improvement should be done in another pull request by whoever owns this file.

include/builtins/symbolic-calculus.k
```diff
@@ -28,7 +28,7 @@ module SYMBOLIC-CALCULUS
   This should be defined in a more generic way using ASTs.
 */
 
-  syntax K ::= K "[" K "/" K "]"
+  syntax KItem ::= K "[" K "/" K "]"
```
grosu: Not sure about this, we probably want substitution to be a function and return a K.  Or even have one for KItem and one for K.  Traian?

include/builtins/array.k
yilongli, line 5: OK, never mind. Please keep doing this. I will fix them in the next commit.

samples/prog-equiv/imp/peq.k
```diff
@@ -342,7 +342,7 @@ equality is required, only the maps and conditions are given as parameters.
   rule isInDom(I:K, (X replBy A:K ;; Rest:K)) => isInDom(I,  Rest)
        when I=/=K X
 
-  syntax K ::= valOf(K, K)
+  syntax KItem ::= valOf(K, K)
 
   rule valOf(I:K, I replBy A:K) => A
```
grosu, line 347: Not sure whether A can have the sort K here, it should probably have sort KItem.  I somehow feel that we should not even review this, we should just merge if the tests pass and then wait until the new parser will run into problems and fix them then.

include/builtins/array.k
grosu, line 5: Agreed.  As I said in my previous note, I think we should just merge this if all the tests pass and then have Radu report errors when he gets the new parser done.  These can actually _help_ Radu tune his parser, so it could be beneficial to keep them :)

Grigore

include/modules/unification.k
```diff
@@ -204,7 +204,7 @@ constraint.
 
 // alpha conversion moved here by grigore
  syntax K ::= renameVariables(K)  [function, hook(MetaK:renameVariables)]
-            | applyAlpha(K)
+//            | applyAlpha(K)
```
dwightguth, line 207: should this have been commented out?

yilongli, line 207: It's not used anywhere. And I don't want to make a decision regarding whether it should be changed to have sort KItem.

src/javasources/KTool/src/org/kframework/compile/checks/CheckSyntaxDecl.java
```diff
@@ -61,12 +62,18 @@ else if (!(pi instanceof Terminal)) {
...
+            String sort = node.getSubsort().getName();
             if (Sort.isBasesort(sort) && !context.isSubsorted(node.getSort(), sort)) {
-                String msg = "Extending  built-in sorts is forbidden: K, KResult, KList, Map,\n\t MapItem, List, ListItem, Set, SetItem, Bag, BagItem, KLabel, CellLabel";
+                String msg = "Subsorting built-in sorts is forbidden: K, KResult, KList, Map,\n\t MapItem, List, ListItem, Set, SetItem, Bag, BagItem, KLabel, CellLabel";
                 GlobalSettings.kem.register(new KException(KException.ExceptionType.ERROR, KException.KExceptionGroup.COMPILER, msg, getName(), node.getFilename(), node.getLocation()));
             }
+        } else if (!node.containsAttribute(Constants.FUNCTION)
+                && (node.getSort().equals(KSorts.K) || 
+                    node.getSort().equals(KSorts.KLIST))) {
+            String msg = "Extending sort K is forbidden:\n\t" + node + "\n\tConsider extending KItem instead.";
```
dwightguth: or KList?

dwightguth:
Overall looks good to me except for the minor things I noted. If you answer everyone else's concerns, you can merge as far as I'm concerned.

include/builtins/symbolic-calculus.k
traiansf: So if we make it `KItem` it needs to be 

```
syntax KItem ::= KItem "[" KItem "/" KItem "]"
```

because we don't want .K in the first position (`.K [ 3 / X ]` should evaluate to `.K`) or in the second position (`X[.K/X]` should evaluate to `.K`), and we don't ever want .K in the third position.

I would personally be in favor of that, and disallow applying substitution on .K

grosu: OK with me, at least until we run into limitations.

Grigore

yilongli:
I have addressed the comments from code review. Further polish of the definitions will be delayed until the new parser is ready. Merged.

PR #715, opened 2014-07-13
Flyweight pattern for sorts in java backend

yilongli:
ready for review


andreistefanescu:
Look good. I'm not sure about having one static cache, since sorts could potentially depend on definitions. Maybe a better place would be in context, but that would complicate things...

cos:
I agree with Andrei about the caching. We already have enough things that are "global". Let's not add another. If we have to add it, it should be attached to the definition somehow. Caching makes things messy, any way it is done. Is it worth it? How much do we gain by caching the sorts?

yilongli:
@andreistefanescu The cache is initially implemented as a map from definition to map. So that is not a problem. Right now, I change it to guava's Table, which looks nicer. But I will probably attach it to definition instead.

@cos I think it is nice and clean and It is definitely worth it.

yilongli:
After second thought, I still prefer to use Guava's Table for now because the code is only used by the constructor of KItem. And I believe the current solution is easiest to understand.

cos:
@yilongli, what do you mean it is nice and clean --  your implementation or caching in general? what speedups do you get from this implementation?

yilongli:
Significant speedup in constructing KItem, which is a heavily used operation during rewriting. Around 10~15% overall speedup in my medium-size example.

dwightguth:
I suggest you compare the speedup against one map per definition. Guava's
Table is likely not optimized for only one or two entries along one axis.

cos:
which is the expensive part of determining the sort? `context.getGLBSort(...)`? If so, caching could be moved there instead. This way it will also have a stronger effect as it will be shared between `KItem` instantiations.

yilongli:
@dwightguth I started with an incorrect implementation which assumes only one definition exist. The speedup has no difference. Although Guava's Table is not optimized for one or two entries along the row, it is not the bottleneck of the constructor of KItem at all. The hashCode() and equals() method of Definition are all inherited from Object.

@cos The expensive part of determining the sort is the part after cache miss. It's not about a single operation such as `context.getGLBSort(...)` or `context.getCommonSort`. Besides, such operations are pretty cheap.

cos:
I was guessing the expensive part is after the cache miss -- that's why we're talking about caching in the first place. But which operation in particular is expensive? Or, if several of them are, which ones? Is it the `for (Production production : productions) {` loop? 

It's nice it is faster. But why is it? Indiscriminately adding caching will hurt us in the long run. The `KItem` constructor has 130 lines now.

yilongli:
The part after cache miss is basically just the `for` loop. The fact that the `KItem` constructor has 130 lines is not because of the newly added code for caching.

cos:
Yes, two `for` loops (one just for compilation) and two `getGLBSort`. Shall I conclude from all this that the `for` loop is expensive and not the `getGLBSort`?

Actually, it does not matter that much. It seems to me that the sort computation is a function: (KLabelConstant, List[Sort]) -> Sort. Going back to the original argument, why can't we cache this in the `Definition` instead? Again, this will allow sharing between `KItem` instantiations.

Yes, the 130 lines are not from caching -- caching is just 10-20 lines overall. But I'm very skeptical about **adding** code -- the codebase is already severely bloated.

cos:
Furthermore, should the code for determining sorts be living in the `KItem` constructor in the first place?

yilongli:
Yes.

The sort computation that is suitable for caching is a function: (Definition, KLabelConstant, KList) -> Sort, though not all information of KList is used. Of course we can, but I think it complicates stuff. I have to make private inner class visible, etc.

What I am trying to say is that the right way to keep the constructor of KItem small is to extract some existing code to separate function, but not to refuse adding the code for caching.

yilongli:
I have extracted the sort computation to a helper method to keep the KItem constructor small. I still prefer to keep the cache as it is for now.

Now, it's ready for review. @andreistefanescu @dwightguth

cos:
> The sort computation that is suitable for caching is a function: (Definition, KLabelConstant, KList) -> Sort, though not all information of KList is used. 

What information from KList is actually used. Is it only the sorts? If so, moving to Definition makes even more sense.

> Of course we can, but I think it complicates stuff. I have to make private inner class visible, etc.

Is the private inner class anything more than a tuple. If it is, does it need to be?

> I have extracted the sort computation to a helper method to keep the KItem constructor small. I still prefer to keep the cache as it is for now.

I'm more convinced now that the right place for this functionality is in the `Definition`. There are several advantages:
- caching is shared among `KItem` instantiations so there is a potential for better speedup
- no need for the `Map<Definition, ...>` indirection
- no `static`, a big plus in my book -- especially if we ever parallelize the implementation

The only disadvantage seems to be the extra effort required to disentangle the sort computation from the rest of `KItem`'s guts.

src/javasources/KTool/src/org/kframework/backend/java/indexing/pathIndex/visitors/TermVisitor.java
```diff
@@ -140,7 +140,7 @@ public void visit(KSequence kSequence) {
         if (kSequence.size() > 0) {
             //TODO (OwolabiL): This is too messy. Restructure the conditionals
             if (kSequence.get(0) instanceof KItem) {
-                boolean isKResult = context.isSubsorted(K_RESULT, (kSequence.get(0)).sort());
```
dwightguth, line 144: why not definition.subsorts()?

dwightguth, line 144: actually you know what, never mind, this is a deprecated code path, it's okay if it doesn't work with parametric sorts later.

src/javasources/KTool/src/org/kframework/backend/java/kil/BuiltinMap.java
```diff
@@ -73,7 +73,7 @@ public boolean isExactSort() {
     }
 
     @Override
-    public String sort() {
+    public Sort sort() {
         // TODO(AndreiS): track the original sort from the grammar
         return KSorts.MAP;
```
dwightguth: I don't like that you have both Sort and KSorts containing Sort constants.

src/javasources/KTool/src/org/kframework/backend/java/kil/KItem.java
```diff
@@ -51,47 +51,29 @@
...
+            if (kLabelConstant.isListLabel()) {
+                return kLabelConstant.getListTerminator();
             }
         }
         
         return new KItem(kLabel, kList, termContext);
     }
     
-    private KItem(Term kLabel, Term kList, String sort, boolean isExactSort) {
+    KItem(Term kLabel, Term kList, Sort sort, boolean isExactSort) {
```
dwightguth, line 76: why no longer private?

src/javasources/KTool/src/org/kframework/backend/java/kil/KLabelConstant.java
```diff
@@ -25,7 +26,7 @@
 public class KLabelConstant extends KLabel implements MaximalSharing {
 
     /* KLabelConstant cache */
-    private static final HashMap<String, KLabelConstant> cache = new HashMap<String, KLabelConstant>();
+    private static final PatriciaTrie<KLabelConstant> cache = new PatriciaTrie<>();
```
dwightguth, line 29: I'd feel better about you adding Apache Collections 4.0 if you removed the references to the old org.apache.commons.collections15 from sourceforge.net.

src/javasources/KTool/src/org/kframework/backend/java/kil/Sort.java
```diff
@@ -0,0 +1,137 @@
...
+     */
+    Object readResolve() throws ObjectStreamException {
+        Sort sort = cache.get(name);
+        if (sort == null) {
+            /* do not use Sort#of to cache this sort; we need to
+             * preserve the original ordinal */
+            sort = this;
+            cache.put(name, sort);
+        } else {
+            assert this.ordinal == sort.ordinal : "ordinal of sort " + name
```
dwightguth, line 132: this method is called by the JVM. Can you please test to make sure that if this AssertionError is thrown, it actually bubbles up to the user?

src/javasources/KTool/src/org/kframework/backend/java/symbolic/FastDestructiveRewriter.java
```diff
@@ -58,7 +58,7 @@ public Term rewrite(Term subject, int bound) {
                 }
                 subject = result;
             } else {
-                computeRewriteStep(subject, 1);
+//                computeRewriteStep(subject, 1);
```
dwightguth, line 61: is this supposed to be here?

src/javasources/KTool/src/org/kframework/backend/java/util/KSorts.java
```diff
@@ -1,18 +1,20 @@
 // Copyright (c) 2013-2014 K Team. All Rights Reserved.
 package org.kframework.backend.java.util;
 
+import org.kframework.backend.java.kil.Sort;
+
 /**
  * @author AndreiS
  */
 public class KSorts {
```
dwightguth: seems to me like we should either deprecate or remove this class.

tests/regression/java-rewrite-engine/multiple_subsorts/program.test.out
```diff
@@ -1,3 +1 @@
-<k>
-    isKResult(_45:Sort) ~> _46:KResult ==K _45
-</k>
+isKResult\((.*):Sort\) ~\> (.*):KResult ==K \1
```
dwightguth: take a look at what I did with LOGIK. currently this matches `isKResult(_45:Sort) ~> _45:KResult ==K _45`. Is that really what you want? If not, you can use lookahead assertions to forbid matching both variables with the same name.

src/javasources/KTool/src/org/kframework/backend/java/kil/KItem.java
yilongli, line 76: Because I refactored the method that build list terminator to `KLabelConstant#buildListTerminator` and that method needs a direct way to construct the list terminator (which is a KItem). We have the KLabelConstant, we have the KList, and we know the sort for sure. There is no need to re-compute it. Besides, if I call `KItem.of` in `KLabelConstant#buildListTerminator`, it results in an infinite loop.

src/javasources/KTool/src/org/kframework/backend/java/kil/KLabelConstant.java
yilongli, line 29: I was about to do that.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/FastDestructiveRewriter.java
yilongli, line 61: not really, it is used to test why the rewriting gets stuck at some final configuration. I forgot to remove it previously, and none of us noticed that.

src/javasources/KTool/src/org/kframework/backend/java/util/KSorts.java
yilongli: yes I will remove it

src/javasources/KTool/src/org/kframework/backend/java/kil/Sort.java
yilongli, line 132: Yes, it is.

tests/regression/java-rewrite-engine/multiple_subsorts/program.test.out
yilongli: done

yilongli:
@dwightguth I couldn't completely eliminate the old apache collections15 because the JUNG library contains certain classes from collections15 in their API's! For example, http://jung.sourceforge.net/doc/api/edu/uci/ics/jung/visualization/util/PredicatedParallelEdgeIndexFunction.html#setPredicate%28org.apache.commons.collections15.Predicate%29.

And some of our code under krun/gui use these API's.

yilongli:
@cos Regarding your first two questions, could you take a look at my code? It's just a few lines. I don't know how to explain better. As for the advantages, the speedup is negligible. The indirection `Table<Definition, ...>` is not too bad. Besides, I don't see any problem with parallelization regarding  the current implementation. The key and value classes for the cache are really local to KItem class. I don't like to expose them. The way I see it, this optimization is really local to KItem and has nothing to do with the outside world.

cos:
@yilongli I was asking the questions because I suggested the function is `(KLabelConstant, List<Sort>) -> Sort`. and you said it is `(Definition, KLabelConstant, KList) -> Sort`. From the code, it seems to me that only sorts are needed (`List<Sort>`), but, as you suggested `KList`, I wanted to be sure. So, again, is there anything more than sorts used there?

Regarding the tuple, `CacheTableColKey` is essentially: (KLabelConstant, List<(Sort, Boolean)>). Its constructor is tainted by `KItem` stuff but that could be eliminated/extracted into a method that generates a clean tuple/helper class. 

The problem with parallelization is that the threads will race on `Table<Definition>,` so you would have to replace it with a thread-safe construct. The contention might not be bad (but it is worth measuring), but why introduce it in the first place considering that we already have a reference to the `Definition` in hand?

I'm not arguing that the optimization is not local to `KItem`. I'm arguing the algorithm for determining sorts doesn't belong to `KItem` in the first place. The need to use `static` for this optimization just reveals the problem.

yilongli:
I don't understand. Are you saying computing the sort of a KItem doesn't belong to the KItem class? If you look at KLabelConstant, it's also cached with a static map. Implementing singletons is the perfect use case for static variables. Seriously, I am open to change, but I just don't see any benefit of over-complicating this thing. If Andrei and Dwight do not complain, I am not going to change this.

cos:
> I don't understand. Are you saying computing the sort of a KItem doesn't belong to the KItem class? 

Yes. It seems to me that it belongs to the `Definition`. I may be wrong. @andreistefanescu, @dwightguth,  please pitch in here.

> If you look at KLabelConstant, it's also cached with a static map. 

Hm... is that even correct (`private static final HashMap<String, KLabelConstant> cache`)? Aren't `KLabelConstant`s also `Definition` specific?

> Implementing singletons is the perfect use case for static variables.

Why is this a "singleton"? Which class has a singleton object?

> Seriously, I am open to change, but I just don't see any benefit of over-complicating this thing. If Andrei and Dwight do not complain, I am not going to change this.

I think what I'm proposing is a simplification in terms of end result, but it does involve a bit of work disentangling of the sort computation. I agree that there are more pressing problems to be solved in K. But I am not, by far, convinced this `static` is the best solution.

yilongli:
Not singleton. My bad. I mean flyweight or hash-consing pattern.

src/javasources/KTool/.classpath
```diff
@@ -5,8 +5,8 @@
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
 	<classpathentry kind="src" path="lib/resources"/>
 	<classpathentry kind="lib" path="../../../lib/java/com.microsoft.z3.jar"/>
-	<classpathentry kind="lib" path="../../../lib/java/collections-generic-4.01.jar"/>
```
dwightguth, line 8: You should probably add this back since it's a transitive dependency.

dwightguth:
I would tend to side with Cosmin that static state, even for something like caching, is undesirable when there are alternatives. However, I also don't want to spend too much time on this now, because I'm not sure it's a priority at the moment. The cache is private so it doesn't expose static state directly outside the class. Therefore if, at a later date, we decide we want to change this, we should be able to do so without too much work done in the way of detangling interconnected code.

Cosmin, does that seem like a reasonable compromise to you? Am I missing some way in which this choice could come back as pain on us later? If so, I suggest you give an example scenario, because as it is, your point, while well made, doesn't give me reason to believe we gain anything by doing this now instead of later. And I'd like to be able to get Yilong unblocked so he can start working on creating a generic kil Sort class, which we need for our next NASA deliverable at the end of August.

dwightguth:
Jenkins: test this please

yilongli:
Can I merge this if tests pass? @dwightguth

cos:
I agree there are more important things to focus on now. We can get back to this later. It can bite us if we decide to allow rewriting on multiple definitions in parallel (e.g., for tests), or it can cause code duplication if somebody decides to compute sorts in another part of the codebase, and doesn't know something similar is already in `KItem`.

Again, I agree we can adjust this later.

dwightguth:
Yes, you can merge if the tests pass.

yilongli:
Computing the sort of what will be similar to KItem?

cos:
@yilongli, not necessarily of what, as simply in another place. E.g., this sort computation might not be that dissimilar to part of @radumereuta's sort inference algorithm. Do they share any code?

radumereuta:
There are parts that are shared, like computing GLB or LUB. One main difference is related to the classes that I'm applying the sort inference, and since these are subject to change in the near future, i would like to have the freedom of doing that quicker and not worry about the backend. So I'm not really eager to merge these two code parts right now.

PR #716, opened 2014-07-14
fix kast --directory

dwightguth:
Fix a regression where kast --directory and kompile --directory got out of sync


dwightguth:
@radumereuta please review

radumereuta:
lgtm

Issue #718, opened 2014-07-16
Java backend doesn't know how to put back kresult in to a HOLE in this case


liyili2:
I have the following definition:

```
module TEST

     configuration <k> $PGM:K </k>

     syntax KItem ::= "toKRList'" "(" SubKRList ")" [strict]

     syntax KItem ::= toKRList(List)

     syntax KResult ::= krlist(List)

     syntax DeclType ::= Prototype(K, K, Bool)
          [strict(1, 2)]

     //might not need this after accosicate matching 
     syntax SubKRItem ::= sItem(K) [strict]
     syntax SubKRList ::= List{SubKRItem,"ss::"} [strict]
     syntax SubKRList ::= toSubKRList(List) [function]

     syntax SubResult ::= srItem(K)
     syntax SubKRItem ::= SubResult
     syntax KResult ::= SubResult

     syntax KItem ::= "abc" | "start"
     syntax Type ::= "skip"
     syntax KResult ::= "good" | Type | Int

     rule toSubKRList(.List) => .SubKRList
     rule toSubKRList(ListItem(K:K) L:List) => sItem(K) ss:: toSubKRList(L)

     rule sItem(V:KResult) => srItem(V) [structural]

     rule toKRList(L:List) => toKRList'(toSubKRList(L)) ~> krlist(.List)
          [structural]

     rule toKRList'((srItem(V:KResult) ss:: S:SubKRList) => S) 
          ~> krlist(_:List (.List => ListItem(V))) 
          when getKLabel(V) =/=KLabel 'krlist
          [structural]
     // Somewhat contrived feature: nested lists are flattened.
     rule toKRList'((srItem(krlist(L:List)) ss:: S:SubKRList) => S) 
          ~> krlist( _:List (.List => L)) 
          [structural]
     rule toKRList'(.SubKRList) => .
          [structural]

     rule abc => skip [structural]

     rule start => Prototype(abc,toKRList(ListItem(1) ListItem(2) ListItem(3)),true)

     rule Prototype(T:Type, krlist(L:List), true) 
          => good [structural]

endmodule

```

After I run the program "start", I get the result in Maude backend as:

```
<k>
    good
</k>

```

However, Java backend gives me the result of:

```
<k>
    krlist ( ListItem ( 1 ) ListItem ( 2 ) ListItem ( 3 ) ) ~> Prototype ( 
      skip , HOLE , true )
</k>

```

[...]

PR #728, opened 2014-07-17
subsorted KResult to KItem instead of K

yilongli:
Fixed https://github.com/kframework/k/issues/718

### DIFF
--- a/include/k-prelude.k
+++ b/include/k-prelude.k
@@ -28,7 +28,7 @@ module K-RESULT
 	imports K-TECHNIQUE
 	imports BOOL-HOOKS
 
-	syntax K ::= KResult
+	syntax KItem ::= KResult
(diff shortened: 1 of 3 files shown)


dwightguth:
I'm worried about what will happen now if I write something like:

```
rule isKResult(foo ~> bar) => true
```

dwightguth:
Actually I guess that's not any different from what already happens if I say:

```
rule isExp(foo ~> bar) => true
```

Okay, go ahead and merge this if tests pass.

grosu:
That's why I suggested in my previous message that all membership predicates should take KItem arguments.  Radu's parser can enforce that.

Grigore

yilongli:
I created a new issue to keep track of it. https://github.com/kframework/k/issues/729

Issue #720, opened 2014-07-16
CompileDataStructures not called from RuleCompilerSteps


dwightguth:
I found a bug in the unified-builtins work that Traian did today. You can't match on data structures in a search pattern because they are not compiled correctly. I tried to add the pass to RuleCompilerSteps but I couldn't because CompileDataStructures expects a rule and search patterns are just that: patterns. They don't have a rewrite in them.

Traian, when you get some time can you try to resolve this?

dwightguth:
Please see the related pull request. Completing this task should also involve modifying one example in the tutorial which is currently blocked on this bug.

PR #730, opened 2014-07-17
Fix for Issue #720

traiansf:
@dwightguth please review


dwightguth:
lgtm. merge if tests pass

traiansf:
@dwightguth All tests finally pass, but we are in a merge freeze.  Please merge it after the unfreeze.

dwightguth:
Jenkins: test this please

dwightguth:
@traiansf looks like you need to update your editor settings to take care of whitespace in .k files.

traiansf:
Ok, I'm puzzled.  I'm using a variety of editors, on a Windows machine.  Of course they tend to put windows line endings esp. to newly created files to make them readable through not-so-smart-but-neverhteless-usefull Windows editors like Notepad.  But I sincerely thought that git is taking care of converting between line endings to make everybody happy by keeping the server with one kind of line endings and converting them appropriately on each OS.

Now, w.r.t. trailing whitespace:  I'm mostly using vim for k.  Can't a line be added to the vim configuration for K to automagically remove trailing whitespace when saving?  If so, can somebody who is more proficient with vim configurations do this?

radumereuta:
Git indeed takes care of converting windows line endings to linux when commiting. The problem is that we already had so many windows terminators from SVN, which are being annoying now. I actually support what Dwight is doing right now.

traiansf:
I'm not saying I don't support what he's doing. Just that I need assistance for complying.

dwightguth:
This is what I found on a casual google search. Should give you several options.

http://stackoverflow.com/questions/356126/how-can-you-automatically-remove-trailing-whitespace-in-vim

dwightguth:
As far as git goes, I don't exactly know what it's capable of but I do know that it doesn't seem to be smart enough to stop IDEs from messing with the newline character inappropriately and causing newlines to get inconsistent. I decided to value code base conformity over ease of use on windows, because I don't want to have to worry about misconfigured editors screwing things up.

PR #732, opened 2014-07-18
remove extraneous io exceptions + improve generic infrastructure for log...

dwightguth:
...ging errors

@cos please review

### DIFF
--- a/src/javasources/KTool/src/org/kframework/backend/kore/KoreBackend.java
+++ b/src/javasources/KTool/src/org/kframework/backend/kore/KoreBackend.java
@@ -45,6 +45,7 @@ import org.kframework.krun.ColorSetting;
 import org.kframework.main.FirstStep;
 import org.kframework.utils.ColorUtil;
 import org.kframework.utils.Stopwatch;
+import org.kframework.utils.general.GlobalSettings;
 
 import java.awt.Color;
 import java.io.IOException;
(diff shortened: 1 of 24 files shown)


src/javasources/KTool/src/org/kframework/backend/kore/KoreBackend.java
```diff
@@ -74,10 +75,15 @@ public void run(Definition toKore) throws IOException {
 
           if(!fileTable.containsKey(((toKore.getItems().get(i)).getFilename()))){
 
+              String filename = ((toKore.getItems().get(i)).getFilename().substring(0,
+                      (toKore.getItems().get(i)).getFilename().length()-2))+".kore";
```
cos, line 79: minor: extract `toKore.getItems().get(i)).getFilename()` into a variable

cos:
This looks good. It gets us stack traces in most cases we want them. Not sure about this, but it may still be good for `KException` to piggyback the original exception. It could be useful when, for example, we're debugging without the verbose option and end up with a few exceptions in `kem`. The exceptions could be inspected for their original source. Would that be hard?

src/javasources/KTool/src/org/kframework/backend/kore/KoreBackend.java
dwightguth, line 79: I'd rather not... this is code that is on a deprecated code path and I don't really particularly want to mess with it any more than I have to...

dwightguth:
@cos please re-review

dwightguth:
Cosmin approved this via IM to me so I am merging this.

Issue #738, opened 2014-07-20
A strange behavior for a computation other than K cell when I want to get keys of a map 


liyili2:
Actually, This problem belongs to both Java backend and Maude backend. If I have this set of rules:

```
module TEST

     configuration <T><k>$PGM:K</k>
     <goto-map-calc color="cyan">
          <goto-calc multiplicity="*" color="cyan">
               <computation color="cyan"> .K </computation>
          </goto-calc>
     </goto-map-calc>
     <control>
          <curr-function color="lightgray">
               file-scope
          </curr-function>
     </control>
    <translation-units>
    <tu multiplicity="*">
    <goto-map color="lightgray"> .Map </goto-map>
    </tu>
    </translation-units>
    </T>

     syntax KItem ::= "start"

     syntax PureStatement ::= "Label" "(" CId "," K ")" [avoid, klabel('Label)]
     syntax PureStatement ::= "Return" "(" K ")" [klabel('Return)]
     syntax KItem ::= calculateGotoMap(CId, K)

     syntax KItem ::= "waitingOnGotoMap"
     syntax CId ::= Identifier(String) | funLabel(CId) | "file-scope"

     syntax Value ::= "emptyValue"

     syntax KItem ::= map(Map)

     syntax KItem ::= abc(Set)

     rule <k> start => calculateGotoMap(Identifier("isprint"),
        Label(funLabel(Identifier("isprint")),Return(emptyValue))) ...</k>
      <translation-units>
       (. =>  <tu>... <goto-map> .Map </goto-map> ...</tu>)
      ...</translation-units>

     rule <computation> Return(emptyValue) => .K ...</computation>
     rule <computation> abc(.Set) => .K ...</computation>
     rule <computation> abc(SetItem(Identifier("isprint"))) => .K ...</computation>

     rule <k> waitingOnGotoMap => .K ...</k>
      <goto-map-calc> .Bag </goto-map-calc>

     rule <k> calculateGotoMap(F:CId, K:K) => waitingOnGotoMap ...</k>
          <curr-function> _ => F </curr-function>
          <goto-map> M:Map => M[F <- map(.Map)] </goto-map>
          <goto-map-calc>
               . => (<goto-calc>... 
                         <computation> 
                              K ~> Return(emptyValue)
                         </computation>
                    ...</goto-calc>)
          </goto-map-calc>
          [structural] 

     rule <computation> (Label(Target:CId, S:K) => abc(keys(M))) ~> K:K </computation>
[...]

PR #743, opened 2014-07-21
fixing issue 738

yilongli:
fixed https://github.com/kframework/k/issues/738

### DIFF
--- a/src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/ExtendedSubstitution.java
+++ b/src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/ExtendedSubstitution.java
@@ -58,4 +58,9 @@ class ExtendedSubstitution {
     void addWriteCell(Cell<?> cell) {
         writeCells.add(cell);
     }
+
+    @Override
+    public String toString() {
+        return "extSubst(" + subst + ", " + writeCells + ")";
+    }
 }
--- a/src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
+++ b/src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
@@ -238,9 +238,11 @@ public class KAbstractRewriteMachine {
         if (content instanceof CellCollection) {
             return ((CellCollection) content).cellMap().get(label);
         } else if (content instanceof Cell) {
-            return Collections.singleton((Cell) content);
+            return cell.getLabel().equals(label) ?
+                    Collections.singletonList((Cell) content) :
+                    Collections.<Cell>emptyList();
         } else {
-            assert false;
+            assert false : "Wrong instruction!";
             return null;
         }
     }
--- a/src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
+++ b/src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
@@ -234,7 +234,15 @@ public class KAbstractRewriteMachine {
     }
 
     private Collection<Cell> getSubCellsByLabel(Cell<?> cell, String label) {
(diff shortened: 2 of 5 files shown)


yilongli:
I could create a regression test to catch exactly this bug, but I am reluctant to do that because, in my opinion, we already have many small regression tests that are too weak to be useful. I believe the right way to avoid similar problems in the future is to have different checkers at the end of kompilation to validate certain properties of the kompiled definition. For example, in this case, we should have a checker saying that there should be no KLabels representing data structure constructors in the definition.

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/ExtendedSubstitution.java
```diff
@@ -58,4 +58,9 @@ void setSubst(Map<Variable, Term> subst) {
     void addWriteCell(Cell<?> cell) {
         writeCells.add(cell);
     }
+
+    @Override
+    public String toString() {
```
dwightguth, line 63: I don't understand how the changes to these three files relate to the change that fixes this issue.

yilongli, line 63: having toString method makes debugging easy. I made the change during debugging.

dwightguth:
Much of the problem with our small regression tests is that they are very slow to execute because they are essentially end-to-end tests. I suggest you make a JUnit unit test for CompileDataStructures which tests the basic functionality of the pass, including this use case. The last time I did something similar for one of the compiler passes, I found a number of latent bugs that had not even yet been detected with the code in question.

src/javasources/KTool/src/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
```diff
@@ -234,7 +234,15 @@ private Term getWriteCellRHS(String cellLabel) {
     }
 
     private Collection<Cell> getSubCellsByLabel(Cell<?> cell, String label) {
-        return ((CellCollection) cell.getContent()).cellMap().get(label);
```
dwightguth, line 237: right okay fine so toString I kind of get. But what's this?

yilongli, line 237: The bug is reported for Java backend without --pattern-matching option. I tried to test --pattern-matching as well and found a class cast exception.

yilongli:
Jenkins: retest this please

yilongli:
Writing a unit test to properly cover all use cases of `CompileDataStructures` is not easy and very time-consuming. Right now, I simply write a very specific unit test to catch the bug exposed in issue#738. I will probably come back some time to write unit tests for all important compilation passes (at the same time, develop some common infrastructure used frequently in building mock KIL objects).

yilongli:
I think the time-out is due to the slight change in `CompileDataStructures` which results in a little more time spent in compilation. The time limit is already 40 mins. We should probably discuss how to reduce the build time effectively.

yilongli:
@andreistefanescu please take a look at my changes

andreistefanescu:
lgtm. I'm not sure what happens with the time. This change should not increase the compilation time by that much. Maybe one of the examples is not terminating now?

yilongli:
Nope. The last example before time out is `bf` and it works on my laptop.

dwightguth:
Jenkins: test this please

dwightguth:
Next time this happens please ssh into the jenkins machine and run `top` to check the cpu usage. I am having trouble figuring out what triggers this, but sometimes processes get orphaned by jenkins and so the cpu gets hammered and builds time out. The fix is to just kill the processes. I will try to look into what is causing this today though to see if I can find a permanent solution.

yilongli:
So after you kill the processes, the build time drops from 40mins to 24mins?

dwightguth:
The reason it timed out was that it was trying to execute one process per thread when there were already processes from a previous build using up the cpus.

Issue #733, opened 2014-07-19
horrible error message when the definition includes strings with bachslashes


dlucanu:
If one tries to compile the following definition

```
module SPEC 
  syntax Ids ::= List{Id, "/\\"}
endmodule
```

then the following error is obtained:

```
$ kompile spec.k
Exception in thread "main" java.lang.AssertionError: SDF2Table returned errors: sdf-load:error: parse error, 'I' unexpected
--- input string [file: ./Program.sdf line: 13 col: 30]

    at org.kframework.parser.utils.Sdf2Table.run_sdf2table(Sdf2Table.java:36)
    at org.kframework.parser.DefinitionLoader.parseDefinition(DefinitionLoader.java:180)
    at org.kframework.parser.DefinitionLoader.loadDefinition(DefinitionLoader.java:86)
    at org.kframework.kompile.KompileFrontEnd.genericCompile(KompileFrontEnd.java:181)
    at org.kframework.kompile.KompileFrontEnd.kompile(KompileFrontEnd.java:161)
    at org.kframework.kompile.KompileFrontEnd.main(KompileFrontEnd.java:72)
    at org.kframework.main.Main.main(Main.java:49)
```

A confusing error is given also given when the string is "/\":

```
$ kompile spec.k
[Error] Outer Parser: Lexical error at line 45, column 32.  Encountered: "\n"
(10), after : "\"/\\\"}"
    File: /Users/dlucanu/homade/homade/khomade/fshi/arch-spec/spec.k
```

I think that the message should say something how the backslash should be handled in such definitions.

PR #744, opened 2014-07-22
Properly escaping list separators when printing to SDF.

radumereuta:
Fixing #733 
@dwightguth please review.


dwightguth:
If this is the correct method to call to escape things in sdf then that needs to be documented either by calling it escapeSDF, or with a javadoc.

radumereuta:
You were right. I'm not doing anything special for SDF.
That's why I removed StringUtil.escape/unescape and replaced all their usages with StringUtil.quoteString/unquoteString.
@dwightguth please re-review.

src/javasources/KTool/src/org/kframework/backend/maude/MaudeFilter.java
```diff
@@ -707,7 +707,7 @@ public Void visit(FloatBuiltin token, Void _) {
 
     @Override
     public Void visit(StringBuiltin token, Void _) {
-        result.append("#_(\"" + StringUtil.escape(token.stringValue()) + "\")");
+        result.append("#_(" + StringUtil.enquoteString(token.stringValue()) + ")");
```
dwightguth: No no no. You need to escape for maude here. EnquoteString escapes K strings.

src/javasources/KTool/src/org/kframework/parser/generator/Definition2SDF.java
```diff
@@ -125,7 +125,7 @@ public static StringBuilder getSdfForDefinition(Definition def, Context context)
         for (Production p : psdfv.outsides) {
             if (p.isListDecl()) {
                 UserList si = (UserList) p.getItems().get(0);
-                sdf.append("    " + StringUtil.escapeSortName(si.getSort()) + " \"" + si.getSeparator() + "\" " + StringUtil.escapeSortName(p.getSort()) + " -> "
+                sdf.append("    " + StringUtil.escapeSortName(si.getSort()) + " " + StringUtil.enquoteString(si.getSeparator()) + " " + StringUtil.escapeSortName(p.getSort()) + " -> "
```
dwightguth: and here you need to escape using sdf strings, not K strings

dwightguth:
Please bear in mind that different languages have different syntaxes for strings. When outputting a string for a particular language, you must escape that string _for that language_. Thus you need an escapeK, escapeMaude, escapeSDF, etc

radumereuta:
I gave it another try.
I don't have the patience to look over every single place where escape and unescape are used.
If there are any more problems they will show up in the future.

dwightguth:
Listen, if you want to keep an "escape" method around for all the call sites that you don't want to change, and make that function deprecated, you can, but changing even the call sites which are clearly Maude specific to call a function "escapeSDF" is clearly not a good solution either. I need you to be more conscious about the diffs you're making. Don't just rush in and make the change. Stop and think: what is the impact of this change on the quality of the code base? How _should_ a feature be designed? Does this change even make sense?

If you had done this, you would have immediately realized that both of the changes you just submitted to me refactoring the call sites of this function are obviously incorrect.

radumereuta:
I've renamed the StringUtil escape functions into:
enquoteCString and equoteKString, and I've added more detailed javadoc for each.
Also, I've implemented the missing unescape for octal values.
Added junit tests for quote unquote.

src/javasources/KTool/src/org/kframework/utils/StringUtil.java
```diff
@@ -11,9 +11,22 @@
 import com.beust.jcommander.ParameterDescription;
 
 public class StringUtil {
-    public static String unescape(String str) {
+    /**
+     * Unescape the textual representation of a string specific tu SDF.
```
dwightguth: and maude

src/javasources/KTool/src/org/kframework/utils/StringUtil.java
```diff
@@ -111,11 +147,15 @@ public static int indexOfAny(String str, String search, int offset) {
     }
 
     /**
-     * Removes the first and last double-quote characters and unescapes special characters.
-     * @param str double-quoted string
+     * Removes the first and last double-quote characters and unescapes special characters
+     * that start with backslash: newline, carriage return, line feed, tab and backslash.
+     * Characters between 127 and 255 are stored as \xFF
+     * Characters between 256 and 65535 are stored as \uFFFFFF
```
dwightguth: \uFFFF

src/javasources/KTool/src/org/kframework/utils/StringUtil.java
```diff
@@ -111,11 +147,15 @@ public static int indexOfAny(String str, String search, int offset) {
...
+     * Characters above 65536 are stored as \UFFFFFFFFFFFF
```
dwightguth: \u0010FFFF

src/javasources/KTool/src/org/kframework/utils/StringUtil.java
```diff
@@ -168,11 +208,15 @@ public static String unquoteString(String str) {
     }
 
     /**
-     * Adds double-quote at the beginning and end of the string and escapes special characters.
+     * Adds double-quote at the beginning and end of the string and escapes special characters
+     * with backslash: newline, carriage return, line feed, tab and backslash.
+     * Characters between 127 and 255 are stored as \xFF
```
dwightguth, line 213: see above comments

src/javasources/KTool/test/org/kframework/utils/StringUtilTest.java
```diff
@@ -0,0 +1,50 @@
...
+        String enquoted = StringUtil.enquoteCString(str);
+        String backToAll = StringUtil.unquoteCString(enquoted);
+        char[] all2 = backToAll.toCharArray();
+
+        Assert.assertEquals("Different sizes after running quote unquote.", all.length , all2.length);
+        for (int i = 0; i < 256; i++)
+            Assert.assertEquals("Different values at position: " + i, all[i] , all2[i]);
+    }
+
+    @Test
```
dwightguth: look into what you can do with `@Test(expected=IllegalArgumentException.class)`

It doesn't really matter, because style isn't as important for tests, but you should be aware of it.

dwightguth:
looks much better. Just a few tiny comments about the comments, and you can merge this.

radumereuta:
@dwightguth please re-review.

dwightguth:
Merge if tests pass

PR #749, opened 2014-07-23
serialize context instead of trying to recreate it

dwightguth:
**please review** @traiansf please review


dwightguth:
@traiansf please review

traiansf:
looks ok, as far as I can tell.

PR #750, opened 2014-07-24
remove dependency on jopt-simple

dwightguth:
@traiansf please review


traiansf:
Looks good, but I wonder why were there option to begin with?

dwightguth:
Long ago, this was put this way by the original authors of krun way back in 2011 when the entire tool was still in Maude. It's sat there ever since because it kept working. I just wanted to get rid of the unnecessary jar from the distribution, or I wouldn't have touched it.

PR #758, opened 2014-07-25
initial commit for Google Guice

dwightguth:
**please review**
- Use dependency injection to bootstrap the tools
- Add static injections for code that will no yet be initialized by Guice, and deprecate global static state
- reorganize the way the tool boots by creating a common FrontEnd class.


dwightguth:
Jenkins: test this please

dwightguth:
@cos please review. Anyone who knows anything about dependency injection, please review also. Others can review too since this is a cross cutting change.

I suggest strongly that anyone who reviews this find me and talk to me unless they are already familiar with dependency injection or Google Guice.

cos:
It looks good overall. One minor concern are the @First, @Second annotations which could have more meaningful names.
I have a long flight today and will give it a closer look offline.

cos:
aside from the @First, @Second, it looks good overall. from my perspective, it is good to merge.

PR #760, opened 2014-07-26
use tiered compilation to decrease jvm startup cost

dwightguth:
@yilongli please review

### DIFF
--- a/bin/kagreg
+++ b/bin/kagreg
@@ -2,7 +2,7 @@
 
 ulimit -s `ulimit -H -s`
 if [ -z "$K_OPTS" ];
-  then export K_OPTS="-Xms64m -Xmx1024m -Xss32m"
+  then export K_OPTS="-Xms64m -Xmx1024m -Xss32m -XX:+TieredCompilation"
 fi
 if "$(dirname $0)/../lib/scripts/checkJava"; then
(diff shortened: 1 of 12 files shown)


yilongli:
Are you going to also change `kast.bat`, `krun.bat`, etc.?

dwightguth:
Good catch. Will do.

yilongli:
Maybe it's a dumb question, but am I supposed to observe any decrease in the Jenkins build time?

dwightguth:
It's not a dumb question. I'm looking into it. Whenever I try it on my local machine, I see decreases in wall time, but they're not replicable when I run Jenkins. Something further needs tuning I believe.

dwightguth:
After playing around with it some, I've concluded the reason we don't see speedups in ktest is that ktest is currently running on jenkins using 12 threads, but the test suite becomes cpu bound around only 4 or 5 threads. After that point, adding new threads does not decrease wall time, it only increases user time as threads wait to be scheduled. The change in question decreases the cost of running kompile and krun as a single process, but it does not appreciably decrease the load on the cpu, because it relies primarily for its speedup on JVM daemon threads performing compilation and profiling. So when all CPUs are at 100% utilization, the code does not actually run any faster.-

Can I merge this? The speedup from this diff is actually very appreciable.

yilongli:
Yes, please merge if all the tests pass.

PR #769, opened 2014-07-30
Add support for anywhere rules

yilongli:


src/javasources/KTool/src/org/kframework/backend/java/symbolic/SubstituteAndEvaluateTransformer.java
```diff
@@ -25,6 +25,8 @@
  */
 public class SubstituteAndEvaluateTransformer extends CopyOnWriteTransformer {
 
+    private static final boolean ENABLE_DEBUG = false;
```
andreistefanescu: we should use either -v or introduce a --debug flag

yilongli:
I am sure not sure why the CIL semantics fails on type checking, but, anyway, it is ready for review. @andreistefanescu @dwightguth 

@grosu I changed the definitions of SIMPLE and KOOL in the tutorial to use [anywhere] rules as well as some comments. Please take a look.

andreistefanescu:
I merged a cil commit that only works with Radu's branch, so now it is
broken.

src/javasources/KTool/src/org/kframework/backend/java/kil/Definition.java
```diff
@@ -96,6 +98,15 @@ public void addRule(Rule rule) {
             }
         } else if (rule.containsAttribute(Attribute.MACRO_KEY)) {
             macros.add(rule);
+        } else if (rule.containsAttribute(Attribute.ANYWHERE_KEY)) {
+            if (!(rule.leftHandSide() instanceof KItem)) {
+                GlobalSettings.kem.registerCriticalWarning(
```
dwightguth, line 107: Can you make an overload of the registerCriticalWarning method that takes only the non-null arguments here?

src/javasources/KTool/src/org/kframework/backend/java/kil/Rule.java
```diff
@@ -175,13 +177,26 @@ public void visit(Cell cell) {
...
+            if (sortPredArg == null) {
+                isSortPredicate = false;
+
+                /*
+                 * YilongL: the right-hand side of the sort predicate rule
+                 * is not necessarily {@code BoolToken#True}? for example:
+                 *     rule isNat(I:Int) => '_>=Int_(I:Int,, Int(#"0"))
+                 */
+                // TODO(YilongL): properly re-implement support for sort predicate rules
+//                GlobalSettings.kem.registerCriticalWarning("Unexpected sort predicate rule: " + this, null, this);
```
dwightguth, line 198: This todo worries me. Can you tell me what you mean by it? have you disabled some kind of functionality that was working before?

dwightguth:
Lgtm other than the things I mentioned. @grosu can you review?

dwightguth:
Jenkins: test this please

src/javasources/KTool/src/org/kframework/backend/java/kil/Rule.java
yilongli, line 198: No, everything that worked before still works now. I am just excluding some new cases which I couldn't handle. Specifically, I used to assume that sort predicate rule is defined on KItem only, i.e., `rule isVal(cons V:Val) => true` in `fun-untyped.k`. However, now I see `rule isNat(I:Int) => I >=Int 0 [function, anywhere]` in `builtins/int.k` after including all `[anywhere]` rules. I need to exclude this rule until I properly re-implement the support for sort predicate later.

src/javasources/KTool/src/org/kframework/backend/java/kil/Definition.java
yilongli, line 107: I have addressed this comment, but the change is actually on the next line. So the comment doesn't disappear.

yilongli:
@dwightguth the build time increases to 30mins again. Is it because our recent changes to the k framework or the Jenkins issue you mentioned last time?

@grosu please review the changes to tutorial

dwightguth:
The build seems to have spiked in time by about 3 minutes between builds
http://office.runtimeverification.com:8080/rv-jenkins/job/k-framework-pull-requests/224/
and
http://office.runtimeverification.com:8080/rv-jenkins/job/k-framework-pull-requests/225/

The logical conclusion is that either something changed on that box between
those two builds, or the change I made to machine integers is a performance
problem. I don't really understand how either of those things could be
true though... I will investigate on Monday.

tutorial/2_languages/1_simple/1_untyped/simple-untyped.k
```diff
@@ -566,17 +566,15 @@ precise location; we prefer to do it this way to avoid locking the store.
...
-// would be underused, because it would only be used at the top of the
-// computation or inside the lvalue wrapper.  So it may not be worth,
-// or we may need to come up with a special notation allowing us to
-// enumerate contexts for [anywhere] rules.
-  rule V:Val[N1:Int, N2:Int, Vs:Vals] => V[N1][N2, Vs]            [structural]
-  rule lvalue(V:Val[N1:Int, N2:Int, Vs:Vals] => V[N1][N2, Vs])    [structural]
-
-  rule array(L,_)[N:Int] => lookup(L +Int N)            [structural]
-  rule lvalue(array(L,_)[N:Int] => lookup(L +Int N))    [structural]
+// The [anywhere] feature would be underused, because it would only
```
grosu: Here I would use present tense: "The [anywhere] feature is underused ...".

Other than this, all the changes to the tutorial look great.  I am very happy to see the anywhere rules back.  Thanks, Yilong!

PR #770, opened 2014-07-31
remove cons attribute everywhere but frontend

dwightguth:
@radumereuta please review


dwightguth:
@radumereuta please review

include/builtins/string.k
```diff
@@ -47,7 +47,7 @@ module STRING-HOOKS
 
 
   syntax String ::= "KLabel2String" "(" KLabel ")"  [function, klabel(KLabel2String)]
-  syntax KLabel ::= "String2KLabel" "(" String ")" [function, cons(KLabel1String2KLabelSyn)]
+  syntax KLabel ::= "String2KLabel" "(" String ")" [function]
```
radumereuta: I suggest you use the short form:
syntax KLabel ::= String2KLabel(String) [function]

include/io/uris.k
```diff
@@ -92,10 +92,10 @@ module URIS
...
   rule #statTypes(K:K) => K
 
-  syntax Stream ::= "#parseInput" "(" String "," String ")"   [cons(Stream1ParseSyn), klabel(#parseInput)]
-  syntax Stream ::= "#buffer" "(" K ")"           [cons(Stream1IOBufferSyn), klabel(#buffer)]
-  syntax Stream ::= "#istream" "(" Int ")"        [cons(Stream1InputStreamSyn)]
-                    | "#ostream" "(" Int ")"        [cons(Stream1OutputStreamSyn)]
+  syntax Stream ::= "#parseInput" "(" String "," String ")"   [klabel(#parseInput)]
+  syntax Stream ::= "#buffer" "(" K ")"           [klabel(#buffer)]
+  syntax Stream ::= "#istream" "(" Int ")"        [klabel(#istream)]
+                    | "#ostream" "(" Int ")"        [klabel(#ostream)]
```
radumereuta: Short form here too...
syntax Stream ::= #buffer(K)
would be equivalent to:
syntax Stream ::= "#buffer" "(" K ")" [klabel('#buffer)]

Notice that the klabel is automatically defined as <quote> + <name>
I think the definition would look better if you do it this way.

include/k-prelude.k
```diff
@@ -71,11 +71,11 @@ module K-WRAPPERS
...
-module K-EVAL
-  imports K-TECHNIQUE
-  syntax Bag ::= "#eval" Map          [cons(Bag1EvalSyn)]
-  syntax Bag ::= "#evalHelper" Map    [cons(Bag1EvalHelperSyn)]
-endmodule
+//module K-EVAL
+//  imports K-TECHNIQUE
+//  syntax Bag ::= "#eval" Map          [cons(Bag1EvalSyn)]
+//  syntax Bag ::= "#evalHelper" Map    [cons(Bag1EvalHelperSyn)]
+//endmodule
```
radumereuta: Why not remove it?
If you comment it, at least write why, and when it is ok to remove it.

src/javasources/KTool/src/org/kframework/backend/BasicBackend.java
```diff
@@ -77,8 +77,6 @@ public boolean autoinclude() {
         steps.add(new AddStreamCells(context));
         steps.add(new AddSymbolicK(context));
         steps.add(new AddSemanticEquality(context));
-        // steps.add(new ResolveFresh());
-        steps.add(new FreshCondToFreshVar(context));
```
radumereuta, line 81: Is this related?

src/javasources/KTool/src/org/kframework/backend/kore/KoreBackend.java
```diff
@@ -131,8 +130,6 @@ public String getDefaultStep() {
         steps.add(new AddStreamCells(context));
         steps.add(new AddSymbolicK(context));
         steps.add(new AddSemanticEquality(context));
-        // steps.add(new ResolveFresh());
-        steps.add(new FreshCondToFreshVar(context));
```
radumereuta, line 135: I think this is obsolete code. Correct me if I'm wrong.

src/javasources/KTool/src/org/kframework/compile/transformers/DesugarStreams.java
```diff
@@ -51,46 +47,33 @@ private Term makeStreamList(String stream, Cell node) {
 //            eq evalCleanConf(T, "stdin") = mkCollection(List, (T, ioBuffer(stdinVariable), noIOVariable, stdinStream)) .
             addAtBeginning = contents;
 //            syntax List ::= "#buffer" "(" K ")"           [cons(List1IOBufferSyn)]
```
radumereuta: Why keep the cons attribute? It's just going to be confusing in the future.

src/javasources/KTool/src/org/kframework/kil/KLabelConstant.java
```diff
@@ -117,20 +119,21 @@ private KLabelConstant(String label, Context context) {
     public KLabelConstant(Element element, Context context) {
         super(element);
         label = StringUtil.unescapeMaude(element.getAttribute(Constants.VALUE_value_ATTR));
-        productions = Collections.unmodifiableList(context.productionsOf(label));
+        // context.productionsOf returns a view into a multimap, which is not serializable. So we copy it
+        productions = ImmutableSet.copyOf(context.productionsOf(label));
     }
 
     @SuppressWarnings("unchecked")
```
radumereuta: Is the warning suppressor still needed?

src/javasources/KTool/src/org/kframework/kil/TermCons.java
```diff
@@ -16,18 +16,16 @@
  */
 public class TermCons extends Term implements Interfaces.MutableList<Term, Enum<?>> {
     /** A unique identifier corresponding to a production, matching the SDF cons */
-    protected final String cons;
```
radumereuta, line 19: Hmm... I'm not sure about this.
Remember that we have that issue with caching. You said we should make production transient and don't store it in the cache. The safest way to do that, would be to store something else, namely the cons, which is safe when considering cache invalidation. So you may want to still keep this field.

radumereuta:
Overall looks reasonable, aside from just a few concerns that I posted in the code.

src/javasources/KTool/src/org/kframework/backend/BasicBackend.java
dwightguth, line 81: yes. this is dead code that I removed in order to avoid having to maintain this class.

src/javasources/KTool/src/org/kframework/kil/TermCons.java
dwightguth, line 19: So if you want to add a uniqueId field to the Production class, and base it on whatever algorithm you used to compute the cons, you're welcome to. But not all TermConses are constructed from SDF anymore. I don't think keeping this field around is the right move.

radumereuta, line 19: I think we're gonna miss an opportunity for an easy fix for the caching problem, but fine... we can add it back when we get to it.

radumereuta:
lgtm, merge if tests pass.

dwightguth:
Jenkins: test this please

PR #774, opened 2014-08-01
fix some bugs to make stuff more explicitly bound

dwightguth:
@cos please review


cos:
looks good, merge if it passes

PR #779, opened 2014-08-03
removed obsolete test generation code

yilongli:
This part of the code was highly experimental and fragile. Right now, I don't have the time to maintain it whatsoever. Besides, since I have a much better understanding of the entire Java backend now, I am convinced that most of the things I did back then were simply wrong. I will redo the whole thing from scratch in the future.


grosu:
I wish everybody did the same, that is, remove their own experimental/fragile code to save others time.  Maintaining such code eats up a lot of time, which we don't have.

Grigore

dwightguth:
Did you forget to delete GroupProductionsBySort?

PR #782, opened 2014-08-04
remove krun dependence on parser code

dwightguth:
@radumereuta @bmmoore please review

### DIFF
--- a/src/javasources/KTool/src/org/kframework/kil/Variable.java
+++ b/src/javasources/KTool/src/org/kframework/kil/Variable.java
@@ -54,7 +54,6 @@ public class Variable extends Term {
         this.name = name;
         this.freshVariable = freshVariable;
         this.freshConstant = freshConstant;
-        this.userTyped = true;
     }
 
     public Variable(String name, Sort sort) {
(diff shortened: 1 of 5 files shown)


dwightguth:
@radumereuta @bmmoore please review

src/javasources/KTool/src/org/kframework/kil/Variable.java
```diff
@@ -54,7 +54,6 @@ public Variable(String name, Sort sort, boolean freshVariable, boolean freshCons
         this.name = name;
         this.freshVariable = freshVariable;
         this.freshConstant = freshConstant;
-        this.userTyped = true;
```
radumereuta, line 57: Explain this please.

dwightguth, line 57: I put it in a while back, but it isn't really appropriate. It was just a mistaken fix for a bug that was better fixed another way. Brandon asked me to clean it up, so I did with this diff.

radumereuta:
Ok, then if Brandon agrees too, lgtm.

src/javasources/KTool/src/org/kframework/kil/Variable.java
bmmoore, line 57: This is the only thing I wanted/needed from this fix, so as long as tests still pass I'm fine with this pull request.

PR #787, opened 2014-08-04
Refactor location, filename, and klabel attributes

dwightguth:
**please review**


dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
@radumereuta @cos please review

src/javasources/KTool/src/org/kframework/compile/transformers/DataStructureToLookupUpdate.java
```diff
@@ -120,8 +122,8 @@ private void registerError(Rule node, String message) {
     private Map<Variable, Integer> concreteSize = new HashMap<>();
     private ArrayList<VariableCache> queue = new ArrayList<>();
     private Status status;
-    private String location;
-    private String filename;
+    private Location location;
+    private File filename;
```
radumereuta: Why is `filename` a File now?
Keeping it as a String has the advantage that is easier to serialize and deserialize.
Plus. How many times are do you need it to actually open a file? Most of the time this is a description, information for the users.

src/javasources/KTool/src/org/kframework/kast/KastOptions.java
```diff
@@ -53,11 +53,11 @@ public String stringToParse() {
      * {@link #stringToParse()}.
      * @return A textual description of the source of the string to parse.
      */
-    public String source() {
+    public File source() {
         if (expression != null) {
-            return "Command line";
+            return null;
```
radumereuta: Are you sure this is right? Please double check.

src/javasources/KTool/src/org/kframework/kil/ASTNode.java
```diff
@@ -19,9 +21,12 @@
      */
     private static final long serialVersionUID = 1L;
     /**
-     * Used on any node for metadata such as location, also used on {@link Rule} and {@link Production} for the attribute list.
+     * Used on any node for metadata also used on {@link Rule} and {@link Production} for the attribute list.
      */
-    protected Attributes attributes;
+    private Attributes attributes;
+
+    private File filename;
```
radumereuta: Again. I don't think File is a good solution. What happens when I send a command line argument?
That has no file representation. The filename should be "Command line argument". If it is generated through the code during one of the stages, it should contain "generated by X"... You should think of `filename` as 'origin'.

src/javasources/KTool/src/org/kframework/kil/ASTNode.java
```diff
@@ -81,23 +94,18 @@ public ASTNode() {
...
-        protected void setLocation(Element elem) {
-        setLocation(getElementLocation(elem));
-        setFilename(getElementFile(elem));
-    }
-
     /**
      * Retrieves the location of the current ASTNode.
      *
-     * @return recorded location or Constants.GENERATED_LOCATION if no recorded location found.
+     * @return recorded location or null if no recorded location found.
```
radumereuta, line 101: I agree with this.

src/javasources/KTool/src/org/kframework/kil/ASTNode.java
```diff
@@ -105,17 +113,17 @@ public String getLocation() {
...
-        putAttribute("location", loc);
+    public void setLocation(Location loc) {
+        location = loc;
     }
 
     /**
      * Retrieves the filename of the current ASTNode.
      *
-     * @return recorded filename or Constants.GENERATED_FILENAME if no recorded location found.
+     * @return recorded filename or null if no recorded location found.
```
radumereuta: Not so much with this. I think `filename` should always have a value. Something other than null and the empty string.

src/javasources/KTool/src/org/kframework/kil/Attributes.java
```diff
@@ -20,31 +24,31 @@
...
     public Attributes(Attributes c) {
         super(c);
         this.contents = c.contents;
     }
 
-    public Attributes(String location, String filename) {
+    public Attributes(Location location, File filename) {
         super(location, filename);
-        contents = new ArrayList<Attribute>();
+        contents = new LinkedHashMap<String, Attribute>();
```
radumereuta, line 37: The reason I've been using ArrayList instead of linked lists, is because it is way way waaaay easier to debug in the IDE. The speed benefit from this change is no way near the benefit of being able to quickly spot the term that you want in the lookup window. I would appreciate it if you would choose something similar.

src/javasources/KTool/src/org/kframework/parser/ProgramLoader.java
```diff
@@ -88,75 +88,70 @@ public static Term processPgm(String content, String filename, String startSymbo
...
+                    "Parse error: unexpected character '" + content.charAt(perror.position) + "'.";
+                Location loc = new Location(perror.line, perror.column,
+                                            perror.line, perror.column + 1);
+                throw new ParseFailedException(new KException(
+                        ExceptionType.ERROR, KExceptionGroup.INNER_PARSER, msg, filename, loc));
             }
-            Stopwatch.instance().printIntermediate("Parsing Program");
-
-            return (Term) out;
-        } catch (IOException e) {
```
radumereuta, line 157: Why did you remove the try/catch?

src/javasources/KTool/test/org/kframework/parser/basic/BasicParsingTests.java
```diff
@@ -19,7 +19,7 @@ public void testLexicalRules() throws Exception {
         // TODO: remove once the new parser is fully functional
         String def = "module TEST syntax Str ::= Token{((~[\\'\\n\\r\\\\])|([\\\\]~[\\n\\r]))*} endmodule";
 
-        List<DefinitionItem> defItemList = Basic.parse("UnitTest", def, null);
+        List<DefinitionItem> defItemList = Basic.parse(null, def, null);
```
radumereuta: Again about `filename` being File and not String.
In this case you are completely ignoring the fact that the AST produced by this parser could have location information. Don't underestimate the importance of good error messages. It can save huge amounts of time.

radumereuta:
Done reviewing, but I don't approve of it, since it reduces the possibility of reporting reporting error messages in certain cases.

src/javasources/KTool/src/org/kframework/kil/ASTNode.java
dwightguth: Filename is a file in order to strongly type it. A filename should be the name of a file.

If you like, I can add a third field however:

```
enum Source {
FILE,
COMMAND_LINE
}
```

Then we would make an assertion fail if you specified an incorrect combination of Source and filename.

Does this meet your approval?

src/javasources/KTool/src/org/kframework/kil/ASTNode.java
dwightguth: see my comment above about a non-null enum Source.

src/javasources/KTool/src/org/kframework/kil/Attributes.java
dwightguth, line 37: This is necessary because otherwise I can have two attributes with the same key and different values. it is still very easy to locate attributes in a LinkedHashMap. And it is a fallacy to introduce runtime dependencies on test code.

src/javasources/KTool/src/org/kframework/parser/ProgramLoader.java
dwightguth, line 157: Because nothing threw it.

src/javasources/KTool/src/org/kframework/kil/ASTNode.java
cos: It the field is called `filename`, then it can and should be a `File`. In this case, it seems to describe an origin that is not necessarily a file. Then it should be renamed to `origin`, and be of an appropriate abstract type. I propose:

```
interface Origin {
  String toString(); // for the sake of having a human-readable description
}
class OriginFile extends Origin {
  File filename;
}
class OriginSystemCommand extends Origin {
 ...
}
```

cos: better yet, as we're describing an ASTNode, the origin could also contain the line number directly.

dwightguth:
@radumereuta @cos please re-review

src/javasources/KTool/src/org/kframework/kil/Source.java
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
+package org.kframework.kil;
+
+import java.io.File;
+import java.io.Serializable;
+
+public abstract class Source implements Comparable<Source>, Serializable {
```
radumereuta: Why do you need to extend Comparable?

dwightguth: Because the maude backend sorts rules by their location and filename

radumereuta: ok

radumereuta:
1. There were a few places where you have to write `Source.of(new File(<string>))`. You could consider making a function `Source.fromFile(String)`.
2. There was a discussion at one point about making the 'Compilation Phase' appear in error messages only when you have `-v` or `--debug`. Right now it's just garbage information for users. It would be a good time to take care of that since you modified so much in the error reporting files.

lgtm
The too points above are just suggestions. You can merge even if you consider is not worth doing right now.

src/javasources/KTool/src/org/kframework/kil/Source.java
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    public static Source of(String optionName) {
+        return new CommandLineSource(optionName);
+    }
+
+    @Override
+    public int compareTo(Source o) {
+        return toString().compareTo(o.toString());
+    }
```
cos: a bit hackish, we could leave the implementation for clients

src/javasources/KTool/src/org/kframework/kil/Source.java
cos: `Source` ca be an interface, no need to restrict its usage to inheritance

PR #791, opened 2014-08-05
fixed issue with AddInjections not changing the productions properly; fi...

andreistefanescu:
...xed SpecificationCompilerSteps to use AddInjections


andreistefanescu:
@dwightguth please review. Also, don't forget to update the transformer for productions.

PR #792, opened 2014-08-05
Replace literal CR with escape sequence

bmmoore:
The fix_whitespace script used to include a literal carriage return
character in one of the replacements, which it quite confusing if
you just cat it. This instead uses bash's support for escape sequences
in $-prefixed string literals.


bmmoore:
I don't think this script would fix old Mac OS-style plain CR line endings, but I doubt we will run into those either.

dwightguth:
Fun fact, we did actually! But I fixed that a while ago and it was only one file I did manually.

PR #797, opened 2014-08-06
Fix cygwin for k 3.4

dwightguth:
I think we need to include this pull request in the release: Maude is currently broken on my version of windows.


dwightguth:
@traiansf please review

traiansf:
Looks good to me.  I wonder though why I did not experience the bug.

dwightguth:
I think it might only affect users who also have the 64 bit version of Cygwin installed on their system? Not sure.

PR #801, opened 2014-08-06
Coq backend

bmmoore:
Java side of the Coq backend. A statically linked binary is further ahead on bmmoore/k@cbdb4f79fdbc


src/javasources/KTool/src/org/kframework/kompile/KompileOptions.java
```diff
@@ -51,7 +52,8 @@ public KompileOptions(Void v) {}
         UNPARSE(false, false, false, UnparserBackend.class, null),
         UNFLATTEN(false, false, false, UnflattenBackend.class, null),
         UNFLATTEN_JAVA(false, true, false, UnflattenJavaBackend.class, null),
-        SYMBOLIC(false, false, true, SymbolicBackend.class, JavaSymbolicKRun.class);
+        SYMBOLIC(false, false, true, SymbolicBackend.class, JavaSymbolicKRun.class),
+        COQ(false, true, false, CoqBackend.class, null);
```
dwightguth, line 56: false, false, true

tests/regression/coq-backend/README
```diff
@@ -0,0 +1,13 @@
+Test(s) for Coq backend.
+
+This is not yet covered by into ktest.
```
dwightguth, line 3: can you make it into a junit integration test actually?

src/javasources/KTool/src/org/kframework/backend/coq/CoqBackend.java
```diff
@@ -0,0 +1,112 @@
...
+                kem.registerCriticalError("Error generating Coq rules definition");
+                return;
+            }
+        } catch (IOException e) {
+            kem.registerCriticalError("Error generating Coq rules definition", e);
+        }
+    }
+
+    private int waitFor(Process p) {
+        boolean interrupted = false;
```
dwightguth: this makes the child process immune to ^C which is probably not what we want

src/javasources/KTool/src/org/kframework/kompile/KompileOptions.java
bmmoore, line 56: I set the java flag because I'm aiming to support features closer to the Java backend than the Maude backend, and the flag seems to be used in various places to type things as KItem rather than K and so on.
The last flag is set to false because I don't make a `-kompiled` directory, I just put the generated `.v` files next to the `.k` files. I wouldn't terribly mind changing it, but I think it makes sense not to hide them away because you will presumably have other Coq files proving things based on the definition, and also for now you need to manually supply a bit of scaffolding to go with the generated files.

dwightguth, line 56: okay sounds good

tests/regression/coq-backend/README
bmmoore, line 3: As you pointed out, we'd have to commit the binary.

PR #813, opened 2014-08-09
added some kast syntax: variables, configurations, basic terms

grosu:


dwightguth:
@grosu was this reviewed? I don't see immediate signs of it.

grosu:
Note that this is in a small and experimental branch, kast-in-K.  Radu and I are working on it only.  I don't think we need to review these branches, do we?  The only reason I made a pull request and then merged instead of pushing directly is because I had this problem with git that I told you about, which did not allow me to push to kframework.

PR #818, opened 2014-08-11
Adding the 'noAutoReject' tag for Tokens. Also:

radumereuta:
- fixing the check for empty productions to also consider the combination of notInRules and notInGround tags
- adding back the <delete> line from ant clean that removes the /basic directory from the code to avoid cleanup issues in git.

@grosu @dwightguth please review.


dwightguth:
I don't like having some special attribute that changes the semantics of reject. We have enough issues with special attributes that are undocumented and do some minor thing. Why was this done?

radumereuta:
It was added because it is needed.
@grosu started to work on kast, and found that there are cases where you just don't want the auto reject. For example in the outer parser, I need to define the bubbles for rules. The lexer rule matches on anything that isn't whitespace, therefore will reject every single keyword in K. Besides "rule", "syntax", "configuration", "context" and "endmodule", it would reject also "require", "module", "imports", and "true" and "false" which were automatically imported. So Grigore and I decided that we need something to modify that default behavior, and we had two options for the default behavior:
1. automatically reject all terminals that could match with the current token declaration (just like we have now)
2. reject all terminals, only if a special attribute/syntax is specified by the user.
The second one involved changing a lot of definitions, so I preferred to implement something that is backwards compatible.

Note that I'm changing the semantics of Token, not reject. Placing the attribute there is correct.

dwightguth:
Okay so the issue seems to be that we have no way to specify an infinite
set of tokens /not/ to reject, because the reject semantics kicks in
recursively. This is surprisingly tricky. Can you please add a note to the
wiki documentation to be processed in the manual?

radumereuta:
@dwightguth  Wiki updated.

grosu:
Just tried this and got this error:

```
[Error] Compiler: Cannot declare empty terminals in the definition.
            Use attribute 'onlyLabel' or 'notInRules' and 'notInGround' paired
with 'klabel(...)' to limit the use to programs.
        Source: File: D:\github.grosu\k\samples\kast\kast.k
        Location: (96,25,96,38)
```

The error is great, but doesn't the text message exceeds 79 characters?  Can we do something to avoid having this problem over and over and over again?

grosu:
Besides, 'notInRules' does not work.  I got the error above because I didn't have a klabel attribute.  Once I added the klabel attribute, I still the error with 'notInRules' but I get no error with 'onlyLabel'.

Why do we have two different attributes that do the same thing, anyway?

Issue #825, opened 2014-08-11
Java backend cannot union two sets


liyili2:
If I have the following set of rules:

```
module TEST


     syntax KItem ::= abc(Set,Set) | final(Set) | "start"

     rule abc(S:Set,S':Set) => final(S S')

     rule start => abc(SetItem(1) SetItem(2),SetItem(2))

endmodule

```

If I run the program "start", I will get the following result:

```
<k>
    final ( S:Set S':Set )
</k>

```

instead of the correct result:

```
<k>
    final ( SetItem ( 1 ) SetItem ( 2 ) )
</k>

```

yilongli:
fixed via https://github.com/kframework/k/pull/826

PR #826, opened 2014-08-11
fixing issue 825

yilongli:
fixing https://github.com/kframework/k/issues/825


yilongli:

andreistefanescu:
Looks good. Please add the example provided by @liyili2 as a unit test.

yilongli:
But the error shown in the example is far from the root cause, which is the visitor implementation. I don't think including that example would not be very useful in the future. Instead, if we always use `baseTerms()` instead of `collectionPatterns()/Functions()/Variables()` separately, we won't likely have this kind of issue anymore.

dwightguth:
@andreistefanescu I'm persuaded by Yilong's logic. Are you okay if I merge this?

andreistefanescu:
agreed

PR #828, opened 2014-08-12
added kernelc semantics

andreistefanescu:


andreistefanescu:
@grosu @yilongli please review. I did not have time to add comments today. I can do that tomorrow morning, if you guys think it is needed.

samples/kernelc/examples/01 undefined/1 division_by_zero.c
```diff
@@ -0,0 +1,19 @@
+/*
```
grosu: In the tutorial, we called this folder `tests` instead of `examples`, and we had a `ktest` file `config.xml` there.  I would do the same here, for consistency.

samples/kernelc/examples/06 call_stack/1 only_g_calls_f.c
```diff
@@ -0,0 +1,43 @@
...
+  g();
+}
+
+int main()
+{
+  h();
+}
+
+
+//@ var S : ListItem
```
grosu: I hope you will not need these var declarations anymore.

samples/kernelc/kernelc-semantics.k
```diff
@@ -0,0 +1,397 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
+
+module KERNELC-SEMANTICS
+  imports KERNELC-SYNTAX
+
+  /*
+   * Macros
+   */
+  rule if (E:Expression) S:Statement => if (E) S else ;     [macro]
```
grosu: These 6 macros would look nicer to be in the syntax module.

Actually, only 4 of them, those referring only to the KernelC syntax.

grosu: Shouldn't the RHS be `if (E) S else {}`  ?

grosu:
The definition looks good overall, so as far as the programs pass I am OK with it.

My only high-level comment would be to add brief Latex comments, so that the language looks good as a poster.  You can make the comments you have now into sections/subsections, and only add one sentence or two wherever you think something interesting is going on.  Also, one paragraph at the beginning explaining what this language is about would also be great.  And, of course, add a title and you as an author.  See simple and other languages in the tutorial for example.

Thanks!

yilongli:
lgtm

samples/kernelc/examples/06 call_stack/1 only_g_calls_f.c
andreistefanescu: We should not need this declarations anymore.

samples/kernelc/kernelc-semantics.k
andreistefanescu: both `;` and `{}` are skip statements in C, but `;` is faster

dwightguth:
You can merge it if Yilong and Grigore are happy.

dwightguth:
@andreistefanescu did you want to do more work on this pull request? If not go ahead and merge it.

PR #830, opened 2014-08-12
Prove simple list head

yilongli:
This will be the last example done in simple-untyped. I will start proving examples in KernelC.


yilongli:

dwightguth:
lgtm. @andreistefanescu merge if you're happy.

yilongli:
@andreistefanescu what do you think? Can I merge this?

samples/java_rewrite_engine/tutorial/2_languages/1_simple/1_untyped/verification/head_spec.k
```diff
@@ -0,0 +1,57 @@
...
+
+module HEAD-SPEC
+  imports SIMPLE-UNTYPED
+
+rule
+<T>
+    <threads>
+        <thread>
+            <k>
+               return (x [ 0 ]) ; => return Y:Int ;
```
andreistefanescu: This is correct, but I would prefer a rule similar to the one you wrote for sum, like `head(X:Val) => Y:Int`.

andreistefanescu:
Looks good overall. You can merge after changing the specification rule.

samples/java_rewrite_engine/tutorial/2_languages/1_simple/1_untyped/verification/head_spec.k
yilongli: Yeah, I can do that. But I decided to write it this way because it is closer to the MatchC specification style, namely `$ => return Y:Int`. Besides, since SIMPLE do not do garbage collection, I will have to write an additional `.Map => ?M:Map` in the `store` cell. 

Should I write the function specification in this function call style in the future?

andreistefanescu: Yes, I think the call stype is better than the one in matchc. Also kernelc garbage collects its variables, so the second issue won't matter.

andreistefanescu:
lgtm

PR #832, opened 2014-08-13
Fixing the 80 character per line issue.

radumereuta:
The error message now looks like this:

```
d:\work\k\samples\test>kompile test.k
[Error] Compiler: Cannot declare empty terminals in the definition.
        Use attribute 'onlyLabel', or both 'notInRules' and 'notInGround'
        paired with 'klabel(...)' to limit the use to programs.
        Source: File: d:\work\k\samples\test\test.k
        Location: (3,19,3,20)
```

@grosu please review.

### DIFF
--- a/src/javasources/KTool/src/org/kframework/compile/checks/CheckSyntaxDecl.java
+++ b/src/javasources/KTool/src/org/kframework/compile/checks/CheckSyntaxDecl.java
@@ -136,8 +136,8 @@ public class CheckSyntaxDecl extends BasicVisitor {
                     (node.containsAttribute("notInRules") && node.containsAttribute("notInGround")))
                     && node.containsAttribute("klabel"))) {
                 String msg = "Cannot declare empty terminals in the definition.\n";
-                msg += "            Use attribute 'onlyLabel' or 'notInRules' and 'notInGround' paired with 'klabel(...)' to limit the use to programs.";
-
+                msg += "        Use attribute 'onlyLabel', or both 'notInRules' and 'notInGround'\n";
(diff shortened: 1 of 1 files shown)


src/javasources/KTool/src/org/kframework/compile/checks/CheckSyntaxDecl.java
```diff
@@ -136,8 +136,8 @@ else if (!(pi instanceof Terminal)) {
                     (node.containsAttribute("notInRules") && node.containsAttribute("notInGround")))
                     && node.containsAttribute("klabel"))) {
                 String msg = "Cannot declare empty terminals in the definition.\n";
-                msg += "            Use attribute 'onlyLabel' or 'notInRules' and 'notInGround' paired with 'klabel(...)' to limit the use to programs.";
-
+                msg += "        Use attribute 'onlyLabel', or both 'notInRules' and 'notInGround'\n";
+                msg += "        paired with 'klabel(...)' to limit the use to programs.";
                 GlobalSettings.kem.registerCompilerError(msg, this, node);
```
grosu, line 141: Can we have this registerCompilerError check the length of the error message, and if larger to print an additional message telling the user to report the error to us?

dwightguth:
This is the wrong solution. KExceptionManager should be dealing with this already, you just have to trace it back to the splitLines method. I'm not going to merge this because it doesn't actually fix the problem in a reliable way for all cases.

radumereuta:
Wait a second. I had a rephrasing in there too, which made the error message easier to understand.
You should still merge this.

Issue #543, opened 2014-06-03
Cooling rules are not executed when LTL model checker is imported


dlucanu:
If the following module 

```
require "builtins/model-checker.k"
module TEST
  imports LTL-HOOKS
  imports MODEL-CHECKER-HOOKS
  syntax Exp ::= Int | Exp "+" Exp [strict] | "(" Exp ")" [bracket]
  syntax KResult ::= Int
  rule I1:Int + I2:Int => I1 +Int I2
  rule <k> .K => (X + 1) + 2 </k> <in> ListItem(X:Int) </in>
  configuration
     <k> .K </k>
     <in> $IN:List </in>
endmodule
```

is compiled with the symbolic backend, 

```
$ kompile test --backend symbolic
```

and then executed with a symbolic input, we see that the cooling rules are not applied:

```
$ krun -cIN "ListItem(#symInt(n))" -cPC true
<k>
    #symInt(n) +Int 1 ~> (HOLE + 2)
</k>
<path-condition>
    true
</path-condition>
<in>
    ListItem(#symInt(n))
</in>
```

If we comment all lines referring to the model-checker and compile again, then everything is ok:

```
$ krun -cIN "ListItem(#symInt(n))" -cPC true
<k>
    #symInt(n) +Int 1 +Int 2
</k>
<path-condition>
    true
</path-condition>
<in>
    ListItem(#symInt(n))
</in>
```

dlucanu:
@radumereuta Radu, I guess that the bug is related to parsing. Can you take a look? Unfortunately there are some examples that are not working and they are reported in papers.

PR #833, opened 2014-08-13
Fixing issue #543

andreiarusoaie:


andreiarusoaie:
@dlucanu @dwightguth please review.

dlucanu:
The modification looks good to me, but I think that we can use this opportunity and clean up the file symbolic-k.k. It has now many commented productions/rules, which in my opinion can be removed.

andreiarusoaie:
@dlucanu I agree we should cleanup, but I think  that should be done in a different pull request.

PR #839, opened 2014-08-14
fixed a few issues with kernelC; added examples with main functions

andreistefanescu:


yilongli:
lgtm

samples/kernelc/patterns/tree_pattern.k
```diff
@@ -0,0 +1,38 @@
...
+  rule tree_keys(leaf) => .Set
+  syntax Int ::= "tree_height" "(" Tree ")"   [function]
+  rule tree_height(node(I:Int, TL:Tree, TR:Tree))
+    => 1 +Int maxInt(tree_height(TL), tree_height(TR))
+  rule tree_height(leaf) => 0
+
+  syntax Map ::= "tree" "(" Pointer ")" "(" Tree ")"    [pattern(1)]
+  rule tree(object(P:Int))(node(I:Int, TL:Tree, TR:Tree))
+    => object(P) |-> (
+          value |-> tv(int, I)
```
yilongli, line 22: You might have noticed already, but I just found out that in the struct declaration of `treeNode` in kernelc programs, the field used to store value is actually declared as `val`.

PR #842, opened 2014-08-14
SMTLib translation

andreistefanescu:


andreistefanescu:
I made a mess of these changes and the changes for the kernelc examples. I will make a one pull request for both of them.

PR #855, opened 2014-08-18
add interface for k module system

dwightguth:
@cos please review


cos:
the copyright is missing -- I cannot fix it as it is in your account

cos:
otherwise, good to merge

cos:
can we have an interface more specific than `Module`? The `ServiceLoader` is nice and retrieves the classes based on their interfaces. I need to know which interfaces to look for, and there is no point in losing this typing info downstream.

PR #844, opened 2014-08-14
Remove code that calls KompileOptions.Backend.java()

dwightguth:
**please review**

PR #858, opened 2014-08-19
fixed the tricky non-deterministic tests

grosu:
But this is not completely done yet, as we need to fix the alpha equivalence problem for outputs with symbolic variables.  That is a separate problem, which we will address later.


dwightguth:
I had to make some more changes so I am moving this back to #844

PR #862, opened 2014-08-20
refactor java backend to use more dependency injection

dwightguth:
**please review**


dwightguth:
@cos @andreistefanescu @yilongli  please review

src/javasources/KTool/src/org/kframework/backend/java/builtins/RealBuiltinIOOperations.java
```diff
@@ -0,0 +1,157 @@
...
+import com.google.inject.Inject;
+
+import java.io.IOException;
+import java.nio.charset.CharacterCodingException;
+
+
+/**
+ * Table of {@code public static} methods for builtin IO operations.
+ */
+public class RealBuiltinIOOperations implements BuiltinIOOperations {
```
andreistefanescu: rename this class

src/javasources/KTool/src/org/kframework/backend/java/kil/Term.java
```diff
@@ -47,20 +43,20 @@ protected Term(Kind kind) {
...
-        KILtoBackendJavaKILTransformer transformer
-                = new KILtoBackendJavaKILTransformer(definition.context());
-        Term term = transformer.transformTerm(kilTerm, definition);
-
-        if (definition.context().javaExecutionOptions.indexingStats){
-            IndexingStatistics.kilTransformationStopWatch.stop();
-        }
-        return term;
-    }
+//    public static Term of(org.kframework.kil.Term kilTerm, Definition definition) {
```
andreistefanescu: Please do not leave code commented out. Remove it!

src/javasources/KTool/src/org/kframework/backend/java/symbolic/PatternMatcher.java
```diff
@@ -661,8 +661,8 @@ public void match(CellCollection cellCollection, Term pattern) {
         CellCollection otherCellCollection = (CellCollection) pattern;
 
         if (cellCollection.hasFrame()) {
-            assert !termContext.definition().context().javaExecutionOptions.concreteExecution() :
-                "the subject term should be ground in concrete execution";
+//            assert !termContext.definition().context().javaExecutionOption/*s.concreteExecution() :
```
andreistefanescu, line 663: why is this assertion commented out?

andreistefanescu:
This is a change of style, and you need to document it somewhere. Other than that, it looks good.

src/javasources/KTool/src/org/kframework/backend/java/symbolic/PatternMatcher.java
dwightguth, line 663: I commented out two assertions in this diff because they introduced test dependencies on production code that were otherwise not present. It seemed to me the better solution than to significantly expand the size of this diff in order to support the assertions. Once each of those two classes is constructed by the injector it should be trivial to uncomment the assertions.

dwightguth:
@andreistefanescu I created the following page that we can link to somewhere if we want: https://github.com/kframework/k/wiki/K-Framework-Architecture

Do you want something additional more specific to guice somewhere in the code? I'm hesitant to write a bunch of documentation about guice when I can refer people to its official documentation, but if you have thoughts about how it specifically applies to K that you want me to document somewhere I can.

src/javasources/KTool/src/org/kframework/backend/java/indexing/IndexingTable.java
```diff
@@ -31,8 +36,20 @@
...
+    public static class Data implements Serializable {
+        public final TopIndex TOP_INDEX = new TopIndex();
+        public final IndexingPair TOP_INDEXING_PAIR = new IndexingPair(TOP_INDEX, TOP_INDEX);
+        public final BottomIndex BOTTOM_INDEX = new BottomIndex();
+        public final IndexingPair BOTTOM_INDEXING_PAIR = new IndexingPair(BOTTOM_INDEX, BOTTOM_INDEX);
+    }
+
+    @Inject
+    public IndexingTable(Provider<Definition> definition, Data data) {
+        this.definition = definition.get();
```
cos, line 51: why is a `Provider<Definition>` needed here? I guess you will have to decide on a `Definition` when there is more than one loaded. But how that the provider help here considering the `Definition` is simply `.get()`?

dwightguth, line 51: This is a standard pattern for resolving circular dependencies: Definition
-> IndexingTable -> Definition in this case. If you have a suggestion for
how to completely eliminate the cycle, that would be better though.

cos:
a more general question regarding injections: how do I look for the place where a particular class is constructed? The best I could come up with so far is to search the entire project for the string `ClassName.class` but this is far from ideal.

cos:
If I want to prune unnecessary dependencies, how do I look for all the places where a particular object is injected?

dwightguth:
So I haven't had a chance to try it out yet but you might look into the
code available to visualize the object graph. I believe it may be possible
to use that code to get the same locator traces that are output when
provision fails. But moreover, most of the bindings currently are either
methods in classes extending AbstractModule which have the @Provides
annotation, or constructors annotated with the @Inject keyword.

dwightguth:
So I did a quick search and came up with:
https://code.google.com/p/guice-plugin/

I'm going to try it out tomorrow, see if it provides the binding tracing
functionality you're talking about

yilongli:
Sometimes I am a little worried about the performance impact of such big change to the fast execution part of the rewrite engine. Can you help me add a performance test using `collatz` in `dynamic-typed-kool` with, say, n = 200?

src/javasources/KTool/src/org/kframework/backend/java/symbolic/PatternMatcher.java
yilongli, line 663: make a TODO item which contains your explanation above?

yilongli:
lgtm. But please make TODO items for the two assertions you commented out. I would like to have them back eventually.

dwightguth:
@yilongli I tested and this change does not seem to have appreciable impact on the speed of the rewrite engine.  Using pattern matching on the program you name, performance before and after went from 23 seconds to 21 seconds. Using unification, performance went from 209 seconds to 215 seconds. Running the code a second time yielded consistent times, so if I had to guess, I'd say there is a slight slowdown associated with initialization, but that peak performance time is unaffected.

Do you want me to try to do performance tuning? I might be able to improve these numbers on symbolic execution if I dig deeper into them.

yilongli:
@dwightguth Thanks. I thought using pattern matching (which now uses the new algorithm under the hood) would be around 20x faster than using unification. I will look into this after the deadline. Anyway, It's not urgent to improve the performance of symbolic execution. I don't want to distract you from more important things.

yilongli:
You can merge after adding the TODO items for assertions you commented out.

PR #868, opened 2014-08-21
SMTLib translation for sort and function declaration and lemmas

andreistefanescu:


andreistefanescu:
@yilongli this should add support for the iterative versions of length and sum. To declare a function in the prelude, use the attribute `smtlib(<smtlib_id>)`. Sorts are automatically declared if the appear in any of the declared functions. Unfortunately, Seq is reserved in Z3, so we need to rename is to IntSeq.

yilongli:
Thanks, Andrei!

PR #875, opened 2014-08-25
Safely borrowing SDF lexical syntax into the new parser.

radumereuta:
@dwightguth please review.
Should be a small fix. Some regular expressions from SDF are not supported by Java, throwing an exception even though the user doesn't care about that. In that case, just give a silent warning, and the definition would still kompile safely.

### DIFF
--- a/src/javasources/KTool/src/org/kframework/parser/concrete2/KSyntax2GrammarStatesFilter.java
+++ b/src/javasources/KTool/src/org/kframework/parser/concrete2/KSyntax2GrammarStatesFilter.java
@@ -5,6 +5,7 @@ import java.util.HashSet;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
 
 import org.kframework.kil.Configuration;
 import org.kframework.kil.KSorts;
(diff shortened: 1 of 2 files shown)


src/javasources/KTool/src/org/kframework/parser/concrete2/KSyntax2GrammarStatesFilter.java
```diff
@@ -185,13 +187,24 @@ public Void visit(Production prd, Void _) {
...
+            Pattern p;
+            if (!prd.containsAttribute(Constants.REGEX)) {
+                // try to use the regular expression from SDF.
+                // If it's not compatible give a warning
+                try {
+                    p = Pattern.compile(lx.getLexicalRule());
+                } catch (PatternSyntaxException ex) {
+                    p = Pattern.compile("NoMatch");
+                    String msg = "Lexical pattern not compatible with the new parser.";
+                    GlobalSettings.kem.registerCompilerWarning(msg, lx);
```
dwightguth: pass the exception to this function

dwightguth:
If you fix my comment you can merge this

src/javasources/KTool/src/org/kframework/parser/concrete2/KSyntax2GrammarStatesFilter.java
radumereuta: Didn't know about that. Nice.

radumereuta:
This is so strange, why didn't it fail the first time?

radumereuta:
@dwightguth please re-review. I also found an improper warning message returned by SyntaxDeclCheck.

PR #877, opened 2014-08-25
Removing copying of attributes from parent to child when eliminating a node.

radumereuta:
Let's see what happens. Don't review yet.


dwightguth:
This isn't the change I told you to make... I asked you to move it to CopyOnWriteTransformer, not delete it...

radumereuta:
I'm not done. I'm trying to understand what you've done with the new visitors. I've seen some inconsistencies regarding the attributed casts. I will follow up on this later today.

radumereuta:
Ok, I give up. I will make an issue with the things I found that are wrong and wait for @dwightguth  to fix it.

PR #882, opened 2014-08-26
Adding Constant class and ProductionReference as a superclass of TermCons

radumereuta:
This allows for better error reporting in the new parser when having ambiguities.
It will also allow for more simple disambiguation filters.

@dwightguth please review. We talked about this change a while back, and now I got to it.
Basically now the new parser outputs only two types of objects: TermCons and Constant, bot having a reference to the production from which they come from.

### DIFF
--- a/src/javasources/KTool/src/org/kframework/compile/transformers/FlattenTerms.java
+++ b/src/javasources/KTool/src/org/kframework/compile/transformers/FlattenTerms.java
@@ -6,6 +6,7 @@ import org.kframework.compile.utils.MaudeHelper;
 import org.kframework.compile.utils.MetaK;
 import org.kframework.kil.*;
 import org.kframework.kil.Collection;
+import org.kframework.kil.loader.Constants;
 import org.kframework.kil.loader.Context;
 import org.kframework.kil.visitors.CopyOnWriteTransformer;
 
@@ -47,13 +48,29 @@ public class FlattenTerms extends CopyOnWriteTransformer {
 
     /**
      * Flattens this TermCons if it has sort K, KItem, or any sort other than
-     * those defined in {@link org.kframework.kil.KSort}.
+     * those defined in {@link org.kframework.kil.Sort}.
      */
     @Override
-    public ASTNode visit(TermCons tc, Void _)  {
+    public ASTNode visit(TermCons tc, Void _) {
         if (tc.getSort().isComputationSort())
             return kTrans.visitNode(tc);
         return super.visit(tc, _);
+    }
+
+    @Override
+    public ASTNode visit(Constant constant, Void _) {
+        Term rez;
+        if (constant.getSort().equals(Sort.KLIST)) {
+            rez = new KLabelConstant(constant.getValue());
+        } else {
+            // builtin token or lexical token
+            rez =  Token.kAppOf(constant.getSort(), constant.getValue());
+        }
+        rez.setLocation(constant.getLocation());
+        rez.setSource(constant.getSource());
+        rez.copyAttributesFrom(constant);
+
+        return this.visitNode(rez);
     }
 
     class FlattenKSyntax extends CopyOnWriteTransformer {
--- a/src/javasources/KTool/src/org/kframework/compile/transformers/FlattenTerms.java
+++ b/src/javasources/KTool/src/org/kframework/compile/transformers/FlattenTerms.java
@@ -59,18 +59,7 @@ public class FlattenTerms extends CopyOnWriteTransformer {
 
     @Override
     public ASTNode visit(Constant constant, Void _) {
-        Term rez;
-        if (constant.getSort().equals(Sort.KLIST)) {
-            rez = new KLabelConstant(constant.getValue());
-        } else {
-            // builtin token or lexical token
-            rez =  Token.kAppOf(constant.getSort(), constant.getValue());
-        }
-        rez.setLocation(constant.getLocation());
-        rez.setSource(constant.getSource());
-        rez.copyAttributesFrom(constant);
-
-        return this.visitNode(rez);
+        return kTrans.visitNode(constant);
     }
 
     class FlattenKSyntax extends CopyOnWriteTransformer {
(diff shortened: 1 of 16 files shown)


radumereuta:
Tests passed, ready to review.

src/javasources/KTool/src/org/kframework/compile/transformers/FlattenTerms.java
```diff
@@ -82,6 +88,22 @@ public ASTNode visit(Freezer node, Void _)  {
         }
 
         @Override
+        public ASTNode visit(Constant constant, Void _) {
+            Term rez;
+            if (constant.getSort().equals(Sort.KLIST)) {
+                rez = new KLabelConstant(constant.getValue());
```
dwightguth, line 94: what the heck is going on here? since when are klabel constants of sort KList??

src/javasources/KTool/src/org/kframework/kil/AbstractVisitor.java
```diff
@@ -766,9 +766,19 @@ public R visit(Rewrite node, P p) throws E {
     }
 
     @Override
+    public R visit(Constant node, P p) throws E {
+        return visit((Term) node, p);
```
dwightguth: this is wrong. In AbstractVisitor, you need to call visit on the parent class. So ProductionReference for Constant and TermCons

src/javasources/KTool/src/org/kframework/kil/Constant.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.Map;
+
+/**
+ * The AST representation of tokens. This should be used only in the front end,
+ * and should be flattened into KIL classes after FlattenSyntax.
+ */
+public class Constant extends ProductionReference {
+    protected String value;
+
+    public Constant(Sort sort, String value) {
```
dwightguth, line 29: I'm not comfortable with this constructor. It should be an absolute that all ProductionReferences have a non-null production

src/javasources/KTool/src/org/kframework/kil/Constant.java
```diff
@@ -0,0 +1,79 @@
...
+    public String getValue() {
+        return value;
+    }
+
+    public void setValue(String value) {
+        this.value = value;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
dwightguth, line 80: please autogenerate these two methods. They are not correct in this form and there are many many cases to consider.

dwightguth:
I suggest you move the code that flattens Constant into NormalizeASTTransformer, that way it does not persist throughout the initial stages of compilation which are currently implemented wrongly over concrete syntax.

src/javasources/KTool/src/org/kframework/kil/Constant.java
radumereuta, line 29: Well... I think this is open to discussion.
1. I want to try and add this to the SDF parser, and do the disambiguation on Constants and TermConses alone. The other thing is that for each Terminal in a grammar, a Constant will be produced. SDF doesn't give me the possibility to do so.
2. Ex: for the production `Exp ::= Exp "+" Exp`, the new parser will generate a TermCons node with two other TermConses, and a Constant for `+`. This will not be present in the final AST because I'm deleting it pretty quickly. In this case it doesn't make too much sense to make the constant have a reference to a Production. This field should be viewed as metadata useful for error reporting.

src/javasources/KTool/src/org/kframework/kil/Constant.java
radumereuta, line 80: This is autogenerated, and I specifically want to test to see if only `value` and `sort` are the same. I'm not interested in the Production. That's metadata.

src/javasources/KTool/src/org/kframework/kil/Constant.java
dwightguth, line 29: So the entire reason you don't want all Constants to have a Production associated with them is because you want terminals to be Constants? Sorry, this just doesn't feel clean to me.Why create an object at all for terminals if they're going to get immediately deleted? This isn't the first time we've special cased this one nasty hack, and it won't be the last. Better for a constant to be a constant, period. Don't use one class to represent two different things.

src/javasources/KTool/src/org/kframework/kil/Constant.java
dwightguth, line 80: okay fine, autogenerate it with something else then, because that's not a valid equals method.

src/javasources/KTool/src/org/kframework/kil/Constant.java
radumereuta, line 29: They are actually the same thing. The parser has two main objects: NonTerminalState, and a PrimitiveState. The first one calls another NonTerminal, the second one matches on something from the text. Doesn't matter if it is a terminal from the middle of the production, or if it is a lexical production.

src/javasources/KTool/src/org/kframework/kil/Constant.java
radumereuta, line 80: Can you be more specific?
I don't see anything wrong with the way IntelliJ generated them.
Note that I'm expecting `sort` and `value` to not be null. I'm fine with getting a NullPointerException. It means it is something wrong in some other part of the code.

src/javasources/KTool/src/org/kframework/kil/Constant.java
dwightguth, line 29: Oh right, I remember, the parser does some funky stuff internally with the Token and KList classes. Okay. I am persuaded that it makes sense to go this route for now. Please make sure it is clear that developers should never call this constructor from outside the parser kernel, then.

radumereuta, line 29: With that I agree. Is there any way to enforce this?

src/javasources/KTool/src/org/kframework/kil/Constant.java
dwightguth, line 80: Okay. This is technically correct I believe if sort and value are non null. But if want them to be non-null then you need to throw a NullPointerException from the constructor if you try to construct with null arguments.

src/javasources/KTool/src/org/kframework/kil/Constant.java
dwightguth, line 29: Okay it looks like the parser kernel doesn't actually call this constructor. It calls the 3-arg constructor with a null production. Maybe you can delete this constructor and explicitly pass null to the constructor calls where null is valid? Then at least the developer would be forced to think what they should pass as the production, and decide for themselves if it should be null. That's probably the best you can enforce. Beyond that you just have to add a comment to the constructor explaining the circumstances under which it's allowed to be null.

src/javasources/KTool/src/org/kframework/kil/Constant.java
radumereuta, line 80: When we switch to Java8, we should have a look at something like: https://blogs.oracle.com/java-platform-group/entry/java_8_s_new_type
Might save us a lot of trouble if we annotate things that we can expect them to be null or not.

dwightguth:
Closing. Please reopen on master

PR #831, opened 2014-08-12
Maven

dwightguth:
This is going to fail on Jenkins, so we'll deal with Jenkins separately, right now I just want feedback on the diff. I suggest you review the two commits separately as they are logically isolated. The only exception is pom.xml.

PR #889, opened 2014-08-27
Maven

dwightguth:
**please review**


dwightguth:
See #831.

dwightguth:
@radumereuta does this new README provide all the information you need, or am I missing something?

radumereuta:
Comments regarding README for develpers:
- "You can test if it works by calling `ant` in a Terminal."  -> I'm not sure we're using `ant` anymore.
- Visual C++ 2012 Redistributable Package - say why we require this. Because z3 fails. If someone has this problem, could look this up and recognize that he/she is missing this.
- MAVEN_OPTS - what is with the backspace? Does the variable really have a backspace? Ah, I think it's because of .md. Hmm... but I'm not using a md visualizer. can you think of something that will make it more clear?
- You can delete the Spoofax related instructions.
- is `mvn verify` something standard? We had `ant test` before. If that is the Maven convention then this is fine, otherwise, I would like to keep the name.
- "artifact" - is that correctly spelled? My editor says that it isn't.
- "compiation" - typo
- "This should not normally be necessary, however." - I think this could be rephrased to make it simpler to understand.
- Could you also add instructions for how to generate the project for IntelliJ? There are quite a few developers right now which are using it.
- How do I run only the junit tests? How do I run checkstyle?

I think it's enough for now.

dwightguth:
The remaining of your comments (mvn verify, artifact) were correct as is.

radumereuta:
Looks better. You should bug some more people to have a look and test everything.

dwightguth:
I will be merging this Friday afternoon if nobody raises any more issues. @traiansf I fixed the issue on windows. Anyone else who might be worried that this change will break there stuff should test it now!

radumereuta:
Before that, I think more people should test this, and look over the READMEs and comment if something doesn't look right. @grosu you should definitely have a say in this. It will affect the way you teach, and the way you present the tool to new users.

dwightguth:
I agree that more people should test this, but I can't afford to wait
around and badger every last person into doing so. Setting a fixed window
leaves the responsibility on others if it is insufficiently tested and
allows us to make progress on something that is already taking far too long.

grosu:
I agree.  Let's switch to it for good.  Then we fix the problems :)  I'll definitely do my best to try it out before Friday and provide feedback.  Anybody who cannot afford to possibly have to interrupt what they are doing for a week or so starting with Friday should do the same.

Grigore

owolabileg:
I tested this and it seems to work well for the most parts. My only problem
so far is that pdf generation always fails on my machine with the error
below. Note that pdf generation works well for me on master and on older
versions of K, but not when I text this "dwightguth maven" branch.  Did I
miss something? Note too that the suggestion in the error message is not so
clear (i.e. where may I find such a latex version of the definition?).

Minors:
1. should we make it explicit in the README that "mvn test" still works,
   but will only run the unit tests (i.e.) no integration tests?
2. developers may want to know exactly what the integration test in K do.
   Maybe we should say in the documentation. Note that I am not talking about
   the idea of integration test in general, but the nature of the intergration
   tests in K.

Thanks,
Owolabi

---

 [exec] ERROR: ['/home/owolabi/projects/tmp/k/target/release/k/bin/kompile'
'--backend' 'pdf'
'/home/owolabi/projects/tmp/k/tutorial/1_k/2_imp/lesson_5/imp.k'] failed
with error (time: 7631 ms)
     [exec] error was: [Error] Critical: Cannot generate the pdf version of
the definition. It seems
     [exec] that `pdflatex` is not installed or is not in your path. To
generate the pdf
     [exec] version you can run `pdflatex` having as argument the latex
version of the
     [exec] definition.

osa1:
> Note too that the suggestion in the error message is not so
> clear (i.e. where may I find such a latex version of the definition?).

I'm responsible for that error message and I admit that it's horrible for several reasons:
1. A LaTeX file may not be generated at all.
2. We actually have very little ideas about whether it's because `pdflatex` is missing or because of something else. If `pdflatex` is missing for real then this message is still printed though :)
3. This is printed as a result of an `IOException` and we don't print `e.getMessage` even when in debug mode. We may need to print that all the time because otherwise it's very hard to guess what went wrong. (although we can do it to some extent -- for example we can check if there's an executable named `pdflatex` before printing an error message etc.)

I'll open an issue about this.

Note that if `--backend latex` is used, then LaTeX file is generated in `--directory` (it's directory of definition by default). If PDF backend is used and a LaTeX is really generated, then it's generated in `.k`, which is under `--directory`.

osa1:
What happened to executable scripts? Are we going to run it using `k3.jar` from now on?

I had a problem with installation. On my system I can't generate `k3.jar` because `test` step is failing:

http://lpaste.net/110791

Since it doesn't generate `k3.jar`, I tried to run it on generated class files but failed that too:

```
➜  classes git:(maven) java -cp . org.kframework.main.Main 
Exception in thread "main" java.lang.NoClassDefFoundError: com/google/inject/ProvisionException
        at java.lang.Class.getDeclaredMethods0(Native Method)
        at java.lang.Class.privateGetDeclaredMethods(Class.java:2688)
        at java.lang.Class.getMethod0(Class.java:2937)
        at java.lang.Class.getMethod(Class.java:1771)
        at sun.launcher.LauncherHelper.validateMainClass(LauncherHelper.java:544)
        at sun.launcher.LauncherHelper.checkAndLoadMain(LauncherHelper.java:526)
Caused by: java.lang.ClassNotFoundException: com.google.inject.ProvisionException
        at java.net.URLClassLoader$1.run(URLClassLoader.java:372)
        at java.net.URLClassLoader$1.run(URLClassLoader.java:361)
        at java.security.AccessController.doPrivileged(Native Method)
        at java.net.URLClassLoader.findClass(URLClassLoader.java:360)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
        at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:308)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
        ... 6 more
```

Maybe we should document how to run it when `k3.jar` generation is failed(probably because of a test failure).

dwightguth:
Omer: as per the developer readme, the executable scripts are in
target/release/k/bin. I will add documentation explaining how to skip unit
tests. I will also look into the test failure you provided.

bmmoore:
The only thing I dislike about this branch is losing the ability to make a single distribution that works on all platforms. That's not deeply tied to moving to Maven, it's actually that making platform-specific releases would have been too hard with ant.

dwightguth:
I'm pulling the trigger on this, since it's Friday and nobody else has reported that something is broken. If something breaks, please IM me and I will make fixing it my top priority.

PR #890, opened 2014-08-27
strongly type attributes

dwightguth:
**please review**


dwightguth:
Jenkins: test this please

dwightguth:
@cos please review

src/javasources/KTool/src/org/kframework/backend/html/HTMLFilter.java
```diff
@@ -503,9 +503,9 @@ public Void visit(Attribute entry, Void _) {
             return null;
 
         // The latex and/or html attributes are processed in the HTMLPatternVisitor, not here.
-        if (entry.getKey().equals("latex"))
+        if (entry.getKey().equals(Attribute.keyOf("latex")))
             return null;
-        if (entry.getKey().equals("html")) {
+        if (entry.getKey().equals(Attribute.keyOf("html"))) {
```
cos: the two `if`s can be merged with an `||`

also, why does this return `null`. I guess `null` encodes not-visited, but it seems ugly. We'll probably get to it along with the AST refactoring.

src/javasources/KTool/src/org/kframework/kil/Attribute.java
```diff
@@ -69,28 +78,48 @@ public Attribute(Attribute attribute) {
...
+            return "@" + key.getTypeLiteral().toString();
+        }
+    }
+
+    public static String toString(Annotation annotation) {
+        if (annotation == null) return null;
+        if (annotation instanceof Named) {
+            return ((Named)annotation).value();
+        }
+        return annotation.toString();
```
cos: shouldn't `toString` be a non-static method on `Key` and `Annotation`?

src/javasources/KTool/src/org/kframework/kil/ASTNode.java
```diff
@@ -165,6 +171,18 @@ public boolean containsAttribute(String key) {
         if (attributes == null)
```
cos, line 171: `hasAttribute` sounds nicer than `containsAttribute`

cos:
I would really like to see `Attribute` disappear as a class. It is redundant, leading to stuff like `Map<Key, Attribute>`, where `Attribute` also has-a `Key`. Again, we'll probably (slowly) get to it along with refactoring KAST.

src/javasources/KTool/src/org/kframework/kil/Attribute.java
dwightguth: Yes but they're not my classes. I am implementing a version of toString specific to printing attributes, that's all.

src/javasources/KTool/src/org/kframework/kil/ASTNode.java
dwightguth, line 171: I can make a separate pull request for this if you want, but I don't really want to mix together large-scale refactorings with local modifications if I can help it.

src/javasources/KTool/src/org/kframework/backend/html/HTMLFilter.java
dwightguth: it returns null because the return type of the method is Void and java does not autobox void types. I did this to prevent the need for having separate Visitor and Transformer interfaces. This is quite a common code pattern for visitors and is used in the javac api as well.

src/javasources/KTool/src/org/kframework/kil/ASTNode.java
cos, line 171: ok

src/javasources/KTool/src/org/kframework/kil/Attribute.java
cos: I really don't like this. The rest of the code becomes littered with `Attribute.toString(attribute.getKey())` just because `Key` happens to be implemented by a third-party class. There must be a cleaner solution. Our interfaces shouldn't be spoiled by our choice for the underlying classes. Can you subclass `Key`? Otherwise, maybe wrap it in our own `Key` class. Worst case, a `keyString()` method on `Attribute` will do.

dwightguth: Okay. I will subclass Key.

cos:
great, this also got rid of the static from the serialization methods.

cos:
looks good to merge

dwightguth:
Jenkins: test this please

dwightguth:
Not really sure what's going on with this because I can't reproduce the failures this build describes. If it fails a second time, I'm going to try modifying ktest to get better debugging output on the command line.

PR #892, opened 2014-08-27
fix race condition on kompiled directory

dwightguth:

### DIFF
--- a/samples/kernelc/tests/binary_search_tree/config.xml
+++ b/samples/kernelc/tests/binary_search_tree/config.xml
@@ -8,6 +8,7 @@
         extension="c"
         >
     <kompile-option name="--backend" value="java" />
+    <kompile-option name="-d" value="binary_search_tree" />
     <program name="find.c">
       <krun-option name="--prove" value="tests/binary_search_tree/find_spec.k" />
     </program>
--- a/samples/kernelc/tests/list/config.xml
+++ b/samples/kernelc/tests/list/config.xml
@@ -9,6 +9,7 @@
             extension="c"
             >
         <kompile-option name="--backend" value="java" />
+        <kompile-option name="-d" value="list" />
         <program name="head.c">
           <krun-option name="--prove" value="tests/list/head_spec.k" />
         </program>
--- a/samples/kernelc/tests/simple/config.xml
+++ b/samples/kernelc/tests/simple/config.xml
@@ -9,6 +9,7 @@
             extension="c"
             >
         <kompile-option name="--backend" value="java" />
+        <kompile-option name="-d" value="simple" />
         <program name="average.c">
           <krun-option name="--prove" value="tests/simple/average_spec.k" />
         </program>
--- a/samples/kernelc/tests/tree/config.xml
+++ b/samples/kernelc/tests/tree/config.xml
@@ -9,6 +9,7 @@
             extension="c"
             >
         <kompile-option name="--backend" value="java" />
+        <kompile-option name="-d" value="tree" />
         <program name="height.c">
           <krun-option name="--prove" value="tests/tree/height_spec.k" />
         </program>


andreistefanescu:
The fix looks good, but I think this is a limitation of ktest. It should not compile the same definition twice.

dwightguth:
I agree. I will ask @osa1 if he wants to fix this, but for now I'm going to merge this, if that's okay.

osa1:
How can I reproduce this problem? IIRC ktest should refuse to run with a helpful error message when some definitions will be compiled multiple times to same directory.

dwightguth:
That does not seem to be the case. Refusing to run is also not the best
behavior. It would be better if you detected identical command strings and
only ran one of them. That way users can specify their tests across
multiple files even for a single definition.

grosu:
One problem which I often face with ktest is that I do not know what commands it is going to run.  The --dry option "almost" shows a list of commands, but they are more complicated than needed, including also the overriding of the .out files.  It would be really nice to have a sort of --dry option (maybe "--show-commands"?), which should simply list all the kompile and krun commands that will be executed, in a way that one can event copy and paste them and thus execute them manually.  I think will help both the user to better understand what ktest does, and us to better debug ktest.

Grigore

osa1:
What you asked is exactly what dry does -- printed commands are exactly
what krun executes. Grigore, what makes you think that it is runs different
commands? If that's really tbe case then  it means --dry has a bug and it
needs to be fixed. Also, what do you mean by "More complicated than
needed"? (Sorry for typos, sent from mobile)

grosu:
I think --dry shows commands of the form

```
krun ... > program.simple.out
```

that is, it also redirects the output of the executed program to the .out file.  I don't know whether this has been implemented or not, Daejun can tell better, but the idea was to not even generate certain such krun commands if the `.out` file is not needed.  Daejun?

Grigore

grosu:
Forgot to tell: I have to run now, which is why I didn't get a chance to test it myself before opening my mouth.  I will tonight.

Grigore

osa1:
@grosu 

I just checked and there's one problem with `--dry`: it generates commands with `> output_file` part even when `--generate-out` is not provided. Other than that it prints exactly what `ktest` runs. I'll fix this and it'll only print `> output_file` when `--generate-out` is provided. (IIRC this was working before, I don't remember why and when this is changed)

@dwightguth 

> Refusing to run is also not the best behavior. It would be better if you detected identical command strings and only ran one of them.

Running identical command strings only once doesn't really avoid compiling to same folder. Different `kompile` commands can generate same files. I'm open to "best behavior" advices.

grosu:
Excellent!

Thanks,
Grigore

dwightguth:
Ideally, the long term behavior is as follows:
- compile same definition into same directory: run command only once,
  execute both sets of programs on same definition
- compile different definitions into same directory: config error at startup
- compile same definition into different directories with different
  options: no need to specify -d to krun, ktest infers it automatically from
  the kompile option.

These are listed in decreasing order of priority, so do the first one
first, then the second, then the third.

daejunpark:
Now I remembered why the output redirection (i.e., `krun ... > program.simple.out`) always appears, as I reviewed @osa1's pull request #913. It appears to inform which output file is to be compared to check correctness of the current output. It was a kind of "meta" information. But I agree that now it is misleading, especially when we have `--generate-out`/`--update-out` options. I also agree with the new suggestion that the output redirection should appear only when the `--generate-out`/`--update-out` options are given. However, I think we should also consider the "meta" information; how to show the meta information? The thing is that we were often wondering why the output comparison failed, and sometimes it turned out that `ktest` attempted to compare a different output file from what we expected.

osa1:
@daejunpark can't you use `--update-out` to see what file is used for output comparisons?

daejunpark:
@osa1 Oh, you're right. Do you mean the combination of `--dry --update-out`? Yes, I think it will be sufficient. Then, I have no further concern. Thanks, Omer!

PR #894, opened 2014-08-27
Add Constant class.

radumereuta:
Continuation from #882 
Let's see what happens.

### DIFF
--- a/src/javasources/KTool/src/org/kframework/kil/Constant.java
+++ b/src/javasources/KTool/src/org/kframework/kil/Constant.java
@@ -17,15 +17,32 @@ import java.util.Map;
  * and should be flattened into KIL classes after FlattenSyntax.
  */
 public class Constant extends ProductionReference {
+
     protected String value;
 
+    /**
+     * Represents a lexical production from the parser.
+     * Should be used only in the 'parser' package.
+     * @param sort Sort of the production.
+     * @param value String representation of the constant.
+     */
     public Constant(Sort sort, String value) {
         super(sort, null);
+        assert value != null;
         this.value = value;
     }
 
+    /**
+     * Represents a lexical production from the parser.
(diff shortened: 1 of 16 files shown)


radumereuta:
Jenkins: test this please

src/javasources/KTool/src/org/kframework/kil/Constant.java
```diff
@@ -0,0 +1,96 @@
...
+
+    protected String value;
+
+    /**
+     * Represents a lexical production from the parser.
+     * Should be used only in the 'parser' package.
+     * @param sort Sort of the production.
+     * @param value String representation of the constant.
+     */
+    public Constant(Sort sort, String value) {
```
radumereuta: I'm using this in JUnit tests. The code won't look too good if I delete this constructor.

radumereuta:
Putting the code that flatten constants in `NormalizeAST` would be wrong. That's the job of `FlattenSyntax`.
At the moment the only place I'm creating Constants, is in the new parser, and I'm calling `FlattenSyntax` before the AST gets to the compilation stages.

radumereuta:
I think I've addressed all the comments.
@dwightguth please review.

src/javasources/KTool/src/org/kframework/kil/Constant.java
dwightguth: It's not generally a good idea for production code to depend on the details of your tests. I'm more concerned about developers calling this constructor by mistake and then getting a NullPointerException than I am about what tests look like. The purpose of a test is to execute cleanly and efficiently, and test the right functionality, not to look pretty.

dwightguth:
Okay. I am fine with FlattenSyntax then. We can always change it later if we run into problems.

radumereuta:
I removed the constructor.
@dwightguth please review.

Issue #893, opened 2014-08-27
Include timestamp in kompile and krun temporary directories.


radumereuta:
Right now, when I kompile something with the option --debug, a temporary directory called `krun[somethin unique]` is created in .k. Could you include the timestamp in the format yyyy-mm-dd-HH-mm-ss?

My current development practice involves multiple runs where a bunch of files gather there and I can't distinguish between them. The timestamp would help greatly.

@dwightguth I will assign this to you since you modified file creation the last time, but if you point me to where I can make such modifications myself, I could take are of it.

dwightguth:
I believe it's in FileUtil. Look for a reference to a guid. And make sure
to append the guid after the timestamp, because timestamps aren't unique
enough.

PR #895, opened 2014-08-28
Adding a timestamp to generated files in .k to make it easier to track the last version.

radumereuta:
@dwightguth please review.


radumereuta:
After merging, close #893.

PR #904, opened 2014-08-29
Making the list terminator incorporate the label instead of just the separator.

radumereuta:
@bmmoore please review.

### DIFF
--- a/src/javasources/KTool/src/org/kframework/compile/transformers/FlattenSyntax.java
+++ b/src/javasources/KTool/src/org/kframework/compile/transformers/FlattenSyntax.java
@@ -2,7 +2,6 @@
 package org.kframework.compile.transformers;
 
 import org.kframework.compile.utils.KilProperty;
-import org.kframework.compile.utils.MetaK;
 import org.kframework.kil.*;
 import org.kframework.kil.loader.Context;
 import org.kframework.kil.visitors.CopyOnWriteTransformer;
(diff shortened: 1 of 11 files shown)


src/javasources/KTool/src/org/kframework/compile/transformers/FlattenSyntax.java
```diff
@@ -32,13 +31,13 @@ public ASTNode visit(Definition node, Void _)  {
...
+        listUnits.clear();
         node = (Module) super.visit(node, _);
-        if (listSeparators.isEmpty())
+        if (listUnits.isEmpty())
             return node;
 
-        for (String sep : listSeparators) {
-            node.addConstant(Sort.KLABEL, MetaK.getListUnitLabel(sep));
+        for (String listUnit : listUnits) {
+            node.addConstant(Sort.KLABEL, listUnit);
```
dwightguth, line 40: what is happening here?

src/javasources/KTool/src/org/kframework/compile/transformers/FlattenTerms.java
```diff
@@ -96,7 +96,7 @@ public ASTNode visit(TermCons tc, Void _)  {
             }
             String label;
             if (tc.isListTerminator())
-                label = tc.getProduction().getListDecl().getTerminatorKLabel();
+                label = tc.getProduction().getTerminatorKLabel();
```
dwightguth, line 99: this doesn't feel like a good api to me. Stuff specific to user lists should be on the UserList class.

src/javasources/KTool/src/org/kframework/compile/transformers/FlattenSyntax.java
radumereuta, line 40: I just made a mechanical transformation, but I assume is needed to declare all the list terminators as KLabels to make maude parse it. Traian wrote this code, so if you are interested in knowing more, you should ask him.

src/javasources/KTool/src/org/kframework/compile/transformers/FlattenTerms.java
radumereuta, line 99: That's the problem I need information from the Production (namely the klabel). I just need to know that I have a UserList inside, but I'm not interested in any of it's information.
This is the right place to put it.

PR #912, opened 2014-08-31
[ktest] improve error reporting when config file has invalid format (also fixes a bug)

osa1:
See commit logs.


src/javasources/KTool/src/org/kframework/utils/StringUtil.java
```diff
@@ -529,18 +528,29 @@ public static String splitLines(String str, int col) {
         return builder.toString();
     }
 
-    private static String splitLine(String str, int col) {
-        if (str.length() < col)
+    private static String splitLine(String str, final int col) {
```
osa1, line 531: @dwightguth where to add unit tests for this function?

PR #913, opened 2014-08-31
--dry should not print commands with output redirections by default (see details inside)

osa1:
ktest's `--dry` was generating commands with `... > output_file` parts. That was wrong because ktest generates output files only when `--generate-out` or `--update-out` arguments are provided.

(Also discussed at #892)

With this patch, `--dry` only generates `... > output_file` parts when `--generate-out` and `--update-out` are provided, and printed paths are same as what ktest would use when that arguments are used.

@daejunpark Small patch but I still don't understand `--generate-out` in all details so it'd be nice if you could review.

### DIFF
--- a/src/javasources/KTool/src/org/kframework/ktest/Test/TestCase.java
+++ b/src/javasources/KTool/src/org/kframework/ktest/Test/TestCase.java
@@ -156,8 +156,8 @@ public class TestCase {
         List<String> stringArgs = new ArrayList<String>();
         stringArgs.add(ExecNames.getKompile());
         stringArgs.add(getDefinition());
-        for (int i = 0; i < kompileOpts.size(); i++) {
-            stringArgs.addAll(kompileOpts.get(i).toStringList());
+        for (PgmArg kompileOpt : kompileOpts) {
+            stringArgs.addAll(kompileOpt.toStringList());
         }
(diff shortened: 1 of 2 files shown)


daejunpark:
Thanks Omer! I think it looks good, but could you please also consider together the issues what I've just raised in https://github.com/kframework/k/pull/892#issuecomment-54079417. What do you think about that? I think you have a lot of experience of showing additional information of ktest (especially in verbose mode).

daejunpark:
Ignore the previous comment. I fully agree with this.

osa1:
Test is failed with this:

```
ERROR: ['/var/lib/jenkins/jobs/k-framework-pull-requests/workspace/k/bin/kompile' '/var/lib/jenkins/jobs/k-framework-pull-requests/workspace/k/samples/kernelc/kernelc.k' '--backend' 'java'] failed with error (time: 31701 ms)
error was: SSL_read_term_from_stream: null
rewriting failed, trace:
    importTbl_0_0
    ReadFromFile_0_0
[Error] Internal: Stratego rewriting failed
```

This is a backend related problem so this is safe to merge.

I can't change target branch of a pull request so I'll reopen this and my other pull requests, this time using master for target branch.

PR #915, opened 2014-09-01
Fixing several bugs in java backend

yilongli:
This pull request is separated from my work on removing redundant configuration cells. Several bugs were revealed after the optimization. It's interesting that none of the semantics in our regression test suite except simple-typed-static was able to catch the bug.


yilongli:
This pull request has been here for a while...

andreistefanescu:
lgtm

PR #918, opened 2014-09-01
#912 reopened for master branch

osa1:
#912

### DIFF
--- a/src/javasources/KTool/src/org/kframework/ktest/Config/ConfigFileParser.java
+++ b/src/javasources/KTool/src/org/kframework/ktest/Config/ConfigFileParser.java
@@ -10,6 +10,7 @@ import org.kframework.ktest.Test.ProgramProfile;
 import org.kframework.ktest.Test.TestCase;
 import org.kframework.utils.OS;
 import org.kframework.utils.file.KPaths;
+import org.kframework.utils.general.GlobalSettings;
 import org.w3c.dom.*;
 import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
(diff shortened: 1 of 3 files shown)


src/javasources/KTool/src/org/kframework/ktest/Config/ConfigFileParser.java
```diff
@@ -44,11 +44,18 @@ public ConfigFileParser(File configFile, KTestOptions cmdArgs) throws IOExceptio
...
         Source xmlFile = new StreamSource(configFile);
-        SchemaFactory schemaFactory = SchemaFactory.newInstance(XMLConstants
-                .W3C_XML_SCHEMA_NS_URI);
+        SchemaFactory schemaFactory = SchemaFactory.newInstance(XMLConstants.W3C_XML_SCHEMA_NS_URI);
         Schema schema = schemaFactory.newSchema(schemaFile);
         Validator validator = schema.newValidator();
-        validator.validate(xmlFile);
+        try {
+            validator.validate(xmlFile);
+        } catch (SAXException e) {
```
dwightguth, line 53: again, is this really a SAX exception now that you've recaught it? Also, this is a not very good error Message. The link might become fragile at any time, and we are likely to forget this is here. Say something like "refer to the K manual" instead.

dwightguth:
Add unit tests to src/javasources/KTool/test/org/kframework/utils/StringUtilTest.java

src/javasources/KTool/src/org/kframework/ktest/Config/ConfigFileParser.java
osa1, line 53: This is unrelated with the other SAX exception. This is thrown by the XML library when XML file is not in format defined in `ktest.xsd`.

I think even inside the XML library `SAXException` is overloaded for different purposes.

I'll edit error message.

dwightguth, line 53: Right, I agree that SAXException is being thrown by the XML parser and caught here. What I am wondering is if we actually have to rethrow it as a SAXException or if it's basically just a wrapper for a K error from this point upward.

osa1, line 53: Your call. Tell me if you want me to convert this to a K error.

dwightguth, line 53: Well I'm asking a domain-specific question here. Are you actually catching the exception thrown in this catch block before KTestFrontEnd? If the answer is yes, probably better to leave it as a SAXException. If the only thing you are doing that is catching the rethrown exception is to pass it to the KExceptionManager, probably better to use the kem here.

osa1, line 53: Answer is yes, `KTestFrontend` is always last to catch.

dwightguth, line 53: Okay. Well in general the guidance I would give is that if a checked exception should be considered completely fatal (ie you never recover from it but always crash with an error message), then you should always catch it and throw a KEMException rather than allowing it to propagate further up the call stack. Reserve rethrowing or propagating checked exceptions solely for cases when the code needs to catch exceptions of that specific type further up the call stack.

osa1:
@dwightguth I addressed your comments:
- Used KEM instead of propagating the exception to upper levels.
- Implemented unit tests for `splitLines`.
- Changed error message to "refer to the K manual".

dwightguth:
Merge if tests pass.

PR #920, opened 2014-09-01
[ktest] simplification and refactoring

osa1:
I finally made the changes that I had in mind for a while.
- Instead individually passing settings, `KTestOptions` object is passed between classes. This became a problem after we had more and more global settings.
- I removed string constants. Since they can't be used in annotations, most of them were left unused after moving to JCommander. For uniformity I removed the rest.
- Some various unused stuff removed.
- ~~Injected `KExceptionManager` is used instead of the one from `GlobalSettings`.~~

Just to make sure I'll run test suite on Windows manually.


src/javasources/KTool/src/org/kframework/ktest/CmdArgs/KTestOptions.java
```diff
@@ -36,11 +36,9 @@
  */
 public class KTestOptions {
 
-    public KTestOptions() {}
+    public KExceptionManager kem;
 
-    //TODO(dwightguth): remove in Guice 4.0
-    @Inject
-    public KTestOptions(Void v) {}
```
dwightguth, line 43: You can't do this. It will interfere with the ability to statically detect that the KTestOptions dependency is not satisfied. Please leave this constructor.

osa1, line 43: Okay. I removed this because of `remove in Guice 4.0` line. I'm putting it back.

src/javasources/KTool/src/org/kframework/ktest/Config/ConfigPreProcessor.java
```diff
@@ -122,8 +121,7 @@ private String resolveEnvVars(String str) {
                 return resolveEnvVars(m.replaceFirst(val));
             } else {
                 String msg = "The variable is not defined in the system environment: " + var;
-                GlobalSettings.kem.registerCriticalError(msg);
-                return null; // unreachable code
+                throw new SAXException(msg);
```
dwightguth: is this actually a SAX error? it seems wrong to me to throw this exception if it's not actually appropriate, unless there is some specific reason to catch this exception and do something SAX-specific.

src/javasources/KTool/src/org/kframework/ktest/CmdArgs/KTestOptions.java
```diff
@@ -36,11 +36,9 @@
  */
 public class KTestOptions {
 
-    public KTestOptions() {}
+    public KExceptionManager kem;
```
dwightguth: this is the wrong place to put a KExceptionManager. If you want access to it from a class that is not injected, use GlobalSettings.kem, or refactor the code to inject the class. If you want access to it from a class that is injected, you should inject it directly.

src/javasources/KTool/src/org/kframework/ktest/KTestFrontEnd.java
```diff
@@ -45,7 +45,6 @@
     }
 
     private final KTestOptions options;
-    private final KExceptionManager kem;
```
dwightguth, line 48: revert this. It is better to only store direct dependencies of a class.

src/javasources/KTool/src/org/kframework/ktest/Config/ConfigPreProcessor.java
osa1: While I'm not 100% sure, I think this may be considered a SAX error and we can re-use the class. From [the documentation](http://docs.oracle.com/javase/7/docs/api/org/xml/sax/SAXException.html):

> This class can contain basic error or warning information from either the XML parser or the application: a parser writer or application writer can subclass it to provide additional functionality.

Do you know any better exception class to use here? All I need is to pass a message -- I can't provide any other information.

src/javasources/KTool/src/org/kframework/ktest/CmdArgs/KTestOptions.java
osa1: Isn't `GlobalSettings.kem` deprecated? I can't use injection because I think `KTestOptions` is not injected. (I'm not sure if it's possible to inject to it, I'll check Guice documentation)

src/javasources/KTool/src/org/kframework/ktest/Config/ConfigPreProcessor.java
dwightguth: pass the message to what though? If you're passing the message just so it can be handled by the KExceptionManager, it's better to just throw a KEMException by calling registerCriticalError, which is what it was doing before. Am I misunderstanding the requirements here?

src/javasources/KTool/src/org/kframework/ktest/CmdArgs/KTestOptions.java
dwightguth: It's deprecated in the sense that we want to get away from it, but not in the sense that references to it should be removed blindly. Currently it should be continued to be used in places where dependency injection is not available yet.

osa1:
@dwightguth please review again, I addressed your comments.

PR #929, opened 2014-09-04
remove temp directory .k

### LINKED ISSUE #798: kv3.4---unnecessary .k folder
If I run `kompile` in some folder, without any arguments, it creates a `.k` folder.  On Unix that may be OK, because it is not visible, but on Windows that looks strange.

osa1:
Resolves #798.

I remember some files in `.k` were actually used for something useful(caching etc.) but apparently those files are now moved to `-kompiled` directory and `.k` can be removed without any side effects.


osa1:
~~Hm, apparently `.k` is still used for something useful. Some `--pdf`  steps are failed while some others are not. I'll try to find what's going wrong.~~

`.k` looks unused. @dwightguth were there any changes on the test server?

osa1:
Since #920 passed, I think there's a problem related with `.k`.

EDIT: I found the problem. Will write a follow-up.

PR #931, opened 2014-09-05
Maude list terminators from label also

bmmoore:
Change also the Maude name of a list terminator to be based off the label of the separator (or the label attribute of the production) instead of the concrete syntax, extending pull request #904.


radumereuta:
lgtm.
Didn't realize that maude did something different.

PR #932, opened 2014-09-05
fix ltrim function of string.k

daejunpark:
fix simple bug of `ltrim` function. please review @andreiarusoaie.


dwightguth:
Jenkins: add to whitelist

PR #944, opened 2014-09-11
Subsort fixes

bmmoore:
The subsort information in Context and the isSubsort method on Productions were designed for parsing purposes. In particular, they consider list elements to be subsorts of the list, and treat a production consisting of a single nonterminal as a subsorting even if it has a label. This makes a lot of sense if you think about concrete syntax, and what you are allowed to write at places where a particular sort is expected, but it's not the right subsort relationship to use in most of the compiler.

Overall, these commits rename the existing subsort related things to have "Syntactic" in the name, change the previous nicer names to give the correct semantic subsorting relationship, and change non-parser things to use the semantic relationships. This caused a few bugs where the last element in user lists were not wrapped into lists, which were mostly concealed (at least in the Java engine) by the sort checks accepting bare items in place of lists.

@radumereuta please review.

### DIFF
--- a/samples/kernelc/tests/execution/tests/undefined.c.out
+++ b/samples/kernelc/tests/execution/tests/undefined.c.out
@@ -1,6 +1,8 @@
 <config>
     <struct>
-        listNode |-> next |-> struct listNode * value |-> int
+        listNode |-> 
+            next |-> struct listNode *
+            value |-> int
     </struct>
     <fun>
         main |-> functionBody (  , int , {  { struct listNode * x ; x = (( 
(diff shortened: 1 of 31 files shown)


src/javasources/KTool/src/org/kframework/compile/transformers/AddEmptyLists.java
```diff
@@ -31,30 +21,81 @@ public AddEmptyLists(Context context) {
...
+        }
+        return null;
+    }
+
+    public ASTNode visit(Rewrite r, Void _) {
+        Term left = r.getLeft();
+        final Sort target;
+        if (left instanceof TermCons) {
+            Production p = ((TermCons) left).getProduction();
+            if (p.containsAttribute("function") || p.containsAttribute("predicate")) {
```
radumereuta: Is there a constant somewhere so you don't have to use strings?

src/javasources/KTool/src/org/kframework/kil/Module.java
```diff
@@ -71,6 +71,14 @@ public String toString() {
         return mkl;
     }
 
+    /**
+     * Return a list of the sorts declared by
+     * items in this module, plus a few builtin
+     * sorts.
+     * Result will contain duplicates if there
```
radumereuta: *are

radumereuta:
lgtm, but it would be good if @dwightguth also has a look.

dwightguth:
lgtm. You can merge if you fix Radu's comments.

bmmoore:
Rebasing onto maven...

PR #946, opened 2014-09-12
use debian x64 z3 binary because it has an older glibc

dwightguth:
@bmmoore please review


bmmoore:
Works for me (passes the kernelc --prove tests).

dwightguth:
Closing. I will fix this by modifying the maven repo.

bmmoore:
When will that be done, will I need to manually clear any cache, and how can I substitute a working version until then?

PR #947, opened 2014-09-12
update z3 to snapshot dependency (we may have further upgrades before ne...

dwightguth:
...xt release)

@bmmoore please review


bmmoore:
Seems good. I've only run the kernelc tests, but it seems to be the same version of the library that worked before - and certainly has no symbol versioning problems according to objdump/ldd. Also, if I understand maven correctly we can update the snapshot version of the files in the repository without further changes to files.

PR #948, opened 2014-09-12
first steps towards a generic transformation pipeline for krun

dwightguth:
**please review**


cos:
(@dwightguth, I'm traveling today so I'll only have intermittent internet access -- we can discuss here)

I think the critical aspect of this refactoring is to eliminate the propagation of program configuration through means other than the `Transformation` interface. Another way to say this is that the `Transformation` should be a pure function. 

```
public interface Transformation<P, R> {
  public R run(P p, Attributes attrs);
  public String getName();
}
```

Is the `attrs` parameter meant to do (part of) this? In this case we should probably also make the return of the `run` be a pair `(R, Attributes)`.

dwightguth:
Jenkins: test this please

dwightguth:
@cos please review

cos:
Some aspects of this refactoring may not be ideal:
- the `Transformation` interface is not yet there (see comment above)
- I am not sure that injection is the right mechanism for discovery of `Transformation`s
- if injection is the right mechanism, it may be possible to simplify it by removing the double indirection through `Provider`s. This may also solve `--debug` issue, right?
- the refactoring deletes 1.6k lines and adds 3.2k lines. This means that, overall, we increased the code base significantly. This is usually not a good sign, especially when we aim to simplify, unify interfaces. From our discussion, part of this is due to decoupling things into new classes. I think another part of it is from the complexity of the injection system. Overall, worrisome.

Having said all this, I do think this is a refactoring in the right direction, especially as many classes now conform to a single interface. We can fix the other problems later. From my perspective, you can merge. You might want a second opinion from someone who is more familiar with the original code. They could check details better.

dwightguth:
@radumereuta can you take a look? You're probably more familiar with krun than anyone else currently, although that isn't saying a lot.

src/main/java/org/kframework/kcheck/RLBackend.java
```diff
@@ -274,17 +269,17 @@ public void run(Definition javaDef) {
...
                         + ") ...");
                 // System.out.println("PGM: " + pgm);
-                KRunResult<SearchResults> result = mkr.search(null, null,
-                        SearchType.FINAL, defaultPattern, pgm,
-                        defaultPatternInfo);
-                System.out.println("Result: " + result + "\n\n");
-            } catch (KRunExecutionException e) {
-                e.printStackTrace();
-            }
+                //KRunResult<SearchResults> result = mkr.search(null, null,
```
radumereuta, line 276: Why keep the comments?

src/main/java/org/kframework/kil/Variable.java
```diff
@@ -101,7 +101,7 @@ public String fullName() {
     }
 
     public String toString() {
-        return name + ":" + sort + " ";
+        return name + ":" + sort;
```
radumereuta, line 104: Why?

src/main/java/org/kframework/kcheck/RLBackend.java
dwightguth, line 276: no particular reason, since this is all going to be deleted in the imminent future. I just am holding off on deleting code from Iasi entirely until we finish the modularization preparations, which this is a part of.

src/main/java/org/kframework/kil/Variable.java
dwightguth, line 104: Because this method is now being called by the code which handles the substitution of search results, and you get a nullpointerexception otherwise. I can probably revert it if you feel strongly about production code relying on the toString implementation though.

src/main/java/org/kframework/parser/ProgramLoader.java
```diff
@@ -115,7 +117,7 @@ public static Term processPgm(byte[] content, Source source, Sort startSymbol,
             }
             out = ((Rule) out).getBody();
         } else if (whatParser == ParserType.BINARY) {
-            try (ByteArrayInputStream in = new ByteArrayInputStream(content)) {
+            try (ByteArrayInputStream in = new ByteArrayInputStream(Base64.decode(content))) {
```
radumereuta, line 120: Why are you switching to Base64?

src/main/java/org/kframework/kil/Variable.java
radumereuta, line 104: The reason I added the space at the end, is because I very often use the toString representation of a term in order to visualize the AST. Adding the whitespace at the end prevents it from merging with other tokens.
Can't you use some other way of getting the desired representation? IDEs use by default the toString() function.

radumereuta:
Done looking over the diff. One thing that I saw as a common problem, is the complete lack of comments, everywhere. Please add more, especially since you, @dwightguth, are pretty much the only one that truly knows this part of the code. And, javadoc is not enough. Add single line comments from time to time, on loops (// iterating over these objects...), on long and complicated ifs/switches... And you should add comments to code that you think would be removed soon to. Just say //TODO: this should be removed because of this and such...

This is pretty much all I can say about this diff. I would have to spend a lot more time to check if it is actually doing the intended purpose.

src/main/java/org/kframework/parser/ProgramLoader.java
dwightguth, line 120: Because I wanted all parsing to happen over strings, and binary files are
not strings.

radumereuta, line 120: ok.

Issue #937, opened 2014-09-08
The `--debug` flag has no effect (i.e., no stack trace is displayed) when the exception happens before the flag's injection


cos:

radumereuta:
I've encountered this a few times, it's really annoying.

osa1:
:+1: I'm also suffering from this. Thankfully it's not very hard to modify the source temporarily to print messages all the time.

dwightguth:
I'm not sure it's possible to fix this issue without introducing behavior that is even more undesirable than the behavior we are trying to fix. I made a stab at it, but once I had an implementation, what I realized was, if the user has a typo while typing the options, then the options won't parse, including the --debug flag, which will cause a giant nasty stack dump to appear to users even when it's a perfectly ordinary user error. Since the debug flag is intended to be of use only to K developers, I believe it is better to provide good user experience to our users by disabling the debug flag until it is explicitly enabled, and taking the hit that some developers will sometimes need to modify the code in order to generate a stack dump.

dwightguth:
Closing this since I haven't heard any response from anyone about my answer to the issue of whether this is possible. Feel free to reopen if you can come up with a solution that does not have the issue I described.

cos:
I fully agree with trimming down the issues and making sure someone is assigned to each. Still, this does not mean we should close down issues which have no resolution yet, or there is nobody available to fix them. I think this issue is one of them. It is annoying so we should find a solution eventually.

PR #949, opened 2014-09-13
[WIP] plugin loading

osa1:
I'm creating this to get some comments and discuss what to do next. This is a very basic and crude implementation of plugin loader. I'll clean the implementation and squash commits before merging.

As an example, I created this kompile plugin: https://github.com/osa1/kompile-plugin. This is the output I get when I run the patched version of K:

```
➜  k git:(plugin-wip) ✗ ./target/release/k/bin/kompile  
found a kmodule
running my kompile frontend
➜  k git:(plugin-wip) ✗ 
```

Second line is printed by the plugin which works as kompile.

Now I have some questions to improve:

1) I'd like to know if I understand the use case correctly. @dwightguth can you please review https://github.com/osa1/kompile-plugin too just to make sure I'm not doing something wrong?
2) How should I get path of the JAR or the directory to search JARs in from user?
3) What should the framework do if there are multiple plugins for same K part? For example, multiple kompile implementations.
4) What should the framework do if there are multiple plugins for different K parts? For example, there may be plugins for kompile and krun. Should we load both? Should we allow user to choose what to load so that he/she can tell the framework to use only kompile plugin?

### DIFF
--- a/src/main/java/org/kframework/main/Main.java
+++ b/src/main/java/org/kframework/main/Main.java
@@ -2,8 +2,12 @@
 package org.kframework.main;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
+import java.util.ServiceLoader;
 
+import org.apache.commons.io.FilenameUtils;
 import org.fusesource.jansi.AnsiConsole;
 import org.kframework.kagreg.KagregFrontEnd;
 import org.kframework.kast.KastFrontEnd;
@@ -18,6 +22,7 @@ import com.google.inject.Injector;
 import com.google.inject.Module;
 import com.google.inject.ProvisionException;
 import com.google.inject.spi.Message;
+import org.kframework.utils.file.JarInfo;
 
 public class Main {
 
@@ -29,36 +34,71 @@ public class Main {
     public static void main(String[] args) {
         AnsiConsole.systemInstall();
 
-        Module[] modules;
+        KPluginClassLoader loader = new KPluginClassLoader();
+        loader.addPath(
+                FilenameUtils.concat(JarInfo.getKBase(false),
+                        FilenameUtils.concat("lib", "plugins")));
+
+        ServiceLoader<KModule> kLoader = ServiceLoader.load(KModule.class, loader);
+        List<KModule> kModules = new ArrayList<>();
+        for (KModule m : kLoader) {
+            kModules.add(m);
+        }
+
+        List<Module> modules = new ArrayList<>(0);
         if (args.length >= 1) {
             String[] args2 = Arrays.copyOfRange(args, 1, args.length);
                 switch (args[0]) {
                     case "-kompile":
-                        modules = KompileFrontEnd.getModules(args2);
+                        modules.addAll(Arrays.asList(KompileFrontEnd.getModules(args2)));
+                        for (KModule kModule : kModules) {
+                            List<Module> ms = kModule.getKompileModules();
+                            if (ms != null) {
+                                modules.addAll(ms);
+                            }
+                        }
                         break;
                     case "-kagreg":
-                        modules = KagregFrontEnd.getModules(args2);
+                        modules = Arrays.asList(KagregFrontEnd.getModules(args2));
                         break;
                     case "-kcheck":
                         assert false : "kcheck no longer supported";
                         return;
                     case "-ktest":
-                        modules = KTestFrontEnd.getModules(args2);
+                        modules.addAll(Arrays.asList(KTestFrontEnd.getModules(args2)));
+                        for (KModule kModule : kModules) {
+                            List<Module> ms = kModule.getKTestModules();
+                            if (ms != null) {
+                                modules.addAll(ms);
+                            }
+                        }
                         break;
                     case "-kast":
-                        modules = KastFrontEnd.getModules(args2);
+                        modules.addAll(Arrays.asList(KastFrontEnd.getModules(args2)));
+                        for (KModule kModule : kModules) {
+                            List<Module> ms = kModule.getKastModules();
+                            if (ms != null) {
+                                modules.addAll(ms);
+                            }
+                        }
                         break;
                     case "-krun":
-                        modules = KRunFrontEnd.getModules(args2);
+                        modules.addAll(Arrays.asList(KRunFrontEnd.getModules(args2)));
+                        for (KModule kModule : kModules) {
+                            List<Module> ms = kModule.getKRunModules();
+                            if (ms != null) {
+                                modules.addAll(ms);
+                            }
+                        }
                         break;
                     case "-kpp":
-                        modules = KppFrontEnd.getModules(args2);
+                        modules = Arrays.asList(KppFrontEnd.getModules(args2));
                         break;
                     default:
                         invalidJarArguments();
                         return;
             }
-            if (modules == null) {
+            if (modules.size() == 0) {
                 //boot error, we should have printed it already
                 System.exit(1);
             }
--- a/src/main/java/org/kframework/main/Main.java
+++ b/src/main/java/org/kframework/main/Main.java
@@ -50,7 +50,7 @@ public class Main {
             String[] args2 = Arrays.copyOfRange(args, 1, args.length);
                 switch (args[0]) {
                     case "-kompile":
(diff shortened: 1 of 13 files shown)


dwightguth:
You should not need the getModuleType method. If getKompileModules() returns a module, it should be added to the dependency injector /in addition to/ the modules already present in the K framework core. So essentially the injector will contain all the kompile modules in the core and all the kompile modules in the loaded plugins. Doing it this way should eliminate your need to ask me questions 3 and 4.

Take a look at how the backends are bound using a MapBinder in KompileModule in order to see how to use this to add a backend without needing to disable the existing configuration.

I believe you should search every jar in the lib/plugins directory of the distribution.

osa1:
@dwightguth please review again, both this patch and kompile-plugin repo.

Now it's running like this:

```
➜  k git:(plugin-wip) ✗ ./target/release/k/bin/kompile --backend MyKompileBackend tutorial/1_k/1_lambda/lesson_1/lambda.k 
found a kmodule
running MyKompileBackend
```

Backend is injected here https://github.com/osa1/kompile-plugin/blob/master/Main/src/myKModule/MyKompile.java

Does it look correct now?

dwightguth:
This is closer to what I had in mind, but bear in mind that there might be more than one KModule per jar and more than one jar in the directory. You also need to clean up the error reporting code so that all exceptions are handled appropriately. Since we don't have access to the KExceptionManager or the GlobalOptions classes in the boot process, you should take a look at FrontEnd.printBootError, although it should be modified to make sure that you still print the stack trace.

osa1:
I made another update. Loading multiple JARs and KModules should be working now. (I'll test that in detail).

Two questions:
1. How should I handle conflicting modules? For example, what happens if multiple krun modules are found? Or multiple providers of same backend?
2. How should I add plug-in loading to test suite? Should I create some JARs manually and try to load them in some unit tests? (regenerating the JARs when API is changed may be too painful though, unless we implement maven projects for that job)

> you should take a look at FrontEnd.printBootError, although it should be modified to make sure that you still print the stack trace.

We don't know about `--debug` argument at that point, do you want to print stack traces all the time? (even when `--debug is not provided) There's also #937 about this.

(Don't worry about lots of commits, I'll squash them to make it ready for merging)

src/main/java/org/kframework/main/Main.java
```diff
@@ -29,36 +32,69 @@
     public static void main(String[] args) {
         AnsiConsole.systemInstall();
 
-        Module[] modules;
+        KClassLoader loader = new KClassLoader();
+        loader.addPath("target/lib/plugins");
```
osa1: This relative path is only sometimes correct. Is there a way to get correct path at runtime?

dwightguth: KPaths.getKBase

dwightguth:
Yes, during boot all errors should print a stack trace in addition to a friendly error message.

Don't worry about testing, the change to make the java backend a module will add test coverage. Don't worry about module conflicts either, the dependency injector will detect these and print an error. Only module developers should see such an error so it doesn't need to be pretty.

src/main/java/org/kframework/main/KClassLoader.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.io.File;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLClassLoader;
+
+/**
+ * A ClassLoader that allows adding JARs in directories.
+ */
+public class KClassLoader extends URLClassLoader {
```
dwightguth: call this KPluginClassLoader?

src/main/java/org/kframework/main/Main.java
```diff
@@ -29,36 +34,71 @@
...
+        for (KModule m : kLoader) {
+            kModules.add(m);
+        }
+
+        List<Module> modules = new ArrayList<>(0);
         if (args.length >= 1) {
             String[] args2 = Arrays.copyOfRange(args, 1, args.length);
                 switch (args[0]) {
                     case "-kompile":
-                        modules = KompileFrontEnd.getModules(args2);
```
dwightguth, line 37: maybe you should make these static methods return a List instead of an array, for consistency?

dwightguth:
This looks good if you clean things up and take care of the two suggestions I just made. You tested it and it works as expected?

dwightguth:
Looks like you need two more things:
1. termination of the application if plugins fail to load correctly.
2. A way to automatically generate the lib/plugins folder in the file system (placeholder file?)

osa1:
I just finished addressing previous comments.

Multiple JARs and classes seem to work fine. One thing that may be a bit annoying is that if multiple classes provide same backend, it just works without any errors. What's happening is that in `Map<String, Backend>` same key is being overwritten.

> termination of the application if plugins fail to load correctly.

I just tried that but it worked fine here. Are you having a problem with termination? How can I reproduce that? I tried on a plugin that fails at different stages of plugin loading and every case worked fine.

> A way to automatically generate the lib/plugins folder in the file system (placeholder file?)

I guess that folder should be created by build system, right? I'm not very experienced with that, any ideas how to do that?

dwightguth:
Take a look at the output of the Jenkins build: you will see that an error was printed but it then continues on to execute the application.

I suggest you test if it actually works if you attempt to provide two different backends with the same key. I suspect what is happening is that you tried to map the same key to the same value twice, which silently succeeds because Guice knows they are actually the same binding. I have had errors before when the same key was bound to two /different/ values however. If that is no longer happening, I would like to know that, please.

As for the build system, I suggest you edit src/main/assembly/bin.xml and add a new entry which maps a blank file in src/main/resources to be copied to the lib/plugins directory of the assembly.

osa1:
> Take a look at the output of the Jenkins build: you will see that an error was printed but it then continues on to execute the application.

Tests are now passing.

> I suggest you test if it actually works if you attempt to provide two different backends with the same key. I suspect what is happening is that you tried to map the same key to the same value twice, which silently succeeds because Guice knows they are actually the same binding. I have had errors before when the same key was bound to two /different/ values however. If that is no longer happening, I would like to know that, please.

Just tried that again and you're right:

```
➜  k git:(plugin-wip) ✗ ./target/release/k/bin/kompile --backend MyKompileBackend tutorial/1_k/1_lambda/lesson_1/lambda.k
Exception in thread "main" com.google.inject.CreationException: Guice creation errors:

1) Map injection failed due to duplicated key "MyKompileBackend"
  at com.google.inject.multibindings.MapBinder$RealMapBinder$1.initialize(MapBinder.java:355)
  at org.kframework.kompile.KompileModule.configure(KompileModule.java:56)

1 error
        at com.google.inject.internal.Errors.throwCreationExceptionIfErrorsExist(Errors.java:435)
        at com.google.inject.internal.InternalInjectorCreator.injectDynamically(InternalInjectorCreator.java:175)
        at com.google.inject.internal.InternalInjectorCreator.build(InternalInjectorCreator.java:109)
        at com.google.inject.Guice.createInjector(Guice.java:95)
        at com.google.inject.Guice.createInjector(Guice.java:72)
        at org.kframework.main.Main.main(Main.java:105)
```

Apparently, in my previous test, I copied the JAR to `/target/lib` instead of `/target/release/k/lib` and thus my JAR is ignored. Sorry for the confusion.

> As for the build system, I suggest you edit src/main/assembly/bin.xml and add a new entry which maps a blank file in src/main/resources to be copied to the lib/plugins directory of the assembly.

I handled this in a different way, is that OK? ( https://github.com/kframework/k/commit/4146312b5fb44553f28847215f47eb89cf09cde4 )

I also squashed commits. Unless you have more comments, I think this is ready for merge.

dwightguth:
I don't think using antrun mkdir is a good solution for creating a placeholder directory. If I were to take target/release/k and put it in an archive, the lack of files in lib/plugins would mean that the directory would not get created by the user when they decompress the archive. Since this is what we will likely eventually be doing with the K release, it would be better if that directory had a file of some kind in it. (Maybe some kind of really basic README?)

osa1:
> If I were to take target/release/k and put it in an archive, the lack of files in lib/plugins would mean that the directory would not get created by the user when they decompress the archive.

Why? I tried compressing/decompressing with empty `plugins` directory and it seemed to work fine. What is the problem with empty directories? Do some zip/rar/tar programs do that while decompressing?

I can still put a README saying "This is the directory that K searches for plugins."

PR #955, opened 2014-09-16
Delete unused files from repo

dwightguth:
@traiansf @cos @yilongli @owolabileg Please review


andreistefanescu:
Please leave the files in tutorial/1_k/5_types/experiments.

dwightguth:
What are we using them for?

grosu:
Just a general principle, without checking the details (I cannot check them now, due to urgent matters):

Everything under tutorial should be either fixed and stay, or otherwise removed!  If you are talking about sample/java-rewrite-engine, that is another story.

owolabileg:
It is OK with me to remove those files from samples/java_rewrite_engine/tutorial/1_k/2_imp/indexing

traiansf:
I'm generally not in favor of removing tests.  Couldn't we revive (at least some) of these tests?  I'm mostly concerned about the regression tests (issues).

traiansf:
I'm ok with it as it is now. Maybe add an issue that those issues should be revisited and either removed or added back to testing so that we don't forget that they are there.

I think some of the issues are testing exceptional behavior, but since we did not have a way to detect that, they were not moved into regression

PR #957, opened 2014-09-16
fix required VC++ Redistributable version number in INSTALL

osa1:


INSTALL.md
```diff
@@ -5,8 +5,9 @@ Users who checked out the sources should follow the instructions in src/README.m
 1. Prerequisites:
   * Java Runtime Edition version 7 or higher (http://java.com/en/download/index.jsp)
     * To make sure java is installed properly, call `java -version` in a terminal.
-  * (On Windows only) Visual C++ 2012 Redistributable Package 
-    (http://www.microsoft.com/en-us/download/details.aspx?id=30679)
+  * (On Windows only) Visual C++ 2010 Redistributable Package
```
radumereuta, line 8: Or more. It works with 2012 too.

osa1, line 8: Are you sure you only have 2012 installed? Because when I try with only 2012, it doesn't work. I need 2010.

radumereuta, line 8: isn't it included?
Anyway, on my Win7 machine I have installed 2005 (which is not enough) and 2012, which made it work when I first run into this problem.
I think on my laptop where I have Win8, I only have 2012, but I will have to get home before I can confirm.

dwightguth:
I suggest you check the actual dependencies of the binary and compare them against information about what version of the redistributable provides those files. That's the most reliable way to confirm this one way or the other.

INSTALL.md
radumereuta, line 8: Ok, on my Win8 laptop I have installed 05, 08, 10, 12. Pretty much everything.
When I first looked for the missing files, I used http://www.dependencywalker.com/
I could have another look tomorrow when I get back to my desktop.

radumereuta, line 8: When I only have 2005 version installed, libz3.dll has these 3 missing files:

```
MSVCR110.DLL
VCOMP110.DLL
MSVCP110.DLL
```

I think the minimum version is 2010, but by searching similar problems, it is possible to require more than one version.
Suggestions?

radumereuta:
That's what I did. I ran a dependency walker and those are the 3 files missing.
I am not sure exactly what version of VC++ distributes them. From what I read on a quick google search is that it mainly depends on which version of VC++ has been used to generate the dlls.

PR #959, opened 2014-09-17
[WIP] switch to Java 8

osa1:
I'm creating this to 1) track progress 2) use Jenkins server to test.

Not ready for merge.


yilongli:
Why is this patch needed? https://gist.github.com/osa1/da85c315b38bc9425ed0 This should have no effect on the maude backend. Is there a bug in the Java backend that is revealed by switching to Java 8?

osa1:
I removed my comment because I realized that that patch is not needed.

> Is there a bug in the Java backend that is revealed by switching to Java 8?

Yes. It's generating different outputs in some cases. `KILToBackendJavaKILTransformer.evaluateRule` is behaving differently when switched to Java 8.

yilongli:
I fixed a bug that could result in incorrect result of `evaluateRule` recently. You may want to check if that solves the problem.

osa1:
I don't think it's fixed. Tried one hour ago with latest master.

---

Ömer Sinan Ağacan
http://osa1.net

2014-09-27 1:01 GMT+03:00 Yilong Li notifications@github.com:

PR #960, opened 2014-09-17
separated Cell2DataStructures to Java-backend-specific version and non-b...

yilongli:
...ackend-specific version

@dwightguth please review

### DIFF
--- a/samples/java_rewrite_engine/tutorial/2_languages/1_simple/1_untyped/tests/config_verification.xml
+++ b/samples/java_rewrite_engine/tutorial/2_languages/1_simple/1_untyped/tests/config_verification.xml
@@ -8,17 +8,22 @@
         extension="simple"
         >
     <kompile-option name="--backend" value="java" />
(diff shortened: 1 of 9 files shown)


yilongli:
@andreistefanescu I have no idea why it also solves your pending problem about printing cell map and cell list... Any idea?

andreistefanescu:
Are you sure the rules are properly transformed? The examples work with cells as well, but the output would be formatted with cells.

yilongli:
Alright, I think I have figured out the problem. This pull request doesn't solve the existing problem about printing cell map and cell list. I made a mistake earlier since I didn't realize Andrei also used this pass to deal with things related to the prover. Therefore, the maude backend doesn't properly transform the rule but somehow still get something looks correct. I have fixed the problem and reverted the output files.

yilongli:

dwightguth:
lgtm. @andreistefanescu merge if you are happy.

PR #961, opened 2014-09-17
fix some of wrong comparisons in MaudeKRun to fix a test

osa1:
I realized this while trying to make tests working under Java 8.

Note that static analyzers are reporting a lot of warnings about compare methods. They'll become a problem sooner or later. I'll open a separate issue for that.

### DIFF
--- a/src/main/java/org/kframework/backend/maude/krun/MaudeKRun.java
+++ b/src/main/java/org/kframework/backend/maude/krun/MaudeKRun.java
@@ -680,15 +680,15 @@ public class MaudeKRun implements KRun {
         assertXML(child.size() == 1);
         String sort = child.get(0).getAttribute("sort");
         String op = child.get(0).getAttribute("op");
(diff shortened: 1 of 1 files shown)


src/main/java/org/kframework/backend/maude/krun/MaudeKRun.java
```diff
@@ -680,15 +680,15 @@ public Transition transform(HyperEdgeMetadata h) {
         assertXML(child.size() == 1);
         String sort = child.get(0).getAttribute("sort");
         String op = child.get(0).getAttribute("op");
-        assertXML(op.equals("_`(_`)") && sort.equals(Sort.KITEM));
+        assertXML(op.equals("_`(_`)") && sort.equals(Sort.KITEM.toString()));
```
yilongli, line 683: You may want to change `String sort` to `Sort sort` instead.

dwightguth:
No, actually this is better as a string because it's not a K sort, it's a maude sort.

PR #962, opened 2014-09-17
fix various findbugs errors categorized CORRECTNESS and BAD_PRACTICE

dwightguth:
**please review**

### DIFF
--- a/src/main/java/org/kframework/parser/concrete2/Grammar.java
+++ b/src/main/java/org/kframework/parser/concrete2/Grammar.java
@@ -288,6 +288,26 @@ public class Grammar implements Serializable {
             final int key;
             public OrderingInfo(int key) { this.key = key; }
             public int compareTo(OrderingInfo that) { return Integer.compare(this.key, that.key); }
+            @Override
+            public int hashCode() {
+                final int prime = 31;
+                int result = 1;
+                result = prime * result + key;
+                return result;
+            }
+            @Override
+            public boolean equals(Object obj) {
+                if (this == obj)
+                    return true;
+                if (obj == null)
+                    return false;
+                if (getClass() != obj.getClass())
+                    return false;
+                OrderingInfo other = (OrderingInfo) obj;
+                if (key != other.key)
+                    return false;
+                return true;
+            }
         }
 
         public NonTerminal(String name) {
@@ -359,6 +379,26 @@ public class Grammar implements Serializable {
             final int key;
             public OrderingInfo(int key) { this.key = key; }
             public int compareTo(OrderingInfo that) { return Integer.compare(this.key, that.key); }
+            @Override
+            public int hashCode() {
+                final int prime = 31;
+                int result = 1;
+                result = prime * result + key;
+                return result;
+            }
+            @Override
+            public boolean equals(Object obj) {
+                if (this == obj)
+                    return true;
+                if (obj == null)
+                    return false;
+                if (getClass() != obj.getClass())
+                    return false;
+                OrderingInfo other = (OrderingInfo) obj;
+                if (key != other.key)
+                    return false;
+                return true;
+            }
             public String toString() { return Integer.toString(key); }
         }
 
--- a/src/main/java/org/kframework/parser/concrete2/Parser.java
+++ b/src/main/java/org/kframework/parser/concrete2/Parser.java
@@ -233,10 +233,30 @@ public class Parser {
                 this.key.stateCall.key.ntCall.exitStateReturns.add(this);
             }
         }
-
+        @Override
         public int hashCode() {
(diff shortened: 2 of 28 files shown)


dwightguth:
@radumereuta @yilongli  please review

yilongli:
lgtm

src/main/java/org/kframework/parser/concrete2/Grammar.java
```diff
@@ -288,6 +288,26 @@ private static void collectNTCallers(State start, Set<State> visited,
             final int key;
             public OrderingInfo(int key) { this.key = key; }
             public int compareTo(OrderingInfo that) { return Integer.compare(this.key, that.key); }
+            @Override
```
radumereuta, line 291: Don't modify this file please.

src/main/java/org/kframework/parser/concrete2/Parser.java
```diff
@@ -233,10 +233,30 @@ public int hashCode() {
                 this.key.stateCall.key.ntCall.exitStateReturns.add(this);
             }
         }
-
+        @Override
         public int hashCode() {
-            return this.key.hashCode();
+            final int prime = 31;
```
radumereuta, line 238: Don't modify this file please.

src/main/java/org/kframework/parser/concrete2/Grammar.java
dwightguth, line 291: Too bad, it's a bug. I'm fixing a bug.

src/main/java/org/kframework/parser/concrete2/Parser.java
dwightguth, line 238: see above

src/main/java/org/kframework/parser/generator/SDFHelper.java
```diff
@@ -91,21 +91,38 @@ public static String getFollowRestrictionsForTerminals(Set<Terminal> terminals)
      * @author RaduFmse
      *
      */
-    private static class Ttuple {
+    private final static class Ttuple {
```
radumereuta: Did I really misspelled Tuple with double t?
Could you take care of this please?

src/main/java/org/kframework/utils/Poset.java
```diff
@@ -19,7 +19,7 @@
 
     private boolean cacheEnabled = false;
 
-    private java.util.Set<Tuple> relations = new HashSet<>();
+    private java.util.Set<Tuple<T>> relations = new HashSet<>();
```
radumereuta, line 23: What? There's another Tuple? This time having a parameter. Could you look and see why we have two?

src/main/java/org/kframework/utils/Poset.java
```diff
@@ -288,22 +288,38 @@ public Tuple(T big, T small) {
...
+            result = prime * result + ((small == null) ? 0 : small.hashCode());
+            return result;
         }
 
+
         @Override
-        public int hashCode() {
-            return big.hashCode() + small.hashCode();
+        public boolean equals(Object obj) {
+            if (this == obj)
```
radumereuta, line 305: Might be ok to assume the tuple elements are not null.

src/main/java/org/kframework/utils/Poset.java
dwightguth, line 23: they're both inner classes, just for different things.

src/main/java/org/kframework/parser/concrete2/Grammar.java
radumereuta, line 291: What bug? I haven't seen any bug reports. I really like to be the one who modifies the parser classes.
The idea for the parser classes is to be as small as possible. I'm don't need hashCode and equals for these classes.
From my perspective, all you did is to increase the size of boilerplate code.

dwightguth, line 291: if you read the title of this pull request, you will see that I ran the tool through findbugs, which is a static analysis tool designed to detect bugs and antipatterns. In this case, these classes implemented Comparable. The contract of Comparable is that it provides a total ordering consistent with equals and hashCode. i.e., if `o1.compareTo(o2) == 0`, then `o1.equals(o2)`, and if `o1.compareTo(o2) != 0`, then `!o1.equals(o2)`

It's bugs like this that lead us into situations where we have to spend long effort migrating from one JVM to another, because we have disobeyed the contract of the Java language.

radumereuta, line 291: Ok, fine.

radumereuta:
lgtm, merge if tests pass.

dwightguth:
Jenkins: test this please

PR #965, opened 2014-09-18
KTest propagate debug flag to processes

osa1:
Do not merge yet. There's a missing feature(I did not update --dry yet).

Some notes:
- Programs are only run twice when --debug is enabled and program is expected to produce error output. (so that we can use --debug output in verbose output and non--debug output to compare program error output with expected error output)
- If --debug is enabled, it's always passed to programs. No matter what the verbosity setting is. (except in second run of programs with expected output)
- If stack trace is printed, it's only used in verbose output and not in generated JUnit reports.
- KTest only cares about error outputs when programs terminate with non-zero exit code. So if a program exists with zero but still prints a stack trace to _stderr_ because of --debug, the test still passes.
- If program prints something different to _stdout_ due to --debug use and exit with non-zero, then the test fails.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -322,6 +322,7 @@
                   <arg value="${config}" />
                   <arg value="-v" />
                   <arg value="--report" />
+                  <arg value="--debug" />
                 </exec>
               </target>
               <skip>${skipKTest}</skip>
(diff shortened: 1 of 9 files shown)


dwightguth:
Jenkins: test this please

dwightguth:
can you please modify the pom so that --debug is passed to the ktest that is spawned by `mvn verify`?

dwightguth:
It looks like the only tests that fail are the prover, because Andrei accidentally committed code that wrote to stdout when you turned --debug on. He says it was a mistake, so I suggest you find the places in the code where these lines are being printed and change them to print on stderr.

PR #967, opened 2014-09-19
Add a built-in IO operator, #external, for executing an external process.

daejunpark:
I've added a new built-in IO operator, `#external`. The usage is as follows. 
@dwightguth @yilongli @andreistefanescu, please review.

Usage: The `#external` takes a string of command as an input and returns the
string of stdout. The command string should use `"\x01"` as a seperator instead
of whitespaces. For example, the following term:

```
  #external("echo" +String "\x01" +String "ok!")
```

rewrites to `"ok!"`.


dwightguth:
I suggest you make this function called #system and return a klist containing the return code, the stdout, and the stderr.

daejunpark:
@dwightguth good suggestion, but, I'm sorry I don't know how to do it in the maude backend. Could you help me to understand the communication with the maude? Or, could we support this feature for only the java backend?

dwightguth:
I suggest you take a look at what the #stat command in the IO server does.
You should be able to adapt it for this purpose. However, if you have no
personal need to support the Maude backend, it should not be necessary for
you to do so. Just leave the hook unimplemented in Maude.

daejunpark:
Thanks, @dwightguth ! 
I've updated for both maude and java backend, as you suggested. Please review.

andreistefanescu:
lgtm

yilongli:
lgtm too

Issue #976, opened 2014-09-23
A function rule is not applied in pattern-matching backend, especially when using K-FUNCTIONAL-VISITOR


daejunpark:
@yilongli 

A function rule is not applied in `--pattern-matching` backend.

In the following example, I've used `K-FUNCTIONAL-VISITOR` to replace `Id`s in a given term. It seems that a reducible functional term generated by the replacement is not considered as a function rule evaluation.

Suppose that `a.k` is given as follows: (sorry, but it's my best to minimize the example.)

```
require "modules/k-functional-visitor.k"

module A



imports K-FUNCTIONAL-VISITOR

syntax PseudoCodes ::= PseudoCode PseudoCodes
                     | PseudoCode
syntax PseudoCode  ::= "Let" Id "=" K ";" [seqstrict(2)]
                     | "Return" K ";" [seqstrict(1)]

rule PC:PseudoCode PCs:PseudoCodes => PC ~> PCs

rule Let X:Id = V:KResult; ~> PCs:PseudoCodes 
  => #visit(PCs, '@Replace`(_`,_`), V, '_==K_, X)
//                   
syntax KItem ::= "@Replace" "(" K "," K ")" [function]
rule @Replace(_, K) => K




syntax KResult ::= Val

syntax Exp ::= Val
             | Id

syntax Var ::= String
syntax Val ::= Int

syntax Val ::= "@desc" "(" Map ")"

syntax Exp ::= Exp "." Var [function]
rule @desc(X |-> V:Val _) . X => V



syntax KItem ::= "run"

syntax Id ::= "$desc"
rule run
  => Let $desc = foo;
     Return $desc."x";

syntax KItem ::= "foo"
rule foo => @desc("x" |-> 1 "y" |-> 2)



endmodule

```

Then, the `pattern-matching` backend gets stuck at the reducible functional term: `@desc( "x" |-> 1 "y" |-> 2 ) . "x"`, which should have been reduced to `1`.

```
$ cat 1.a
run

$ krun --pattern-matching 1.a
[7, 32.31 ms]
28.48 ms(mc=3.185 ms, eval=9.411 ms[0.000 ns, 9.304 ms], rew=14.50 ms) + 0.000 ns
870.0 μs
0.000 ns
<k>
    @desc ( 
        "x" |-> 1
        "y" |-> 2 ) . "x" ~> Return HOLE ;
</k>
```

[...]

PR #978, opened 2014-09-23
fixing issue #976

yilongli:
fixed https://github.com/kframework/k/issues/976

@dwightguth please review


src/main/java/org/kframework/backend/java/builtins/BuiltinVisitorOperations.java
```diff
@@ -63,6 +63,18 @@ public ASTNode transform(Term term) {
...
                 }
             });
+            postTransformer.addTransformer(new LocalTransformer(context) {
+                @Override
+                public ASTNode transform(KItem kItem) {
+                    /* TODO(YilongL): we need a way to pass the
+                     * `copyOnShareSubstAndEval` argument of
+                     * KItem#evaluateFunction to the evaluation of builtin
+                     * function that needs such information; or even better, a
+                     * new mechanism for invoking builtin functions */
```
dwightguth, line 73: if you don't want to do this now, it's not necessary, but I should point out that BuiltinVisitorOperations is constructed by Guice, so if you want to bind a boolean value in the configuration, you can pass it to the constructor of this class and it will automatically propagate from the part of the code that knows whether you have passed the --pattern-matching flag.

dwightguth:
Lgtm other than my minor comment which you can address or not as you choose. You can merge if the tests pass.

yilongli:
Jenkins: retest this please

yilongli:
@dwightguth I don't understand why the CIL test failed during kompilation for my two different unrelated pull requests. And I couldn't reproduce the problem on my machine. Thanks.
http://office.runtimeverification.com:8080/rv-jenkins/job/k-framework-pull-requests-maven/106/console

dwightguth:
I have seen this error before, it seems to be intermittant. I suggest you rerun the test suite and hopefully it will pass this time. Next time you see it, please let me know, because I enabled the stack trace in the hopes that it will give me enough information to debug the problem.

yilongli:
Jenkins: retest this please

Issue #979, opened 2014-09-23
Java backend fails to pull out the correct list in a data structure


liyili2:
If I have the following definition:

```
module TEST

     syntax KItem ::= "start"

     syntax KItem ::= "calculateNewBytes-aux" "(" DataList "," DataList "," List ")"
        syntax KItem ::= joinBitsToBytes(List) [function]
     syntax KItem ::= "joinBitsToBytes-aux" "(" K "," K ")" [function]

     rule start => calculateNewBytes-aux
      (dataList(.List),dataList(ListItem ( piece ( 0 , 1 ) )),ListItem(piece(1,1)) )

     syntax DataList ::= dataList(List)
     syntax Bits ::= piece(Bits, Int)
     syntax Bits ::= Bits "bit::" Bits
     syntax Bits ::= Int

     rule calculateNewBytes-aux(dataList(.List), dataList(L:List), Result:List)
          => joinBitsToBytes(Result L)
          [structural]

     rule joinBitsToBytes(L:List)
          => joinBitsToBytes-aux(dataList(L), dataList(.List))


     rule joinBitsToBytes-aux(
               dataList(ListItem(piece(N:Bits, Len:Int)) ListItem(piece(M:Bits, 1)) 
                  L:List), dataList(R:List))
          => joinBitsToBytes-aux(
               dataList(ListItem(piece(piece(N, Len) bit:: piece(M, 1), Len +Int 1)) L),
               dataList(R))
          requires Len <Int 8
     rule joinBitsToBytes-aux(
               dataList(ListItem(piece(N:Bits, Len:Int)) L:List), dataList(R:List))
          => joinBitsToBytes-aux(
               dataList(L), dataList(R ListItem(piece(N, Len))))
          when Len ==Int 8 
     rule joinBitsToBytes-aux(dataList(.List), dataList(R:List))
          => dataList(R) 

endmodule
```

When I execute the program "start", the java backend gives me a result like:

```
<k>
    dataList ( .List )
</k>

```

Which is wrong, the correct result from the Maude backend is:

```
<k>
    joinBitsToBytes-aux ( dataList ( ListItem ( piece ( piece ( 1 , 1 ) bit:: 
      piece ( 0 , 1 ) , 2 ) ) ) , dataList ( .List ) )
</k>

```

More interestingly, If I remove the function "joinBitsToBytes" and ask the function "calculateNewBytes-aux" to rewrite to "joinBitsToBytes-aux" function directly as:

```
module TEST

[...]

PR #989, opened 2014-09-25
fixed ConcreteCollectionVariable#matchConcreteSize

yilongli:
fixed https://github.com/kframework/k/issues/979

@dwightguth please review


dwightguth:
I don't understand what this function is supposed to do, so I can't independently verify that the change you made is correct. @andreistefanescu can you review?

yilongli:
This function checks if a ConcreteCollectionVariable, i.e., a variable representing collection with fixed number of elements, can _match_ a given term, that is, the given term should not be narrowed.

dwightguth:
So a ConcreteCollectionVariable always knows exactly how many elements it has, and that's what the concreteSize field is? But then a variable with size 5 should be able to match a set with 4 elements and a variable. So why was this wrong?

yilongli:
Because in your example a variable with size 5 is the pattern, while a set with 4 elements and a variable is the subject, you actually require the variable in the subject to have only one element. The subject is then narrowed.

dwightguth:
lgtm

andreistefanescu:
@yilongli I think this change makes SymbolicConstraint.EqualityOperations#isFalse and SymbolicUnifier#unify incorrect. The change in behavior should only apply in the PatternMatcher.

yilongli:
@andreistefanescu please review again.

andreistefanescu:
lgtm

yilongli:
Jenkins: retest this please

PR #992, opened 2014-09-25
printed Java rewrite engine statistics only when --statistics is on

yilongli:
fixed https://github.com/kframework/k/issues/987
@dwightguth please review


dwightguth:
Can you please show me an example of the output you get when you run some examples with the --statistics flag?

yilongli:
When the option --statistics is on, it shows exactly the same information as before. Usually a single line such as `[209611, 7.203 s]` printed at the beginning, showing the number of rewrite steps and the total rewriting time. When using the faster rewrite engine, more specific profiling information will be printed after that line. Anyway, this pull request only hides profiling information by default to allow people comparing the results from different backends directly.

src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -594,41 +593,11 @@ private ConstrainedTerm applyRule(ConstrainedTerm constrainedTerm, List<Rule> ru
         }
 
         stopwatch.stop();
-        System.err.println("[" + visited.size() + "states, " + step + "steps, " + stopwatch + "]");
-
-        return searchResults;
-    }
-
-
-    public List<ConstrainedTerm> prove(List<Rule> rules, FileSystem fs, TermContext context) {
```
dwightguth: why was this entire function deleted?

dwightguth:
You should be able to fix the merge conflicts without needing to do anything as long as you use the recursive merge strategy.

src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
yilongli: I think it's been dead code for a long time

dwightguth:
You can merge this if you fix the conflicts and the tests pass.

PR #999, opened 2014-09-26
some improvements in java backend strategy rule iterators

osa1:
While trying to debug something, I realized that rewrite engine is trying to iterate through rewrite rules, but strategy is just returning empty list of rules. I made some simplifications in the iterator methods to prevent this. This should come with a minor performance improvement. (no more iterating through empty lists)

Note that I'm just learning Java backend, so I may have done something wrong. @yilongli @dwightguth please review.

### DIFF
--- a/k-core/src/main/java/org/kframework/backend/java/strategies/CompositeStrategy.java
+++ b/k-core/src/main/java/org/kframework/backend/java/strategies/CompositeStrategy.java
@@ -3,10 +3,7 @@ package org.kframework.backend.java.strategies;
 
 import org.kframework.backend.java.kil.Rule;
 
-import java.util.Collection;
-import java.util.LinkedHashSet;
-import java.util.LinkedList;
-import java.util.Iterator;
+import java.util.*;
 
 /**
  * A CompositeStrategy is used to combine multiple to strategies together in
@@ -35,7 +32,7 @@ import java.util.Iterator;
 
 public class CompositeStrategy implements Strategy {
     public CompositeStrategy() {
-        stack = new LinkedList<Strategy>();
+        stack = new LinkedList<>();
     }
 
     public void push(Strategy s) {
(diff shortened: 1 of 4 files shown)


dwightguth:
lgtm. Yilong merge if you are happy and tests pass.

yilongli:
lgtm though I am not familiar with the implementation of strategy. However, the cil test goes from 33s to 38s. @osa1 can you test it on your machine to make sure your change wouldn't slow down the rewrite engine?

osa1:
Interesting. This should only make it faster. @yilongli how do I run cil tests?

yilongli:
Check out the c-semantics repository and switch to the branch java_rewrite_engine. Go to the directory  cil-semantics and run ktest as usual.

osa1:
@yilongli you should only run krun command to test:

```
$ krun 'tests/ktest/test1.cil.c' '--smt' 'none' '--pattern' '<result-value> ###result-value(I:Int) </result-value>' '--directory' 'cil-semantic'
```

When I tried with `time`, I had 2.19s with this build vs. 2.24 current master.

I don't thing this will make it slower and I don't think performance difference won't be significant in either way. Think this patch as 1) simplification(why so complicated `hasNext` and `next` methods?) 2) removing some redundant operations (which really helps for debugging, try debugging applied rules, you'll have to manually step through empty list of rules)

yilongli:
Sorry, I don't understand how it can be so fast? The Jenkins machine takes half a minute to finish this test. I don't think your change will make it slower either but just to be safe.

k-core/src/main/java/org/kframework/backend/java/strategies/CompositeStrategy.java
```diff
@@ -76,48 +73,35 @@ public void reset(Collection<Rule> rules) {
...
-        // will pull it back up the recursion, applying it to each successive
-        // strategy on the way back up.
-        if (!stack.peekFirst().hasNext()) {
-            // Pop the first element so the subsequent recursive call to next()
-            // will be operating on the next strategy in the stack.
-            Strategy top = stack.pollFirst();
-            // Apply the result of the recursive call to the top strategy.
-            top.reset(next());
-            stack.addFirst(top);
+        while (!stack.isEmpty()) {
```
yilongli, line 81: In the past, I found that sometimes this method return duplicate rules. But I have never tried to identify the root cause. Can you make sure this is no longer the case? Actually, I don't really like returning a collection of rules in the first place.

osa1:
@yilongli test suite is running only one krun process, and I'm only testing it. No point in testing PDF and kompile steps, which take a long time but not related with this patch. This is strictly krun related.

k-core/src/main/java/org/kframework/backend/java/strategies/CompositeStrategy.java
osa1, line 81: Duplicate rules are still happening, but it's not related with this method. Some rules are added multiple times to index table. As an example, when I compile this file:

```
module TEST
  syntax Sort ::= Int | Bool
  syntax KResult ::= Int | Bool
  syntax KItem ::= "test"
  rule test => isKResult(?S:Sort) ~> ?KR:KResult ==K ?S:Sort
endmodule
```

These rules are added to index table:

```
rule <generatedTop><k>'test(.KList) ~> _3:K</k>_0:Bag</generatedTop> => <generatedTop><k>isKResult(_1:Sort) ~> '_==K_(_2:KResult,, _1:Sort) ~> _3:K</k>_0:Bag</generatedTop> requires [] ensures [] [Location: (5,3,5,60), File: /home/omer/K/k_new_dist/k/issue696/test.k]
rule <generatedTop><k>'test(.KList) ~> _3:K</k>_0:Bag</generatedTop> => <generatedTop><k>isKResult(_1:Sort) ~> '_==K_(_2:KResult,, _1:Sort) ~> _3:K</k>_0:Bag</generatedTop> requires [] ensures [] [Location: (5,3,5,60), File: /home/omer/K/k_new_dist/k/issue696/test.k]
rule <generatedTop><k>'test(.KList) ~> _3:K</k>_0:Bag</generatedTop> => <generatedTop><k>Bool(#"true") ~> '_==K_(_2:KResult,, _1:Sort) ~> _3:K</k>_0:Bag</generatedTop> requires [] ensures [] [Location: (5,3,5,60), File: /home/omer/K/k_new_dist/k/issue696/test.k]
```

First two rules are same. I'll fix this too, but that'll be another patch.

yilongli:
Of course, but this is the console output from your last build:

```
Running /var/lib/jenkins/jobs/k-framework-pull-requests-maven/workspace/c-semantics/cil-semantics/cil.k programs... (1 in total, 0 with input, 1 with output, 0 with error)
Running ['/var/lib/jenkins/jobs/k-framework-pull-requests-maven/workspace/k/k-distribution/target/release/k/bin/krun' '/var/lib/jenkins/jobs/k-framework-pull-requests-maven/workspace/c-semantics/cil-semantics/tests/ktest/test1.cil.c' '--smt' 'none' '--pattern' '<result-value> ###result-value(I:Int) </result-value>' '--directory' '/var/lib/jenkins/jobs/k-framework-pull-requests-maven/workspace/c-semantics/cil-semantics' '--debug'] [output: /var/lib/jenkins/jobs/k-framework-pull-requests-maven/workspace/c-semantics/cil-semantics/tests/test1.cil.c.out]
Done with ['/var/lib/jenkins/jobs/k-framework-pull-requests-maven/workspace/k/k-distribution/target/release/k/bin/krun' '/var/lib/jenkins/jobs/k-framework-pull-requests-maven/workspace/c-semantics/cil-semantics/tests/ktest/test1.cil.c' '--smt' 'none' '--pattern' '<result-value> ###result-value(I:Int) </result-value>' '--directory' '/var/lib/jenkins/jobs/k-framework-pull-requests-maven/workspace/c-semantics/cil-semantics'] (time 38400 ms)
```

I think it's counting only krun time, isn't?

osa1:
Right, sorry for confusion.

old:

```
90.04s user 1.17s system 157% cpu 57.830 total
```

new:

```
84.94s user 0.92s system 181% cpu 47.267 total
```

k-core/src/main/java/org/kframework/backend/java/strategies/CompositeStrategy.java
yilongli, line 81: OK, so the problem is the indexing table.

yilongli:
Alright, I am merging it.

Issue #995, opened 2014-09-25
Java Search does not work with IO


dwightguth:
```
$ cd k-distribution/tutorial/1_k/4_imp++/lesson_6
$ kompile imp.k --backend java --transition "lookup increment assignment read print"
$ echo 23 | krun ../lesson_1/tests/spawn.imp --search
```

This should output a variety of search solutions. Instead, it gets stuck during IO.

```
[3states, 15steps, 74.05 ms]
Search results:

Solution 1:
B:Bag -->
<T>
    <threads>
        <thread>
            <k>
                (read ( )) ~> x = HOLE ; ~> spawn x = (x / 2) ; ~> spawn x
                   = (x + 10) ; ~> print ( x, "\n" ) ;
            </k>
            <env>
                x |-> 1
            </env>
        </thread>
    </threads>
    <store>
        1 |-> 0
    </store>
    <in>
        ListItem ( #parseInput ( "Int" , " \n\t\r" ) )
        ListItem ( #buffer ( "23\n" ) )
    </in>
    <out>
        ListItem ( #buffer ( "x = " ) )
    </out>
</T>
```

PR #1000, opened 2014-09-26
fixing issue 995

yilongli:
fixed https://github.com/kframework/k/issues/995
@dwightguth please review


yilongli:
So this is not the perfect solution, I have created an issue to keep track of the problem. https://github.com/kframework/k/issues/1002 I have talked to Dwight about this. I am merging this since all tests pass.

Issue #997, opened 2014-09-25
Wrongly applied rule in java backend


dwightguth:
```
$ cd k-distribution/tutorial/1_k/5_types/lesson_1
$ krun ../../2_imp/lesson_1/tests/sum.imp
```

This should not evaluate the rule on line 71 of imp.k, however, for some reason it does, resulting in the following stuck configuration:

```
[5, 33.63 ms]
<T>
    <k>
        block
    </k>
    <tenv>
        HOLE (n = 100) ; (sum = 0) ; while ( (! (n <= 0)) ) { (sum = (sum 
          + n)) ; (n = (n + -1)) ; }
    </tenv>
</T>
```

Issue #1003, opened 2014-09-26
loss of variable sort information during symbolic execution


yilongli:
This is a long known problem of our current implementation of symbolic execution. Symbolic constraint between a variable and a term is stored as a substitution. When you substitute that variable in the RHS of a rule to build the new subject term, the variable disappears and the sort information is gone. It could lead to incorrect result eventually.

PR #1004, opened 2014-09-26
Fixing issue 997

yilongli:
fixed https://github.com/kframework/k/issues/997

There are three problems here:
1) KItem built from KLabelInjection and empty KList should have exact sort `KItem`;
2) when we use the unification-based rewrite engine to execute programs, usually the resulting constraint should be a proper order-sorted substitution (except `logik`). However, I cannot know whether the user is trying to just execute programs;
3) see https://github.com/kframework/k/issues/1003.

Basically, I fixed the first one and created an issue for the third one. I can also fix the second problem if there is an option that user can specify to tell the unification-based rewrite engine to check whether the resulting constraint should be a proper substitution. However, this is kind of redundant because the user can always just specify `--pattern-matching` to use the fast rewriter based on pattern matching.


dwightguth:
lgtm. @andreistefanescu merge if tests pass and you are happy

andreistefanescu:
lgtm

PR #1006, opened 2014-09-26
fix build number

dwightguth:
@yilongli please review. This fixes a bug causing an exception to be thrown when running kompile/kast/krun/ktest --version


yilongli:
lgtm

Issue #985, opened 2014-09-25
An exception error when using .K in #if #then #else #fi rule


liyili2:
If I have the following definition:

```
module TEST

     syntax KItem ::= "start"

     syntax KItem ::= Call(List,K)

     rule start => Call(.List,1)

     rule Call(P:List,A:K) => #if P ==K .List #then .K #else 1 #fi 

endmodule
```

When I compile it by Java backend and run the program "start", I will get the following error:

```
File: /home/qangel/project/k-unified/bin/test.k,(7,6,7,32)
[Error] Internal: Uncaught exception thrown of type AssertionError.
Please file a bug report at https://github.com/kframework/k/issues

```

If I run it in Maude backend, it will be fine:

```
<k>
    .K
</k>

```

traiansf:
Isn't this because you are not sort-decreasing?  You are attempting to transform a KItem into .K which is not KItem.  I thought the Java engine is not happy with this.

liyili2:
I use a constructor KItem ::= "dotK" and the rule for dotK is like:

```
rule dotK => .K [structural]
```

This works in Java backend, tell me the reason I need to add this step.

yilongli:
The full stacktrace is the following:

```
java.lang.AssertionError: unexpected kind K of term .K;expected kind KItem instead
    at org.kframework.backend.java.kil.KItem$KItemOperations.evaluateFunction(KItem.java:320)
    at org.kframework.backend.java.kil.KItem$KItemOperations.resolveFunctionAndAnywhere(KItem.java:287)
    at org.kframework.backend.java.kil.KItem.resolveFunctionAndAnywhere(KItem.java:253)
    at org.kframework.backend.java.symbolic.SubstituteAndEvaluateTransformer.transform(SubstituteAndEvaluateTransformer.java:139)
    at org.kframework.backend.java.kil.KItem.accept(KItem.java:584)
    at org.kframework.backend.java.symbolic.CopyOnWriteTransformer.transform(CopyOnWriteTransformer.java:226)
    at org.kframework.backend.java.symbolic.SubstituteAndEvaluateTransformer.transform(SubstituteAndEvaluateTransformer.java:165)
    at org.kframework.backend.java.kil.KSequence.accept(KSequence.java:140)
    at org.kframework.backend.java.rewritemachine.KAbstractRewriteMachine.rewrite(KAbstractRewriteMachine.java:103)
[...]

PR #1007, opened 2014-09-26
fixing issue 985

yilongli:
fixed https://github.com/kframework/k/issues/985

The failing assertion seems to be too strong. Besides, this assertion is only made for builtin functions and not user-defined functions. So I guess it's ok to remove it.


yilongli:
As suggested by Andrei, the assertion should be checked at kompile time in the pass `AddInjections`. However, I am not very motivated to fix the `AddInjection` pass right now, as the code probably needs to be changed completely for the new KORE/KAST. I would like to make an issue to keep track of it and get this pull request merged to unblock the work on c semantics.
@dwightguth

dwightguth:
Go ahead and make the issue.

PR #1008, opened 2014-09-27
kompile should only create a directory if definition exists

### LINKED ISSUE #986: kompile creates "-kompiled" directory even when no compilation is done
This is becoming annoying when you make a typo while writing definition file name or call a wrong command while running commands using bash history.

```
➜  lesson_6 git:(master) ✗ kompile no-such-file.k
[Error] Critical: Could not find 'required' file: no-such-file.k given at
console.
➜  lesson_6 git:(master) ✗ ls | grep no-such-file
no-such-file-kompiled
➜  lesson_6 git:(master) ✗ 
```

This is annoying because now I have to manually remove this directory to compile correct file:

```
➜  lesson_6 git:(master) ✗ kompile lambda.k 
[Error] Critical: Creating multiple kompiled definition in the same directory
is not allowed. Found no-such-file-kompiled and lambda-kompiled.
```

osa1:
(Resolves #986 )

Note that this may not be flexible enough for new features. A more advanced implementation might provide an API to access to "-kompiled" directory and then generate the directory only when a request to put a file in that directory is happened. (e.g. we can create the directory lazily)

If you want I can also implement that.

### DIFF
--- a/k-core/src/main/java/org/kframework/kompile/KompileFrontEnd.java
+++ b/k-core/src/main/java/org/kframework/kompile/KompileFrontEnd.java
@@ -110,7 +110,8 @@ public class KompileFrontEnd extends FrontEnd {
         if (backend.generatesDefinition()) {
                 context.kompiled = new File(options.directory, FilenameUtils.removeExtension(options.mainDefinitionFile().getName()) + "-kompiled");
                 checkAnotherKompiled(context.kompiled);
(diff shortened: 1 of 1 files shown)


dwightguth:
I think having a file system api would be a good thing because it assists with some things we might be able to do in the long run to significantly improve the performance of the tool. But we don't need to do it right now I don't think... This should be good enough for the issue you are having.

PR #1016, opened 2014-09-29
cleanup matching logic prover step 1

yilongli:
ready for review


yilongli:
There is SO MUCH more that need to be done along this line. However, I have to switch to work on the verification of floating point numbers for now.

One of the problem of this pull request is that some of the changes break the matching logic prover for some reason that I don't know yet. However, I have at least been able to locate the code that causes the failure and comment them out temporarily. I didn't remove them because I think that is the way they should be eventually.

yilongli:

dwightguth:
lgtm although @andreistefanescu should review the more detailed changes to Equality and SymbolicConstraint.

andreistefanescu:
Looks good, you can merge. Sorry it took me a while, but I was quite large.

yilongli:
@andreistefanescu thanks a lot!

PR #1018, opened 2014-09-30
remove old outdated unparser

dwightguth:
This is a bit tricky to review, but it helps to know that all I did was delete UnparserFilter, and rename UnparserFilterNew to UnparserFilter.

@radumereuta please review


dwightguth:
This is going to be tricky to fix with the unparser in the state it's currently in. I'm going to work on making the UnparserFilterNew more manageable, and then come back to this.

PR #1019, opened 2014-09-30
Non recursive pattern matcher

yilongli:
I was hoping to work more on this. However, I am busy with the matching logic prover so I would like to merge what I have right now.

The speedup of pattern matching gained by eliminating visitor pattern is consistent with the experiment in http://higherlogics.blogspot.com/2008/10/vtable-dispatching-vs-runtime-tests-and.html as well as the paper written by Scala guys http://lampwww.epfl.ch/~emir/written/MatchingObjectsWithPatterns-TR.pdf

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/symbolic/NonACPatternMatcher.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/symbolic/NonACPatternMatcher.java
@@ -1,0 +1,365 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
+package org.kframework.backend.java.symbolic;
+
+import org.apache.commons.lang3.tuple.Pair;
+import org.kframework.backend.java.kil.Cell;
+import org.kframework.backend.java.kil.CellCollection;
+import org.kframework.backend.java.kil.CellLabel;
+import org.kframework.backend.java.kil.ConcreteCollectionVariable;
+import org.kframework.backend.java.kil.Hole;
+import org.kframework.backend.java.kil.KCollection;
+import org.kframework.backend.java.kil.KItem;
+import org.kframework.backend.java.kil.KLabelConstant;
+import org.kframework.backend.java.kil.KLabelInjection;
+import org.kframework.backend.java.kil.KList;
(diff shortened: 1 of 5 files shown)


java-backend/src/main/java/org/kframework/backend/java/symbolic/NonACPatternMatcher.java
```diff
@@ -0,0 +1,499 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
```
dwightguth, line 1: My only question is, how much, if any of this file contains code that came originally from PatternMatcher.java? because if they contain code in common, it needs to be factored so that it isn't duplicated.

yilongli, line 1: I think only one static helper method actually contains largely duplicate code from PatternMatcher. The plan is to have only one pattern matcher eventually, however, I don't have the time to work on that right now.

dwightguth, line 1: can I get some idea then of what your roadmap is in order to get to only one PatternMatcher from here?

yilongli, line 1: Basically, the current non-recursive pattern matcher only deals with non-AC matching. I need to extend it to support AC matching as well. It's good to have one general-purpose pattern matcher, even though in the case of fast execution, AC matching will be eliminated before sending two terms to the pattern matcher.

dwightguth, line 1: so then the old PatternMatcher which was recursive is just going to go away entirely once this work is completed, is that the idea?

yilongli, line 1: And the new one will be renamed to PatternMatcher.

dwightguth:
Okay, it looks good to me. If you get Andrei to review it, and the tests pass, I am okay to merge this.

andreistefanescu:
lgtm

Issue #1011, opened 2014-09-28
[Java backend] name of the sort is effecting subsort calculations (and some other problems in subsort calculations)


osa1:
Here's a demonstration:

```
➜  issue696 git:(master) ✗ cat test.k
module TEST
  syntax Sort ::= Int | Bool
  syntax KResult ::= Int | Bool
  syntax KItem ::= "test"
  rule test => isKResult(?S:Sort) ~> ?KR:KResult ==K ?S:Sort
endmodule
```

With Java 7:

```
➜  issue696 git:(master) ✗ java -version
java version "1.7.0_21"
Java(TM) SE Runtime Environment (build 1.7.0_21-b11)
Java HotSpot(TM) 64-Bit Server VM (build 23.21-b01, mixed mode)

➜  issue696 git:(master) ✗ rm -rf test-kompiled && kompile test.k --backend java && krun empty
<k>
    isKResult ( _33:Sort ) ~> _36:KResult ==K _33
</k>
```

With Java 8:

```
➜  issue696 git:(master) ✗ java -version
java version "1.8.0"
Java(TM) SE Runtime Environment (build 1.8.0-b132)
Java HotSpot(TM) 64-Bit Server VM (build 25.0-b70, mixed mode)

➜  issue696 git:(master) ✗ rm -rf test-kompiled && kompile test.k --backend java && krun empty
<k>
    true ~> _35:KResult ==K _36:Sort
</k>
```

The problem is with subsort calculations. `org.kframework.backend.java.util.Subsorts` is generating different subsorting graphs when I run it using Java 7 and Java 8.

Here's what it's generating when run with Java 7:

```
this.sorts:
[Bag, Bool, MInt, TCPError, #Bot, KResult, K, #Float, TCPAnswer, #String, #Int, Set, Int, List, #Bool, Float, Sort, Nat, KList, Map, Char, KItem, Stream, String, KLabel, BagItem, Id]

ordinal of KResult: 4
ordinal of Sort: 27

subsort graph:
[...]

PR #1020, opened 2014-09-30
implement equivalent sorts

osa1:
Very crude first implementation. Some tests are expected to fail. Not ready for merge.

@yilongli @andreistefanescu can you have a look?


k-core/src/main/java/org/kframework/utils/Poset.java
```diff
@@ -2,19 +2,14 @@
...
-import java.util.Set;
-import java.util.Stack;
+import java.util.*;
 
 import org.apache.commons.collections4.set.UnmodifiableSet;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ArrayTable;
 import com.google.common.collect.Table;
+import org.kframework.kil.Sort;
```
yilongli: A `Poset` should be generic enough to have nothing to do with `Sort`.

osa1: It's not used anywhere, just forgot to remove import statement.

k-core/src/main/java/org/kframework/utils/Poset.java
```diff
@@ -22,6 +17,7 @@
 
     private java.util.Set<Tuple<T>> relations = new HashSet<>();
     private java.util.Set<T> elements = new HashSet<>();
+    private Map<T, Set<T>> equivalentElems = new HashMap<>();
```
yilongli, line 27: an alternative is to have one representative for each equivalence class, that is, a map from T to T

k-core/src/main/java/org/kframework/utils/Poset.java
```diff
@@ -2,19 +2,14 @@
...
 
 import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Set;
-import java.util.Stack;
+import java.util.*;
```
yilongli: I suggest you set a larger limit in your IDE, we agreed in the past that we should not collapse import statements like these

osa1:
This is good. As far as I can see there's only one failure and that's the problematic case I mentioned in #1011 . Current expected output of it is wrong.

k-core/src/main/java/org/kframework/compile/transformers/CompleteSortLattice.java
```diff
@@ -92,20 +84,46 @@ public ASTNode visit(Module node, Void _)  {
...
-                    for (Production production : productions) {
+                    for (Production production : sort1Prods) {
                         if (!context.isSubsorted(sort2, production.getChildSort(0))) {
                             continue sort2_loop;
                         }
                     }
 
-                    transformedNode.addSubsort(sort2, sort1, context);
-                    change = true;
+                    if (context.isSubsorted(sort1, sort2)) {
```
yilongli, line 95: This is not enough, I think. For example, you have `syntax A ::= X | Y` and `syntax KResult ::= X | Y`, you have to somehow compare the subsorts of A with subsorts of KResult to conclude that they are equivalent sorts. I don't think such information is already computed in context?

yilongli:
The current regression test suite is not very strong. I suspect there are many places that use equivalent sorts. It would be nice to write a few tests about this.

k-core/src/main/java/org/kframework/compile/transformers/CompleteSortLattice.java
osa1, line 95: To be honest, I don't understand how subsorts are declared*. The removed code was declaring `sort1 < sort2` and I replaced that code with a code that assumes `sort1 < sort2`. So if it was previously correct, it should still be correct. (see also the else block, which is same as old code)

*: The code checks if `childNode`s of all productions of `sort1` are subsorts of `sort2` and if that's the case, declares `sort1 < sort2`. I don't know what's meant by `childNode` and it's not documented anywhere. I assume it's RHS of the production, and if that's really the case then looks correct to me.

osa1:
I managed to generate a problematic definition file for this patch:

```
module TEST

  syntax S1 ::= Int
  syntax S2 ::= Bool | S1
  syntax KResult ::= S2
  syntax S3 ::= Int | Bool
  syntax KItem ::= "test"

endmodule
```

When I run `kompile --backend java` using Java 8, this is compiled in 20 seconds, but when I run same command using Java 7, it's getting stuck in an infinite loop(or taking huge amounts of time for some other reason -- killed the process after 114 seconds which is definitely not normal)

EDIT: I found the bug.

osa1:
@yilongli @andreistefanescu I updated one test and created two new tests: https://github.com/osa1/k/commit/88d407ea71ab71b4add77f9369b8360959799ec0 https://github.com/osa1/k/commit/277819b199b8a079d43462dd5ff1125147fa9c41 can you have a look and tell me if outputs look correct?

k-core/src/main/java/org/kframework/utils/Poset.java
```diff
@@ -47,7 +50,27 @@ public void addRelation(T big, T small) {
...
+        if (bigEqs == null) {
+            bigEqs = new HashSet<>();
+            bigEqs.add(big);
+        }
+        if (smallEqs == null) {
+            smallEqs = new HashSet<>();
+            smallEqs.add(small);
+        }
+
+        for (T s1 : bigEqs) {
```
yilongli, line 69: This looks too expensive to me. Can you precompute some information and eliminate this nested loop? For example, have one representative for each equivalent class.

k-core/src/main/java/org/kframework/compile/transformers/CompleteSortLattice.java
yilongli, line 95: alright, then I am fine with it

yilongli:
The tests look good to me.

k-core/src/main/java/org/kframework/utils/Poset.java
andreistefanescu, line 69: I agree with @yilongli . An alternative is to keep for each set of equivalent sorts a representative, and then have a `HashMap` from each sort to the representative of the equivalence class. The `relations` should only contain pairs of representatives of equivalence classes.

osa1:
Found another bug with subsorting and added a test for that.

osa1:
We need to handle transitive closure generation and cycle detection too.

PR #1029, opened 2014-10-02
update pom to allow cross platform compilation

dwightguth:
@cos  please review

I have tested on ubuntu 12.04, mac os x 10.9, and windows with various default jvms.


cos:
looks good. deleting the arch classifier is cleaner.

PR #1030, opened 2014-10-03
Remove more obsolete code

yilongli:
Removed some obsolete code for z3 interaction and path indexing. Path indexing was good work, however, we eventually decided to go a different route. Therefore, the code related to path indexing has not been maintained for a while and is probably broken.

@dwightguth please review


dwightguth:
This looks good but I think @andreistefanescu should approve what you did to KILtoZ3

yilongli:
@andreistefanescu what do you think?

andreistefanescu:
I think we could remove that code, and if we ever want to use it again, we can get it from the history.

PR #1031, opened 2014-10-03
modularize maude backend as test case for the AbstractKModule class

dwightguth:
**please review**


dwightguth:
@cos I know this isn't your favorite implementation, but it works pretty well and does the trick of hiding dependency injection from the user mostly. Can you look it over and see if you see any easy improvements or fixes?

cos:
It looks good. While I think it is still more involved for the plugin developer then it could be, not requiring Guice knowledge is a very significant improvement. I have no detail suggestion. From my perspective, merge.

grosu:
Cosmin (and others), if you have ideas on how to do things better, you are encouraged to try them out!  However, make sure of two things:

1) You do not slow the team down in the short term (remember the golden rule of advising: anything takes double the next category of time than the student says it takes; e.g., if the student says it takes 2 days, then I should plan for 4 weeks); and even more importantly,

2) You should not slow Dwight down with what he's doing in order to try your ideas out.  He is responsible for a large project, as well as refactoring another one, and has lots of things to take care of.  so if you are convinced something different from his approach works, then try it out yourself and bring something tangible on the table.  We all have ideas, too many, what we don't have is the time to work them out.

Cheers,
Grigore

PR #1036, opened 2014-10-06
fix bug with profiles in maven 3.2.2

dwightguth:
@traiansf please review


traiansf:
I've tested it on both Windows (maven 3.2.1) and on my Scientific Linux VM (maven 3.2.3) and seems ok.  So... lgtm.

PR #1040, opened 2014-10-07
Fix a bug in circularity detection of poset.

radumereuta:
I'm not sure how we didn't detect this one until now, but it was a bug in the actual algorithm.
@dwightguth please review.


dwightguth:
Looks good to me. Can you please write a unit test?

radumereuta:
Right... sorry, forgot about it. I've added a test that fails without the fix.

dwightguth:
Merge if tests pass

PR #1044, opened 2014-10-08
further improve code related to PatternMatcher

yilongli:
Nothing really interesting, mostly refactoring.


yilongli:

dwightguth:
lgtm. @andreistefanescu merge if you are happy

andreistefanescu:
Looks good. There a lot of code movement thought, so I might have missed something... My only question is whether the recursive pattern matcher is still used anywhere.

yilongli:
Yes, it's still used in searching because the non-recursive one cannot handle AC matching yet.

PR #1052, opened 2014-10-09
a ktest bug

osa1:
This fixes a bug.

 KTest tries to compile the file if `--skip kompile` is provided, but definition is not kompiled. (e.g. `-kompiled` directory doesn't exist)

Previously it was ignoring `-d` argument specified in `<test ... />` in config files and was looking to the directory that definition file resides. With this patch it's now reading `-d` and checking correct place to see if definition is not compiled yet.


dwightguth:
Nice! does this also make it so that the log strings output by ktest have absolute directory paths when the config xml specifies a directory?

osa1:
Yes, that's happened to be a nice side effect. (it was previously relative path which is not as useful)

dwightguth:
lgtm. If you can confirm that you tested the behavior you just mentioned, and that you have observed the log output of the Jenkins build to make sure that it really is doing what you just described, you can merge this.

PR #1053, opened 2014-10-09
replace IntSeq with List

andreistefanescu:


andreistefanescu:
@yilongli please review
@dwightguth please see why checkstyle fails for java files with lines ending in "->"

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinListOperations.java
```diff
@@ -13,8 +14,21 @@
  */
 public class BuiltinListOperations {
 
-    public static Term constructor(BuiltinList term1, BuiltinList term2, TermContext context) {
-        return BuiltinList.concatenate(term1, term2);
+    public static Term constructor(Term list1, Term list2, TermContext context) {
+        if (list1.sort() != Sort.LIST || list2.sort() != Sort.LIST) {
```
yilongli, line 18: I am not sure about this check. Should we let `BuiltinList#concatenate` to decide what is illegal?

java-backend/src/main/java/org/kframework/backend/java/kil/AssociativeCommutativeCollection.java
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) 2014 K Team. All Rights Reserved.
 package org.kframework.backend.java.kil;
 
+import java.util.List;
```
yilongli: useless?

java-backend/src/main/java/org/kframework/backend/java/kil/BuiltinMap.java
```diff
@@ -238,6 +264,9 @@ public Term build() {
                     patternsBuilder.build(),
                     functionsBuilder.build(),
                     variablesBuilder.build());
+            //return builtinMap.baseTerms().size() == 1 && builtinMap.concreteSize() == 0 ?
```
yilongli, line 278: why not use the new code? It looks right to me

java-backend/src/main/java/org/kframework/backend/java/kil/Collection.java
```diff
@@ -41,6 +44,30 @@ protected Collection(Variable frame, Kind kind) {
         this.frame = frame;
     }
 
+    public Term toLabelRepresentation(TermContext context) {
```
yilongli: Javadoc

java-backend/src/main/java/org/kframework/backend/java/kil/Collection.java
```diff
@@ -41,6 +44,30 @@ protected Collection(Variable frame, Kind kind) {
...
+        for (int i = components.size() - 2; i >= 0; --i) {
+            result = KItem.of(
+                    KLabelConstant.of(sort.constructorLabel(), context.definition().context()),
+                    KList.concatenate(components.get(i), result),
+                    context);
+        }
+        return result;
+    }
+
+    abstract protected List<Term> getLabelRepresentationComponents(TermContext context);
```
yilongli: javadoc

java-backend/src/main/java/org/kframework/backend/java/kil/KCollection.java
```diff
@@ -105,6 +105,11 @@ public String toString() {
     }
 
     @Override
+    protected List<Term> getLabelRepresentationComponents(TermContext context) {
```
yilongli: ``` final```?
```

File: java-backend/src/main/java/org/kframework/backend/java/symbolic/NonACPatternMatcher.java [hunk 7]
```diff
@@ -88,7 +89,16 @@ public NonACPatternMatcher(boolean matchOnFunctionSymbol, TermContext context) {
         taskBuffer.clear();
         tasks.addFirst(Pair.of(subject, pattern));
         failed = false;
-        return match() ? substitution : null;
+        if (match()) {
+            // TODO(AndreiS): this is an ad-hoc evaluation
+            if (termContext.definition().context().krunOptions != null
```
yilongli, line 95: it is not related to your change but it seems unnatural to have to check if `krunOptions` is null

java-backend/src/main/java/org/kframework/backend/java/symbolic/NonACPatternMatcher.java
```diff
@@ -88,7 +89,16 @@ public NonACPatternMatcher(boolean matchOnFunctionSymbol, TermContext context) {
...
+                    && termContext.definition().context().krunOptions.experimental.prove() != null) {
+                PatternMatcher.evaluateSubstitution(substitution, termContext);
```
yilongli, line 97: can we move this to somewhere else such as `KItem`? BTW, when do we need to evaluate substitution?

java-backend/src/main/java/org/kframework/backend/java/symbolic/NonACPatternMatcher.java
```diff
@@ -417,6 +433,11 @@ private void match(KLabelInjection kLabelInjection, KLabelInjection pattern) {
         addMatchingTask(kLabelInjection.term(), pattern.term());
     }
 
+    private void match(BuiltinList builtinList, BuiltinList pattern) {
```
yilongli, line 437: why do we have to convert the two builtin lists to label representation in order to match them? for convenience?

java-backend/src/main/java/org/kframework/backend/java/symbolic/PrePostTransformer.java
```diff
@@ -276,8 +276,7 @@ public ASTNode transform(BuiltinList builtinList) {
         }
         assert astNode instanceof BuiltinList : "preTransformer should not modify type";
         builtinList = (BuiltinList) astNode;
-        builtinList = (BuiltinList) super.transform(builtinList);
-        return builtinList.accept(postTransformer);
+        return ((JavaSymbolicObject) super.transform(builtinList)).accept(postTransformer);
```
yilongli, line 279: I never like `PrePostTransformer`. It's slow and too complicated. I think it's only good for some quick experiment. Do you use it for substitution and/or evaluation? If so, I would like to reimplement the transformer deriving `CopyOnWriteTransformer` directly.

yilongli:
lgtm except for the small things I mentioned. The specifications look much nicer.

andreistefanescu:
Jenkins: test this please

java-backend/src/main/java/org/kframework/backend/java/kil/BuiltinMap.java
andreistefanescu, line 278: there are still problems with multiple representations of maps (so I left it commented out for now).

java-backend/src/main/java/org/kframework/backend/java/symbolic/NonACPatternMatcher.java
andreistefanescu, line 95: well, this code can be executed from outside krun; maybe @dwightguth want to weight in?

java-backend/src/main/java/org/kframework/backend/java/symbolic/NonACPatternMatcher.java
andreistefanescu, line 437: yes, just for convenience; we will change this when we implement support for A and AC

java-backend/src/main/java/org/kframework/backend/java/symbolic/NonACPatternMatcher.java
andreistefanescu, line 97: this is because of converting to KLabel/KList format for A matching, so an evaluation is required after to convert back to builtin representation

java-backend/src/main/java/org/kframework/backend/java/symbolic/NonACPatternMatcher.java
dwightguth, line 95: yes, I will weigh in. This is fine with me for now. Eventually, as the code base mutates and evolves, I expect I will want to see the boolean value of whether we are proving something be an injected dependency, and then we would just inject false into kompile.

andreistefanescu:
@yilongli @dwightguth please review

dwightguth:
lgtm. You can merge if you address Yilong's comments.

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinListOperations.java
yilongli, line 18: what do you think about this? I think it makes sense to concat a List and a ListItem.

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinMapOperations.java
```diff
@@ -28,7 +29,10 @@
  */
 public class BuiltinMapOperations {
 
-    public static Term constructor(BuiltinMap map1, BuiltinMap map2, TermContext context) {
+    public static Term constructor(Term map1, Term map2, TermContext context) {
+        if (map1.sort() != Sort.MAP || map2.sort() != Sort.MAP) {
+            throw new IllegalArgumentException();
```
yilongli, line 26: same as my comment in `BuiltinListOperations`

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinSetOperations.java
```diff
@@ -15,7 +16,10 @@
  */
 public class BuiltinSetOperations {
 
-    public static Term constructor(BuiltinSet set1, BuiltinSet set2, TermContext context) {
+    public static Term constructor(Term set1, Term set2, TermContext context) {
+        if (set1.sort() != Sort.SET || set2.sort() != Sort.SET) {
+            throw new IllegalArgumentException();
```
yilongli, line 21: same in `BuiltinListOperations`

java-backend/src/main/java/org/kframework/backend/java/compile/DataStructureToLookupUpdate.java
```diff
@@ -125,7 +125,9 @@ public ASTNode visit(Rule node, Void _)  {
                + "DataStructureToLookupUpdate pass should be applied after ResolveRewrite pass.";
 
         if (context.krunOptions != null && context.krunOptions.experimental.prove() != null
-                || node.containsAttribute(Attribute.PATTERN_FOLDING_KEY)) {
+                || node.containsAttribute(Attribute.PATTERN_KEY)
```
yilongli, line 128: maybe it's time to make a category for rules that are used only in the matching logic prover so that you don't have to modify the check every time you have some new type of rule

java-backend/src/main/java/org/kframework/backend/java/kil/BuiltinList.java
```diff
@@ -439,7 +487,9 @@ public Term build() {
                     elementsRightBuilder.build(),
                     baseTermTypesBuilder.build(),
                     listVariablesBuilder.build());
-            return builtinList.hasFrame() && builtinList.concreteSize() == 0 ? builtinList.frame : builtinList;
+            return builtinList.baseTerms().size() == 1 && builtinList.concreteSize() == 0 ?
```
yilongli, line 490: I think it's still problematic. If `builtinList.isConcreteCollection() && builtinList.concreteSize() == 1`, then we should return that only element.

java-backend/src/main/java/org/kframework/backend/java/kil/BuiltinMap.java
yilongli, line 278: But it's at least more correct than the existing code, right? I write a lot of code that depends on the correctness of canonical form of terms. Could you take a look at my comment in `BuiltinList.Builder#build` and do the same thing for it?

java-backend/src/main/java/org/kframework/backend/java/kil/BuiltinSet.java
```diff
@@ -201,7 +228,9 @@ public Term build() {
                     patternsBuilder.build(),
                     functionsBuilder.build(),
                     variablesBuilder.build());
-            return builtinSet.hasFrame() && builtinSet.elements.isEmpty() ? builtinSet.frame : builtinSet;
+            return builtinSet.baseTerms().size() == 1 && builtinSet.concreteSize() == 0 ?
```
yilongli, line 231: same comment as `BuiltinList.Builder#build()`

java-backend/src/main/java/org/kframework/backend/java/kil/Collection.java
```diff
@@ -42,6 +45,40 @@ protected Collection(Variable frame, Kind kind) {
     }
 
     /**
+     * Returns a KORE (KLabel/KList) representation of this collection. The returned representation
+     * is not unique (due to associativity/commutativity).
+     * {@link org.kframework.backend.java.kil.Term#evaluate} is the inverse operation.
+     */
+    public Term toLabelRepresentation(TermContext context) {
```
yilongli: This looks like the `toK` function in the K builtin documentation, right? If so, I think it's better to keep the name conform to that documentation.

java-backend/src/main/java/org/kframework/backend/java/kil/Collection.java
```diff
@@ -42,6 +45,40 @@ protected Collection(Variable frame, Kind kind) {
...
+        }
+        return result;
+    }
+
+    /**
+     * Returns a list aggregating the base terms and the elements/entries of this collection.
+     * Each collection is responsible for representing its elements/entries in KLabel and KList
+     * format.
+     */
+    abstract protected List<Term> getLabelRepresentationComponents(TermContext context);
```
yilongli: maybe a shorter name related to `toK`?

java-backend/src/main/java/org/kframework/backend/java/kil/BuiltinMap.java
```diff
@@ -62,6 +65,10 @@ public boolean isUnifiableByCurrentAlgorithm() {
         return collectionFunctions.isEmpty() && collectionVariables.size() <= 1;
     }
 
+    public boolean hasGroundKeys() {
+        return entries.keySet().stream().allMatch(Term::isGround);
```
yilongli, line 76: the method name is a little bit confusing at first sight because I thought it only checked whether there is at least one ground key

java-backend/src/main/java/org/kframework/backend/java/symbolic/NonACPatternMatcher.java
yilongli, line 97: can you add your comment above in the code?

java-backend/src/main/java/org/kframework/backend/java/symbolic/PatternMatcher.java
```diff
@@ -770,4 +779,10 @@ public String getName() {
         return this.getClass().toString();
     }
 
+    public static void evaluateSubstitution(Map<Variable, Term> substitution, TermContext context) {
+        for (Map.Entry<Variable, Term> entry : substitution.entrySet()) {
```
yilongli, line 782: since this can be done in one line using `forEach`, I prefer to not add another public static helper method to the `PatternMatcher` class.

java-backend/src/main/java/org/kframework/backend/java/symbolic/PatternMatcher.java
```diff
@@ -289,6 +289,15 @@ public void match(BuiltinList builtinList, Term pattern) {
             this.fail(builtinList, pattern);
         }
 
+        if (matchOnFunctionSymbol) {
+            builtinList.toLabelRepresentation(termContext).accept(
+                    this,
+                    ((BuiltinList) pattern).toLabelRepresentation(termContext));
+            // TODO(AndreiS): this is an ad-hoc evaluation
+            evaluateSubstitution(fSubstitution, termContext);
```
yilongli, line 296: what does it mean? can you be more specific about the "ad-hoc evaluation" in the TODO item?

java-backend/src/main/java/org/kframework/backend/java/symbolic/UseSMT.java
```diff
@@ -44,7 +44,7 @@ public Term checkSat(Term term, TermContext termContext) {
             com.microsoft.z3.Context context = new com.microsoft.z3.Context();
             Solver solver = context.mkSolver();
             BoolExpr query = context.parseSMTLIB2String(
-                    KILtoSMTLib.translateConstraint(constraint),
+                    "(declare-sort IntSeq)" + KILtoSMTLib.translateConstraint(constraint),
```
yilongli, line 48: why not do it in the smt prelude file?

yilongli:
lgtm overall. Since I am in the process of cleaning up the ml prover, I am being a bit more conservative about the changes than usual.

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinListOperations.java
andreistefanescu, line 18: There is no ListItem sort. So I think this is the only way concatenate would typecheck.

java-backend/src/main/java/org/kframework/backend/java/kil/BuiltinList.java
andreistefanescu, line 490: The list element sort is not subsorted to the list sort. So you want to return `ListItem(X)` (as an instance of BuiltinList with just one element), not just `X`.

java-backend/src/main/java/org/kframework/backend/java/kil/Collection.java
andreistefanescu: Sure, I can rename it.

java-backend/src/main/java/org/kframework/backend/java/kil/BuiltinList.java
yilongli, line 490: yeah, right

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinListOperations.java
yilongli, line 18: I see, but why not just pass the arguments to `BuiltinList#concatenate` and let it to do the typecheck? `BuiltinList.Builder#concatenate(Term)` already does the check.

PR #1057, opened 2014-10-10
Kast in k

dwightguth:
I created this diff for use in managing the discussion on the KORE syntax. We may choose to close it rather than merge it, I just need a place to put comments.


samples/kast/kore.k
```diff
@@ -0,0 +1,146 @@
...
+
+// Known issue: The attributes do not yet parse as K terms
+
+// Recall that in KORE productions must have one of the attributes klabel, token, bracket
+
+
+// Module defining only the sorts K and KString, useful for modularity
+module BASIC-K
+  syntax K   //[...]   // no constructors at this stage
+  syntax KString ::= r"[\\\"]([^\\\"\\n\\r\\\\]|[\\\\][nrtf\\\"\\\\])*[\\\"]"   [token, noAutoReject, onlyLabel, hook(org.kframework.kore.KString)]
```
dwightguth: this needs support for unicode. I suggest you take a look at the current syntax for the String builtin in K.

samples/kast/kore.k
```diff
@@ -0,0 +1,146 @@
...
+  syntax K     ::= KItem                                 
+
+  syntax KItem  ::= "#token" "(" KString "," KString ")"  [ klabel(#KToken), hook(org.kframework.kore.KToken) ]
+  // @cos proposes to eliminate the next injection, and instead to allow
+  // the use of meta-level KAST in definitions: #KApply(#token("_+_","KLabel"),#KList(1,3))
+  // @grosu is skeptic: that would imply to allow KAST syntax in programs; not worth the price
+  syntax KItem ::= "#klabel" "(" KLabel ")"              [ klabel(#WrappedKLabel), hook(org.kframework.kore.meta.WrappedKLabel) ]
+  syntax KItem ::= KLabel "(" KList ")"                  [ klabel(#KApply), hook(org.kframework.kore.KApply) ]
+
+  syntax KLabel ::= r"`(\\\\`|\\\\\\\\|[^`\\\\\\s])+`"   [ klabel(#KLabel), token ] // everything except `, unless escaped \`
```
dwightguth: I don't see us using single quotes for anything here. Might we use single quotes to delimit KLabel in order to avoid the ambiguity with the K bracket?

samples/kast/kore.k
```diff
@@ -0,0 +1,146 @@
...
+  syntax KLabel ::= r"[#a-z][a-zA-Z0-9]*"                [ klabel(#KLabel), token, hook(org.kframework.kore.KLabel) ] // something that doesn't collide with meta-variables
+// KLabel brackets not needed if @cos' approach above works
+//  syntax KLabel ::= "`" KLabel "`"   [ bracket, hook ]
+
+  syntax KList ::= K                
+  syntax KList ::= ".::KList"       [klabel(#EmptyKList), hook(org.kframework.kore.EmptyKList) ]
```
dwightguth: I still think it's better to have one implementation class for both this and the following case. An empty KList should be, in the implementation, simply a KList without any elements.

samples/kast/kore.k
```diff
@@ -0,0 +1,146 @@
...
+  syntax KList ::= KList "," KList  [klabel(#KList), left, assoc, unit(#EmptyKList), hook(org.kframework.kore.KList) ]
+endmodule
+
+
+
+// To be used when pretty-printing stuck computations
+module KSEQ
+  imports KAST
+  syntax K     ::= ".::K"                                [ klabel(#EmptyK), hook(org.kframework.kore.EmptyK) ]
```
dwightguth: see above

samples/kast/kore.k
```diff
@@ -0,0 +1,146 @@
...
+  syntax K     ::= ".::K"                                [ klabel(#EmptyK), hook(org.kframework.kore.EmptyK) ] 
+  syntax K     ::= K "~>" K                              [ klabel(#KSequence), left, assoc, unit(#EmptyK), hook(org.kframework.kore.KSequence) ]
+endmodule
+
+
+
+// To be used to parse semantic rules
+module K
+  imports KSEQ
+  syntax KVariable ::= r"(?<![A-Za-z0-9\\_])(\\$|\\!|\\?)?([A-Z][A-Za-z0-9']*|_)" [ hook(org.kframework.kore.KVariable) ] // same like KSort?
```
dwightguth: @radumereuta can you confirm you're okay with this regular expression and that it's sufficiently compatible with the one we currently use to support everything we care about?

samples/kast/kore.k
```diff
@@ -0,0 +1,146 @@
...
+  syntax KItem     ::= KVariable        
+  syntax KLabel    ::= KVariable        
+  syntax K         ::= K "=>" K         [ klabel(#KRewrite), hook(org.kframework.kore.KRewrite) ]
+  // K bracket needed, e.g., Cxt[V ~> `env(E) => .`]
+  syntax K         ::= "`" K "`"        [ bracket ]
```
dwightguth: I'd still feel better with a matched delimiter for brackets. It would make it much easier for editors to deal with, for one thing.

samples/kast/kore.k
```diff
@@ -0,0 +1,146 @@
...
+  syntax KImport       ::= "imports" KModuleName    [ klabel(#KImport), hook(org.kframework.kore.outer.Import) ]
+  syntax KImportList   ::= KImport KImportList      [ klabel(#KImportList), assoc, comm, unit(#emptyKImportList) ]
+  syntax KImportList   ::= ""                       [ klabel(#emptyKImportList), onlyLabel]
+
+  syntax KSentenceList ::= KSentence KSentenceList      [ klabel(#KSentenceList), assoc, comm, unit(#kemptyKSentenceList) ]
+  syntax KSentenceList ::= ""                           [ klabel(#emptyKSentenceList), onlyLabel]
+
+  syntax KSentence ::= "syntax" KSort                             KAttributesDeclaration   [ klabel(#KSyntaxSort), hook(org.kframework.kore.outer.SyntaxSort) ]
+  syntax KSentence ::= "syntax" KSort "::=" KProduction           KAttributesDeclaration   [ klabel(#KSyntaxProduction), hook(org.kframework.kore.outer.SyntaxProduction) ] ]
+  syntax KSentence ::= "syntax" "priority" KString ">" KString    KAttributesDeclaration   [ klabel(#KSyntaxPriority), hook(org.kframework.kore.outer.SyntaxPriority) ]
```
dwightguth: it's not entirely clear to me how you intend to use this to capture more complex priority declarations like \* / > + -

And you're still missing associativity declarations

samples/kast/kore.k
```diff
@@ -0,0 +1,146 @@
...
+  syntax KImportList   ::= ""                       [ klabel(#emptyKImportList), onlyLabel]
+
+  syntax KSentenceList ::= KSentence KSentenceList      [ klabel(#KSentenceList), assoc, comm, unit(#kemptyKSentenceList) ]
+  syntax KSentenceList ::= ""                           [ klabel(#emptyKSentenceList), onlyLabel]
+
+  syntax KSentence ::= "syntax" KSort                             KAttributesDeclaration   [ klabel(#KSyntaxSort), hook(org.kframework.kore.outer.SyntaxSort) ]
+  syntax KSentence ::= "syntax" KSort "::=" KProduction           KAttributesDeclaration   [ klabel(#KSyntaxProduction), hook(org.kframework.kore.outer.SyntaxProduction) ] ]
+  syntax KSentence ::= "syntax" "priority" KString ">" KString    KAttributesDeclaration   [ klabel(#KSyntaxPriority), hook(org.kframework.kore.outer.SyntaxPriority) ]  
+  
+  syntax KSentence ::= "configuration" K "requires" K             KAttributesDeclaration   [ klabel(#KConfiguration), hook(org.kframework.kore.outer.Configuration) ]
```
dwightguth: why no ensures? what if I want the definition to always start out assuming something is true?

dwightguth:
kore.k is still missing several pieces of syntax used by K. Cells, freezers, the hole... If you don't want to include these in this syntax, you need to provide a detailed and exhaustive mechanical translation from all of this syntax into kore. I'd also like to see some kind of indication that you've thought through how you intend to deal with Map, List, Set, Bag, String, Int, Float, etc, and their attendant details. Is there going to be room in this API to convert certain KLabels and tokens into privileged data structures in the implementation?

samples/kast/kore.k
grosu: What do you mean by "one implementation class"?  Are you referring to the hook?  If yes, that is @cos' territory.  If not, it is mine and I'd like to better understand what you mean.

samples/kast/kore.k
grosu: @dwightguth We know there are still details like this to be taken care of; in the interest of time-to-market, we decided to pick whatever regexes were good enough to parse the three examples we have in this folder, and be prepared to refine them later.  Also, I really dislike Java's approach to write double backslashes in regexes, only because it treats them as ordinary strings.  Scala has something like this `raw"a\nb"`, which is equivalent to the string `"a\\nb"`.  We may also adopt something like that.  But, again in the interest of time, we decided that these are relatively small issues that can be fixed after the deployment of KORE.

samples/kast/kore.k
dwightguth: I mean the hook yes

samples/kast/kore.k
grosu: Which one?  Parentheses and brackets are taken, almost all languages use them.  Backtick is typically available, and we only reserve one character now; hard to improve over that.  Brackets are going to be rarely used in general, anyway.  The analogy to markdown's "bubbles" is also nice and intuitive.

samples/kast/kore.k
grosu: priorities form a partial order, and we can define any partial order one pair at a time.

@radumereuta How about assoc declarations, what would a good, minimal KORE syntax for them be?

samples/kast/kore.k
grosu: @dwightguth good point.  I'll add it.

samples/kast/kore.k
dwightguth: I'm not sure you understand what I'm saying. It's _really bad_ to use unmatched delimiters like this because unlike quotation marks, brackets nest. Therefore you have an unresolvable ambiguity in the following term:

`` Foo ` Bar ` Baz ``

Is this `(` Foo `) Bar (` Baz `)` or `(` Foo (` Bar`) Baz`)` ?

There is literally no way for an editor to decide something like this, so it simply cannot guess where the matching end bracket is for a begin bracket. And that is _really important_. We are completely destroying the editor experience of programming in K if you can't do that, and suddenly complex rules with lots of brackets become completely impossible to decipher.

samples/kast/kore.k
dwightguth: Yes, using a potentially quadratic amount of syntax. I guess this is good
enough for now, but we need to be aware that we might have to change it if
the performance becomes too degraded in cases where the polynomial
explosion is prohibitive

samples/kast/kore.k
grosu: @dwightguth that is not captured in this syntax, but it was discussed extensively in meetings and on the K wiki of KORE: we will use the markdown approach for nesting: ``` `foo` bar `baz` `` abc `` `foo` bar `baz` ``` is the same as `((foo) bar (baz) ) abc ((foo) bar (baz))`.  But such cases of nested K brackets are extremely rare, so it is OK for people to write a bit more when the case arises. 

So the correct syntax here should be something like this:

``````
syntax K         ::= "```" K "```"  [bracket, prefer]
                   |  "``" K "``"   [bracket, prefer]
                   |   "`" K "`"    [bracket] 
``````

samples/kast/kore.k
cos: I want to keep the hooks named as the `KLabel`s. `EmptyKList` is simply an alternate constructor for an empty `KList`.

dwightguth: Do they create the same resulting object? If so, I'm okay with the synonym. I just don't want there to be a difference between an EmptyKList and an empty KList

samples/kast/kore.k
cos: @radumereuta ?

samples/kast/kore.k
cos: Yes, both generate a `KList` object. An overloaded `KList` constructor would be slightly better, but we're following the convention that hooks have the same name as the `KLabel`s.

samples/kast/kore.k
cos: Yes, things would be cleaner if we allow a priority list instead of just a pair. Shall we make this change now and settle it for the future?

radumereuta: I like this idea. We don't need the absolute minimal representation. We also want the definition to be somewhat readable. At one point we have to debug it. I would like to be able to make sense of what's inside, especially if we're talking about quadratic behaviors.

samples/kast/kore.k
radumereuta: I have the same concerns as @dwightguth, but I am willing to give it a try with backticks.

samples/kast/kore.k
radumereuta: We can take care of the details later. Let's make a minimal functioning prototype first.

samples/kast/kore.k
grosu: ok then.  let's make a distinction between minimality where the asymptotic size of the code does not change, and minimality at the expense of non-liniar code explosion.

dwightguth:
Closing since discussion is complete.

Issue #1059, opened 2014-10-10
Should different backends use different temporary directories?


osa1:
I'm currently working on improving KTest's parallelism. One problem that I'm currently having is when I run `kompile definition.k` and `kompile definition.k --backend pdf`at the same time, they're having a race, whichever terminates first is removing the `.k` directory and other one is failing.

This is not a PDF backend specific problem though. If Java backend also uses temporary directory then `kompile definition.k` and `kompile definition.k --backend java` should also be a problem.

I'd like to discuss what to do about this. I think one easy way to fix this is to make backends generate temp directories with different names.

We briefly discussed about a filesystem API in #1008 . Maybe we can implement that too. Using the API we can make sure that backends won't be able to read from/write to arbitrary files, they'll be restricted with their temporary directory, which is generated by filesystem API by reading some fields of backend(to decide on a directory name). This will also help cleaning up after backends are done.

Any ideas?

radumereuta:
Well... the parsing is common for all backends, and it would be best if it's stored in the same place for caching reasons.
Maybe we should implement a mutex at the file level...

dwightguth:
Since Grigore wants us to do something about the temp directory in the
first place, here's what I propose you do:
1. make the temp directory an actual temp directory, ie made via
   Files.createTempDirectory
2. If --debug is not passed, do the same thing as before and delete the
   temp directory.
3. If --debug is passed, do not delete the temp directory, but output the
   path to it on the command line.

osa1:
> If --debug is passed, do not delete the temp directory, but output the
> path to it on the command line.

I don't have a better alternative but this part will lead to more verbose ktest output.

PR #1060, opened 2014-10-10
More parallel ktest

osa1:
[WIP]

This turned out to be more involved then I expected, because ktest was designed for this. We knew that it's possible to improve concurrency of it but we decided to go with simpler approach at the time.

This is blocked by #1059 .

TODO:
- [x] KTest should check for required `-kompiled` directories when `--skip kompile` is provided, otherwise krun tests will fail. (this was working before this patch)
- [x] Posix-only steps are not implemented.

I'll organize commits before merging this.


osa1:
Hmm, build failed with:

```
ERROR: ['/var/lib/jenkins/jobs/k-framework-pull-requests-maven/workspace/k/k-distribution/target/release/k/bin/krun' '/var/lib/jenkins/jobs/k-framework-pull-requests-maven/workspace/k/k-distribution/samples/java_rewrite_engine/tutorial/2_languages/1_simple/1_untyped/verification/comm_assoc.simple' '--prove' 'verification/comm_assoc_spec.k' '--directory' 'verification'] failed with error (time: 80528 ms)
error was: Aborted (core dumped)
```

Does this look familiar to anyone? @yilongli ?

yilongli:
The z3 library seems to crash again. Adding `--z3-executable` to the krun option should solve the problem.

osa1:
This is ready for reviews.

The problem with Z3 is disappeared, but some others appeared randomly. I think we'll have that Z3 problem randomly again in the future, maybe we can have a different issue for tracking that.

I'll try to figure why some other tests randomly failed after some completely unrelated commit.

I think we had good improvement. Not a very precise benchmark, but Jenkins took 41 min before that and 34 min with this patch.

Another improvement is that this is now safe: https://github.com/osa1/k/commit/8653e49ae11c153b42c795b02cc4fe89fda6c0d8 When detected multiple test cases that compile to same directory, ktest compiles only once. So we can now distribute tests for same definition to multiple files.

I guess only disadvantage of this approach is that we have slightly worse ktest output. This is because we're not running different steps(scripts, ktest, pdf, krun) in order, they now all interleaved. (only ordering is that in a test case kompile process runs before kruns) This also effects statistics. For example, we can't know how long kompile steps took because kompile steps are interleaved with PDFs and kruns etc.

For reviewers:

There is actually two things going on:
1. A new class `TaskQueue` which does most of the job. I recommend reviewing it from the file instead of the diff.
2. Rest is just moving things from `TestSuite` to `TestCase` and minor things like passing an extra argument to `KRunProgram`, introducing some new fields etc.

It'd also be nice if I could get some feedback on current KTest output (you can see it in Jenkins page). It's mostly same, a bit less verbose but also less informative as mentioned above.

osa1:
Again, I can't reproduce errors in both rvwork and my laptop. @yilongli do they look familiar to you by any chance?

yilongli:
So there are two failures. The first one is another z3 library crash. Adding `--z3-executable` should solve the problem. The second failure seems like an insufficient timeout limit of z3 because the formula in the final configuration is clearly unsat. You can try to set a larger `--z3-cnstr-timeout` for this example (default is 50ms).

osa1:
Thanks Yilong.

This is now working and ready for reviews/merge.

dwightguth:
Jenkins: test this please

dwightguth:
Since this test concerns both ktest and threading, I decided to run it a second time just in case it catches some nondeterministic errors. If the tests pass the second time, go ahead and merge this. We'll deal with further issues if they arise.

PR #1061, opened 2014-10-10
switch to new universal distribution format

dwightguth:
FYI @bmmoore @grosu 

@yilongli @cos please review

### DIFF
--- a/k-core/pom.xml
+++ b/k-core/pom.xml
@@ -85,12 +85,38 @@
       <artifactId>mpfr_java</artifactId>
       <version>1.0.1</version>
     </dependency>
+
     <dependency>
       <groupId>org.kframework.mpfr_java</groupId>
       <artifactId>mpfr_java</artifactId>
       <version>1.0.1</version>
-      <classifier>${native.classifier}</classifier>
+      <classifier>linux64</classifier>
     </dependency>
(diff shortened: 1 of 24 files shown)


grosu:
@dwightguth Is this about packaging K as _one_ distribution for all platforms, instead of a different distribution for each platform?

dwightguth:
Yes

grosu:
Great!  Then it LGTM if it works.

yilongli:
lgtm though I don't think I am very qualified to review these changes.

k-core/pom.xml
```diff
@@ -307,12 +356,6 @@
         </os>
```
cos, line 356: we're installing cygwin for windows users, that's customer service! :-)

pom.xml
```diff
@@ -199,224 +199,91 @@
     </plugins>
   </build>
   <profiles>
```
cos, line 201: if we only generate one distribution, why do we still need profiles here?

dwightguth, line 201: there is a universal distribution, but not a universal _build_. A universal build is actually impossible because we have to run native executables to generate the sdf sources.

cos:
lgtm.

PR #1062, opened 2014-10-11
Cleanup ml prover 3

yilongli:
I have been able to track down the root cause that makes the merge sort example fails from time to time. However, it is a bit hard to fix it before cleaning up several important classes. This pull request is the first step towards this direction. It consists of many bug fixes and small refactorings. I also try to make the prover's behavior more deterministic so that the merge sort example fails less often.


yilongli:

yilongli:
@cos I also had to remove some `Data` classes you added because they made my refactoring too complicated. However, when I am done with the cleanup of the prover in a month or so, many things will be much simpler and, hopefully, you don't even need these `Data` classes.

cos:
I agree the `Data` classes are ugly but they are an intermediate step to a purpose: separate "stuff that is specific to the term/constraint" from "stuff that is there because it was easier that way at some point", e.g., Context. The latter "stuff" prevents some otherwise simple things like returning to a previous state after a few rewrite steps -- you end up in an inconsistent state. If your refactoring will eventually achieve this purpose then I am all for it.

yilongli:
Yes, I think so.

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -507,11 +474,14 @@ private ConstrainedTerm applyRule(ConstrainedTerm constrainedTerm, List<Rule> ru
...
+//                    for (Rule rule : appliedRules) {
+//                        System.err.println(rule.getLocation() + " " + rule.getSource());
+//                    }
+
                     /* add helper rule */
                     HashSet<Variable> ruleVariables = new HashSet<Variable>(initialTerm.variableSet());
                     ruleVariables.addAll(targetTerm.variableSet());
-                    Map<Variable, Variable> freshSubstitution = Variable.getFreshSubstitution(
-                            ruleVariables);
+                    Variable.getFreshSubstitution(ruleVariables);
```
andreistefanescu: does `Variable#getFreshSubstitution` have side effects?

andreistefanescu:
Looks good.

java-backend/src/main/java/org/kframework/backend/java/kil/KilFactory.java
```diff
@@ -1,31 +1,25 @@
 // Copyright (c) 2014 K Team. All Rights Reserved.
```
dwightguth: if this is how you intend to use this class, I think you can just delete it and inject KILtoBackendJavaKILTransformer directly.

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
yilongli: right, this is the leftover of the automatic refactoring. I will remove it.

PR #1063, opened 2014-10-12
Kompile temp dirs

### LINKED ISSUE #1059: Should different backends use different temporary directories?

osa1:
Resolves #1059 .

### DIFF
--- a/k-core/src/main/java/org/kframework/kompile/KompileFrontEnd.java
+++ b/k-core/src/main/java/org/kframework/kompile/KompileFrontEnd.java
@@ -4,6 +4,7 @@ package org.kframework.kompile;
 import java.io.File;
 import java.io.FilenameFilter;
 import java.io.IOException;
+import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -21,7 +22,6 @@ import org.kframework.parser.DefinitionLoader;
 import org.kframework.utils.BinaryLoader;
 import org.kframework.utils.Stopwatch;
 import org.kframework.utils.errorsystem.KExceptionManager;
-import org.kframework.utils.file.FileUtil;
 import org.kframework.utils.file.JarInfo;
 import org.kframework.utils.inject.JCommanderModule;
 import org.kframework.utils.inject.JCommanderModule.ExperimentalUsage;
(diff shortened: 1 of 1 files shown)


k-core/src/main/java/org/kframework/kompile/KompileFrontEnd.java
```diff
@@ -104,12 +116,13 @@ public void run() {
...
 
         if (backend.generatesDefinition()) {
-                context.kompiled = new File(options.directory, FilenameUtils.removeExtension(options.mainDefinitionFile().getName()) + "-kompiled");
+            context.kompiled = new File(options.directory, FilenameUtils.removeExtension(
+                    options.mainDefinitionFile().getName()) + "-kompiled");
+            if (context.kompiled.exists()) {
                 checkAnotherKompiled(context.kompiled);
-                if (options.mainDefinitionFile().exists()
-                        && !context.kompiled.exists() && !context.kompiled.mkdirs()) {
-                    kem.registerCriticalError("Could not create directory " + context.kompiled);
```
osa1, line 111: This was printing "Could not create directory ..." when definition file is missing.

k-core/src/main/java/org/kframework/kompile/KompileFrontEnd.java
```diff
@@ -104,12 +116,13 @@ public void run() {
         }
 
         if (backend.generatesDefinition()) {
-                context.kompiled = new File(options.directory, FilenameUtils.removeExtension(options.mainDefinitionFile().getName()) + "-kompiled");
+            context.kompiled = new File(options.directory, FilenameUtils.removeExtension(
+                    options.mainDefinitionFile().getName()) + "-kompiled");
+            if (context.kompiled.exists()) {
                 checkAnotherKompiled(context.kompiled);
```
osa1, line 117: This was checked even when parent directory is missing.

osa1:
This is ready for reviews.

radumereuta:
lgtm, but you should expect at least another review.

PR #1064, opened 2014-10-12
remove shared .k generation in maude backend

osa1:
If this works as expected this means that we finally get rid of all dotk directories left on the filesystem.

### DIFF
--- a/k-core/src/main/java/org/kframework/utils/inject/DefinitionLoadingModule.java
+++ b/k-core/src/main/java/org/kframework/utils/inject/DefinitionLoadingModule.java
@@ -33,14 +33,6 @@ public class DefinitionLoadingModule extends AbstractModule {
 
         sw.printIntermediate("Loading serialized context");
 
-        context.dotk = new File(
-                options.definition().getParent() + File.separator
-                        + ".k");
-        if (!context.dotk.exists()) {
-            if (!context.dotk.mkdirs()) {
-                kem.registerCriticalError("could not create directory " + context.dotk);
-            }
-        }
         context.kompiled = options.definition();
 
         sw.printIntermediate("Initializing definition paths");
--- a/maude-backend/src/main/java/org/kframework/backend/maude/krun/MaudeExecutor.java
+++ b/maude-backend/src/main/java/org/kframework/backend/maude/krun/MaudeExecutor.java
@@ -122,7 +122,8 @@ public class MaudeExecutor implements Executor {
         this.maudeOptions = maudeOptions;
 
 
-        krunTempDir = new File(context.dotk, FileUtil.generateUniqueFolderName("krun"));
+        krunTempDir = new File(System.getProperty("user.dir"),
+                FileUtil.generateUniqueFolderName(".krun"));
         inFile = new File(krunTempDir, "maude_in");
         outFile = new File(krunTempDir, "maude_out");
         errFile = new File(krunTempDir, "maude_err");


dwightguth:
I'm not sure I understand what this is doing? I thought the idea was to use Java's built in support for temporary directories, ie, /tmp?

osa1:
@dwightguth if we do that in krun how can we find generated file when it runs in debug mode? We can't print the path because that'd cause problems while testing since ktest always pass --debug.

osa1:
Ready for reviews.

radumereuta:
I also dislike the idea of letting java choose the temporary directory.
I would like to have a place where I can always find the latest temporary files if I use --debug.

osa1:
@radumereuta You can run `ls -a` before and after running `krun --debug` and see which directory is created. Alternatively, if you clean all `.krun___` directories and run `krun --debug`, there will be only one `.krun___` created and that'll be your directory.

traiansf:
I think that letting the OS choose the temporary directory is fine, as long as one has the option to have it displayed on the screen... The worst that could happen is that you would need to run a kompilation/execution twice if you didn't think about adding the option in the first place.

However, it this avoids worst problems such as race conditions on the temporary directory, I would say we should go for it.

osa1:
> However, it this avoids worst problems such as race conditions on the temporary directory, I would say we should go for it.

Yes, the goal is to avoid all kinds of race conditions in both kompile and krun. (for krun, see my other pull request)

osa1:
In addition to avoiding race conditions, all temporary directories are cleaned after these two patches. Previously that was not easily\* possible because removing `.k` was not safe since other krun/kompile processes may be using it.

*: I'm saying "easily" because I think we could still find a way, like checking contents of `.k` to see if it only has current process' files etc. That's still not a great way though, because once we run a krun/kompile with `--debug` there will be at least one other file, so we just keep filling `.k` with more temp files.

radumereuta:
Well... sometimes I want to keep multiple runs. Especially in kompile when I want to check if I'm generating the correct SDF files, and I need to compare with previous versions.
I did a small change a few weeks back, where I would prepend the timestamp on the generated files so I have them in order.
I don't think there is a need to do something clever about the files generated during --debug, as long as they have unique names. It is the users responsibility to delete them. If he/she knows about --debug, then should know about the generated files too.

osa1:
> I don't think there is a need to do something clever about the files generated during --debug, as long as they have unique names.

I'm not doing anything clever in debug mode. Temp files are left on the file system with same names as before(except they're not in `.k` now). You can still have multiple runs and collect temp files as before.

osa1:
Guys, really, I think we're doing bikeshedding here. This is really the most trivial patch this repo has ever seen in it's entire life, but we discussed this for literally days. All it does is that it moves temp files from `.k/temp` to `temp` and that's it. I don't understand what's being a problem. Please tell me if 2 lines of code in this patch is not clear. I explained the motivation for doing so above.

grosu:
Sorry I didn't jump in, I'm overly swamped with stuff for school and other projects.  I don't even know if I'm answering to the right concern/question, so I apologize in advance if I don't.

Just a quick comment: one of the reason we had .k was because we did not want to pollute our user's eye with lots of generated files which are needed only to make things faster.

The idea was that the -kompiled folder contains the actual kompiled language definitions, for example the kompiled C definition, which may take a lot of options and time to tune and generate.  Then we can pass such kompiled definitions to other people to krun, or even package in Valgrind-like tools, who don't care about K at all, but only about the language model that we give them.  So the -kompiled folder is important to be visible to the user's eye.

On the other hand, the .k folder contains caching and stuff that can be generated any time, so we don't need to or want to show it to users.  It is all for tool perfomance reasons, and may even disappear eventually if the tool becomes fast enough.

Dwight, can you please add this to the to-be manual, to make sure we address it there eventually?  Or otherwise delegate it to somebody else.

Thanks,
Grigore

dwightguth:
This looks fine, but I should note that I am probably going to change this code again soon in a diff I am preparing about the file system.

PR #1065, opened 2014-10-13
removed duplicate rules in indexing table

yilongli:
@dwightguth please review

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/indexing/IndexingTable.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/indexing/IndexingTable.java
@@ -1,8 +1,9 @@
 // Copyright (c) 2014 K Team. All Rights Reserved.
 package org.kframework.backend.java.indexing;
 
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Lists;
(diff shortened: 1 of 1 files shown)


yilongli:
Since `ImmutableSet` also has fixed iteration order, it 's ok to replace `List` with it.

dwightguth:
I'd feel better if  you made the variables expect an ImmutableSet instead of a Set, if you are relying on the details of a particular Set implementation.

yilongli:
Well, if the rewrite engine is bug-free, then it doesn't matter if the rules are stored in an ImmutableSet or HashSet. I am only using ImmutableSet for easy debugging.

dwightguth:
Well we actually want a deterministic ordering of rule applications, don't
we? it protects us against issues arising from non confluent definitions as
the framework evolves.

yilongli:
I am not sure that we always want deterministic ordering of rule applications. In the case of non-confluent definitions, shouldn't we expose and fix them as early as possible?

dwightguth:
Well, yes, I agree that we should be able to expose non confluent
definitions. However, I think it will be more effective to do this by means
of a flag to kompile that randomizes the execution order of the rules than
to try to cling to the notion of nondeterministic set iteration, when it's
always the same for any given JVM version.

yilongli:
Sure.

yilongli:
Done. @dwightguth

PR #1066, opened 2014-10-13
separated SymbolicUnifier and SymbolicConstraint

yilongli:
OK, I finally managed to cut the complex interdependence of `SymbolicUnifier` and `SymbolicConstraint`. `SymbolicUnifier` no longer takes a `SymbolicConstraint` in its constructor and `SymbolicConstraint` no longer relies on the `SymbolicUnifer` to update itself. This is quite an important step towards further refactoring of the ml prover.


yilongli:
Alright, not ready for review. It's a little bit tricky here...

yilongli:

Some examples become much slower after https://github.com/yilongli/k/commit/0f4d53da36038a2ca32fcfdb2614cb96264c5f5c. I have inspected the order of rules that are applied and their results. Since the logged traces before and after that commit are the same, I would like to not worry about the performance degradation for now.

k-distribution/samples/tests/config.xml
```diff
@@ -2,81 +2,6 @@
 <!-- Copyright (c) 2013-2014 K Team. All Rights Reserved. -->
 
 <tests>
-  <!-- QUINE -->
```
dwightguth, line 5: revert this?

k-distribution/tests/config.xml
```diff
@@ -2,15 +2,7 @@
 <!-- Copyright (c) 2012-2014 K Team. All Rights Reserved. -->
 
 <tests>
-  <!-- Tutorial Languages and Programs. -->
```
dwightguth, line 5: and this?

k-distribution/samples/tests/config.xml
yilongli, line 5: Sure, but I might need to add a few more commits to make this pull request complete. However, I would like to start getting reviews.

dwightguth:
Okay. I don't have the capacity to review this in detail because I don't really understand the rewriter well enough, but the interface of these two classes does look better to me now.

yilongli:
Alright, this pull request is now complete and I have reverted all changes to the test config files.

yilongli:
@andreistefanescu can I merge this pull request?

andreistefanescu:
looks good

yilongli:
@andreistefanescu I think I have managed to fix the bug revealed by merging with your latest change. Please review my last two commits.

yilongli:
Besides, I notice that performance of the prover on example such as `quicksort` is significantly improved.
Before replacing `IntSeq` with `List` https://github.com/yilongli/k/commit/a75b94e030c3a2bf96ea3fc0f6ef2c6e53d95ac4:

```
real    0m56.031s
user    1m6.480s
sys 0m7.496s

```

Now:

```
real    0m12.810s
user    0m32.050s
sys 0m1.336s
```

andreistefanescu:
Looks good! Please rename `simplify2`.

PR #1070, opened 2014-10-15
add unified interface for accessing file system

dwightguth:
**please review**


dwightguth:
@osa1 @yilongli @cos please review

yilongli:
lgtm

java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaSymbolicKRunModule.java
```diff
@@ -96,22 +100,24 @@ public SimulationModule(List<Module> definitionSpecificModules) {
...
+                    public <T> Void visit(Binding<T> binding) {
+                        Key<T> key = binding.getKey();
+                        if (key.getAnnotation() == null && key.getAnnotationType() == null) {
+                            bind(key.getTypeLiteral()).annotatedWith(Spec.class).to(key.getTypeLiteral());
+                            expose(key.getTypeLiteral()).annotatedWith(Spec.class);
+                        }
+                        return null;
+                    }
+                });
+            }
```
cos, line 115: cool. this is quite generic. could it be pulled up to `AbstractModule`?

cos:
Frankly, I don't like it. I would like to see components depend on fewer things, not more. For example, now, every time I want to create a `JavaSymbolicBackend`, I have to instantiate an entire `FileUtil`, which know about many other things. The backend only needs the definition file. Actually, more precisely, it only needs a definition. Why should it care where that definition comes from in the first place? Even worse, why should it care about temporary directories, working directories, and all other things the `FileUtil` provides? And this happens in several other places. We're mixing configuration with real functionality.

k-core/src/main/java/org/kframework/backend/html/HTMLFilter.java
```diff
@@ -660,15 +658,15 @@ private Color alter(Color a ) {
 
     private void loadProperties() {
         try {
-            FileUtil.loadProperties(Latex2HTMLzero, getClass(), includePath + "Latex2HTMLzero.properties");
+            FileUtil.loadProperties(Latex2HTMLzero, getClass(), "Latex2HTMLzero.properties");
         } catch (IOException e) {
-            System.out.println("error loading " + includePath + "Latex2HTMLzero.properties");
+            GlobalSettings.kem.registerCriticalError("error loading " + "Latex2HTMLzero.properties", e);
```
osa1: minor thing: no need to concatenate strings here.

k-core/src/main/java/org/kframework/main/KPluginClassLoader.java
```diff
@@ -1,43 +0,0 @@
...
-                        addURL(f.toURI().toURL());
-                    } catch (MalformedURLException e) {
-                        e.printStackTrace();
-                        FrontEnd.printBootError(e.getMessage());
-                    }
-                }
-            }
-        }
-    }
-}
```
osa1, line 43: What happened to this module? I can't see a new one replacing this.

osa1:
> For example, now, every time I want to create a JavaSymbolicBackend, I have to instantiate an entire FileUtil, which know about many other things.

As far as I understand you don't need to instantiate `FileUtil`, it's injected by previous parts which is what's good about this patch. Previously every backend had to handle their files by themselves and 1) there were duplicated functionalities everywhere 2) every backend had to handle conventions like temp dir names etc. 3) as I tried to solve in my several previous patches cleanup was harder etc.

This basically moves file handling from backend parts to frontend which I think is right thing to do.

> The backend only needs the definition file. Actually, more precisely, it only needs a definition. Why should it care where that definition comes from in the first place? Even worse, why should it care about temporary directories, working directories, and all other things the FileUtil provides? 

It needs to handle temp files, cache files, provide meaningful error messages etc. there are some places where it needs file paths and some other stuff. I guess those may be further abstracted to some other shared module but that's not a problem with this patch, I think this only improves stuff.

osa1:
lgtm :+1:

cos:
@osa1,

> As far as I understand you don't need to instantiate FileUtil, it's injected by previous parts which is what's good about this patch. 

_Somebody_ does need to instantiate it. The fact it is injected doesn't make it disappear, it only changes who constructs it. If you want to make a unit test you have to either a) wake up the dependency injection framework (with GuiceBerry or something), or b) instantiate the class yourself. This is quite the overhead for a measly unit test.

> every backend had to handle conventions like temp dir names etc.
> By the way, what is wrong with `Files.createTempDirectory` for temp directories?
> 
> It needs to handle temp files, cache files, provide meaningful error messages etc. there are some places where it needs file paths and some other stuff. 

Where precisely does the `JavaSymbolicBackend` need to do all this?

osa1:
> Somebody does need to instantiate it.

How else could it be?

> The fact it is injected doesn't make it disappear, it only changes who constructs it.

That's exactly the point. Current method instantiate it in one place and passes it to other parts. Previous approach was instantiating it in use sites.

> Where precisely does the JavaSymbolicBackend need to do all this?

I don't know. If it's not needed why don't you remove the argument from the constructor? That should solve it, I don't understand what's the big deal.

.gitignore
```diff
@@ -1,4 +1,7 @@
-.k
+.kompile*
+.krun*
+.kast*
+.ktest*
```
osa1, line 4: If we weren't running tests using --debug we could use this info to check whether tests left temp files on the filesystem using something like this:

```
! find . \( -iname ".kompile*" -o -iname ".krun*" -o -iname ".kast*" -o -iname ".ktest*" \) | egrep ".*"
```

cos:
> I don't know. If it's not needed why don't you remove the argument from the constructor? That should solve it, I don't understand what's the big deal.

That's exactly what I'm trying to say :-) We should remove a dependency whenever possible, not move it around.

java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaSymbolicBackend.java
```diff
@@ -69,8 +71,7 @@ public Definition lastStep(Definition javaDef) {
 
         definition.setIndex(index.get());
 
-        loader.saveOrDie(new File(context.kompiled,
-                JavaSymbolicBackend.DEFINITION_FILENAME).toString(),
+        loader.saveOrDie(files.resolveKompiled(JavaSymbolicBackend.DEFINITION_FILENAME),
```
dwightguth, line 74: @cos as you can see the reason why you need to construct a FileUtil for this class is it needs access to the directory associated with the compiled definition. In order to be able to make this dependency more granular, it would be necessary to construct four separate FileUtil classes which each do very similar things with different directories. That seemed like a less clean solution to me than the unfortunate side effect of making this dependency slightly more centralized. Especially since you can simply choose to pass dummy values to the directories you don't want to use, if you really want to. If you can provide me with an alternate solution which is cleaner than this one, or provide me a compelling argument why the solution I just mentioned is worth the increase in code complexity, I am more than willing to change it. I just picked the API that seemed the cleanest for what I needed to be able to do, which was centralize the code which resolves, reads, and writes files in the framework.

java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaSymbolicKRunModule.java
dwightguth, line 115: Can you make an issue for me on this one? You have made a very good suggestion here, but I'd like to do it in a separate pull request to keep this one cleaner. It's a bit of an involved change which affects parts of the code this diff doesn't have to modify. All this is really doing is copy-pasting code already present in KRunModule into JavaSymbolicKRunModule.

k-core/src/main/java/org/kframework/main/KPluginClassLoader.java
dwightguth, line 43: I deleted it because I ended up not using it a while ago. The code as written had issues with the classloaders that meant that the two different classloaders used by the application didn't share classes from the k core. It made the code as written unusable and since I don't know anything about classloaders, I elected to make plugins be placed in the main application's classpath instead and use only a single classloader. I can put this in a separate pull request if you want because it's not really related, but it didn't seem worth the effort to me.

osa1, line 43: OK. I was just wondering what happened.

java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaSymbolicKRunModule.java
cos, line 115: done: #1082

java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaSymbolicBackend.java
cos, line 74: > Especially since you can simply choose to pass dummy values to the directories you don't want to use, if you really want to. 

This is very, very ugly because it is not if I want to, I have to, there is no way around it.

> @cos as you can see the reason why you need to construct a FileUtil for this class is it needs access to the directory associated with the compiled definition. In order to be able to make this dependency more granular, it would be necessary to construct four separate FileUtil classes which each do very similar things with different directories.

I'm starting to think the problem is Guice. If you're thinking about the easiest way to please Guice, yes, you need extra classes to annotate a simple value. One could argue we want to this for better type safety. Ok, but then we should embrace it, create the extra classes, not pass around the `FileUtil` intersection type.

>  If you can provide me with an alternate solution which is cleaner than this one, or provide me a compelling argument why the solution 

The problem here is that anybody who has something to do with the kompilation directory will now have to know about all the stuff in FileUtil -- `FileUtil.kompiledDir` is private. The solution:

The JavaSymbolicBackend doesn't know about `FileUtil`, it only receives what it needs, i.e., kompiledDir.

```
JavaSymbolicBackend(
             Stopwatch sw,
             Context context,
             BinaryLoader loader,
             Provider<RuleIndex> index,
             Provider<KILtoBackendJavaKILTransformer> transformer,
             File kompiledDir) {
```

`loader.saveOrDie(files.resolveKompiled(JavaSymbolicBackend.DEFINITION_FILENAME),`

is replaced with

`loader.saveOrDie(new File(kompiledDir, JavaSymbolicBackend.DEFINITION_FILENAME),`

dwightguth, line 74: Cosmin, the problem with this approach is I'd like to reach a point where FileUtil or classes like it are the only objects constructing File instances. I definitely can do what you describe, but it provides a less nice API to the user who wants to read and write to files. We end up back in the situation we started with where very little is stopping the user from unknowingly breaking the paradigm by constructing their paths manually.

I know you hate Guice, but it has nothing to do with this. I could very easily do as you describe and construct the API so that each call site calls the File constructor. Nothing having anything to do with Guice constrains me here one way or the other. The question is whether I want to expose an API to our developers which encourages the raw manipulation of File objects. I decided I wanted to discourage that in order to keep the code from fragmenting (which as you can see it had done in the past over time). That means having strongly typed methods like "resolve path in temp dir" and "copy file to kompiled directory". This is literally the nicest API I could come up with for accomplising that goal. I really don't think it's that big of a loss to have to spend a little more effort to construct this object outside production code, especially because I'm pretty sure that's something you should not be doing anyway. Test code should be mocking dependencies like this which have side effects on the file system. And production code should instantiate this class via the injector, making this not an issue.

I'm going to give you a chance to respond again to this to note your objections, but then I'm going to merge this. It's good to express your opinions, but it's also good to accept that sometimes the project decides to go in a different direction than you see as ideal yourself.

cos, line 74: > Cosmin, the problem with this approach is I'd like to reach a point where FileUtil or classes like it are the only objects constructing File instances.

Why would you even want this?

> I definitely can do what you describe, but it provides a less nice API to the user who wants to read and write to files. 

So `new File(kompiledDir, foo)` is uglier then `files.resolveKompiled(foo)`? If you don't like it, you could even provide some helper methods/classes. But don't make them into a monolithic dependency that needs to be passed around everywhere.

> We end up back in the situation we started with where very little is stopping the user from unknowingly breaking the paradigm by constructing their paths manually.

`kompiledDir/mystuff` is not rocket science. What I'm reading here is that if we entangle this enough, and make it hard enough for the developer, he won't make mistakes anymore. It might be true, because he won't develop anything new -- he will spend his time writing mocking code for all the dependencies. 

Another way to put this, if you want a more engineerish explanation, is that your approach increases the coupling of the system. And with very little gain. 

> I know you hate Guice

You know wrong. I don't hate Guice. I see Guice as a nice thick rug. It can be comfortable, and thus useful. But, being thick, you can also shove a lot of dirt under it. The dirt is the high coupling.

> I'm going to give you a chance to respond again to this to note your objections, but then I'm going to merge this. It's good to express your opinions, but it's also good to accept that sometimes the project decides to go in a different direction than you see as ideal yourself.

Thank you for the opportunity to object without effect! I've been doing this quite a bit lately.

dwightguth, line 74: But Cosmin, creating the monolithic dependency is the entire purpose of
this code review. I need this because the direction I want to take the
performance of the application in requires that no code except FileUtil
have any knowledge of the current working directory. And all four of the
paths in this class ultimately are relative to the current directory. And
if we don't establish a clear paradigm for accessing files, it becomes too
likely that bugs will slip through as a result of this and lead to
situations that are very difficult to correct and fix. Sometimes having a
very clear API /is/ desirable for exactly the reason you are so disdainful
of. To make it easy to avoid easy-to-make mistakes.

Okay, I'm gonna merge this. I'm sorry you feel so frustrated that people
aren't listening to you. Believe it or not, though, we are. I am just
considering other factors in addition to the ones that you have yourself
mentioned.

cos, line 74: This is not about feelings. It's about engineering. 

> creating the monolithic dependency is the entire purpose of
> this code review

For what you gain out of it, this is bad engineering.

grosu, line 74: OK, I'd like to understand this as well, all the pros and cons of the current approach and what alternative we have, with its pros and cons.  Cosmin, when you come back from OOPSLA, let us get together in a room with Dwight and discuss this.

Thanks,
Grigore

PR #1076, opened 2014-10-17
minor cleanup of rewriters

yilongli:
Since this may affect not only the prover, I separate it from my working branch of refactoring matching logic prover.

Basically, I change the code to use pattern matching instead of unification to match the search pattern.

@dwightguth @andreistefanescu  please review


andreistefanescu:
lgtm

dwightguth:
Yay red diffs!

PR #1077, opened 2014-10-17
remove dependency-injected static state

dwightguth:
**please review**


dwightguth:
@cos @osa1 @yilongli please review

java-backend/src/main/java/org/kframework/backend/java/builtins/BitVector.java
```diff
@@ -206,7 +205,7 @@ public static BitVector fromDigits(List<BitVector> digits) {
     public static int getBitwidthOrDie(ASTNode t) {
         Integer bitwidth = getBitwidth(t);
         if (bitwidth == null) {
-            GlobalSettings.kem.registerCriticalError("Expected machine integer variable to declare a bitwidth." +
+            throw KExceptionManager.criticalError("Expected machine integer variable to declare a bitwidth." +
```
osa1, line 208: I think this change is really good. No more `assert false; // unreachable`s.

k-core/src/main/java/org/kframework/backend/BasicBackend.java
```diff
@@ -48,62 +37,6 @@ public boolean autoinclude() {
...
-        steps.add(new Cell2DataStructure(context));
-        steps.add(new ResolveSupercool(context));
-        steps.add(new AddStrictStar(context));
-        steps.add(new AddDefaultComputational(context));
-        steps.add(new AddOptionalTags(context));
-        steps.add(new DeclareCellLabels(context));
-        steps.add(new LastStep(this, context));
-        return steps;
-    }
-
```
osa1, line 106: Just yesterday I made this change but hesitated to send a pull request ...

maude-backend/src/main/java/org/kframework/backend/maude/KompileBackend.java
```diff
@@ -2,10 +2,61 @@
...
+import org.kframework.compile.transformers.ResolveOpenCells;
+import org.kframework.compile.transformers.ResolveRewrite;
+import org.kframework.compile.transformers.ResolveSupercool;
+import org.kframework.compile.transformers.ResolveSyntaxPredicates;
+import org.kframework.compile.transformers.SetVariablesInferredSort;
+import org.kframework.compile.transformers.StrictnessToContexts;
+import org.kframework.compile.utils.CheckVisitorStep;
+import org.kframework.compile.utils.CompileDataStructures;
+import org.kframework.compile.utils.CompilerSteps;
+import org.kframework.compile.utils.InitializeConfigurationStructure;
```
osa1, line 55: These could be simplified to:

``` java
import org.kframework.compile.transformers.*;
import org.kframework.compile.utils.*;
```

k-core/src/main/java/org/kframework/utils/XmlLoader.java
```diff
@@ -15,6 +15,10 @@
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.transform.*;
+import javax.xml.transform.dom.DOMSource;
+import javax.xml.transform.stream.StreamResult;
```
osa1: Are those used? Rest of the file is unchanged.

maude-backend/src/main/java/org/kframework/backend/maude/KompileBackend.java
dwightguth, line 55: I will leave this as is because it was just generated automatically by Eclipse

k-core/src/main/java/org/kframework/utils/XmlLoader.java
dwightguth: fixed. I took the easy path here when I was resolving a merge conflict, that is why.

osa1:
(Haven't read line by line but) lgtm :+1:

cos:
@dwightguth, if someone else would submit this pull request, would you review it? there are over 3000 changes in almost 200 files.

there are parts of it that I really like, like throwing exceptions without using a kem instance (`throw KExceptionManager.criticalError(...)`)

and there are parts I don't, like the whole passing around KItemOperations -- do we really need that? it's only purpose seem to be knowing which `Tool` to pass to the KItem constructor. Do we really want to do this whole twisted thing just for one argument? `KItem.of` was much cleaner? Also, it seem to me the construction of a `KItem` should not depend on the tool doing it. The fact it does now is an artifact. Let's not introduce this much extra complexity just for this. 

I haven't looked over the whole thing. It is much more than I can chew.

dwightguth:
I will break this up into pieces based on the four fields that are being refactored out. That should make this much easier to read.

PR #1084, opened 2014-10-22
Context refactoring -- prelude

osa1:
This is only the most basic stuff. I'm working on figuring where to move rest of the attributes/methods.

About removed `CollectBracketsVisitor` visitor: Generated data was never read so I just removed the visitor and the related attribute in `Context`.


dwightguth:
if canonicalBracketForSort is not being used that is probably a bug. I suggest you go into AddBracketsVisitor and AddBracketsVisitor2 and find the places where we are constructing new brackets and use that field to determine the correct bracket production for each instance. And if no bracket can be found, then use the existing builtin bracket.

osa1:
@dwightguth I don't understand. How do you want me to use that unused/removed field?

dwightguth:
Note how in the two classes I mention we call the constructor to the Bracket class in several places. What you should do is take a look at the sort of the term which is having a bracket put around it, and consult the map to see if there is a production which is the bracket for that sort. If there is, instead of creating a Bracket, create an instance of that production. Otherwise do the same thing that was being done before.

dwightguth:
Essentially this field, like several others, forms an index of information about the definition. Eventually we want this information to be precomputed and updated by the Definition object as it is transformed.

osa1:
I think the visitor is essentially useless:

``` java
public Void visit(Production node, Void _) {
    if (node.isBracket()) {
        context.canonicalBracketForSort.put(node.getSort(), node);
    }
    return null;
}
```

For every bracket it sees it's updating the map without replacing any of previously encountered brackets. Also, passes you mentioned are not part of this pipeline so to do what you just described I also need to incorporate that into the Java backend pipeline.

Another point to consider is that since everything is super fragile at this point, I'm pretty sure that work will be super involved and maybe even wrong. Maybe @yilongli can tell something about this. (since this is part of Java backend transformations)

EDIT: I pinged wrong person again... Sorry...

dwightguth:
It's true it's updating the map without concern for what is already there,
but for the purpose that is needed here, which is that we need to have
access to a bracket production which is valid in a particular position in
the AST, it ought to be fine. I'm not sure what you think this has to do
with the Java backend though. It should be a modification that only affects
the unparser. The larger modification I'm talking about is long-term, not
something to be done immediately. I am just describing the migration path
that we will have to take to get rid of the Context entirely.

osa1:
The removed transformation was used in two places:
- kil.Definition.preprocess
- JavaSymbolicBackend

What I'm trying to say is that if I want to do what you described then I need to add `AddBracketsVisitor` transformers to the pipeline in both of them and that'll cause the universe to collapse. (e.g. every single test in the framework to fail until lots of unrelated bugs are fixed)

> I'm not sure what you think this has to do with the Java backend though

Becuase removed transformer was used in Java backend too.

> It should be a modification that only affects the unparser.

Oh, so you're suggesting using that only in unparser? I guess that makes sense. I'll ask for help from whoever wrote the unparser.

(I still don't understand how is using same bracket for a sort instead of creating new brackets more useful though)

dwightguth:
As I said, all I am suggesting is using the existing field populated by the existing visitor CollectBracketsVisitor in the existing transformers AddBracketsFilter and AddBracketsFilter2.

dwightguth:
I was the one who wrote those two classes btw, you are more than welcome to ask me questions.

traiansf:
I think the unparser was written by Stefan @ciobaca

dwightguth:
Can you please put back the calls to this visitor as well?

osa1:
Done.

dwightguth:
lgtm. Merge if tests pass

PR #1087, opened 2014-10-22
remove static state for Tool.instance

dwightguth:
@cos is this a better solution for you? Eventually, as you say, we should break this dependency, so I opted to go for the local modification instead.


dwightguth:
@cos @yilongli please review

yilongli:
fine with me

cos:
lgtm

PR #1089, opened 2014-10-22
refactor ProgramLoader, RunProcess to remove static state

dwightguth:
**please review**


dwightguth:
@cos @radumereuta please review

k-core/src/main/java/org/kframework/krun/api/KRunState.java
```diff
@@ -56,51 +55,47 @@ public KRunState(Term rawResult) {
...
-                        ExceptionType.WARNING,
-                        KExceptionGroup.INTERNAL,
-                        "Could not load parser: brackets may be unsound"));
+                    result = newResult;
                 }
+            } catch (IOException e) {
+                GlobalSettings.kem.register(new KException(
+                    ExceptionType.WARNING,
+                    KExceptionGroup.INTERNAL,
+                    "Could not load parser: brackets may be unsound"));
```
radumereuta, line 97: Is this error message correct?
It doesn't make too much sense to me.

dwightguth, line 97: fun fact, this block is actually unreachable. I delete it in another pull request I have out right now. This one is just changing indentation though. You should view this block of code while ignoring whitespace.

k-core/src/main/java/org/kframework/parser/DefinitionLoader.java
```diff
@@ -90,7 +91,8 @@ public DefinitionLoader(
             OuterParser outer,
             @Backend.Documentation boolean documentation,
             @Backend.Autoinclude boolean autoinclude,
-            FileUtil files) {
+            FileUtil files,
+            ProgramSDF programSDF) {
```
radumereuta: Send the loader as a parameter to getSdfForPrograms instead of making it as an object.
This function doesn't need to be instantiated to work. Static is fine in this case.

dwightguth: There is a very good reason for doing it this way which is that it allows mocking of this very large dependency. In general, the paradigm of dependency injection is that wherever possible, a class should be constructed with only its direct dependencies. There are places in some of my diffs where doing that would be unmanageable with the code as it currently exists, but this is not one of those cases.

k-core/src/main/java/org/kframework/parser/DefinitionLoader.java
```diff
@@ -90,7 +91,8 @@ public DefinitionLoader(
             OuterParser outer,
             @Backend.Documentation boolean documentation,
             @Backend.Autoinclude boolean autoinclude,
-            FileUtil files) {
+            FileUtil files,
```
radumereuta: Please add javadoc to FileUtil.
Once for the entire class, and for each method individually. I have no idea why it is needed here.

k-core/src/main/java/org/kframework/parser/ProgramLoader.java
```diff
@@ -34,23 +33,43 @@
...
 public class ProgramLoader {
 
+    private final BinaryLoader loader;
+    private final Stopwatch sw;
+    private final KExceptionManager kem;
+    private final GlobalOptions globalOptions;
+
+    @Inject
+    ProgramLoader(
+            BinaryLoader loader,
```
radumereuta, line 56: This looks better than `DefinitionLoader`. Do the same as here.

k-core/src/main/java/org/kframework/parser/DefinitionLoader.java
radumereuta: But it makes no sense for `DefinitionLoader` to be dependent on something so weird as `ProgramSdf`. Somebody looking from the outside will not know how to instantiate this because it is so specific to SDF. Also, with the new parser I will have to change the dependency, but not if I only require the loader, which makes sense.

dwightguth: But Radu, it /is/ dependent on this class. The dependency doesn't go away if you hide it inside static methods. If you don't want this dependency to be like this, the solution is to change the dependency structure, not sweep the problem under the rug.

I suggest you do what you should have done a long time ago which is to abstract an interface for generating and using parse tables, and then fork the implementation of the interface based on whether you are using the new or old parser. But that is something for you to do in your own pull request. This is just a procedural refactoring designed to get rid of undesirable static state.

radumereuta: Hmm... ok. Let me make a pull request which removes the static state from this class.

radumereuta: Please have a look at #1102 and tell me if this changes things.

cos:
lgtm

dwightguth:
@radumereuta please re-review

radumereuta:
lgtm

PR #1094, opened 2014-10-23
Cleanup ml prover 5

yilongli:

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
@@ -136,10 +136,6 @@ public class ConstrainedTerm extends JavaSymbolicObject {
         }
 
         unificationConstraint.expandPatternsAndSimplify(false);
-        // TODO(AndreiS): figure out when a map unification provided more data
-        unificationConstraint.expandPatternsAndSimplify(false);
-        unificationConstraint.expandPatternsAndSimplify(false);
-        unificationConstraint.expandPatternsAndSimplify(false);
 
         final Set<Variable> variables = Sets.newHashSet(unificationConstraint.variableSet());
         variables.removeAll(variableSet());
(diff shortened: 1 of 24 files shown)


java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
```diff
@@ -136,16 +136,10 @@ public SymbolicConstraint matchImplies(ConstrainedTerm constrainedTerm) {
...
-        // TODO(AndreiS): figure out when a map unification provided more data
-        unificationConstraint.expandPatternsAndSimplify(false);
-        unificationConstraint.expandPatternsAndSimplify(false);
-        unificationConstraint.expandPatternsAndSimplify(false);
 
         final Set<Variable> variables = Sets.newHashSet(unificationConstraint.variableSet());
         variables.removeAll(variableSet());
-        unificationConstraint.orientSubstitution(variables);
-        if (unificationConstraint.isFalse()
-                || !unificationConstraint.substitution().keySet().containsAll(variables)) {
```
andreistefanescu, line 148: Why did you remove this condition?

java-backend/src/main/java/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
```diff
@@ -519,10 +519,17 @@ public ASTNode visit(org.kframework.kil.Rule node, Void _)  {
...
-        Collection<Variable> freshVariables = new ArrayList<>();
+        Set<Variable> freshConstants = Sets.newHashSet();
         // TODO(AndreiS): check !Variable only appears in the RHS
         for (org.kframework.kil.Variable variable : node.getBody().variables()) {
             if (variable.isFreshConstant()) {
+                freshConstants.add((Variable) this.visitNode(variable));
+            }
+        }
+
+        Set<Variable> freshVariables = Sets.newHashSet();
```
andreistefanescu: you did not refactor this into a lambda :-)

andreistefanescu:
Looks good. This refactoring and the previous one are really useful!

java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
yilongli, line 148: because `orientSubstitution` now returns a boolean to indicate whether the orientation succeeds (the orientation is possible and the resulting constraint is not false). If so, then there is no need to check `unificationConstraint.substitution().keySet().containsAll(variables)`.  If not, then the substitution will remain unchanged.

dwightguth:
lgtm. You can merge if you are happy with this, Yilong.

PR #1096, opened 2014-10-23
update readme with environment variables

### LINKED ISSUE #1093: Update the readme on how to set up a running environment in an IDE
I had some troubles today setting up my IntelliJ to run kompile and krun.
The problem is that recently the running scripts have been updated to include the path to the executables required by K. Basically the resolution of which executable to run depending on the platform has been moved to bash scripts, and now this has to be simulated by the IDE.

I think this should be better explained in readme for developers.

dwightguth:
fixes #1093 

@radumereuta please review

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -4,8 +4,7 @@ This is a readme file for the developers.
 # Prerequisites
 
 ## Java Development Kit (required JDK8 or higher)
-You can follow the instructions here: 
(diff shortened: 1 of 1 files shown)


radumereuta:
This works, but it's still really annoying to set up.
I think it makes it much harder for a new user to begin working on any part of the code.
I think in the end, @grosu has a point in #1093.
But this is ok to merge if you can't find a better solution.

PR #1098, opened 2014-10-23
fixed multi-representation of BuiltinMap

yilongli:
Let's see if the problem has been fixed in my previous two pull requests.


yilongli:
@andreistefanescu The problem is that `Equality#isFalse` incorrectly returns true for `.Map =? list(null)(.List)`.

yilongli:
Hmm... It's actually more complicated than I thought. It turns out that this bug is the reason why the CIL semantics can pass quite a few tests though many of its function are not total.

Now I will have to fix the cil semantics too.

yilongli:
Jenkins: retest this please

PR #1099, opened 2014-10-23
refactor and remove GlobalSettings.kem

dwightguth:
**please review**


dwightguth:
@osa1 @radumereuta please review

java-backend/src/main/java/org/kframework/backend/java/compile/DataStructureToLookupUpdate.java
```diff
@@ -233,9 +232,7 @@ public ASTNode visit(Rule node, Void _)  {
                         ((BuiltinLookup) queue.get(i)).variables(),
                         ((BuiltinLookup) queue.get(j)).variables());
                 if (!commonVariables.isEmpty()) {
-                    GlobalSettings.kem.registerCriticalError("Unsupported map pattern in the rule left-hand side", node);
-                    /* dead code */
-                    return null;
+                    throw KExceptionManager.criticalError("Unsupported map pattern in the rule left-hand side", node);
```
radumereuta, line 235: Why do we have to go through `KExceptionManager`? Why not have our own exception hierarchy and say `throw new SomeException(...)`?

java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaSymbolicBackend.java
```diff
@@ -93,11 +97,11 @@ public String getDefaultStep() {
         steps.add(new RemoveBrackets(context));
         // SetVariablesInferredSort must be performed before AddEmptyLists
         steps.add(new SetVariablesInferredSort(context));
-        steps.add(new AddEmptyLists(context));
+        steps.add(new AddEmptyLists(context, kem));
```
radumereuta, line 100: Is this how you handle warnings?

k-core/src/main/java/org/kframework/backend/BasicBackend.java
```diff
@@ -44,67 +34,18 @@ public Definition firstStep(Definition def) {
...
 
     @Override
-    public boolean autoinclude() {
-        return !options.experimental.noPrelude;
-    }
-
     public CompilerSteps<Definition> getCompilationSteps() {
         CompilerSteps<Definition> steps = new CompilerSteps<Definition>(context);
         steps.add(new FirstStep(this, context));
-        steps.add(new CheckVisitorStep<Definition>(new CheckConfigurationCells(context), context));
```
radumereuta, line 54: What does this have to do with removing kem?

k-core/src/main/java/org/kframework/krun/XmlUtil.java
```diff
@@ -74,15 +74,8 @@ public static void serializeXML(Node doc, String fileName) {
...
-        } catch (TransformerFactoryConfigurationError factoryError) {
-            // factoryError.printStackTrace();
-            org.kframework.utils.Error.report("Error creating TransformerFactory:" + factoryError.getMessage());
-        } catch (TransformerException transformerError) {
-            // transformerError.printStackTrace();
-            org.kframework.utils.Error.report("Error transforming document:" + transformerError.getMessage());
-        } catch (IOException ioException) {
-            // ioException.printStackTrace();
-            org.kframework.utils.Error.report("Error while serialize XML:" + ioException.getMessage());
+        } catch (TransformerFactoryConfigurationError | TransformerException | IOException e) {
```
radumereuta, line 77: Why are you catching an error?
Please rename if indeed this is what you are trying to do.

k-core/src/main/java/org/kframework/krun/tools/Executor.java
```diff
@@ -159,6 +166,35 @@ public SearchPattern(ASTNode pattern) {
...
+            if (pattern == null) {
+                patternToParse = KRunOptions.DEFAULT_PATTERN;
+            }
+            if (patternToParse.equals(KRunOptions.DEFAULT_PATTERN)) {
+                Sentence s = new Sentence();
+                s.setBody(new Cell("generatedTop", new Variable("B", Sort.BAG)));
+                s.addAttribute(Attribute.ANYWHERE);
+                return s;
+            }
+            org.kframework.parser.concrete.KParser
```
radumereuta: Can you put this on a single line?

k-core/src/main/java/org/kframework/krun/XmlUtil.java
dwightguth, line 77: this is the exact same code that was there before, just routed through KExceptionManager instead of Error. I'd rather not touch it at all. I assume this is happening because it's a checked exception thrown by one of the methods in the try block, though.

k-core/src/main/java/org/kframework/krun/tools/Prover.java
```diff
@@ -61,6 +63,7 @@ protected Tool(
             this.kem = kem;
             this.prover = prover;
             this.files = files;
+            this.loader = loader;
```
radumereuta: Please rename this to `definitionLoader`. We currently also use the `loader` name for instances of `FileUtils`. This should avoid some confusion.

java-backend/src/main/java/org/kframework/backend/java/compile/DataStructureToLookupUpdate.java
dwightguth, line 235: in fact this is what I am doing with ParseFailedException. It is a further optimization that could be done, but I think this diff is big enough already. The point of this diff is to remove the static KExceptionManager. Making errors occur by throwing an exception was simply done to facilitate that refactoring without requiring such intrusive changes. If you like, you can add an issue for this and assign it to me though.

java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaSymbolicBackend.java
dwightguth, line 100: yeah

dwightguth, line 100: basically warnings still need to be recorded because unlike errors, they don't stop the compilation process, so they can't be handled on the catch-side.

k-core/src/main/java/org/kframework/backend/BasicBackend.java
dwightguth, line 54: By making this change, which I intended to make a while ago, it was no longer necessary for BasicBackend to take a KExceptionManager to its constructor.

maude-backend/src/main/java/org/kframework/backend/maude/MaudeContextExtractor.java
```diff
@@ -1,45 +0,0 @@
...
-
-import org.kframework.kil.*;
-import org.kframework.kil.loader.Context;
-import org.kframework.kil.visitors.CopyOnWriteTransformer;
-
-/**
- * Initially created by: Traian Florin Serbanuta
- * <p/>
- * Date: 12/18/12
- * Time: 12:59 AM
```
radumereuta, line 12: @traiansf, you should go to bed earlier... :)

java-backend/src/main/java/org/kframework/backend/java/compile/DataStructureToLookupUpdate.java
radumereuta, line 235: :+1:

k-core/src/main/java/org/kframework/backend/BasicBackend.java
radumereuta, line 54: Huh? I was asking where are all the compilation steps...

java-backend/src/main/java/org/kframework/backend/java/compile/DataStructureToLookupUpdate.java
radumereuta, line 235: It's fine for now.

k-core/src/main/java/org/kframework/krun/XmlUtil.java
radumereuta, line 77: Oh, this comes from `javax`. I thought it is our own creation. Yuck.

radumereuta:
Done. Just a few minor things left for you to change. Other than that, lgtm

k-core/src/main/java/org/kframework/backend/BasicBackend.java
osa1, line 54: Those steps are moved to `KompileBackend`. I did this change in the past but hesitated to send a pull request for some reason. For the motivation, see this chart:

```
Backend name         |  Overrides BasicBackend steps
----------------------------------------------------
KompileBackend       |  yes
JavaSymbolicBackend  |  yes
CoqBackend           |  yes
HtmlBackend          |  no
LatexBackend         |  no
PdfBackend           |  no
UnparserBackend      |  no
```

The thing to observe here is that all backends that are meant to execute the code already overrides it, and the ones that don't override actually don't need any steps at all.

So default steps provided by `BasicBackend` was not really a good default.

(Another advantage of moving steps out of `BasicBackend` is performance, because now `HtmlBackend`, `LatexBackend`, `PdfBackend` and `UnparserBackend` won't be running unnecessary steps, although the cost of the steps were probably very small)

k-core/src/main/java/org/kframework/backend/unparser/ConcretizeTerm.java
```diff
@@ -0,0 +1,93 @@
...
+import org.kframework.krun.FlattenDisambiguationFilter;
+import org.kframework.krun.KRunOptions;
+import org.kframework.krun.SubstitutionFilter;
+import org.kframework.krun.api.SearchResult;
+import org.kframework.parser.TermLoader;
+import org.kframework.parser.concrete.disambiguate.TypeInferenceSupremumFilter;
+
+import com.google.inject.Inject;
+
+public class ConcretizeTerm {
```
osa1, line 31: What is this class doing? Can we at least add module-level documentation?

k-core/src/main/java/org/kframework/ktest/Test/TaskQueue.java
```diff
@@ -124,10 +123,8 @@ public void terminate() {
                 }
             }
         } catch (InterruptedException e) {
-            // TODO: I have no idea when this is thrown, and in what condition tpe is left.
-            // For now, I'm assuming that we're done.
-            // Still register a warning to keep track of things.
-            GlobalSettings.kem.registerCompilerWarning(e.getMessage(), e);
+            Thread.currentThread().interrupt();
```
osa1, line 126: This interrupts the main thread, do we really need that? Main thread can still continue waiting for other tasks(running new tasks) and report the results.

k-core/src/main/java/org/kframework/backend/unparser/ConcretizeTerm.java
grosu, line 31: Actually, the original plan was to reject pull requests if they are not at least minimally documented.  That means, at least minimal javadoc, but modules should probably have more than minimal documentation.  @dwightguth, please continue to reject pull requests if they come without minimal documentation,

Grigore

k-core/src/main/java/org/kframework/ktest/Test/TaskQueue.java
dwightguth, line 126: it's not considered correct Java code for a program to catch InterruptedException and do nothing, not even reset the interrupt flag by calling this method. In practice, I don't think this is likely to make much of a difference since the user shouldn't ever interrupt anything unless they want the process to terminate.

dwightguth:
@radumereuta @osa1 please re-review

osa1:
lgtm.

radumereuta:
lgtm

PR #1100, opened 2014-10-23
Context and transformers work

osa1:
I thought it'd be nice to discuss on code, so here's a pull request. This a work in progress.

What I'm hoping to do is to implement part 1.1 described here: https://github.com/kframework/k/wiki/Design-of-the-new-compiler-parser

I'll occasionally ping people to get some reviews/feedback.

### DIFF
--- a/k-core/src/main/java/org/kframework/kil/AbstractVisitor.java
+++ b/k-core/src/main/java/org/kframework/kil/AbstractVisitor.java
@@ -292,9 +292,7 @@ public abstract class AbstractVisitor<P, R, E extends Throwable> implements Visi
 
     @Override
     public R visit(Definition node, P p) throws E {
-        if (currentDefinition == null) {
-            currentDefinition = node;
-        }
+        currentDefinition = node;
         node = genericVisitList(node, p, mutableList(Definition.class), null);
         return visit((ASTNode) node, p);
     }
@@ -311,9 +309,7 @@ public abstract class AbstractVisitor<P, R, E extends Throwable> implements Visi
 
     @Override
     public R visit(Module node, P p) throws E {
-        if (currentModule == null) {
-            currentModule = node;
-        }
+        currentModule = node;
         node = genericVisitList(node, p, mutableList(Module.class), null);
         return visit((DefinitionItem) node, p);
     }
--- a/k-core/src/main/java/org/kframework/kil/AbstractVisitor.java
+++ b/k-core/src/main/java/org/kframework/kil/AbstractVisitor.java
@@ -8,6 +8,7 @@ import org.kframework.kil.visitors.CopyOnWriteTransformer;
 import org.kframework.kil.visitors.Visitor;
 import org.kframework.kompile.KompileOptions;
 import org.kframework.main.GlobalOptions;
+import org.kframework.kil.loader.Context;
 
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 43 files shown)


k-core/src/main/java/org/kframework/kil/AbstractVisitor.java
```diff
@@ -55,24 +56,49 @@
...
+        this.currentDefinition = null;
+        this.currentModule = null;
+        this.kompileOptions = context == null ? null : context.kompileOptions;
+        this.globalOptions = context == null ? null : context.globalOptions;
         this.name = this.getClass().toString();
     }
 
-    public AbstractVisitor(String name, org.kframework.kil.loader.Context context) {
-        this(context);
+    public AbstractVisitor(String name, Context context) {
```
dwightguth, line 70: please use constructor inheritance for these

k-core/src/main/java/org/kframework/kil/AbstractVisitor.java
```diff
@@ -55,24 +56,49 @@
  * return {@code null}.
  */
 public abstract class AbstractVisitor<P, R, E extends Throwable> implements Visitor<P, R, E> {
-    protected org.kframework.kil.loader.Context context;
-    protected KompileOptions kompileOptions;
-    protected GlobalOptions globalOptions;
-    String name;
+    protected final Context context;
+    protected final Module currentModule;
```
dwightguth: sorry I meant protected final getters, not fields. The field should be private but mutable.

k-core/src/main/java/org/kframework/kil/AbstractVisitor.java
```diff
@@ -55,24 +56,49 @@
  * return {@code null}.
  */
 public abstract class AbstractVisitor<P, R, E extends Throwable> implements Visitor<P, R, E> {
-    protected org.kframework.kil.loader.Context context;
-    protected KompileOptions kompileOptions;
-    protected GlobalOptions globalOptions;
-    String name;
+    protected final Context context;
```
dwightguth, line 59: this can stay the way it is though

k-core/src/main/java/org/kframework/kil/AbstractVisitor.java
```diff
@@ -55,24 +56,49 @@
...
+    protected final Module currentModule;
+    protected final Definition currentDefinition;
+    protected final KompileOptions kompileOptions;
```
dwightguth: probably delete these as unnecessary, and refactor the places they're called?

k-core/src/main/java/org/kframework/kil/AbstractVisitor.java
osa1: But a protected getter will have same visibility with the field, is it useful?

dwightguth: the idea is to prevent subclasses of Visitor from setting the field or making its value public. Obviously there are ultimately ways around, but it's still a nice API.

radumereuta:
At a first glance, it looks like a good start.

osa1:
@radumereuta @dwightguth please review last commit (https://github.com/osa1/k/commit/4fd3cd9e7942a3c9fae0f42513b57b83a5f0ffec)

(other commits are not changed, they look new because I rebased the patch)

osa1:
@dwightguth can you review this? I'd like to merge this to keep pull requests small, otherwise keeping them up-to-date with master branch is becoming too much problem.

k-core/src/main/java/org/kframework/parser/DefinitionLoader.java
```diff
@@ -301,7 +301,7 @@ public Definition parseDefinition(File mainFile, String mainModule, Context cont
             // the objects in the sentences are mutable, and we risk altering them and miss
             // warning and error messages when kompiling next time around
             try {
-                def = (Definition) new DisambiguateRulesFilter(context, true).visitNode(def);
+                def = (Definition) new DisambiguateRulesFilter(context, def, true).visitNode(def);
```
dwightguth: This shouldn't need to be passed a reference to the definition since it is also visiting the entire definition. The whole point was that the currentDefinition and currentModule fields would get updated automatically in AbstractVisitor whenever they are visited.

k-core/src/main/java/org/kframework/parser/generator/DisambiguateRulesFilter.java
```diff
@@ -32,12 +33,26 @@ public DisambiguateRulesFilter(Context context, boolean checkInclusion) {
...
+                                   Module currentModule, boolean checkInclusion) {
+        super(DisambiguateRulesFilter.class.getName(), context, currentDefinition, currentModule);
+        this.checkInclusion = checkInclusion;
+    }
 
     @Override
     public ASTNode visit(Module m, Void _) throws ParseFailedException {
-        localModule = m.getName();
-        return super.visit(m, _);
+        if (getCurrentModule() == null) {
```
dwightguth: same deal here, should happen automatically.

k-core/src/main/java/org/kframework/parser/generator/ParseConfigsFilter.java
```diff
@@ -47,15 +48,39 @@ public ParseConfigsFilter(Context context, boolean checkInclusion) {
...
+        }
+    }
 
     @Override
     public ASTNode visit(Module m, Void _) throws ParseFailedException {
-        localModule = m.getName();
-        ASTNode rez = super.visit(m, _);
-        new CollectStartSymbolPgmVisitor(context).visitNode(rez);
-        return rez;
+        if (getCurrentModule() == null) {
```
dwightguth: etc

dwightguth:
Please let me know if it's still not clear the architecture I had in mind. You are more than welcome to ask clarifying questions.

k-core/src/main/java/org/kframework/parser/DefinitionLoader.java
osa1: > The whole point was that the currentDefinition and currentModule fields would get updated automatically in AbstractVisitor whenever they are visited.

I think the only way to do that is by setting the field(instead of creating a new visitor with the field is set) in `AbstractVisitor`. Do you want me to do that?

(I'm asking because I thought we don't want to change any state of abstract visitor classes)

dwightguth: Yes, sorry, that is what I am saying. I may have miscommunicated if you got
that impression. I think what I probably meant for you to understand is
that all changes to the visitor infrastructure should happen so that they
do not require significant increase in the complexity of BasicVisitor,
AbstractTransformer, CopyOnWriteTransformer, etc.

On Mon, Oct 27, 2014 at 9:32 AM, Ömer Sinan Ağacan <notifications@github.com

osa1: Thanks Dwight, that makes sense. I'll make the changes and ping you again.

osa1:
@radumereuta @dwightguth I addressed the comments, can you please tell me if it looks good now?

dwightguth:
This looks amazing! Thank you very much and keep up the good work!

Issue #1110, opened 2014-10-25
the parser doesn't create the proper list terminator for my specification


yilongli:
I wanted to get rid of the `.Exps` in my specification rule:
https://github.com/yilongli/c-semantics/blob/java_rewrite_engine/cil-semantics/tests/verification/sum_spec.k#L11

Without writing the `.Exps`, `tv(unsigned int, I32N:MInt{bitwidth(32)})` remained as a `TypedValue` after the parsing. I expected something like `tv(unsigned int, I32N:MInt{bitwidth(32)}), .Exps` because:
`cil-syntax.k:  syntax FunCall ::= Exp "(" Exps ")"   [klabel('call), strict(2)]`

And since `TypedValue` is not subsorted to `TypedValues` or `Exps`, unification failed and the ml prover got stuck.

A similar case is the following rule:
`cil-exp.k:  rule *tv(T:Type, va_list(tv(_:Type, V:Value))) => tv(downPtr(T), V)`
However, this rule became 
`rule * tv ( T:Type , va_list ( '_,_(tv ( _:Type , V:Value ) , .TypedValues) )  )   => tv ( downPtr ( T:Type )  , V:Value )  [ structural()]`.
Note that there is this `.TypedValues` after `tv(...)`, which is what I would like to have.

@radumereuta @andreistefanescu any idea?

radumereuta:
I can answer this.
The step which tries to add the list terminators work only on TermConses, the classes which have direct access to he production. If you write a rule in the labeled format, I will consider that everything has the sort KItem, and do nothing.

Looking at your examples, I can only see abstract syntax.

yilongli:
Awesome, many thanks! I realized that I had `'tv` in my spec and `tv` in the other rule when I was filing the issue. So I replaced `'tv` with `tv`, failed, and then gave up. Now I convert all abstract syntax to concrete syntax and it works great.

PR #1111, opened 2014-10-26
Symbolic module

andreiarusoaie:
Added back the (basic) symbolic backend as a separate module.
@dlucanu @dwightguth please review


andreiarusoaie:
This is not working. Seems that the tool changed significantly in the last few days.

dlucanu:
This is contrary to the modular approach. If a feature is developed as a separate module, then its functioning shouldn't be affected by the the change of the main tool. 
Did these last changes affected the modular approach?

In this way we cannot progress too much.

andreiarusoaie:
It seems that a method disappeared and some other methods have changed their arguments (they are related to the context object). I have revisited my code and fixed the easy errors, but I still have to fix some error related with the  Maude generation - don't know yet what the problem is. As soon as I finish,  I will create a new pull request.

grosu:
Dorel, that will only happen once KORE will be fully incorporated in the tool; until then, things might still have to change.  For example, as you know, we are in process of eliminating the Context object.  If as part of the module you used some method that gave you the Context object, then there is obviously a problem, which I currently don't know how to solve except for gradual changes in the modules, as well.

Or Cosmin, do you think that your translator KORE2KIL will take care of that, so modules implemented using the old data-structures can just use your converter?

Grigore

dlucanu:
Grigore, thank you for explanations. You are right, these changes can be done only gradually. For us is a bit difficult to keep the trace of all changes and to realise which ones affect our work. But we try to figure it out in order to make progress.

osa1:
> It seems that a method disappeared and some other methods have changed their arguments (they are related to the context object).

@andreiarusoaie, if you need some state/methods in Context, comment at #1110 or at mailing list discussion. I'm hoping to merge Context changes gradually and unless you merge Symbolic backend first, you may have compilation problems with every change.

grosu:
No, the idea is to keep the backends as separate modules, except for the main Java backend.  So some method needs to be found by which the existing backends are modified incrementally, until we finally reach the KORE fixedpoint of the internal representation.  (A message about that will be sent soon.)  One of the reasons for which we separated the backends as modules is exactly because we do not want to maintain all of them each time we do a refactoring.  So, for now, the best we can do is that you guys stay closely in the same loop until the transition happens, hopefully very soon.

Grigore

dwightguth:
I should clarify also that the intent of the modularization is to keep the core of the K framework easy to maintain by putting the onus of ensuring that the module compiles and works correctly on the module developer rather than on the K team. As such, I suggest you work with us to create a separate repository in the kframework organization containing this code which contains the complete information of how to install the module and what version of the K framework it works with.

PR #1115, opened 2014-10-27
Parse in module

radumereuta:
Don't review yet.


radumereuta:
@dwightguth please review.
You can ignore the .k files that were added, so mainly look at the java code.

.gitignore
```diff
@@ -18,4 +19,6 @@ java-backend/.project
 k-distribution/.project
 *~
 junit-reports
+k-distribution/samples/test/
```
dwightguth: delete these two lines

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIOOperationsImpl.java
```diff
@@ -44,12 +48,14 @@ public BuiltinIOOperationsImpl(
             Context context,
             ConfigurationCreationOptions ccOptions,
             KILtoBackendJavaKILTransformer kilTransformer,
+            ProgramLoader programLoader,
```
dwightguth, line 52: as I said to you earlier, this class should not depend directly on the internal parser but instead allow the user to specify an external parser that gets invoked with an environment variable for the module.

.gitignore
radumereuta: I want to keep this one. I always have a /test directory in samples that is connected to the IDE.
Having it always show up in git diff makes it really annoying. Adding it to the ignore list means that I won't commit it by mistake. I will remove /release though. I don't know how that got there.

k-distribution/include/modules/k-parser.k
```diff
@@ -110,4 +110,9 @@ module K-PARSER
   rule parseSKExpInjectedString(S1:String, 0, I:Int, S2:String)
     => parseSKExps(ltrim(substrString(S1, findString(S1, ")", I +Int 1) +Int 1, lengthString(S1))), #parseToken("String", substrString(S1, 1, I -Int 1)))
     when S2 =/=String "\\"
+
```
dwightguth: this is the wrong file. This should go in uris.k next to #parse

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIOOperationsImpl.java
```diff
@@ -148,7 +154,27 @@ public Term parse(StringToken term1, StringToken term2, TermContext termContext)
...
+                    new GeneratedSource(this.getClass()),
+                    Sort.of(startSymbol.stringValue()),
+                    moduleName.stringValue(), context);
+            Term term = kilTransformer.transformAndEval(kast);
+            term = term.evaluate(termContext);
+            return term;
+        } catch (ParseFailedException e) {
+            String line = "" + e.getKException().getLocation().lineStart;
+            String col  = "" + e.getKException().getLocation().columnStart;
+            return processIOException("noparse", StringToken.of(e.getMessage() + " Line: " + line + " Column: " + col), termContext);
```
dwightguth: if you are going to change the syntax of #noparse then you also need to modify the IO server and tcp.k

k-core/src/main/java/org/kframework/compile/checks/CheckListDecl.java
```diff
@@ -37,10 +37,11 @@ public Void visit(Production node, Void _) {
                 String msg = "Inline list declarations are not allowed.";
                 throw KExceptionManager.compilerError(msg, this, pi);
             }
-            if (pi instanceof Lexical && node.getItems().size() > 1) {
-                String msg = "Inline lexical/token declarations are not allowed.";
-                throw KExceptionManager.compilerError(msg, this, pi);
-            }
+            // TODO: (radum) with the new parser lexical declarations are allowed in more places now.
+            // if (pi instanceof Lexical && node.getItems().size() > 1) {
```
dwightguth: if this code no longer applies you should delete it

k-core/src/main/java/org/kframework/compile/sharing/TokenSortCollector.java
```diff
@@ -38,7 +39,7 @@
      *
      * @see TokenSortCollector
      */
-    public static Set<Sort> collectTokenSorts(Definition definition, Context context) {
+    public static Set<Sort> collectTokenSorts(ASTNode definition, Context context) {
```
dwightguth: why?

k-core/src/main/java/org/kframework/compile/sharing/TokenSortCollector.java
```diff
@@ -92,7 +93,7 @@ private void checkIllegalProduction(Production production) {
          *   syntax Id ::= "Main"
          *   syntax Id ::= "String2Id" "(" String ")"  [function, klabel(String2Id)]
          */
-        if (!production.isLexical() && !production.isTerminal()
+        /*if (!production.isLexical() && !production.isTerminal()
```
dwightguth: same here. Do not comment out code. Either keep it or delete it

k-core/src/main/java/org/kframework/parser/concrete2/KSyntax2GrammarStatesFilter.java
```diff
@@ -250,6 +250,19 @@ public Void visit(Production prd, Void _) {
                             prd.getSort() + "-S", nt, grammar.get(srt.getName()), false);
                     previous.next.add(nts);
                     previous = nts;
```
dwightguth: shouldn't there also be a case deleted here of how we handled tokens the old way?

k-core/src/main/java/org/kframework/parser/generator/ParsersPerModule.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+public class ParsersPerModule {
+
+    /**
+     * Generates and saves to disk a Map of moduleName -> programParser for the entire definition
+     * given as argument.
+     * @param def        The definition for which to generate all the program parsers.
+     * @param context    The Context object, with all the helper methods.
+     */
+    public static  Map<String, Grammar> generateParsersForModules(Definition def, Context context, KExceptionManager kem) {
```
dwightguth: you're not going to like this, but this class and QuickParser should be injected dependencies. I eventually want a unit test written for DefinitionLoader and we can't do that as long as it statically depends on a large body of other code.

k-core/src/main/java/org/kframework/parser/utils/QuickParser.java
```diff
@@ -0,0 +1,43 @@
...
+import org.kframework.main.GlobalOptions;
+import org.kframework.parser.ProgramLoader;
+import org.kframework.parser.concrete2.Grammar;
+import org.kframework.parser.concrete2.KSyntax2GrammarStatesFilter;
+import org.kframework.parser.generator.CollectTerminalsVisitor;
+import org.kframework.utils.errorsystem.KExceptionManager;
+import org.kframework.utils.errorsystem.ParseFailedException;
+
+public class QuickParser {
+    public static ProductionReference parse(String program, Sort startSymbol, ASTNode definition, KExceptionManager kem) throws ParseFailedException {
```
dwightguth: again, why ASTNode?

k-core/src/main/java/org/kframework/parser/generator/ParsersPerModule.java
dwightguth: oh never mind I think I misunderstood the purpose of QuickParser. But the point still stands regarding this class

k-core/src/main/java/org/kframework/parser/utils/QuickParser.java
```diff
@@ -0,0 +1,43 @@
...
+        Context context = new Context();
+        context.kompileOptions = new KompileOptions();
```
dwightguth: this looks  really really bad. we should not be constructing junk just to pass it around. If you intend this to be used only by tests, put it in src/test. if you intend it to be used by production code, you need to come up with a more correct source for these dependencies.

k-core/src/main/javacc/Outer.jj
```diff
@@ -443,6 +443,8 @@ PARSER_END(Outer)
                  "\\\"" | "\\n" | "\\r" | "\\t" | "\\\\")* "\"">
 | <STRING_CASE_INSENSITIVE: "\'" (~["\'", "\\", "\n"] |
                  "\\\'" | "\\n" | "\\r" | "\\t" | "\\\\")* "\'">
+| <STRING_REGEX: "r\"" (~["\"", "\\", "\n"] |
```
dwightguth: oh hmm okay I think I understand. You are creating a new syntax for terminals, which is why you are keeping the existing Token syntax. Still, you need to add a warning that appears on definitions that use Token{} telling them that it is deprecated and may be removed in a later version of K. Similar to what we did with List{K}.

dwightguth:
In addition to the existing comments I made, you also need to clean up the stuff you put in the samples directory somehow. It is not appropriate for code that is not clean and presentable to the user to be included in the samples directory. If you want a K definition of KAST in the samples directory, that's fine, but it needs to be clean and usable using the latest version of the K tool.

dwightguth:
I also need you to flatten this history somewhat using rebase. Right now we have commits in the history containing large binary files and I can't merge that. I also can't have confidence in the merge if it has so many commits that I don't have time to go through each one and make sure it isn't screwing up the history somehow.

dwightguth:
It looks like there are several major changes that I need from you in this diff. The code itself looks mostly okay, it's just not currently presented in a very good form. Please let me know if you have questions. And remember, these pull requests would not be so time consuming and full of crap like this if you merged more often.

k-core/src/main/java/org/kframework/compile/sharing/TokenSortCollector.java
radumereuta: For the SDF parser I call it for the entire Definition, but for the new parser I want to call it just for a Module. So I choose the common class of both. In any case, `TokenSortCollector` matches only on `Production` so the starting point doesn't matter.

k-core/src/main/java/org/kframework/compile/sharing/TokenSortCollector.java
radumereuta: Actually, I commented this one out because it is code written by @yilongli and I wanted to ask him why did he needed that. In the front end this restriction doesn't make too much sense.
So.. @yilongli, do you remember why you needed this check?

k-core/src/main/java/org/kframework/parser/concrete2/KSyntax2GrammarStatesFilter.java
radumereuta: You are right, but I'm not there yet. It's going to be an incremental change because I also have to figure out how to reject things in the new implementation. But yes, I will have to eliminate lines 187-224.

k-core/src/main/java/org/kframework/parser/utils/QuickParser.java
radumereuta: I can give it a Module or a Definition.
The grammar generator is a visitor which looks only at Productions, so deciding what goes into the parser is somewhere outside, decided by the user of this function.

k-core/src/main/javacc/Outer.jj
radumereuta: We should do this at one point, but I don't think we're there yet.

k-distribution/include/modules/k-parser.k
radumereuta: Actually, it doesn't make sense to me. Why uris.k? I want a parser, what does that have to do with uris?
I think we have an outdated organization in these two files. If I'm wrong, please explain the logic, because the names are not intuitive at all.

k-core/src/main/java/org/kframework/parser/utils/QuickParser.java
radumereuta: This ugliness comes from `Context`. Once that is sorted out, this should also be cleaner.
This function should in no way depend on `kompileOptions` and `globalOptions`, but this is the only way I can instantiate the `Context` object. When we are going to use `DefinitionContext` and `ModuleContext` we won't have this issue.

k-core/src/main/javacc/Outer.jj
radumereuta: I mean this should still work in parallel with SDF at least for a while longer, and it shouldn't behave differently.

.gitignore
dwightguth: is there a reason you can't just keep the test directory outside of the source controlled directory structure, like I do?

k-distribution/include/modules/k-parser.k
dwightguth: it's just legacy really is why it's in uris.k. If you want to create a new file for it that's fine, but this file is legacy stuff we are no longer using, and is also in modules, whereas this function should be in builtins and be autoincluded.

.gitignore
radumereuta: I guess I could do that too... I actually haven't thought of this.

k-core/src/main/java/org/kframework/compile/sharing/TokenSortCollector.java
dwightguth: okay. Given that we are replacing all the data structures soon this should probably be fine, but I'll talk to Cosmin to see if there's a better solution moving forward.

k-core/src/main/java/org/kframework/parser/utils/QuickParser.java
dwightguth: unfortunately I can't merge this in its current form. The purpose of migration patterns is to create a smooth flow from the old behavior to the new without putting the code in an inconsistent state ever. Dependencies that should not be present in the code base should be included in the migration planning and gradually removed, but they shouldn't simply be faked in production code.

k-core/src/main/java/org/kframework/compile/sharing/TokenSortCollector.java
cos: once we transition to `kore`, I think the parser should only accept a `kore.Module`. It will not have to worry how a definition is flattened into a module (and the module will handle its own dependencies). I know Radu wants the parser to work using only one syntax rule. We can have a helper function converting a syntax rule to a module for that.

k-core/src/main/java/org/kframework/parser/utils/QuickParser.java
cos: actually, let's have this only take a module or a definition already. we can do this by overloading.

k-core/src/main/java/org/kframework/parser/utils/QuickParser.java
cos: @dwightguth, I disagree. the K code base is full of unnecessary dependencies. they slow down development and make unit testing close to impossible. mocking that ugliness to create clean interfaces is a great first step towards getting rid of them. the simple action of parsing a small snippet was unfeasible before -- there was too much junk to initialize. `QuickParser` makes this easy. at some point it won't have to create all that junk underneath. but any clients building upon it now will never know about the junk, and thus they don't have to propagate the it. it is a very nice development, one that should be propagated throughout the code base.

dwightguth: As I said, I have no complaints with this at all if we want it to be a class in src/test/java. It's specifically if we have the intent to use this in production code that I am not satisfied. We have a number of features that we have to support which will not be supported if this code as it currently exists is used in production.

cos: could you give an example?

cos: anyway, let's not dwell on this. @radumereuta, are you ok with moving this to `src/test`?

radumereuta: For now, yes. I don't think I need it somewhere else.

kernel/src/test/java/org/kframework/parser/utils/KoreIT.java
```diff
@@ -0,0 +1,42 @@
...
+import org.kframework.parser.outer.Outer;
+import org.kframework.utils.BaseTestCase;
+import org.kframework.utils.errorsystem.ParseFailedException;
+import org.kframework.utils.file.JarInfo;
+
+import java.io.File;
+
+/**
+ * Test a the kore definition using the new parser.
+ * KoreIT has to be run after the distribution has been built (apparently), therefore the name IT at the end.
```
cos: why does it have to run after the distribution? Is it because `"/samples/kast/` are not available before?

radumereuta: Yup.

cos: This is not really an integration test. I say we move the definitions to `test/resources` and allow the test to run along with the other unit tests. Also, we do want `kore.k` in the main tool, not in some samples directory.

radumereuta: Actually, I would like the kore.k file to be somewhere in the /include
directory.

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIOOperationsImpl.java
cos, line 52: I think we can make it configurable at a later point. Let's have this pull request merged so we can move other things along.

pdaian:
Jenkins: test this please

dwightguth:
Jenkins: test this please

radumereuta:
Do I have to reopen this one?

dwightguth:
Radu, it is actually being tested, you can look at it on Jenkins. I am
asking Philip to look into why the status is not getting updated.

radumereuta:
Ok, then @dwightguth please re-review. I think I've addressed all your comments (you may have to look over the entire thing again, since it's been a while).
I remove the part where I tried to send back a message if the parse failed. I will make a separate pull request with only that change at a later time.

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIOOperationsImpl.java
```diff
@@ -153,6 +159,22 @@ public Term parse(StringToken term1, StringToken term2, TermContext termContext)
     }
 
     @Override
+    public Term parseInModule(StringToken input, StringToken startSymbol, StringToken moduleName, TermContext termContext) {
+        try {
+            org.kframework.kil.Term kast = programLoader.parseInModule(
+                    input.stringValue(),
+                    new GeneratedSource(this.getClass()),
```
dwightguth: Sources.generatedBy

dwightguth:
src/test/resources

kernel/src/test/java/org/kframework/parser/utils/KoreTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Test a the kore definition using the new parser.
+ * KoreTest has to be run after the distribution has been built (apparently), therefore the name IT at the end.
+ */
+public class KoreTest extends BaseTestCase {
+
+    @Test
+    public void testKore() throws Exception {
+        System.out.println(new File("src/test/resources/kast/quote-unquote.kore").getAbsolutePath());
```
cos: leftover println

cos:
Did the tests pass locally?

I guess it failed because Eclipse/Intellij run the tests relative to the project/module, while mvn might (Dwight?) run them relative to the root of everything.

PR #1117, opened 2014-10-28
remove static state in JavaClassesFactory

dwightguth:
@radumereuta please review


radumereuta:
Somehow this solution seems horrible, but still better than before.

lgtm.

dwightguth:
Well ideally JavaClassesFactory would handle all the XML code and then call constructors that set the fields directly, but at this point with this code going away it's not worth the effort. I just did what was minimally necessary to remove the static state.

PR #1119, opened 2014-10-28
make conses thread safe

dwightguth:
@radumereuta please review

### DIFF
--- a/k-core/src/main/java/org/kframework/utils/StringUtil.java
+++ b/k-core/src/main/java/org/kframework/utils/StringUtil.java
@@ -418,20 +418,18 @@ public class StringUtil {
         return StringUtil.unEscapeSortName(ret);
     }
 
-    private static int number = 0;
-
     /**
      * Generate incremental numbers that doesn't contain the number 1
      *
      * @return an integer that doesn't contain the number 1
      */
-    public static int getUniqueId() {
+    public static int getUniqueId(int[] number) {
         boolean valid = false;
         while (!valid) {
-            int nr = number;
+            int nr = number[0];
             while (nr > 0) {
                 if (nr % 10 == 1) {
-                    number++;
+                    number[0]++;
                     break;
                 } else {
                     nr /= 10;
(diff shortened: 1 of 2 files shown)


k-core/src/main/java/org/kframework/utils/StringUtil.java
```diff
@@ -418,20 +418,18 @@ public static String getSortNameFromCons(String str) {
...
 
-    private static int number = 0;
-
     /**
      * Generate incremental numbers that doesn't contain the number 1
      *
      * @return an integer that doesn't contain the number 1
      */
-    public static int getUniqueId() {
+    public static int getUniqueId(int[] number) {
```
radumereuta: Can't you just make this one synchronized? I have no preference for what number I get in the end.

dwightguth: no, I can't make it synchronized because it needs to be consistent across calls to kompile. That means in order to be truly thread safe each definition must have its own counter.

radumereuta: Then how do you establish what number to assign to each definition?

dwightguth: each definition starts at 0 and increments in a particular order.

radumereuta: That particularity is exactly what I'm asking for. I don't see it in this diff. Could you point me to some place in the code base?

dwightguth: each definition has their own context. Each context has their own counter. Each counter starts at 0 and is incremented whenever this function is called. All of these things are in the diff. The only thing that has changed is the fact that the field is now an instance field on Context instead of being static.

radumereuta: Then why not move this function to the context also?
It looks clumsy like this.

dwightguth:
@radumereuta please re-review

radumereuta:
lgtm, merge if tests pass.

PR #1121, opened 2014-10-28
more file io refactoring

dwightguth:
Removes more references to static working directory.

@osa1 @radumereuta Please review


k-core/src/main/java/org/kframework/ktest/Proc.java
```diff
@@ -97,6 +99,7 @@
     private ProcOutput procOutput = new ProcOutput(null, null, -1);
 
     private final KExceptionManager kem;
+    private final Map<String, String> env;
 
     public static final int SIGTERM = 143;
```
radumereuta, line 104: This seems very specific. What is it? Is it portable?

k-core/src/main/java/org/kframework/ktest/Test/TestSuite.java
```diff
@@ -19,10 +22,10 @@
     private final KTestOptions options;
     private final ReportGen reportGen;
 
-    public TestSuite(List<TestCase> tests, KTestOptions options) {
+    public TestSuite(List<TestCase> tests, KTestOptions options, FileUtil files) {
         this.tests = tests;
         this.options = options;
-        reportGen = options.getGenerateReport() ? new ReportGen() : null;
+        reportGen = options.getGenerateReport() ? new ReportGen(files.resolveWorkingDirectory("junit-reports")) : null;
```
radumereuta: "junit-reports" should be a constant defined somewhere else. Just in case you will use it somewhere else and you want to change it in the future.

k-core/src/main/java/org/kframework/parser/DefinitionLoader.java
```diff
@@ -61,7 +62,8 @@ public DefinitionLoader(
             OuterParser outer,
             @Backend.Documentation boolean documentation,
             @Backend.Autoinclude boolean autoinclude,
-            FileUtil files) {
+            FileUtil files,
+            Sdf2Table sdf2Table) {
```
radumereuta, line 65: This injection is wrong. I will follow up with comments when I have the reason why.

k-core/src/main/java/org/kframework/parser/utils/Sdf2Table.java
```diff
@@ -8,16 +8,26 @@
 import org.kframework.utils.OS;
 import org.kframework.utils.ThreadedStreamCapturer;
 
+import com.google.inject.Inject;
+import com.google.inject.Provider;
+
```
radumereuta, line 13: Explain in a few lines the process of this entire file.
Mainly, why do I need a provided ProcessBuilder, and I can't use a new one?

maude-backend/src/main/java/org/kframework/krun/ioserver/commands/CommandSystem.java
```diff
@@ -4,9 +4,10 @@
 import org.kframework.krun.RunProcess;
 import org.kframework.krun.RunProcess.ProcessOutput;
 import org.kframework.krun.api.io.FileSystem;
+
 import java.net.Socket;
-import java.util.Map;
 import java.util.HashMap;
+import java.util.Map;
```
radumereuta, line 8: What happened here?
I think this is worth a revert to preserve the change to a minimum. Could make tracing the original author easier.

radumereuta:
Other than Sdf2table I don't have other comments, but I don't feel confident reviewing changes in krun and ktest, so you should wait for other reviews there.
I think tomorrow I will come to the RV. We can discuss the issue then, because something seems wrong in Sdf2table.

k-core/src/main/java/org/kframework/ktest/Proc.java
dwightguth, line 104: @osa1? I didn't write this, as you can see it was already there.

dwightguth:
I agree what I'm doing with Sdf2Table isn't very nice but it's necessary if we want Sdf2Table to inherit the environment of kompile moving forward. This is being done to make room for the client-server architecture that makes krun >4x faster.

k-core/src/main/java/org/kframework/ktest/Proc.java
osa1, line 104: @radumereuta see #1127 .

radumereuta:
lgtm.
It looks ugly and counter intuitive, but since we are planing to remove SDF I can look the other way for now.

k-core/src/main/java/org/kframework/parser/utils/Sdf2Table.java
osa1, line 13: I can think of two reasons:
- This way we can make `Sdf2Table` testable in unit tests by providing it a mocked `ProcessBuilder` object that doesn't really spawn new processes. (I guess this is one of the classic arguments for factory classes)
- `Sdf2Table` can't create new `ProcessBuilder` because it doesn't know some of the things it needs to know to create a `ProcessBuilder`. (like shell process environment and working dir -- in theory `Sdf2Table` can know all those but for abstraction purposes doing it in one place for the whole framework is better for usual reasons)

osa1:
lgtm.

PR #1122, opened 2014-10-28
Fixing a bug in the require resolution algorithm.

radumereuta:
@dwightguth please review.
You recently did some changes and removed a functionality that people were using.
You assumed that autoinclude.k is always in the /include directory, but the way I implemented the require resolution is to first look relatively to the current working directory, and then in the /include. This way, if someone didn't like the default autoinclude.k could simply create his/her own in the current working directory and control what is included and what not.

### DIFF
--- a/k-core/src/main/java/org/kframework/parser/generator/OuterParser.java
+++ b/k-core/src/main/java/org/kframework/parser/generator/OuterParser.java
@@ -21,6 +21,7 @@ import org.kframework.parser.outer.Outer;
 import org.kframework.utils.errorsystem.KExceptionManager;
 import org.kframework.utils.file.FileUtil;
 import com.google.inject.Inject;
+import org.kframework.utils.file.JarInfo;
 
 public class OuterParser {
     private List<DefinitionItem> moduleItems;
@@ -71,7 +72,7 @@ public class OuterParser {
                 List<DefinitionItem> tempmi = moduleItems;
                 moduleItems = new ArrayList<DefinitionItem>();
 
-                File autoinclude = files.resolveKBase("include/" + autoincludedFile);
+                File autoinclude = buildCanonicalPath(autoincludedFile, file);
                 if (!autoinclude.exists())
                     throw KExceptionManager.criticalError(missingFileMsg + autoinclude + " autoimported for every definition ");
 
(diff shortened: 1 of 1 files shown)


dwightguth:
I don't like this. It introduces unneeded complexity to the process of determining where files are resolved to. If the user doesn't like the autoincludes, they can always turn it off and then manually include whatever they wish.

Can you provide to me a use case that the current implementation is not able to handle?

radumereuta:
Do we still have the --noprelude option? I don't see it in the command line options.
I've also tested it and nothing happens, and I really mean nothing. No error messages, no files created... I just get the prompt back immediately.

dwightguth:
see #1123 

Your mistake is that the option is visible under kompile -X and is --no-prelude with a hyphen.

radumereuta:
Yeah, I saw in the end.
@daejunpark would the solution that @dwightguth proposes work for you?
What you will have to do, is to include the autoinclude-java.k in the main js.k file, and kompile the definition each time with the --no-prelude option.

I don't particularly like this solution since it makes the tool look less pretty. I implemented the previous behavior thinking of the C preprocessor, so it's not such a wild idea.

@grosu any thoughts on this?

dwightguth:
If this is specifically about the javascript semantics, I would be interested to know first of all why you need to not include the autoincludes. That seems like a use case that I would like to know more about the motivation of.

radumereuta:
The problem is related to modularity and autoincluded stuff in the parsers.
@daejunpark needs his own definition for Floats, which conflict with the ones that are predefined.

We talked at one point about not include anything, but it is a big change, since every single definition that we currently have, is based on this.

dwightguth:
So how does he intend to do any type of floating point computation without including the Float module, then? it seems to me that while you may not want Float included in the syntax module, you still need it in the autoinclude that gets included in the definition. I guess I just don't follow why having a different syntax for floats in JavaScript than K led to the decision to change the autoincludes. It doesn't match with my own understanding of how to implement PL data types in K.

radumereuta:
I only know the parsing problem. @daejunpark care to weigh in?

daejunpark:
Sorry for the late reply.

The reason why I need a different syntax of float for the program parser is that `Infinity` and `NaN` are properties of the global object of JS. For example, JS allows `this.Infinity` or `this.NaN`. So, the program parser needs to have its own syntax of float, while the rule parser will have the built-in float syntax. This is the same case for the string syntax, because JS allows the single quoted strings.

So, the program parser of JS needs its own syntax for float and string. That's why I need to modify the auto-included builtin modules.

dwightguth:
I am going to close this in favor of the resolution offered by #1128.

PR #1124, opened 2014-10-28
Context and transformers work - 2

osa1:
Continued from #1100.

Not ready for merge -- opened to ask questions on the code and get some reviews/feedback.


k-core/src/main/java/org/kframework/krun/InitialConfigurationProvider.java
```diff
@@ -50,10 +52,12 @@ public InitialConfigurationProvider(
             ConfigurationCreationOptions ccOptions,
             Stopwatch sw,
             Context context,
+            Definition definition,
```
osa1, line 55: @dwightguth, can you help me understand how is this injected? This code works but I don't understand how. I was actually expecting this code to fail because I didn't write a provider for `Definition`.

dwightguth, line 55: yeah... this has to do with the default behavior of Guice when it discovers 0-args constructors. You are getting a blank definition because it sees the 0-arg constructor and tries to construct it. At one point I considered changing the code to disable this functionality but it didn't seem worth it at that point. Would you find it sufficiently less confusing if it only bound constructors that were annotated? I can show you how to go about changing that functionality, it just requires minor tweaks to every class that we currently inject without an @Inject annotation.

osa1, line 55: > Would you find it sufficiently less confusing if it only bound constructors that were annotated?

Yes, I think that would be great.

I guess we'll need to add a `Void` argument to classes with 0-arg constructors that we don't want to use in injection, right?

osa1:
One tricky thing about this patch is that visitors/transformers are sometimes instantiated from other visitors/transformers and in those cases we need to make sure that `currentDefinition` and `currentModule` are properly set.

k-core/src/main/java/org/kframework/krun/InitialConfigurationProvider.java
dwightguth, line 55: That is actually not the preferred approach. I suggest you add the
following line to the configure method of
org.kframework.utils.inject.CommonModule:

`binder().requireAtInjectOnConstructors();`

Then delete the constructors with the Void argument, and fix all the
creation and provision errors that pop up by adding 0-arg @Inject annotated
constructors.

On Wed, Oct 29, 2014 at 1:53 AM, Ömer Sinan Ağacan <notifications@github.com

dwightguth:
@osa1 I think that is an acceptable compromise given that in the past they had to be instantiated with a Context.

osa1:
@dwightguth @radumereuta I'm making progress and if you're available some reviews would be really helpful.

What's done:
1. Definition object is now serialized in kompile and loaded/injected by krun frontend.
2. Using (1), I moved some state(`configVarSorts`, `numModules`, `numSentences`, `numProductions`, `numCells`) from `Context` to `Definition`.
3. Removed `Context` dependency from some visitors/transformers. (passed `null` instead of `Context`)
4. Added a lot of constructors to be able to create new visitors/transformers from other visitors/transformers. New constructors are needed because otherwise we couldn't pass `currentDefinition` and `currentModule` to created visitor/transformer.
5. Similar to `Context`, `Definition` is mocked in `BaseTestCase`. This `Definition` represents the one that were serialized in kompile and loaded in krun. (e.g. "definition.bin" in `-kompiled` dir)
6. Several warnings are fixed on the way.

radumereuta:
All of these sound amazing. Thanks @osa1

dwightguth:
Unfortunately I don't think we're ready to remove Context from the constructor of visitors yet. Currently it is still required by AbstractVisitor to visit rewrites, which depend on the sort graph. If you want to remove Context from visitors, you will first need to complete the step which moves the sort graph to a field on Module.

k-core/src/main/java/org/kframework/kil/Definition.java
```diff
@@ -29,16 +31,28 @@
     private String mainSyntaxModule;
     private Poset<String> modules = Poset.create();
 
+    public Map<String, Sort> configVarSorts = new HashMap<>();
+
+    /** For statistics. See {@link #printStatistics}. */
+    public int numModules, numSentences, numProductions, numCells;
```
dwightguth: I think these fields are actually used by only one place in the code, so maybe they should be moved there instead?

k-core/src/main/java/org/kframework/kil/Definition.java
```diff
@@ -169,4 +183,13 @@ public Definition shallowCopy() {
     public void setChildren(List<DefinitionItem> children, Enum<?> _) {
         this.items = children;
     }
+
+    public void printStatistics() {
```
dwightguth: same here

k-core/src/main/java/org/kframework/kil/loader/CountNodesVisitor.java
```diff
@@ -11,18 +11,21 @@
 import org.kframework.kil.visitors.BasicVisitor;
 
 public class CountNodesVisitor extends BasicVisitor {
```
dwightguth, line 13: yes, here. Maybe make these fields be on the visitor itself, then make getters for them?

k-core/src/main/java/org/kframework/kompile/KompileFrontEnd.java
```diff
@@ -85,29 +84,29 @@ public boolean run() {
...
+        loader.saveOrDie(files.resolveKompiled("definition.bin"), compiledDef);
 
-        verbose();
+        verbose(compiledDef);
         return true;
     }
 
-    private void verbose() {
+    private void verbose(Definition def) {
         sw.printTotal("Total");
```
dwightguth, line 98: and then move the invocation of CountNodes to here?

k-core/src/main/java/org/kframework/parser/ProgramLoader.java
```diff
@@ -81,13 +78,13 @@ public ASTNode loadPgmAst(String content, Source source, Boolean kappize, Sort s
...
         XmlLoader.reportErrors(doc);
         JavaClassesFactory.startConstruction(context);
-        out = JavaClassesFactory.getTerm((Element) doc.getDocumentElement().getFirstChild().getNextSibling());
+        ASTNode out = JavaClassesFactory.getTerm((Element) doc.getDocumentElement().getFirstChild().getNextSibling());
         JavaClassesFactory.endConstruction();
 
-        out = new PriorityFilter(context).visitNode(out);
-        out = new PreferAvoidFilter(context).visitNode(out);
+        out = new PriorityFilter(context, null, null).visitNode(out);
+        out = new PreferAvoidFilter(context, null, null).visitNode(out);
```
dwightguth: these should actually be initialized. Presumably in this case to the Definition object and the syntax module, since this is the program parser.

k-core/src/main/java/org/kframework/parser/ProgramLoader.java
```diff
@@ -158,19 +154,19 @@ public Term processPgm(String content, Source source, Sort startSymbol,
                 out = new MakeConsList(context).visitNode(out);
                 if (globalOptions.debug)
                     System.err.println("Clean: " + out + "\n");
-                out = new PriorityFilter(context).visitNode(out);
-                out = new PreferAvoidFilter(context).visitNode(out);
+                out = new PriorityFilter(context, null, null).visitNode(out);
```
dwightguth: same for these

dwightguth:
okay, those are my comments so far. I'll have more to say after the diff is more mature.

k-core/src/main/java/org/kframework/parser/ProgramLoader.java
osa1: As far as I can see we don't have a `Definition` at that time to initialize the fields. I think parser is operating on `Term` (or `ASTNode`, which is superclass of `Definition`).

@radumereuta  is this part of the code run by any tests?

osa1:
- Addressed some of the comments.
- Rebased on current master branch.

k-core/src/main/java/org/kframework/parser/ProgramLoader.java
osa1: Actually, this visitor doesn't need `Definition`, `Context` or any other state. Apparently I added the arguments by mistake.

osa1:
I'm considering this as a failed attempt and closing this. I'll open a new pull request about this soon.

PR #1129, opened 2014-10-29
make profiler thread safe

dwightguth:
@yilongli please review

I elected to not use instance state because it would have required a much more invasive change to the code base. Let me know if you're okay with this solution.


yilongli:
lgtm. Thanks, Dwight.

PR #1131, opened 2014-10-29
add Scala KORE inner classes

cos:
@dwightguth, **please review** 
- add kore module to the mvn build system, make kernel (and only kernel) dependent on kore
- add Scala KORE classes


cos:
@dwightguth, have a look

cos:
seems like we have a failure in the .jar file path fix:
http://office.runtimeverification.com:8080/rv-jenkins/job/k-framework-pull-requests-maven/435/org.kframework.k$kernel/testReport/

@dwightguth, what do you think?

PR #1132, opened 2014-10-29
hotfix/rename k core to kernel

cos:
**please review** if ktest passes

@dwightguth, let's make this in two steps. this is the k-core -> kernel rename


cos:
I don't get this failure locally. @dwightguth, could you please have a look?

dwightguth:
This  diff is still deleting the marker file we talked about yesterday, that is why the build is failing.

cos:
ok, I force added it in the previous commit, but didn't after the split. let's see now.

dwightguth:
lgtm

PR #1133, opened 2014-10-29
feature/scala support

cos:
**please review**

just some helper plugins for SBT -- no effect on most users.


cos:
@dwightguth, have a look when you can.

PR #1134, opened 2014-10-29
feature/sketch kore java interface

cos:
**please review**

Mostly sketches of the main data structures. We'll have to use them for a while to make sure everything is in order.

Also, once the interface stabilizes a bit (after we use it for a bit) I will document how to use it on a wiki page. For now, consider the examples as being the documentation.


cos:
Jenkins: test this please

osa1:
Indentations are different from rest of the framework.

radumereuta:
I agree with @osa1 here. https://github.com/kframework/k/wiki/Style-Guidelines

cos:
we're using a different language, so the style can be different. I'm using the official Scala style guide:
http://docs.scala-lang.org/style/indentation.html

osa1:
@cos there some Java files in this patch.

cos:
@osa1, thanks. fixed the indentation.

cos:
@dwightguth, @osa1, please have a look.

cos:
also, the the description in the first comment.

kore/pom.xml
```diff
@@ -24,6 +24,12 @@
       <artifactId>scala-library</artifactId>
       <version>${scala.version}</version>
     </dependency>
+    <dependency>
```
dwightguth: this pom should already have access to junit 4.11. You shouldn't need this.

dwightguth:
one minor comment plus you should also merge against master because this pull request still references the diff that we merged already with sbt.

cos:
removed the redundant dependency and rebased.

dwightguth:
Merge if tests pass.

PR #1140, opened 2014-10-31
refactor ktest to run krun immediately following kompile

dwightguth:
@osa1 please review

This allows the nailgun branch to run many more definitions with less overhead by being able to cache fewer parse tables and classloaders as a result of the fact that the krun programs of a definition are run as soon as possible after the definition is compiled.


osa1:
lgtm.

PR #1141, opened 2014-10-31
make initial config obtained only once via provider instead of singleton

dwightguth:
@yilongli please review

I did this because singletons rely on a static lock in Guice which causes a deadlock if process A spawns process B while provisioning, and then waits for the process to complete. This was causing external parsers to not work.


yilongli:
lgtm

PR #1142, opened 2014-10-31
change krun to normalize anonymous variables in output

dwightguth:
This causes krun to be unaffected by the static counter of anonymous variables.

@yilongli please review


yilongli:
lgtm

PR #1145, opened 2014-11-03
Context refactoring

osa1:
Second part of https://github.com/kframework/k/wiki/Design-of-the-new-compiler-parser .

Some things listed in the wiki link may not be possible:
- ~~`Map<String, ASTNode> locations`: Not possible to move to Maude executor, because generated by `DefinitionLoader` which runs in kompile. We have to serialize this field and load in Maude krun.~~
- ~~`Set<Sort> variableTokenSorts`: This is still used, are we sure that this won't break anyting? Or is it a mistake in the wiki page?~~

These are also missing:
- `SetMultiMap<String, Production> tags`: This should be possible, but it's a bit tricky. The problem is that currently it's collecting while productions are parsed to prevent redundant computations. I need to find a way to generate it after productions are added.

### DIFF
--- a/kernel/src/main/java/org/kframework/compile/checks/CheckSyntaxDecl.java
+++ b/kernel/src/main/java/org/kframework/compile/checks/CheckSyntaxDecl.java
@@ -90,7 +90,7 @@ public class CheckSyntaxDecl extends BasicVisitor {
                 sorts++;
                 NonTerminal s = (NonTerminal) pi;
                 if (!s.getSort().isCellSort()) {
-                    if (!context.definedSorts.contains(s.getSort())) {
+                    if (!context.getAllSorts().contains(s.getSort())) {
                         String msg = "Undefined sort " + s;
                         throw KExceptionManager.compilerError(msg, this, s);
                     }
@@ -103,7 +103,7 @@ public class CheckSyntaxDecl extends BasicVisitor {
             if (pi instanceof UserList) {
                 sorts++;
                 UserList s = (UserList) pi;
-                if (!s.getSort().getName().startsWith("#") && !context.definedSorts.contains(s.getSort())) {
+                if (!s.getSort().getName().startsWith("#") && !context.getAllSorts().contains(s.getSort())) {
                     String msg = "Undefined sort " + s.getSort();
                     throw KExceptionManager.compilerError(msg, this, s);
                 }
--- a/kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
+++ b/kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
@@ -107,7 +107,7 @@ public class KastFrontEnd extends FrontEnd {
             if (env.get("KRUN_SORT") != null) {
                 sort = Sort.of(env.get("KRUN_SORT"));
             } else {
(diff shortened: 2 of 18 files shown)


dwightguth:
I thought we were going to serialize the definition though? that should allow locations to be populated in krun.

I am sure regarding variableTokenSorts. This is leftover code from the maude symbolic backend and should be removed.

Please let me know if you have any questions about how you would like to be able to populate tags. I agree it should be possible.

kernel/src/main/java/org/kframework/kil/Cell.java
```diff
@@ -56,6 +56,12 @@
     String endLabel;
 
     /**
+      * Builtin data structure (list, map, set) for the cell.
+      * Set by {@link org.kframework.compile.transformers.Cell2DataStructure}.
+      */
+    public CellDataStructure dataStructure;
```
dwightguth: I meant an Attribute, not a field on this class.

osa1:
Looks like build is failed because of some of the recent commits. I'll rebase this.

kernel/src/main/java/org/kframework/kil/Cell.java
osa1: Sorry, what do you mean by an attribute?

dwightguth: I mean look at org.kframework.kil.Attribute and the helper functions
associated with that class in ASTNode.

On Mon, Nov 3, 2014 at 11:20 AM, Ömer Sinan Ağacan <notifications@github.com

radumereuta:
I think this is ok.

osa1:
I had a problem with Cell attributes related with Java backend(stack trace of error site is completely unrelated, but it has to be related with Cell attributes because it's only happening with Cell attributes commit) so removed that commit for now.

osa1:
@dwightguth 

> I am sure regarding variableTokenSorts. This is leftover code from the maude symbolic backend and should be removed.

Removing that broke Java backend. (commit: https://github.com/osa1/k/commit/45eb042227c02f748e76f844383eb40c3022d40a) Should I put that back?

dwightguth:
Per our discussion with Cosmin, I suggest you do the minimum amount of work needed to get this cleaned up and passing on some subset of the tasks we intended to complete, and then merge it in that form before moving on to the data structure refactoring itself.

osa1:
There's this problem with moving cell data structure out of Context: Java backend is creating sorts with hard-coded labels like "generatedTop". Previously cell data structures where queried using cell labels this was working, but since we removed CellLabel -> CellDataStructure map it can't find the data structure and(because now it's searching the data structure using fresh Cell with label "generatedTop") I think this is what's making Java backend to fail.

dwightguth:
Yeah just revert that part for now since we intend to address these issues
primarily later.

osa1:
Finally, tests are passing.

@dwightguth please review.

I didn't make any changes about `tags`, `cellDataStructures` and `variableTokenSorts`.

dwightguth:
Excellent. Go ahead and work with Cosmin on the KORE refactorings. Let me know if you run into situations where you don't have something to work on.

PR #1147, opened 2014-11-03
fix ambiguity in substitution.k

dwightguth:
@traiansf you wrote this, can you review?


dwightguth:
@yilongli can you check to make sure that the latest cil semantics is actually working? it seems to be failing in this test and I don't think cil even uses substitution.

yilongli:
Jenkins: retest this please

traiansf:
The code there is pretty ugly, but the change you did is indeed consistent with the original intended meaning, so you can merge once you sort it out with CIL

PR #1155, opened 2014-11-04
feature/outer kore

cos:
**please review**

outer KORE sketch. will probably change as we start using it.


cos:
@dwightguth, @radumereuta, @osa1, have a look.

cos:
review after #1134 is merged.

dwightguth:
I will review this once you clean up this diff.

cos:
rebased

kore/build.sbt
```diff
@@ -2,3 +2,6 @@
 
 net.virtualvoid.sbt.graph.Plugin.graphSettings
 
+libraryDependencies ++= Seq(
+   "junit" % "junit" % "4.8.1" % "test"
```
dwightguth: 4.11?

kore/src/main/scala/org/kframework/kore/outer/KLabelMappings.scala
```diff
@@ -0,0 +1,19 @@
...
+  lazy val labelsToProductions: Map[kore.KLabel, Set[SyntaxProduction]] =
+    sentences collect {
+      case prod: SyntaxProduction => (makeKLabel(prod.items), prod)
+    } groupBy (_._1) mapValues (_ map { _._2 })
+
+  def makeKLabel(items: Seq[ProductionItem]): kore.KLabel = kore.KLabel(
+    items map {
+      case NonTerminal(sort) => "_"
+      case Terminal(string) => string
+      case RegexTerminal(regex) => ???
```
dwightguth, line 17: this looks suspiciously like a case that needs to be taken care of

kore/src/main/scala/org/kframework/kore/outer/outer-to-string.scala
```diff
@@ -0,0 +1,46 @@
...
+}
+
+trait DefinitionToString {
+  self: Definition =>
+  override def toString = modules.mkString("\n\n\n")
+}
+
+trait RuleToString {
+  self: Rule =>
+  override def toString = "  rule " + body + attributes
```
dwightguth: side conditions?

kore/src/main/scala/org/kframework/kore/outer/outer-to-string.scala
```diff
@@ -0,0 +1,46 @@
...
+}
+
+trait ConfigurationToString {
+  self: Configuration =>
+  override def toString = "  configuration " + contents
```
dwightguth: side conditions? attributes?

kore/src/main/scala/org/kframework/kore/outer/outer-to-string.scala
```diff
@@ -0,0 +1,46 @@
...
+}
+
+trait SyntaxProductionToString {
+  self: SyntaxProduction =>
+  override def toString = "" + items.mkString(" ") + attributes
```
dwightguth, line 20: looks like you're missing some cases for some of the types of production items.

kore/src/main/scala/org/kframework/kore/outer/outer-to-string.scala
```diff
@@ -0,0 +1,46 @@
+package org.kframework.kore.outer
+
+trait ModuleToString {
+  self: Module =>
+  override def toString = "module " + name + "\n" + sentences.toList.sortBy(_.toString).reverse.mkString("\n\n") + "\n\nendmodule"
+}
+
+trait DefinitionToString {
+  self: Definition =>
+  override def toString = modules.mkString("\n\n\n")
```
dwightguth: requires?

kore/src/main/scala/org/kframework/kore/outer/outer-to-string.scala
```diff
@@ -0,0 +1,46 @@
+package org.kframework.kore.outer
+
+trait ModuleToString {
+  self: Module =>
+  override def toString = "module " + name + "\n" + sentences.toList.sortBy(_.toString).reverse.mkString("\n\n") + "\n\nendmodule"
```
dwightguth: attributes?

kore/src/main/scala/org/kframework/kore/outer/outer.scala
```diff
@@ -0,0 +1,63 @@
...
+  extends ModuleToString with ParserPiece with KLabelMappings
+// hooked but different from core, Import is a sentence here
+
+trait Sentence { // marker
+  val attributes: Attributes
+}
+
+case class Rule(
+  label: String,
+  body: kore.K,
```
dwightguth, line 30: side conditions?

kore/src/main/scala/org/kframework/kore/outer/outer.scala
```diff
@@ -0,0 +1,63 @@
...
+  attributes: Attributes) extends Sentence
+  with RuleToString
+
+case class Configuration(contents: kore.K, attributes: Attributes = Attributes()) extends Sentence // hooked
```
dwightguth: side conditions?

kore/src/main/scala/org/kframework/kore/outer/outer.scala
```diff
@@ -0,0 +1,63 @@
...
+  with ConfigurationToString
+
+case class ModuleComment(comment: String, attributes: Attributes = Attributes()) extends Sentence
+
+case class Import(what: String, attributes: Attributes = Attributes()) extends Sentence // hooked
+
+case class SyntaxPriority(higher: String, lower: String, attributes: Attributes = Attributes()) extends Sentence with ParserPiece
```
dwightguth, line 40: list of string?

kore/src/main/scala/org/kframework/kore/outer/outer-to-string.scala
dwightguth, line 20: also missing "syntax Sort ::="

dwightguth:
I would like to see unit tests on your string conversions.

kore/src/main/scala/org/kframework/kore/outer/KLabelMappings.scala
cos, line 17: we haven't decided how to encode it in the label yet, and there are no such cases just yet. it will fail fast once we get there.
??? is a valid Scala singleton object with the type bottom.

dwightguth:
I think you need to take a moment to go over https://github.com/kframework/k/blob/kast-in-K/samples/kast/kore.k and ensure that all the cases line up.

kore/src/main/scala/org/kframework/kore/outer/outer-to-string.scala
cos: there was back and forth regarding whether to keep them or not. the decision was to keep them, so I added them now.

kore/src/main/scala/org/kframework/kore/outer/outer-to-string.scala
cos: afaik, it should only have attributes. in any case, I'll leave it for now. the *ToStrings are only meant to help with debugging (at least for now). they are not the final pretty-printer.

kore/src/main/scala/org/kframework/kore/outer/outer-to-string.scala
cos, line 20: I'll add them one by one as we encounter them. I didn't make this complete because I'm not sure I want to keep it like this. I may want to use the main pretty-printer (yet to be developed) for toString.

kore/src/main/scala/org/kframework/kore/outer/outer-to-string.scala
cos: added it.

kore/src/main/scala/org/kframework/kore/outer/outer.scala
cos, line 30: added.

kore/src/main/scala/org/kframework/kore/outer/outer.scala
cos: does it have side conditions?

kore/src/main/scala/org/kframework/kore/outer/outer.scala
cos, line 40: I know we decided that but kore.k is not changed just yet. I'll leave it like this until we get to it.

cos:
regarding string conversions, we'll make unit tests for the unparser. for now, these are just pretty-printers for debugging purposes.

kore/src/main/scala/org/kframework/kore/outer/KLabelMappings.scala
radumereuta, line 17: Consider the content just a string. How I interpret that string is the job o the front-end. All I need from this case, is the wrapper, so I know how to interpret the content.

kore/src/main/scala/org/kframework/kore/outer/outer-to-string.scala
radumereuta: Configuration can have ensures for example. I think we should have a similar pattern on how we handle the three main types of sentences, and the only thing that is different, is the wrapper. We did special cases in the beginning, and we ended up generalizing concepts, and implementation to simplify it...

kore/src/main/scala/org/kframework/kore/outer/outer-to-string.scala
```diff
@@ -0,0 +1,46 @@
...
+//    val assocString = assoc match {
+//      case Unspecified => ""
+//      case Left => "left:\n"
+//      case Right => "right:\n"
+//      case NonAssoc => "non-assoc:\n"
+//    }
+//    assocString + productions.mkString("\n              |")
+//  }
+//}
+
```
radumereuta, line 37: Why do you have comments?

kore/src/main/scala/org/kframework/kore/outer/outer-to-string.scala
```diff
@@ -0,0 +1,46 @@
...
+trait TerminalToString {
+  self: Terminal =>
+  override def toString = "\"" + value + "\""
+}
```
radumereuta: New line at end of file?

kore/src/main/scala/org/kframework/kore/outer/outer.scala
```diff
@@ -0,0 +1,65 @@
...
+
+// will be needed once we figure out how to encode associtivity
+//object Associativity extends Enumeration {
+//  val Left, Right, NonAssoc, Unspecified = Value
+//}
+
+case class SyntaxSort(sort: Sort, attributes: Attributes = Attributes()) extends Sentence with ParserPiece
```
radumereuta, line 47: ParserPiece? What is that? It has no meaning to me.

kore/src/main/scala/org/kframework/kore/outer/outer-to-string.scala
grosu: Guys, there are NO CONFIGURATIONS in KORE!  :-)
Please see the wiki: https://github.com/kframework/k/wiki/KAST-and-KORE.

osa1: > I think we should have a similar pattern on how we handle the three main types of sentences

Where can I learn more about "three main types of sentences"?

kore/src/main/scala/org/kframework/kore/outer/KLabelMappings.scala
cos, line 17: @radumereuta, remember the discussion we had with Grigore at RV. Is the match of the regex made part of the AST or not? If not, we'll have to retain its content in either the label or the attributes of the term.

kore/src/main/scala/org/kframework/kore/outer/outer.scala
```diff
@@ -0,0 +1,65 @@
...
+  ensures: kore.K,
+  attributes: Attributes) extends Sentence
+  with RuleToString
+
+case class Configuration(contents: kore.K, attributes: Attributes = Attributes()) extends Sentence // hooked
+  with ConfigurationToString
+
+case class ModuleComment(comment: String, attributes: Attributes = Attributes()) extends Sentence
+
+case class Import(what: String, attributes: Attributes = Attributes()) extends Sentence // hooked
```
osa1, line 38: What's this? We have both `import` and `require`, but `require` is not a `sentence`.. How does that work? (just checked KORE wiki page but couldn't see anything relevant)

kore/src/main/scala/org/kframework/kore/outer/outer.scala
```diff
@@ -0,0 +1,65 @@
...
+//object Associativity extends Enumeration {
+//  val Left, Right, NonAssoc, Unspecified = Value
+//}
+
+case class SyntaxSort(sort: Sort, attributes: Attributes = Attributes()) extends Sentence with ParserPiece
+
+case class SyntaxProduction(sort: Sort, items: Seq[ProductionItem], attributes: Attributes = Attributes()) extends Sentence with ParserPiece // hooked but problematic, see kast-core.k 
+  with SyntaxProductionToString
+
+sealed trait ProductionItem // marker
```
osa1, line 52: How is this supposed to work? This doesn't have any methods or attributes, how is it useful?

How can we use this from Java side?

kore/src/main/scala/org/kframework/kore/outer/outer-to-string.scala
cos: It is leftover from when we had configurations. I will remove it.

(short parenthesis: eventually, we'll need some sort of special data structure representing the configuration -- all clients need it, and having a special class makes common operations easier. As long as it implements the right interfaces, it will still retain all the benefits listed under 
https://github.com/kframework/k/wiki/KAST-and-KORE#cells-configurations-and-configuration-abstraction
I.e., configuration terms could still be treated like regular terms.)

In any case, I will remove it, and we can have this discussion when we'll be pressed by it.

kore/src/main/scala/org/kframework/kore/outer/outer-to-string.scala
cos, line 37: see the comment of the comment

kore/src/main/scala/org/kframework/kore/outer/outer.scala
cos, line 47: it's a marker trait for anything that can be used for parsing. Remember you wanted to have the `QuickParser` take any `ASTNode` as a definition of a parser. This is ugly and error prone -- the user can send there any kind of junk, e.g., a rule (how do you parse with a rule?). 

`ParserPiece` simply means "this can be used, on its own, for parsing a bit of text". 

Alternatively (and I think this will be cleaner) we could only accept a module to be used for parsing, and maybe have a few overloaded methods which also accept production types and wrap them in a module.

kore/src/main/scala/org/kframework/kore/outer/outer.scala
cos, line 52: > How is this supposed to work? This doesn't have any methods or attributes, how is it useful?

it's a marker for all production items. for example, `SyntaxProduction` only accepts `ProductionItem`s, not any type of object. if you try to send it something else, it will fail at compile time, which is what we want.

> How can we use this from Java side?

it's a Java interface

kore/src/main/scala/org/kframework/kore/outer/outer.scala
cos, line 38: check the K documentation. `require` is similar to a Java `import`, i.e., it tells which files contain modules relevant to this definition. `import` is akin to Java `extends`, i.e., it merges another module into this one.

kore/src/main/scala/org/kframework/kore/outer/outer-to-string.scala
cos: fixed

kore/src/main/scala/org/kframework/kore/outer/outer.scala
osa1, line 52: How is an interface without method useful? Knowing that you have a value with type `ProductionItem` won't allow you to do anything with it.

(If this works as an open type family then I guess this is useful in Scala land but if we think that as a Java interface then this looks useless to me)

cos, line 52: Think in terms of a language which has inheritance. You cannot do something _to_ it, but you can do something _with_ it. You can pass it to someone which accepts `ProductionItem`s. Consider the following examples:

```
SyntaxProduction(Terminal("foo"), RegexTerminal("bar"))
```

```
SyntaxProduction(Import("foo"), NonTerminal("bar"))
```

How do you enforce that the first one compiles and the second one doesn't?

(Nothing above is specific to Scala -- it is the same in Java. And no, I don't think it is really the same as type families. Scala abstract types are more akin to type families)

kore/src/main/scala/org/kframework/kore/outer/KLabelMappings.scala
radumereuta, line 17: I have NO intention to match the content of the regex. The java regex that we currently use is very complex. I don't want to integrate that language into K. I want to use the already pretty good parser that the RegexEngine uses. I will just waste my time. Consider this node the same as a `Terminal`.

kore/src/main/scala/org/kframework/kore/outer/outer.scala
radumereuta, line 47: I am not so sure about this. If the user gives me a rule in the AST, then I will just ignore it. I can make my parser to take an ASTNode representing a `Definition` or a `Module`, and look for pieces of syntax that I'm interested in.
The way I see `QuickParser` is as something independent: hey, look into this AST and see what kind of parser can you generate.

My worry is that you are introducing a concept that we don't need. If something can be used for parsing, only the parser should know about it. Take for example a subsort `syntax Exp ::= Int`. This is a `ParserPiece`, but this also has semantic meaning when generating predicates...

Other opinions?

kore/src/main/scala/org/kframework/kore/outer/KLabelMappings.scala
cos, line 17: It seems I didn't make myself understood. I was not referring to the regex implementation -- I do not care about it. This is just about how the regex match is interpreted. 
What is the klabel for `syntax Foo ::= "foo" r"bar|buz"`? 
Furthermore, how does the KORE/KAST for "foobar" look like?

cos, line 17: We don't necessarily have to figure out the klabel right now as KORE syntax declarations must, by design, have a `klabel` attribute. But we do need to figure out how the "bar or buz" is saved in each term.

kore/src/main/scala/org/kframework/kore/outer/outer.scala
cos, line 47: your parser should expect "downmoduled" definitions, i.e, outer KORE,  not inner KORE/KAST. This means we already have types for them, they are not just ASTs. If we have types for them, why not use them at compile time?

kore/src/main/scala/org/kframework/kore/outer/KLabelMappings.scala
dwightguth, line 17: As I understand it, the value of whether you specified bar or baz is not included, since this production is neither a token nor a nonterminal. I suggest we simply find a similar placeholder to the _ that we use for nonterminals. But I agree this can be decided later for now.

dwightguth:
lgtm.

cos:
@grosu, @radumereuta, @osa1, did I address your concerns?

grosu:
This is a long thread, not sure if I'm addressing something still relevant or not, but I think this is important to clarify:

@cos asked:

> What is the klabel for syntax Foo ::= "foo" r"bar|buz"? 
> Furthermore, how does the KORE/KAST for "foobar" look like?

The short answer is "It is not your business, it is Radu's."  :)

But yes, we will need to rethink our default labels as soon as we are done with the incorporation of KORE, er even earlier if people have bandwidth for it.  But for now, recall that KORE expects _all_ productions (except subsortings) to have explicit klabels.  It is frontends' job to assign them klabels.  For now, Radu can pick whatever convention he wants to.

radumereuta:
lgtm.

dwightguth:
This build timeout is my fault, I accidentally overlapped cpu resources on the jenkins box

dwightguth:
Jenkins: test this please

cos:
@osa1, yours is the last ok needed

osa1:
lgtm.

Issue #1153, opened 2014-11-04
[Possible parser bug] UpdateAssocVisitor is called before CollectProductionsVisitor


osa1:
`UpdateAssocVisitor` is using `tags` field of Context. That field is filled by `addProduction` and `removeProduction`, which are used by `CollectProductionsVisitor`.

The problem is that `UpdateAssocVisitor` is called before `CollectProductionsVisitor` in `Definition.preprocess` and so `productions` and `tags` fields are empty at that point.

So this part of `UpdateAssocVisitor` is basically useless:

``` java
/**
 * Because the block associativity is not reflexive in SDF, I have to add it manually.
 */
@Override
public Void visit(PriorityExtendedAssoc pri, Void _) {
    for (KLabelConstant c : pri.getTags()) {
        Set<Production> prods = SDFHelper.getProductionsForTag(c.getLabel(), context);
        for (Production p : prods) {
            if (!p.containsAttribute(Constants.LEFT) &&
                !p.containsAttribute(Constants.RIGHT) &&
                !p.containsAttribute(Constants.NON_ASSOC)) {
                p.addAttribute(pri.getAssoc(), "");
            }
        }
    }
    return null;
}
```

I'm wondering how is this supposed to work. @radumereuta any ideas?

PR #1156, opened 2014-11-04
Fixing #1153.

radumereuta:
@osa1 you are right in #1153. The dependency between the two is not respected.
Nothing crashed until now since we weren't using that production in any definition, and SDF handled this somehow by default.

So, please review.


osa1:
Thanks Radu, lgtm.

PR #1157, opened 2014-11-04
add option to run K with nailgun

dwightguth:
@yilongli @osa1 please review

This pull request does not actually activate nailgun by default. In order to use nailgun, you must first set the K_NAILGUN environment variable. Once this is committed, I will experiment with the jenkins job to try to get it to run with nailgun.


grosu:
Thanks @dwightguth!

PLEASE: add a short paragraph about how this works to the manual-to-be.

Grigore

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
I believe I have finally fixed the tests.

@yilongli @cos @osa1 please review

dwightguth:
Also: ktest is down to 14 minutes!

kernel/pom.xml
```diff
@@ -126,6 +126,34 @@
       <artifactId>jline</artifactId>
       <version>0.9.94</version>
     </dependency>
+
+    <dependency>
+      <groupId>org.kframework.dependencies</groupId>
+      <artifactId>nailgun-server</artifactId>
+      <version>0.9.2-SNAPSHOT</version>
+    </dependency>
```
cos, line 134: Could nailgun live in the `k-distribution` to keep the `kernel` small? Running the entire `ktest` suite requires the k-distribution anyhow.

if yes and we agree it would be a good idea, I still think the pull request should me merged. The speed advantage is too important to postpone.

kernel/src/main/java/org/kframework/utils/file/TTYInfo.java
```diff
@@ -0,0 +1,13 @@
...
+package org.kframework.utils.file;
+
+public class TTYInfo {
+
+    public final boolean stdin, stdout, stderr;
+
+    public TTYInfo(boolean stdin, boolean stdout, boolean stderr) {
+        this.stdin = stdin;
+        this.stdout = stdout;
+        this.stderr = stderr;
```
cos, line 19: What does this class do? There is no documentation and the name is not eloquent.

kernel/pom.xml
dwightguth, line 134: This is a real compile time dependency. I guess I could move the client, but the kernel won't compile without these classes.

Now we have talked for a while about having a module encapsulating all the frontend details of our console-based tools however. We could hypothetically move the package org.kframework.main into this new module, and that would remove the dependency in the kernel. That's definitely a large enough change to not want to block this merge on though. Maybe something to think about for the long term?

cos, line 134: I agree.

dwightguth:
Jenkins: test this please

yilongli:
lgtm

dwightguth:
Jenkins: test this please

osa1:
lgtm

cos:
lgtm

PR #1162, opened 2014-11-06
Feature/inner kore improvements

cos:
**please review**, @dwightguth, @osa1 
- changed the underlying representation of KList to a linked list
- made most other KORE classes be treatable as a KList (at least from Scala)
- changed the signature of the Transformers to match Java 8 lambdas
- added many tests

### DIFF
--- a/kore/src/main/scala/org/kframework/kore/KList.scala
+++ b/kore/src/main/scala/org/kframework/kore/KList.scala
@@ -1,0 +1,80 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
+
+package org.kframework.kore
+
+import collection.{ AbstractSeq, LinearSeq, LinearSeqOptimized, Seq, generic, mutable }
+
+abstract class KList extends KListLike[KList] {
+  def copy(l: LinearSeq[K]) = KList(l: _*)
+}
+
+final case object EmptyKList extends KList with Serializable {
+  override def isEmpty = true
+  override def head: Nothing = throw new NoSuchElementException("head of empty klist")
+  override def tail: KList = throw new UnsupportedOperationException("tail of empty klist")
+  override def equals(that: Any) = that match {
+    case that: scala.collection.GenSeq[_] => that.isEmpty
+    case _ => false
+  }
+}
+
+final case class ConsKList(override val head: K, override val tail: KList) extends KList {
+  override def isEmpty = false
+}
+
+object KList extends CanBuildKListLike[KList] {
+  def apply(l: K*): KList = l.foldLeft(EmptyKList: KList) { (l: KList, h: K) => new ConsKList(h, l) }
+
+  implicit def inject(k: K): KList = KList(k)
+  implicit def seqOfKtoKList(s: Seq[K]) = KList(s: _*)
+  def fromJava(l: Array[K]) = apply(l: _*)
+  val fromList = seqOfKtoKList _
+
+  def unapplySeq(l: KList): Option[Seq[K]] = Some(l)
+}
+
+/**
+ * Describes objects which contain K's and can be iterated like a KList
+ */
+trait KListLike[+This <: KListLike[This]] extends LinearSeq[K] with LinearSeqOptimized[K, This] {
+  self: This =>
+  override def newBuilder: mutable.Builder[K, This] =
+    new mutable.ListBuffer mapResult copy
+
+  def copy(l: LinearSeq[K]): This
+}
+
+/**
+ * A KList-backed implementation of KList
+ *
+ * Implementing classes only need to provide an implementation of:
+ * def copy(LinearSeq[K]): This
+ *
+ * @see org.kframework.kore.Attributes for an example
+ *
+ * KList is not based on this class because it cannot be backed by itself
+ *
+ */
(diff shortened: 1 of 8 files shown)


cos:
Jenkins, test this

osa1:
lgtm.

dwightguth:
KList is supposed to be associative, isn't it? Why does this data structure seem to be a cons list then?

grosu:
I just to strengthen that: KList is associative!  Also, in KORE, we want the "assoc" attribute to make it clear that any K label having it is associative, that is, 

```
syntax KLabel ::= "abc" [assoc]
```

means that I can freely write `abc(.KList)` or `abc(K1)` or `abc(K1,K2,K3)` in rules, and do associative matching on them.  Moreover, if the operation is also `comm`, then we will want both associative and commutative matching on its arguments (given as a KList).  Of course, this has nothing to do with parsing right now, but let's keep it in mind for later.

cos:
Could somebody explain how the underlying implementation influences the way the data structures behave during matching? Which would be the best data structure type and why?

grosu:
Yes, we should not worry about execution at parsing stage.  But, you will still need to know something about it in order to parse rules relying on associative matching.  For example, if `abc` is some vararg multiplication operation, then I want to write the rule:

```
rule abc(L1:Klist, 0, L2:Klist) => 0
```

How do you store it in KORE if KList is a cons list?

cos:
Right now it is stored as:

```
KApply(KLabel("abc"), KList(KVariable("L1"), KToken(Int, 0), KVariable("L2")))
```

The `KList` above generates internally `ConsKList(Kvariable("L1"), ConsKList(KToken(Int, 0), ConsKList(KVariable("L2"), EmptyKList)))`, but this has nothing to do with any of the operations this entire data structure responds to. It has exactly the same interface as it would have it would be backed by a different data structure. The only difference is in the performance characteristics. For the common operation in the compiler (i.e., traversal), the linked list is likely the fastest choice. It is only slower on indexed access, but do we ever do that?

I am not arguing that it is irrelevant that `KList` is associative. I just think it is irrelevant in the choice of underlying data structure. Whether we use a an array-backed list or a cons list to hold the `KList` doesn't seem to make any difference. Both (and likely any other) will also require some form of normalization to make sure we have good "equals". Backends may make other choices for speeding up matching.

In any case, which data structure do you suggest as a better choice?

dwightguth:
Maybe I'm misunderstanding, but doesn't a cons list have inefficient append
operations?

cos:
@dwightguth, not necessarily. Appending is just linking the tail to the new list/element. The operation itself is O(1). If you don't know which is last element, it can become O(n). It is O(n) right now, but we can easily make it O(1) by remembering the tail. We would pay some price in terms of memory then. The price is only worth it if append is a very common operation in the compiler. I don't think the discussion about optimizations is worth it now, as long as the underlying representation doesn't leak into the interfaces -- and it doesn't.

dwightguth:
Okay. Can you show me where in this diff the interface for this KList is
then?

cos:
Simply consider only the `KList` abstract class. The interface is gathered from several traits. In a nutshell, it has a `KList(K... elements)` constructor and head/tail/apply/find/map/reduce/fold etc. operations (just Scala `List`). 

Again, if we decide to use something else underneath, none of these will disappear, they will just have different performance characteristics.

cos:
Once they settle a bit (and we're close), I will document how to use the classes from Java on the wiki. The operations are very homogenous, so I think a few pages on wiki coupled with the `kore.k` syntax document will be enough to know how to use everything.

kore/src/main/scala/org/kframework/kore/KList.scala
```diff
@@ -0,0 +1,80 @@
...
+  self: This =>
+  val klist: KList
+
+  override def head = klist.head
+  override def tail = copy(klist.tail)
+  override def isEmpty = klist.isEmpty
+}
+
+/**
+ * Should be extended by companion objects of classes extending KListLike
```
dwightguth, line 68: companion object? can you provide an example of what that means?

cos:
Scala doesn't have the concept of static members (fields or methods). Instead, it uses a singleton object with the same name of as the class. This singleton is called the "companion object" of the class.

The `CanBuildKListLike` trait is used for defining all the Scala implicits needed for making operations on `KListLike` collections return the same collection. E.g., in Scala: `KSequence(...).map(x => x)` has the type `KSequence`. There is no special `map` operation defined in `KSequence`. Everything is done via implicits defined by the companion object `KSequence` through extending `CanBuildKListLike` -- no extra code in either the `KSequence` class or its companion object.

For Java, as it doesn't have any similar mechanism, I had to add some extra code to achieve the same interface for `map`, and will probably extends to a few more (e.g., `filter`). 

I intend the "go to" class for extension to be `KCollection`, and it does not require the companion object when extended from Java. Again, the companion object only makes the Scala interfaces a bit nicer.

dwightguth:
okay. this diff seems fine then.

cos:
I think I can make the code needed for the Java `map` interface a bit smaller. Will merge this and will make a PR for that a bit later.

PR #1163, opened 2014-11-06
move scala dependency to the top

cos:
**please review**, @dwightguth 

there are some weird errors in Eclipse otherwise

### DIFF
--- a/kore/pom.xml
+++ b/kore/pom.xml
@@ -14,17 +14,6 @@
 
   <name>K Framework KORE</name>
 
-  <properties>
-    <scala.version>2.10.4</scala.version>
-  </properties>
-
-  <dependencies>
-    <dependency>
-      <groupId>org.scala-lang</groupId>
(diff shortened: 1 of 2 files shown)


dwightguth:
please define "weird errors" for me. This seems like a strange move given the fact that it should only actually be a dependency of that one module.

cos:
`The type scala.runtime.Nothing$ cannot be resolved. It is indirectly referenced from required .class files`

For some reason, if the dependency is not at the top, Eclipse (and mvn from time to time) decide that some Scala classes do not exist.

dwightguth:
Okay. I guess this is fine. I don't really see how that could possibly actually make a difference though. From the perspective of both maven and eclipse, it should be the same either way. It's just weird. But if you say it fixes the problem, I'm okay with this. Just make sure you move it back if it turns out you were wrong and it doesn't actually fix the problem.

PR #1164, opened 2014-11-06
refactor k to generate posters from compiled definitions

dwightguth:
**please review**

### DIFF
--- a/kernel/src/main/java/org/kframework/kompile/KompileOptions.java
+++ b/kernel/src/main/java/org/kframework/kompile/KompileOptions.java
@@ -52,7 +52,7 @@ public final class KompileOptions implements Serializable {
     @Parameter(names={"--directory", "-d"}, description="Path to the directory in which the output resides. An output can be either a kompiled K definition or a document which depends on the type of backend. The default is the current directory.")
     public String directory = ".";
 
-    @Parameter(names="--backend", description="Choose a backend. <backend> is one of [pdf|latex|html|maude|java|unparse|symbolic]. Each of [pdf|latex|html] generates a document from the given K definition. Either of [maude|java] creates the kompiled K definition. 'unparse' generates an unparsed version of the given K definition. 'symbolic' generates symbolic semantics. Experimental: 'doc' generates a .tex document, omitting rules unless specified.")
+    @Parameter(names="--backend", description="Choose a backend. <backend> is one of [pdf|coq|java]. Each creates the kompiled K definition.")
     public String backend = Backends.MAUDE;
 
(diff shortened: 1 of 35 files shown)


dwightguth:
@cos @yilongli @osa1 please review

@grosu : I am proposing to make this change because it significantly improves the performance of generating pdfs from already-compiled definitions (from being almost 2x as slow as compiling once, to being only negligibly slower than compiling once), and because it makes the implementation much simpler conceptually by separating the act of parsing and compiling a definition from generating documentation from a parsed definition.. However, I recognize this is a significant alteration of the tool to add a new frontend which generates posters from compiled definitions, so I'd like your feedback. Please let me know if this is a change you can live with in your teaching.

dwightguth:
Jenkins: test this please

grosu:
ok with me for now.  go ahead and merge if everybody else is OK.

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
```diff
@@ -52,7 +52,7 @@ public void setFiles(FileUtil files) {
     @Parameter(names={"--directory", "-d"}, description="Path to the directory in which the output resides. An output can be either a kompiled K definition or a document which depends on the type of backend. The default is the current directory.")
     public String directory = ".";
 
-    @Parameter(names="--backend", description="Choose a backend. <backend> is one of [pdf|latex|html|maude|java|unparse|symbolic]. Each of [pdf|latex|html] generates a document from the given K definition. Either of [maude|java] creates the kompiled K definition. 'unparse' generates an unparsed version of the given K definition. 'symbolic' generates symbolic semantics. Experimental: 'doc' generates a .tex document, omitting rules unless specified.")
+    @Parameter(names="--backend", description="Choose a backend. <backend> is one of [pdf|coq|java]. Each creates the kompiled K definition.")
```
yilongli: no maude?

kernel/src/main/java/org/kframework/ktest/ExecNames.java
```diff
@@ -32,4 +32,8 @@ public static String getExecutable(String exe) {
         }
         return JarInfo.getKBase(false) + FILE_SEPARATOR + "bin" + FILE_SEPARATOR + exe;
     }
+
+    public static String getKDoc() {
+        return getExecutable("kdoc");
```
yilongli, line 37: extract constant string `"kdoc"`?

yilongli:
looks ok to me

cos:
lgtm

PR #1165, opened 2014-11-07
Debugger Refactoring

msaxena2:
#### Overview

Suggested Refactoring for the debugger's setup. The following changes are suggested:
- A controller class (see  kernel/src/main/java/org/kframework/debugger/controller/DebuggerController.java)   introduced. This class acts as a sandwich layer between the fronted and the backend, contains common logic, and will be a part of the kernel.
- A new interface for the frontend introduced (see  kernel/src/main/java/org/kframework/debugger/frontend/DebuggerFrontend.java ). All frontends will now implement this interface. The ultimate goal is to have a plugin-like setup, where frontends are maintained independently from Kernel. Currently however, for experimental purposes, a textual frontend is included as a part of the Kernel. 
#### Implementation Details

Final aim is to have a plugin like setup, where user can download a frontend jar, and plug it into the kernel. This setup allows for the plugging in process, and makes it easier for others to develop frontends. Currently, the frontend and the backend are injected into the above mentioned controller class. For the purpose of this injection,  kernel/src/main/java/org/kframework/krun/KRunModule.java had to be modified. Please note that these changes aren't meant to be permanent yet. Later, we can have a service loader look up available frontends and inject them into the controller, allowing for the plugin and play.
#### Advantages
- Common logic between the frontend and the backend has been separated and put into the controller class. This allows writing frontends much easier. I'll be soon updating the wiki with details about writing a frontend. Also, the power of Transformations shine through. The frontend writer need not know anything about the State. In order to get a pretty printable result, he just needs to call the transformation's run function
- As mentioned earlier, it allows for plug-and-play styled frontends.
- Better integration with guice. Guice is now used for injecting the dependencies into the controller. Controlling the dependencies has become easier
- The implementation is not truly frontend and backend independent. We only need to ship the controller with the Kernel.
#### Feedback

Please feel free to provide feedback, and ask questions. Also keep in mind that changes are experimental (hence the new branch). I plan to get the required approval from Dwight before putting in more effort. However, around 80% of the refactoring has been completed. Feel free to email me, Manasvi at (msaxena2@illinois.edu).
Thanks!


msaxena2:
Dwight, if you could please look at this request and provide feedback, it'd great.
Thanks

kernel/src/main/java/org/kframework/debugger/backend/Debugger.java
```diff
@@ -0,0 +1,129 @@
...
+     *
+     * @param s The string to append to stdin.
+     */
+    public abstract void readFromStdin(String s);
+
+    /**
+     * Start debugging from a particular term
+     */
+    public abstract void start(Term initialConfiguration) throws KRunExecutionException;
+}
```
dwightguth, line 129: add newline here

kernel/src/main/java/org/kframework/debugger/controller/DebuggerController.java
```diff
@@ -0,0 +1,226 @@
...
+            this.searchPrinter = searchPrinter;
+            this.graphPrinter = graphPrinter;
+            this.transitionPrinter = transitionPrinter;
+        }
+
+        @Override
+        public Void run(Void v, Attributes a) {
+            //entry point into the entire debugger architecture
+            a.add(Context.class, context);
+            try{
```
dwightguth, line 86: you should change your coding style. Braces should be like so:

```
try {
  // body
} catch (Exception e) {
  // body
}
```

kernel/src/main/java/org/kframework/debugger/controller/DebuggerController.java
```diff
@@ -0,0 +1,226 @@
...
+                backend.start(initialConfiguration);
+            }
+            catch(UnsupportedBackendOptionException e){
+                kem.registerCriticalError("Backend \""
```
dwightguth, line 90: looks like this still hasn't kept up to date with the latest version of the framework. This should be:

```
throw KExceptionManager.criticalError(
```

now.

kernel/src/main/java/org/kframework/debugger/controller/DebuggerController.java
```diff
@@ -0,0 +1,226 @@
...
+                throw new AssertionError("unreachable");
+            }
+            catch(KRunExecutionException e){
+                kem.registerCriticalError(e.getMessage(), e);
+            }
+
+            //initialize the frontend
+            FrontendStatus status = frontend.initFrontend();
+            if(status == FrontendStatus.FAILURE){
+                //throw customized exception
```
dwightguth, line 102: you need to throw that friendly error here. Returning null will likely cause a stack trace.

kernel/src/main/java/org/kframework/debugger/controller/DebuggerController.java
```diff
@@ -0,0 +1,226 @@
...
+                                        null,
+                                        statePrinter,
+                                        null,
+                                        ControllerResult.Status.VALID
+                                );
+                                frontend.publishResult(result);
+                                break;
+                            }
+                            else{
+                               break;
```
dwightguth, line 133: this isn't right. step(0) can still technically change the resulting term.  This is because in the maude backend there is a difference between rules and equations. similarly, in the java backend a step(0) would likely evaluate functions.

kernel/src/main/java/org/kframework/debugger/controller/DebuggerController.java
```diff
@@ -0,0 +1,226 @@
...
+            try {
+                backend.start(initialConfiguration);
+                ControllerResult result;
+                KRunState state;
+                //wait for the command
+                boolean exitCondition = false;
+                do {
+                    DebuggerCommand command = frontend.getCommand();
+                    KRunState currentState = backend.getState(backend.getCurrentState());
+                    switch(command.getGenericCommand()){
```
yilongli, line 115: please follow the K coding style and add proper whitespaces

dwightguth:
If you intend to replace the old Debugger class can you please delete it? I'd like to get a better idea how much this diff adds versus how much it removes. It feels a little bit like we're adding complexity for a gain that I can't see. As we said via IM, a graphical debugger would likely not use the Controller class you created regardless, because it needs to be asynchronous.

dwightguth:
Closing as outdated.

PR #1167, opened 2014-11-07
merge regression tests into single definition for faster execution

dwightguth:
@yilongli please review


yilongli:
Looks great! Thanks, Dwight.

PR #1169, opened 2014-11-10
Adding location information back to maude generated files.

radumereuta:
@traiansf or @dwightguth please review


traiansf:
why is this needed, and if it is, why was it removed in the first place?

radumereuta:
@dwightguth removed the location information from the attributes, and created special fields on ASTNode. Now I have to do a special case to print them back. I'm still using the generated maude files to track down problems when I make modifications to the front end, using diff. That's why I need to add it back.

traiansf:
Well,  since it's only affecting the Maude backend, feel free to merge if all tests pass.

PR #1172, opened 2014-11-12
fixed incorrect K Abstract Machine instructions ...

yilongli:
... caused by cells injected under KItems in rules.

For example:`rule <k> loadObj(<threads> G:Bag </threads>) => . ...</k>`

@dwightguth please review


dwightguth:
Merge if tests pass

PR #1173, opened 2014-11-12
more minor C semantics fixes

dwightguth:
**do not review**

### DIFF
--- a/kernel/src/main/java/org/kframework/parser/DefinitionLoader.java
+++ b/kernel/src/main/java/org/kframework/parser/DefinitionLoader.java
@@ -194,6 +194,7 @@ public class DefinitionLoader {
         sw.printIntermediate("File Gen Def");
 
         File cache = files.resolveKompiled("defx-cache.bin");
+        Thread t2 = null;
         if (!oldSdf.equals(newSdf) || !files.resolveKompiled("Rule.tbl").exists()
                 || !files.resolveKompiled("Ground.tbl").exists()) {
             try {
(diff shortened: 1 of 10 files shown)


dwightguth:
@radumereuta @yilongli please review your portions

yilongli:
lgtm

kernel/src/main/java/org/kframework/parser/DefinitionLoader.java
```diff
@@ -254,6 +253,7 @@ public Definition parseDefinition(File mainFile, String mainModule, Context cont
             loader.saveOrDie(cache, clf.getKept());
         }
 
+
```
radumereuta, line 256: why the extra space?

kernel/src/main/java/org/kframework/parser/DefinitionLoader.java
```diff
@@ -262,6 +262,17 @@ public Definition parseDefinition(File mainFile, String mainModule, Context cont
...
+            if (t2 != null) {
+                t2.join();
+                files.copyTempFileToKompiledFile("ground/Concrete.tbl", "Ground.tbl");
+            }
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw KExceptionManager.criticalError(
+                    "Thread was interrupted trying to run SDF2Table");
+        }
+
```
radumereuta, line 275: Can you explain what was the issue? I don't understand why you had to move the second thread.

dwightguth, line 275: All I did was move the join on the thread to after the parsing of configs and rules takes place. This increases the parallelism of kompile by allowing both parsing of rules and the generation of the ground table to take place simultaneously.

dwightguth:
I'm gonna close this. I keep accumulating minor changes so I'd rather hold on to them for the time being and break them up into individual diffs once I get the C semantics closer to working.

kernel/src/main/java/org/kframework/parser/DefinitionLoader.java
radumereuta, line 275: Makes sense, but from my experience, the Ground parser finishes faster than the rule parser, because the grammar is slightly smaller.

dwightguth:
Reopening in order to test, but do not review yet.

dwightguth:
Jenkins: test this please

PR #1193, opened 2014-11-19
add parallelism to kompile with Ground.tbl

dwightguth:
@radumereuta please review

This is the change I showed you before, just isolated and with your comment addressed.


radumereuta:
lgtm

PR #1195, opened 2014-11-19
fix two unsoundness bugs in prover

dwightguth:

One of these, the code was simply incorrect. The other (the one in Equality.isFalse) is not guaranteed to be correct during partial evaluation.


java-backend/src/main/java/org/kframework/backend/java/symbolic/Equality.java
```diff
@@ -254,14 +254,16 @@ public boolean isFalse(Equality equality) {
                         rightHandSide.sort());
             } else {
                 boolean unifiable = false;
-                if (leftHandSide instanceof Variable && rightHandSide instanceof KItem) {
+                if (leftHandSide instanceof Variable && rightHandSide instanceof KItem
```
yilongli, line 263: Please add some comments here with the rule that makes it incorrect during partial compilation. Most of the functionality in this class will be moved to the `SymbolicUnifier` later.

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicUnifier.java
```diff
@@ -305,14 +305,12 @@ private void unifySet(BuiltinSet set, BuiltinSet otherSet) {
         Multiset<Variable> remainingVariables = Multisets.difference(variables, commonVariables);
         Multiset<Variable> otherRemainingVariables = Multisets.difference(otherVariables, commonVariables);
 
-        if (remainingElements.isEmpty()
-                && remainingPatterns.isEmpty()
```
andreistefanescu, line 309: Why is this incorrect?

dwightguth:
@andreistefanescu @yilongli  please re-review

java-backend/src/main/java/org/kframework/backend/java/symbolic/IncrementalCollector.java
```diff
@@ -0,0 +1,69 @@
...
+                set.addAll(termSet);
+            } else {
+                setStack.push(set);
+                set = new HashSet<>();
+                setValue.accept(set, term);
+            }
+        }
+    }
+
+    private class PostVariableVisitor extends LocalVisitor {
```
andreistefanescu: Why is this class named PostVariableVisitor?

andreistefanescu:
lgtm

java-backend/src/main/java/org/kframework/backend/java/symbolic/IncrementalCollector.java
dwightguth: it was left over from before the rename of the outer class. Good catch.

yilongli:
lgtm

PR #1198, opened 2014-11-20
Adding option for parseInModule since it takes so long for some definitions.

radumereuta:
@dwightguth please review.
I also found the 'loud' option which is not being used anymore, so I removed it.

### DIFF
--- a/kernel/src/main/java/org/kframework/kompile/KompileOptions.java
+++ b/kernel/src/main/java/org/kframework/kompile/KompileOptions.java
@@ -130,7 +130,7 @@ public final class KompileOptions implements Serializable {
         public boolean noPrelude = false;
 
         // TODO: (RaduM) this might be temporary, since I want to generate the parsers by need if the definition is loaded
-        @Parameter(names="--parseInModule", description="Generates all program parsers for each module.")
+        @Parameter(names="--java-parser", description="Generate tables for the new java parser.")
         public boolean parseInModule = false;
 
         @Parameter(names="--documentation", listConverter=StringListConverter.class, description="<string> is a comma-separated list of tags designating rules to be included in the file generated with --backend=doc")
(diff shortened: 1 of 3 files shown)


kernel/src/main/java/org/kframework/kompile/KompileOptions.java
```diff
@@ -129,8 +129,9 @@ public String syntaxModule() {
         @Parameter(names="--no-prelude", description="Do not include anything automatically.")
         public boolean noPrelude = false;
 
-        @Parameter(names="--loud", description="Prints 'Done' at the end if all is ok.")
-        public boolean loud = false;
+        // TODO: (RaduM) this might be temporary, since I want to generate the parsers by need if the definition is loaded
+        @Parameter(names="--parseInModule", description="Generates all program parsers for each module.")
```
dwightguth: call this --java-parser and give it a description accordingly.

radumereuta:
Done.

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
```diff
@@ -129,8 +129,9 @@ public String syntaxModule() {
         @Parameter(names="--no-prelude", description="Do not include anything automatically.")
         public boolean noPrelude = false;
 
-        @Parameter(names="--loud", description="Prints 'Done' at the end if all is ok.")
-        public boolean loud = false;
+        // TODO: (RaduM) this might be temporary, since I want to generate the parsers by need if the definition is loaded
+        @Parameter(names="--java-parser", description="Generate tables for the new java parser.")
+        public boolean parseInModule = false;
```
dwightguth: rename the field too?

dwightguth:
I believe you will need to modify the config for some of the tests in the test suite.

radumereuta:
Somehow I forgot to rename the field. Should be fixed now.
I have no idea why the test failed the first time. Could you have a look?

radumereuta:
Everything passed.

PR #1200, opened 2014-11-21
close some memory leaks involving nailgun

dwightguth:
@yilongli please review

This pull request removes some major dominators that were keeping memory accessible long after the application had quit

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/kil/Definition.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/kil/Definition.java
@@ -10,11 +10,14 @@ import org.kframework.kil.ASTNode;
 import org.kframework.kil.Attribute;
 import org.kframework.kil.loader.Context;
 import org.kframework.utils.errorsystem.KExceptionManager;
+
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import com.google.common.collect.ArrayListMultimap;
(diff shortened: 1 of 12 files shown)


java-backend/src/main/java/org/kframework/backend/java/kil/Definition.java
```diff
@@ -58,6 +61,8 @@
     private RuleIndex index;
     public final IndexingTable.Data indexingData;
 
+    final Map<KItem.CacheTableColKey, KItem.CacheTableValue> SORT_CACHE_TABLE = new HashMap<>();
```
yilongli: since it's no longer static, we should probably change its name to `sortCacheTable`

java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
```diff
@@ -111,7 +107,7 @@ private KItem(Term kLabel, Term kList, TermContext termContext, Tool tool) {
                     && definition.sortPredicateRulesOn(kLabelConstant).isEmpty();
             if (enableCache) {
                 cacheTabColKey = new CacheTableColKey(kLabelConstant, (KList) kList);
-                cacheTabVal = SORT_CACHE_TABLE.get(definition, cacheTabColKey);
+                cacheTabVal = definition.SORT_CACHE_TABLE.get(cacheTabColKey);
```
yilongli: I would prefer to not expose this map directly to KItem

java-backend/src/main/java/org/kframework/backend/java/kil/Definition.java
yilongli: make it private?

dwightguth:
@yilongli please re-review

yilongli:
lgtm

PR #1204, opened 2014-11-22
remove try-catch

cos:


kernel/src/main/java/org/kframework/parser/DefinitionLoader.java
```diff
@@ -247,17 +247,12 @@ public Definition parseDefinition(File mainFile, String mainModule, Context cont
...
-        } finally {
-            // save definition
-            loader.saveOrDie(cache, clf.getKept());
-        }
+        def = (Definition) clf.visitNode(def);
+        cachedSentences = clf.getKept().size();
+        prf = new ParseRulesFilter(context, clf.getKept());
+        def = (Definition) prf.visitNode(def);
+        // save definition
+        loader.saveOrDie(cache, clf.getKept());
```
radumereuta, line 255: This last line has to be in the finally branch, otherwise you won't get the incremental rule parsing.
Also, you didn't remove the try-catch from ParseConfigFilter.

dwightguth:
Closing. This was merged as part of #1211

PR #1205, opened 2014-11-24
simplification in substitution and function evaluation

andreistefanescu:


andreistefanescu:
@yilongli what do you think?

java-backend/src/main/java/org/kframework/backend/java/kil/Term.java
```diff
@@ -131,17 +131,11 @@ public Term substituteWithBinders(Map<Variable, ? extends Term> substitution, Te
...
      */
     public Term substituteAndEvaluate(Map<Variable, ? extends Term> substitution, TermContext context) {
-        // TODO(AndreiS): disable the check below when proving things until this is properly fixed by Cosmin
-        if (context.definition().context().krunOptions == null
-                || context.definition().context().krunOptions.experimental.prove == null) {
-            // TODO(AndreiS): assert that there are not any unevaluated functions in this term
-            if (substitution.isEmpty() || isGround()) {
-                return this;
-            }
+        if ((!substitution.isEmpty() && !isGround()) || !isNormal()) {
```
yilongli: This condition is little hard to read. I would prefer `return isNormal && (substitution.isEmpty() || isGround()) ? this : (Term) this.accept(new SubstituteAndEvaluateTransformer(substitution, context));`

java-backend/src/main/java/org/kframework/backend/java/symbolic/SubstituteAndEvaluateTransformer.java
```diff
@@ -41,19 +41,7 @@ public SubstituteAndEvaluateTransformer(Map<Variable, ? extends Term> substituti
...
-            set1 = set2;
-            set2 = tmp;
-        }
-        for (Variable variable : set1) {
-            if (set2.contains(variable)) {
-                return true;
-            }
-        }
-        return false;
+        return ((!substitution.isEmpty() && !object.isGround()) || !object.isNormal());
```
yilongli: The original version is more precise?

andreistefanescu: Yes it is, but I think it's too expensive.

yilongli: But it also avoids unnecessary recursion. So does the new version show speedup on any example?

andreistefanescu: Well, for example the avl delete goes from 18s to 17.5. I know it's not much, but it helps a little.

yilongli: ok, then I am fine with it. BTW, can you remove the redundant parentheses around the expressions?

dwightguth:
@andreistefanescu you didn't actually address Yilong's comments before you merged. Please perform a more careful reading of this pull request and then recreate the diff.

dwightguth:
I'm sorry, I misread this diff. Never mind. I'll re-merge the change.

PR #1207, opened 2014-11-25
Fixed typos in the documentation

omarzd:
Even though this file will still undergo many edits, and even though fixing typos is usually done in the final revision, I just felt like doing it now, if you don't mind.


dwightguth:
Jenkins: ok to test

dwightguth:
Not sure why this test failed, but it's certainly not because of this diff. I am going to merge this.

PR #1210, opened 2014-11-25
Module context

radumereuta:
Do not review yet, I'm expecting failures.


dwightguth:
I'm not going to go into detail yet because as you say, you're not done yet, but a preliminary look tells me there seems to be at least one significant issue you need to address, namely, that you need to remove the information from Context which you are moving into ModuleContext.

radumereuta:
You touched a good point there.
I think this is too involving for now. I start to understand why @osa1 failed on this refactoring. There are just too many things that depend to the wrong implementation from the global Context.
If @cos plans to reimplement some of the compiler steps, then it will be a good opportunity to fix the middle end. But for now I will limit myself to fixing the front end.

dwightguth:
The fact that significant work has to be done to get somewhere does not
preclude that work being done. It is simply not safe to have two competing
implementations of the exact same set of data in the codebase at the same
time. It's a disaster waiting to happen. If you are concerned about scope,
start by making a ModuleContext that forwards all its fields to the global
Context, and then transfer the implementation of those fields one field at
a time.

I'm simply not willing to compromise about this, not when it's such a clear and present danger to the codebase to do what you describe. I'm sorry.

radumereuta:
Jenkins: test this please

osa1:
Is this complete or are you going to try removing those state from `Context`?

radumereuta:
This is work in progress. Right now I only managed to correctly implement the declared sorts feature.
I will try to gradually implement more and more things that are needed by the front-end or the compiler, but if there are steps which are too dependent to the old implementation of the Context, I will not remove them now. It's going to happen when we switch to the new kore compiler.

kernel/src/main/java/org/kframework/kil/loader/FillInModuleContext.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Queue;
+import java.util.Set;
+
+public class FillInModuleContext extends BasicVisitor {
+
+    private boolean autoinclude;
```
dwightguth: final?

kernel/src/main/java/org/kframework/kil/loader/FillInModuleContext.java
```diff
@@ -0,0 +1,122 @@
...
+    private Poset<String> moduleInclusion = new Poset<>();
```
dwightguth: I think maybe this field should be in the data structures somewhere, rather than making this visitor stateful.

kernel/src/main/java/org/kframework/kil/loader/FillInModuleContext.java
```diff
@@ -0,0 +1,122 @@
...
+        super(FillInModuleContext.class.toString(), null);
+        this.autoinclude = autoinclude;
+    }
+
+    @Override
+    public Void visit(Definition def, Void _)  {
+        super.visit(def, _);
+        if (def.getMainSyntaxModule().equals(def.getMainModule())) {
+            // if the syntax module is the same with the main module, add everything
+            def.getModulesMap().get(def.getMainModule()).getModuleContext().addImportedModule(this.getCurrentDefinition().getModulesMap().get(Constants.AUTO_INCLUDED_MODULE));
```
dwightguth: this looks wrong to me. We already have visitors elsewhere which check the correct information about the autoincludes and add an "imports" statement.

kernel/src/main/java/org/kframework/kil/loader/FillInModuleContext.java
```diff
@@ -0,0 +1,122 @@
...
+                // add everything from the included module to the context of the current one
+                start.getModuleContext().add(next.getModuleContext());
+            }
+        }
+    }
+
+    @Override
+    public Void visit(Import node, Void _)  {
+        Module module;
+        if (!node.getName().startsWith("#")) { // maude legacy: some modules specified with # are builtin
```
dwightguth, line 61: can you add a TODO here with my name on it? I want to try to delete Maude starting Friday, and I need to be able to keep track of this to know to delete it.

kernel/src/main/java/org/kframework/kil/loader/FillInModuleContext.java
```diff
@@ -0,0 +1,122 @@
...
+        moduleInclusion.addRelation(this.getCurrentModule().getName(), node.getName());
+        return null;
+    }
+
+    @Override
+    public Void visit(Module node, Void _) {
+        if (autoinclude) {
+            if (this.getCurrentDefinition().getMainSyntaxModule().equals(node.getName())) {
+                // the syntax module automatically includes Constants.AUTO_INCLUDED_SYNTAX_MODULE
+                node.getModuleContext().addImportedModule(this.getCurrentDefinition().getModulesMap().get(Constants.AUTO_INCLUDED_SYNTAX_MODULE));
```
dwightguth: see above comment on autoincludes

kernel/src/main/java/org/kframework/kil/loader/FillInModuleContext.java
```diff
@@ -0,0 +1,110 @@
...
+    @Override
+    public Void visit(Module node, Void _) {
+        if (autoinclude) {
+            if (this.getCurrentDefinition().getMainSyntaxModule().equals(node.getName())) {
+                // the syntax module automatically includes Constants.AUTO_INCLUDED_SYNTAX_MODULE
+                node.getModuleContext().addImportedModule(this.getCurrentDefinition().getDefinitionContext().getModuleByName(Constants.AUTO_INCLUDED_SYNTAX_MODULE));
+                // TODO: (RaduM) try to figure out how and when to import the autoincluded modules.
+            } else if (!node.isPredefined()) {
+                // every user defined module automatically includes Constants.AUTO_INCLUDED_MODULE
+                node.getModuleContext().addImportedModule(this.getCurrentDefinition().getDefinitionContext().getModuleByName(Constants.AUTO_INCLUDED_MODULE));
```
radumereuta: I have to add something here.
Right now we kind of have two different levels of module auto inclusion.
Because I have to generate clean parsers for modules, I include the `AUTO-INCLUDED-MODULE-SYNTAX` in the main syntax module, and generate a parser for programs. Then I move on to the rule parser, which needs more syntax. Then I include the next module: `AUTO-INCLUDED-MODULE`. The step you described in the previous comment, handles only the second stage. What I'm doing here, is the first stage.

I know this is a really ugly thing. I will try to think of something cleaner for the new parser prototype. Probably I won't include anything in the beginning.

radumereuta:
@dwightguth I made a few modifications. Please re-review.

kernel/src/main/java/org/kframework/kil/loader/FillInModuleContext.java
dwightguth: this still isn't the right way to do this. As I said, if you want the autoincluded syntax module to be included in the main syntax module, you should automatically add that imports statement. It doesn't need to be handled specially like this, it should flow naturally and automatically from the principles of module inclusion.

kernel/src/main/java/org/kframework/kil/Definition.java
```diff
@@ -59,24 +60,9 @@ public String toString() {
         return "DEF: " + mainFile + " -> " + mainModule + "\n" + content;
     }
 
-    public void addModuleImport(String mainModule, String importedModule) {
```
dwightguth, line 62: you need to leave these methods here and forward them to the new implementation in order to prevent yourself from giving Cosmin extra work fixing his converters.

kernel/src/main/java/org/kframework/kil/Definition.java
```diff
@@ -134,14 +121,6 @@ public void preprocess(org.kframework.kil.loader.Context context) {
         context.makeFreshFunctionNamesMap(this.getSyntaxByTag(Attribute.FRESH_GENERATOR, context));
     }
 
-    public Map<String, Module> getModulesMap() {
```
dwightguth, line 137: same here

kernel/src/main/java/org/kframework/kil/loader/FillInModuleContext.java
```diff
@@ -0,0 +1,87 @@
...
+                // add everything from the included module to the context of the current one
+                start.getModuleContext().add(next.getModuleContext());
+            }
+        }
+    }
+
+    @Override
+    public Void visit(Import node, Void _)  {
+        Module module;
+        // TODO (dwight) remove only the condition the same time with maude backend
```
dwightguth: TODO(dwightguth)

exactly like that. I need to be able to grep for this string.

kernel/src/main/java/org/kframework/parser/generator/CollectIncludedModulesVisitor.java
```diff
@@ -38,13 +38,13 @@ public Void visit(Definition def, Void _) {
...
-                Module m = def.getModulesMap().get(mname);
+                Module m = def.getDefinitionContext().getModuleByName(mname);
                 for (ModuleItem s : m.getItems()) {
                     if (s instanceof Import) {
                         Import imp = ((Import) s);
                         String mname2 = imp.getName();
-                        Module mm = def.getModulesMap().get(mname2);
-                        // if the module starts with # it means it is predefined in maude
+                        Module mm = def.getDefinitionContext().getModuleByName(mname2);
+                        // TODO (Dwight) if the module starts with # it means it is predefined in maude
```
dwightguth: see above

dwightguth:
I also suggest you revert all the changes corresponding to the methods you forwarded to DefinitionContext. That will help keep this diff smaller and more readable.

dwightguth:
Jenkins: test this please

PR #1211, opened 2014-11-25
KORE milestone 1

cos:
**please review**
- unit-tested convertors from KORE to KIL and back
- significant changes and simplifications to KORE data structures. They are smaller and more powerful now


dwightguth:
I haven't looked this over in detail, but we agreed that the scala portion of the codebase would be restricted to the data structures, and that's not what I see here. I can't merge this with the rewriter code and so forth written in scala.

cos:
It depends on what you understand by data structures. In any case, I have better things to do then argue with you. There is no pattern matcher, rewriter, or meta-level. Solved.

(if anybody is interested in them, you can find them on my fork: https://github.com/cos/k)

cos:
the tests are failing because the new tests are either not run as integration tests (despite having the the "IT" postfix) or the jenkins paths are somehow special

is this the correct path for the generated sdf2table: 
/var/lib/jenkins/jobs/k-framework-pull-requests/workspace/k-distribution/../k-distribution/target/release/k/lib/native/UNIX/sdf2table
?

dwightguth:
No, it's "linux", not "unix"

cos:
that is generated automatically by `OS.current()`

dwightguth:
Unfortunately there's no guarantee even if we made a 1-1 mapping between os
names in maven and the OS enum in Java, that a particular binary might be
found in the lib/native/<OS>/ directory. If you want to hack it to work
specifically for sdf2table, based on such a 1-1 mapping, though, you can.
It just requires that you rename the values of the enum "WINDOWS", "OSX",
and "LINUX", and then call .name().toLower()

cos:
why should there be a difference between "OS enum in Java" and maven?

cos:
I've started writing a wiki page describing the KORE data structures. Anything not on the wiki is not considered part of the API.

https://github.com/kframework/k/wiki/KORE-data-structures-guide

grosu:
WONDERFUL!  I like particularly the "anything not on the wiki is not part of the API", which I fully concur with.

Grigore

cos:
There is a bit work on the KIL-to-KORE convertor, but we should start agreeing upon the rest. 

Please have a look at the data structures (in kast.scala and outer/outer.scala). Their interfaces remain mostly the same. Most of the changes are in the underlying implementation. The most important part is agreeing upon the API presented here:
https://github.com/kframework/k/wiki/KORE-data-structures-guide
Use it as a guide when reading the code.

cos:
One significant change is that I've made `KList` a `K`. I realized this is needed while working on the pattern matcher. It doesn't make sense for one `K` term to be a `KList` but for a `KList` to not be a `K`. Futhermore, as `KList` is associative, a `KList` can appear within another `KList` (for example, within a pattern). Without making `KList` a `K`, this would be disallowed.

Radu, this doesn't affect you in any way. The `KList` can remain not-a-`K` from a parsing perspective. The change is only in the data structures.

cos:
The tests should pass once we solve the `OS.current()` issue. @dwightguth, please let me know the best way to solve this. I don't understand why we have multiple versions of paths, and I don't want to hack it in to work.

.gitignore
```diff
@@ -18,3 +18,4 @@ k-distribution/.project
 *~
 junit-reports
 target/
+kore/doc
```
radumereuta: New line at the end?

k-distribution/src/test/java/org/kframework/kore/convertors/BaseTest.java
```diff
@@ -0,0 +1,81 @@
...
+        def.setItems(Outer.parse(Sources.generatedBy(TestKILtoKOREIT.class), definitionText, null));
+        return def;
+    }
+
+    private String clean(String definitionText) {
+        return definitionText.replace("// Copyright (c) 2014 K Team. All Rights Reserved.", "")
+                .trim();
+    }
+
+}
```
radumereuta: New line at the end...

k-distribution/src/test/resources/convertor-tests/context-expected.k
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
+
+Require(include.k)
+
+module INCLUDE
+  syntax String ::= #String
+  syntax Map ::= ".Map" [function() hook(Map:.Map)]
+  syntax #String ::=  [regex([\"](([^\"\n\r\\])|([\\][nrtf\"\\])|([\\][x][0-9a-fA-F]{2})|([\\][u][0-9a-fA-F]{4})|([\\][U][0-9a-fA-F]{8}))*[\"])]
```
radumereuta: How is this even possible? Where is the production?

k-distribution/src/test/resources/convertor-tests/ruleWithRequiresEnsures-expected.k
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
+
+module TEST
+  syntax Exp
+  rule A [sort(Exp)]=>A [sort(Exp)] requires true ensures A [sort(Exp)] []
+  rule A [sort(Exp)]=>A [sort(Exp)] requires A [sort(Exp)] ensures true []
+  rule A [sort(Exp)]=>A [sort(Exp)] requires A [sort(Exp)] ensures A [sort(Exp)] []
+endmodule
```
radumereuta: NewLine

kernel/src/main/java/org/kframework/compile/transformers/AddKCell.java
```diff
@@ -41,7 +41,7 @@ public ASTNode visit(Module module, Void _)  {
             returnModule = newModule;
         } else {
             returnModule = newModule.shallowCopy();
-            returnModule = returnModule.addModuleItems(newRules);
+            returnModule = returnModule.setModuleItems(newRules);
```
radumereuta: Explain.

kernel/src/main/java/org/kframework/kil/Module.java
```diff
@@ -111,7 +111,7 @@ public String toString() {
         return visitor.complete(this, visitor.visit(this, p));
     }
 
-    public Module addModuleItems(List<ModuleItem> rules) {
+    public Module setModuleItems(List<ModuleItem> rules) {
```
radumereuta: But this is not set, it's add. I think it was correct before.

kernel/src/main/java/org/kframework/parser/DefinitionLoader.java
```diff
@@ -247,17 +247,12 @@ public Definition parseDefinition(File mainFile, String mainModule, Context cont
...
-        } finally {
-            // save definition
-            loader.saveOrDie(cache, clf.getKept());
-        }
+        def = (Definition) clf.visitNode(def);
+        cachedSentences = clf.getKept().size();
+        prf = new ParseRulesFilter(context, clf.getKept());
+        def = (Definition) prf.visitNode(def);
+        // save definition
+        loader.saveOrDie(cache, clf.getKept());
```
radumereuta: I've commented before on this. If you don't do this in the finally block, you won't save the cached rules.
However, it wasn't correct before either. You will have to catch the error and rethrow it, instead of printing the stack.

kernel/src/main/java/org/kframework/parser/utils/Sdf2Table.java
```diff
@@ -8,40 +8,49 @@
...
 import org.kframework.utils.ThreadedStreamCapturer;
 
+import com.google.common.collect.Lists;
 import com.google.inject.Inject;
 import com.google.inject.Provider;
 
 public class Sdf2Table {
 
-    private final Provider<ProcessBuilder> pb;
+    private final ProcessBuilder pb;
```
radumereuta: @dwightguth is this correct?

dwightguth: yeah, this is fine.

kernel/src/main/java/org/kframework/parser/DefinitionLoader.java
dwightguth: well no, you just need to not have the catch block. A try-finally doesn't need to have a catch block.

dwightguth:
@cos the issue with OS.current is that some binaries have separate versions for 32 and 64 bit, and some binaries are only dependent on the OS. For example, sdf2table has 3 different binaries, one for each OS, but z3 has 5.

I honestly don't know of a good solution for you if your goal is to have a completely in-java mechanism for choosing binaries to run. You could maybe do it if you recompiled sdf2table and nailgun so that they have separate 32 and 64 bit binaries, and then set the path to always be one of the 5 directories of "linux64, linux32, osx, windows32, windows64".

I still think your best bet for a simple solution that solves your immediate problem is to do what I described in the earlier comment and change the names of the enum values to match the maven OS constants.

k-distribution/src/test/resources/convertor-tests/context-expected.k
cos: fixed

kernel/src/main/java/org/kframework/compile/transformers/AddKCell.java
cos: Omer piggybacked this refactoring. He is right, that method's implementation _sets_ the module items, it doesn't _add_ them to existing ones.

kernel/src/main/java/org/kframework/kil/Module.java
cos: It is a _set_. It adds to a new empty list of items. It was wrong before. The refactoring is correct. If _add_ behavior is assumed anywhere, it needs to be fixed.

kernel/src/main/java/org/kframework/parser/DefinitionLoader.java
cos: fixed

kernel/src/main/java/org/kframework/kil/Module.java
```diff
@@ -111,7 +111,7 @@ public String toString() {
...
         Module result = new Module(this);
         List<ModuleItem> items = new ArrayList<>(this.items);
```
radumereuta, line 126: Look at this line. It's an add. New ArrayList takes as argument `this.items`.

kernel/src/main/java/org/kframework/kil/Module.java
dwightguth: I think we're arguing about semantics here. What this method does is return a new module with both the module items of "this", and the module items in "rules". Arguably you could call it either add or set, depending on precisely how you want to name a function which represents an immutable setter.

cos: Right, it's an add. Darn Java syntax.

cos:
Jenkins, test this please

cos:
let's get this started... please review.

k-distribution/src/test/java/org/kframework/kore/convertors/TstKILtoKOREIT.java
```diff
@@ -0,0 +1,89 @@
...
+import org.junit.FixMethodOrder;
+import org.junit.Test;
+import org.junit.runners.MethodSorters;
+
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
+public class TstKILtoKOREIT extends BaseTest {
+
+    @Test
+    public void emptyModule() throws IOException {
+        outerKILtoKORETest();
```
dwightguth, line 16: this feels like a potential source of confusion here. I understand what is going on, but I think it would be better not to have magic method names.

kernel/src/main/java/org/kframework/kore/convertors/KILtoInnerKORE.java
```diff
@@ -0,0 +1,141 @@
...
+        return KBag(KList(contents));
+    }
+
+    KApply cellMarker = org.kframework.kore.outer.Configuration.cellMarker();
+
+    @SuppressWarnings("unchecked")
+    public KApply apply(Cell body) {
+        K x = apply(body.getContents());
+        if (x instanceof KBag && !((KBag) x).isEmpty()) {
+            return KApply(KLabel(body.getLabel()), KList(((KBag) x)), Attributes(cellMarker));
```
dwightguth, line 74: @grosu are you happy with this as a default klabel strategy for cells? I would have expected "'<cellName>_</cellName>", not "cellName", but I admit I have no real strong preference here.

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
```diff
@@ -0,0 +1,261 @@
...
+import com.google.common.collect.Sets;
+
+import static org.kframework.kore.outer.Constructors.*;
+import static org.kframework.kore.Constructors.*;
+
+public class KILtoKORE extends KILTransformation<Object> {
+
+    // we mark productions in same priority blocks with ids unique to that
+    // priority block
+    public int priorityBlockId = 0;
```
dwightguth: why are these public?

kernel/src/main/java/org/kframework/kore/convertors/KILtoInnerKORE.java
```diff
@@ -0,0 +1,141 @@
...
+public class KILtoInnerKORE extends KILTransformation<K> {
+
+    public static final String PRODUCTION_ID = "productionID";
+
+    public K apply(Bag body) {
+        List<K> contents = body.getContents().stream().map(this).collect(Collectors.toList());
+        return KBag(KList(contents));
+    }
+
+    KApply cellMarker = org.kframework.kore.outer.Configuration.cellMarker();
```
dwightguth: private static final?

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    public ModuleComment apply(LiterateModuleComment m) {
+        return new org.kframework.kore.outer.ModuleComment(m.getValue(), inner.apply(m
+                .getAttributes()));
+    }
+
+    public org.kframework.kore.outer.Configuration apply(Configuration kilConfiguration) {
+        Cell body = (Cell) kilConfiguration.getBody();
+        return Configuration(inner.apply(body), inner.applyOrTrue(kilConfiguration.getEnsures()),
```
dwightguth, line 98: don't you need to generate productions for the cells that are declared here?

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
```diff
@@ -0,0 +1,261 @@
...
+        // "left", "right", "non-assoc"
+        switch (assocOrig) {
+        case "left":
+            return Associativity.Left();
+        case "right":
+            return Associativity.Right();
+        case "non-assoc":
+            return Associativity.NonAssoc();
+        default:
+            throw new RuntimeException("Incorrect assoc string: " + assocOrig);
```
dwightguth: I believe the convention we have is to use AssertionError for things like this.

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
```diff
@@ -0,0 +1,261 @@
...
+                if (p.getItems().size() == 1 && p.getItems().get(0) instanceof UserList) {
+                    applyUserList(res, sort, p, (UserList) p.getItems().get(0));
+                } else {
+                    List<ProductionItem> items = new ArrayList<>();
+                    // TODO: when to use RegexTerminal?
+                    for (org.kframework.kil.ProductionItem it : p.getItems()) {
+                        if (it instanceof NonTerminal) {
+                            items.add(NonTerminal(apply(((NonTerminal) it).getSort())));
+                        } else if (it instanceof UserList) {
+                            throw new RuntimeException("Lists should have applyed before.");
```
dwightguth: and here

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
```diff
@@ -0,0 +1,261 @@
...
+                        } else if (it instanceof Lexical) {
+                            items.add(RegexTerminal(((Lexical) it).getLexicalRule()));
+                        } else if (it instanceof Terminal) {
+                            items.add(Terminal(((Terminal) it).getTerminal()));
+                        } else {
+                            throw new RuntimeException("Unhandled case");
```
dwightguth: and here

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
```diff
@@ -0,0 +1,261 @@
...
+                kilProductionId);
+        prod1 = SyntaxProduction(sort,
+                Seq(NonTerminal(sort), Terminal(userList.getSeparator()), NonTerminal(sort)),
+                attrsWithKilProductionId);
+
+        // lst ::= elem
+        prod2 = SyntaxProduction(sort, Seq(NonTerminal(elementSort)), attrsWithKilProductionId);
+
+        // lst ::= .UserList
+        prod3 = SyntaxProduction(sort, Seq(Terminal("." + sort.toString())),
```
dwightguth, line 240: I believe this needs to have a different klabel in order to maintain compatibility.

kernel/src/main/java/org/kframework/kore/convertors/KOREtoKIL.java
```diff
@@ -0,0 +1,528 @@
...
+import static org.kframework.kore.Collections.*;
+
+public class KOREtoKIL {
+
+    private static RuntimeException NOT_IMPLEMENTED() {
+        return NOT_IMPLEMENTED("Not implemented");
+    }
+
+    private static RuntimeException NOT_IMPLEMENTED(String s) {
+        return new RuntimeException(s);
```
dwightguth: and here

kernel/src/main/java/org/kframework/kore/convertors/KOREtoKIL.java
```diff
@@ -0,0 +1,528 @@
...
+    }
+
+    class ListProductionCollector {
```
dwightguth, line 32: javadoc here? it's not clear to me what this is doing exactly

kore/src/main/java/org/kframework/kore/Constructors.java
```diff
@@ -0,0 +1,117 @@
...
+    public static Attributes Attributes(KList klist) {
+        return new Attributes(klist.delegate().toSet());
+    }
+
+    public static KString KString(String s) {
+        return new KString(s);
+    }
+
+    public static KLabel Hole() {
+        return Hole$.MODULE$;
```
dwightguth, line 38: I don't know this syntax, can you explain?

kore/src/main/scala/org/kframework/kore/builtins.scala
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
```
dwightguth, line 1: why in particular these builtins and not the rest?

kore/src/main/scala/org/kframework/kore/kast-to-string.scala
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
```
dwightguth, line 1: you said that this wasn't intended to be used as an unparser, correct?

k-distribution/src/test/java/org/kframework/kore/convertors/TstKILtoKOREIT.java
cos, line 16: otherwise we'll have to be careful to keep method names and test names in sync.

dwightguth:
Can you please add data about the visitors and the other ancillary features of the API which are not on the wiki page yet? I'd like to understand better how they work before I finish reviewing this.

kernel/src/main/java/org/kframework/kore/convertors/KILtoInnerKORE.java
cos, line 74: This is in the spirit of https://github.com/kframework/k/wiki/KAST-and-KORE, where each "cell type" becomes a separate production. Hence, it will have a distinct klabel. 

I have no strong preference either, and we can change representation easily.

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
cos: redundant. removed them.

kernel/src/main/java/org/kframework/kore/convertors/KILtoInnerKORE.java
cos: we're using it from other classes.

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
cos, line 240: can you elaborate?

kore/src/main/java/org/kframework/kore/Constructors.java
cos, line 38: It's bridging Scala singletons. Don't worry about it. From Java, you only see `Hole()`, which gives you the klabel for a `HOLE`.

kore/src/main/scala/org/kframework/kore/builtins.scala
cos, line 1: because these are the ones we needed so far.

kore/src/main/scala/org/kframework/kore/kast-to-string.scala
cos, line 1: correct.

cos:
There isn't much visiting infrastructure because there wasn't much need for it yet. The KORE to KIL convertor required a complete traversal of KORE so visitors wouldn't have helped. In the future the idea is to use pattern matching instead of visitors. 

As for other features, I don't want to document on the wiki any features which should not be considered part of the KORE API. So, if something is not on the wiki, consider it an implementation detail.

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
dwightguth, line 240: user lists with the same separator share a klabel.

kernel/src/main/java/org/kframework/kore/convertors/KILtoInnerKORE.java
grosu, line 74: Just to make it absolutely clear: what KLabel a particular cell production takes should be completely orthogonal to KORE or to any kompilation step.  It should only be a particular decision that the frontend takes in order to desugar the concrete syntax (of the semantics) to KORE.

Since this KIL to KORE translator is only temporary, I would say that it simply does not matter what name you pick for the KLabels corresponding to cells, as far as everything works.

cos, line 74: Yes. The klabel names do not matter. The important part is that cells have the "cell" attribute which will make the compiler treat them specially, i.e., do configuration concretization.

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
cos, line 98: we're not doing that just yet. we want to have, at the very least, the convertors integrated into the pipeline before implementing that.

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
cos, line 240: ideally, it should work like this. we'll fix/adapt when/if it fails.

cos:
let's wrap this stage up. @dwightguth, @radumereuta, @grosu, are you ok with merging this?

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
dwightguth, line 240: I don't know that it actually should work like this. Consider the following case:

rule start => x, 5, 6, 7, .Exps
rule x, I:Ints => end

This should successfully rewrite start to end, but if you put it through the transformer you are describing here, I'm not sure that it will.

kernel/src/main/java/org/kframework/kore/convertors/KILtoInnerKORE.java
dwightguth: are you writing to it? if so, it can still be public static final

kernel/src/main/java/org/kframework/utils/Stopwatch.java
```diff
@@ -19,6 +19,7 @@
     private final GlobalOptions options;
 
     @Inject
+    public
```
radumereuta: Why isn't this on the same line with the constructor?

kernel/src/main/java/org/kframework/utils/file/FileUtil.java
```diff
@@ -162,7 +162,7 @@ private void copyFileToDirectory(File from, File toDir) {
         }
     }
 
-    private void save(File file, String content) {
```
radumereuta, line 165: I know it's not part of this diff, but why isn't this function static?

dwightguth, line 165: it probably can be static now. I think this was done before the code which made KExceptionManager not need to be instantiated to throw an error.

k-distribution/src/test/java/org/kframework/kore/convertors/BaseTest.java
```diff
@@ -0,0 +1,108 @@
...
+            throw new RuntimeException(e);
+        }
+        def.setItems(Outer.parse(Sources.generatedBy(TstKILtoKOREIT.class), definitionText, null));
+        return def;
+    }
+
+    private String clean(String definitionText) {
+        return definitionText.replace("// Copyright (c) 2014 K Team. All Rights Reserved.", "")
+                .replaceAll(" *\n", "\n").trim();
+    }
```
radumereuta, line 106: Why do you need this? Something like this seems odd.

k-distribution/src/test/resources/convertor-tests/simpleSyntax-kilexpected.k
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
+
+DEF: null -> null
+module TEST
+  syntax Foo ::=
+
+
+endmodule
```
radumereuta: New line.

k-distribution/src/test/resources/convertor-tests/syntaxPriorities-kilexpected.k
```diff
@@ -0,0 +1,27 @@
...
+
+  syntax priorities'x
+> 'y 'z
+> 'm 't
+> 'u
+
+  syntax right 'u
+
+
+endmodule
```
radumereuta, line 27: Maybe you could put less new lines after syntax. This looks kind of stretched out.

k-distribution/src/test/resources/convertor-tests/syntaxWithAttributes-kilexpected.k
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
+
+DEF: null -> null
+module TEST
+  syntax Foo ::=
```
radumereuta, line 5: This will give a parse error. Is this what you are trying to do?

k-distribution/src/test/resources/convertor-tests/syntaxWithOR-kilexpected.k
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
+
+DEF: null -> null
+module TEST
+  syntax Foo ::= "x"
+
+  syntax Foo ::= "y"
+
+
+endmodule
```
radumereuta: New Line.

kernel/src/main/java/org/kframework/kil/Definition.java
```diff
@@ -12,15 +12,17 @@
...
 import java.util.List;
 import java.util.Map;
 
-
 /**
- * Represents a language definition.
- * Includes contents from all {@code required}-d files.
+ * Represents a language definition. Includes contents from all {@code required}
+ * -d files.
+ *
```
radumereuta: Please set your editor to not format comments. Then revert this part.

kernel/src/main/java/org/kframework/kil/Definition.java
```diff
@@ -39,7 +41,8 @@ public Definition() {
     }
 
     @Inject
-    public Definition(Void v) {}
+    public Definition(Void v) {
+    }
```
radumereuta: Revert.

kernel/src/main/java/org/kframework/kil/Definition.java
```diff
@@ -126,8 +137,8 @@ public void preprocess(org.kframework.kil.loader.Context context) {
         context.setTokenSorts(TokenSortCollector.collectTokenSorts(this, context));
 
         /* collect the data structure sorts */
-        DataStructureSortCollector dataStructureSortCollector
-                = new DataStructureSortCollector(context);
+        DataStructureSortCollector dataStructureSortCollector = new DataStructureSortCollector(
+                context);
```
radumereuta: Autoformatting again. Please revert. Stuff like this since it will only make it harder to look into the history, and make merges unnecessarily hard.

kernel/src/main/java/org/kframework/kil/Definition.java
```diff
@@ -109,7 +119,8 @@ public String getMainSyntaxModule() {
 
     public void preprocess(org.kframework.kil.loader.Context context) {
         // Collect information
-        // this.accept(new AddSymbolicVariablesDeclaration(context, this.getMainSyntaxModule()));
+        // this.accept(new AddSymbolicVariablesDeclaration(context,
+        // this.getMainSyntaxModule()));
```
radumereuta: Revert autoformat.

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
```diff
@@ -17,7 +17,7 @@
 import com.beust.jcommander.ParametersDelegate;
 import com.google.inject.Inject;
 
-public final class KompileOptions implements Serializable {
+public class KompileOptions implements Serializable {
```
radumereuta, line 20: @dwightguth is this correct?

dwightguth, line 20: Yeah, that's fine. I imagine he wants to mock it and can't because it's
final.

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
cos, line 240: What you're hinging at seems to be related to sorting side conditions. Syntactically, this would rewrite to end. In any case, I want to see it fail first, and will deal with it then.

kernel/src/main/java/org/kframework/kore/convertors/KILtoInnerKORE.java
cos: I don't understand the question. I'm certainly not writing to it as it is final.

radumereuta:
The diff is too big to go into details, but except for the syntactic issues it looks ok.

kernel/src/main/java/org/kframework/kore/convertors/KILtoInnerKORE.java
dwightguth: okay now I'm really confused. You say you're not writing to it beacuse it's final, but I don't see a final marker...

k-distribution/src/test/java/org/kframework/kore/convertors/BaseTest.java
cos, line 106: because checkstyle forces me to add the verbose copyright to every tiny file. I need to strip it before comparing to the actual output.

k-distribution/src/test/resources/convertor-tests/syntaxPriorities-kilexpected.k
cos, line 27: I fully agree. This little beauty is what KIL `toString()` outputs now. Deal with whoever implemented it in the first place.

k-distribution/src/test/resources/convertor-tests/syntaxWithAttributes-kilexpected.k
cos, line 5: Again, this is what KIL's `toString()` outputs.

dwightguth, line 5: Is there a reason you're not calling the unparser? the KIL toString is not
necessarily guranteed to be sound with respect to uniqueness. Two different
terms can have the same toString.

k-distribution/src/test/java/org/kframework/kore/convertors/BaseTest.java
radumereuta, line 106: I'm pretty sure the outer parser has these comments. Why aren't you printing them?

k-distribution/src/test/resources/convertor-tests/syntaxWithAttributes-kilexpected.k
radumereuta, line 5: In any case, toString is wrong in this case. You should fix it now, since you are working on it.

kernel/src/main/java/org/kframework/kore/convertors/KILtoInnerKORE.java
cos: nevermind, I was looking at PRODUCTION_ID, no cellMarker. I can make `cellMarker` private, sure.

kernel/src/main/java/org/kframework/kil/Definition.java
cos: I don't think the syntax before followed conventions.

dwightguth: I'm not sure that's the point. The point Radu is making is that changing lines in a diff for purely stylistic reasons is something that can cause unnecessary problems, and you should get out of the habit of it.

cos: I didn't indent to change it. It got piggybacked by mistake. In any case, I really don't like that files are not formatted using a common auto-formatter. I tiresome to make sure you're only auto formatting what you've changed. It's time wasted for the sake of having nicer code formatting once in every 100 lines when the auto-formatter doesn't produce nice enough code.

In the meantime, I will revert the file to its original.

radumereuta: Looking at the Google Java Style, `An empty block or block-like construct may be closed immediately after it is opened`. Where may is italic. So it is optional, there is no standard in this case. I don't have arguments against, but I would prefer if you don't put your name on that line in the history. It falls better in the category of code ownership.

k-distribution/src/test/resources/convertor-tests/syntaxWithAttributes-kilexpected.k
cos, line 5: I didn't know we already have a better unparser then toString(). The output is not parsed so it doesn't matter if it is incorrect. It only matters that it contains most important information. I will leave it like this for now as there are many assertion files to fix otherwise. I will change to the unparser then if I notice any bug slipping in because of this.

kernel/src/main/java/org/kframework/kil/Definition.java
cos: Ok, Radu. Regarding the original issue, I've reverted the entire fire to its master version, and only added my change. My name is still on four more lines then it should be but all of them are unimportant (comments and an empty constructor).

cos:
Ok, is this good to merge?

k-distribution/src/test/resources/convertor-tests/context-kilexpected.k
```diff
@@ -0,0 +1,25 @@
...
+endmodule
+module TEST
+  context []:KItem  + _:K
+  context _:K + []:KItem
+  imports INCLUDE
+  syntax Exp ::= Exp "+" Exp
+
+
+endmodule
+require "include.k"
```
radumereuta: New Line.

dwightguth:
lgtm

k-distribution/src/test/resources/convertor-tests/userList-expected.k
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
+
+module TEST
+  syntax UserList ::= UserList "," UserList [Stx]
+  syntax UserList ::= Stx [Stx]
+  syntax UserList ::= ".UserList" [Stx]
+  syntax Stx ::= "1"
+endmodule
```
radumereuta: New line.

kernel/src/main/java/org/kframework/utils/file/FileUtil.java
radumereuta, line 165: `load` should also be static.

k-distribution/src/test/java/org/kframework/kore/convertors/BaseTest.java
radumereuta, line 106: You haven't answered this one.

cos, line 106: We're not covering comments yet because it's not obvious how they should be approached. Basically, we cannot just throw the comments in the same bucket with everything else because order matters. Each comment refers to the sentence just below it but that order is not preserved in KORE. We'll have to figure out a smarter way of attaching comments to ASTs but we haven't yet.

kernel/src/main/java/org/kframework/utils/file/FileUtil.java
cos, line 165: Radu, this is not the purpose of this pull request, and I haven't even touched the "load" line. I agree it is good to fix small things along the way but if you notice them, notice them from the beginning, not just before the pull request is ready to be merged. Don't trickle your feedback, do it in one go.

dwightguth:
We seem to have pretty much reached consensus on this, and the remaining issues are minor enough not to be worth blocking the merge on. I am going to merge this.

PR #1212, opened 2014-11-25
removing lookup and update classes form the java backend

andreistefanescu:
not ready for review


cos:
horrey! if this is what I think it is :-)

andreistefanescu:
Yes, it is :-)

cos:
lGREATtm :+1:

dwightguth:
Maybe this can be a separate commit, but I think we should also remove the lookup/update classes from the generic kil.

andreistefanescu:
This is quite hairy, so for now I will focus on the java backend.

cos:
if they are not needed by the backend, it shouldn't be a problem removing them from the generic kil
afaik, there are no other tools using this representation

andreistefanescu:
@yilongli @cos @dwightguth please review
@dwightguth I deleted a unit test for list because mocking the dependencies seems to much. If you want, I can add it back as a K file.

dwightguth:
@andreistefanescu : I am a little concerned about the test being removed because my experience has constistently been that the code we write corresponding to data structures provides very poor code coverage when we test it only using K files, and we end up not covering paths in the bodies of these functions, leading to errors that we catch later when our commits have broken something. I would feel a lot more comfortable if there were unit tests for your additions to BuiltinListOperations, DataStructures, etc.

yilongli:
Looks like a huge improvement. I will need some time to process it.

k-distribution/samples/kernelc/tests/sorting/insertion_sort_spec.k
```diff
@@ -37,10 +37,10 @@ rule [insertion-sort-loop-inv1]:
   =>
     .K
   ...</k>
-  <env>...
```
yilongli, line 40: why remove the `...`?

andreistefanescu, line 40: Because the semantics of map update, which cannot be evaluated if you have a map variable, unless you know the map variable does not contain the key you want to update. We need to improve on how maps work, but that is something for another day.

yilongli, line 40: then I would like to have a TODO comment here to keep track of it

cos:
it looks great. this is a large step forward. but common, don't go 4/5 of the way, delete the now-redundant classes from the compiler.

java-backend/src/main/java/org/kframework/backend/java/kil/DataStructures.java
```diff
@@ -0,0 +1,140 @@
+// Copyright (c) 2013-2014 K Team. All Rights Reserved.
+package org.kframework.backend.java.kil;
+
+import org.kframework.backend.java.builtins.IntToken;
+
+import java.util.Map;
+import java.util.Set;
+
+public interface DataStructures {
```
yilongli, line 15: javadoc

java-backend/src/main/java/org/kframework/backend/java/kil/Rule.java
```diff
@@ -239,29 +239,20 @@ public void visit(Cell cell) {
...
-                assert eq.leftHandSide() instanceof DataStructureLookupOrChoice;
-                if (eq.leftHandSide() instanceof DataStructureLookup) {
-                    DataStructureLookup lookup = (DataStructureLookup) eq.leftHandSide();
-                    Term value = eq.rightHandSide();
-
-                    // do not double count base variable again
-                    lhsVariablesToReuse.addAll(VariableOccurrencesCounter.count(lookup.key()));
-                    lhsVariablesToReuse.addAll(VariableOccurrencesCounter.count(value));
+                if (DataStructures.isLookup(eq.leftHandSide())) {
+                    lhsVariablesToReuse.addAll(VariableOccurrencesCounter.count(DataStructures.getLookupKey(eq.leftHandSide())));
```
yilongli, line 255: add back my comment `// do not double count base variable again`?

yilongli:
lgtm

PR #1213, opened 2014-11-26
add javaBackendHooks method to AbstractKModule + clean up BuiltinFunctio...

dwightguth:
...n code

**do not review yet**


dwightguth:
@andreistefanescu @yilongli please review

java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
```diff
@@ -327,16 +331,20 @@ public Term evaluateFunction(KItem kItem, boolean copyOnShareSubstAndEval, TermC
...
-                    Term result = context.global().builtins.invoke(context, kLabelConstant, arguments);
+                    Term result = builtins.invoke(context, kLabelConstant, arguments);
                     if (result != null) {
                         return result;
                     }
-                } catch (IllegalAccessException | IllegalArgumentException e) {
-                } catch (InvocationTargetException e) {
-                    Throwable t = e.getTargetException();
+                } catch (ClassCastException e) {
+                // DISABLE EXCEPTION CHECKSTYLE
```
yilongli, line 342: I don't know much about checkstyle but is there a better way to suppress the warning?

java-backend/src/main/java/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
```diff
@@ -107,6 +107,7 @@ private boolean rewrite() {
 
                     cell.unsafeSetContent((Term) rightHandSide.accept(substAndEvalTransformer));
                 }
+                context.setTopTerm(subject);
```
yilongli: I am not very comfortable with the idea of context being mutable and I have to set it during rewriting. I think this is error-prone.

java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaSymbolicCommonModule.java
```diff
@@ -1,60 +1,75 @@
...
-            BuiltinCollectionOperations.class,
-            BuiltinListOperations.class,
-            BuiltinMapOperations.class,
-            BuiltinSetOperations.class,
-            BuiltinIOOperations.class,
-            BuiltinVisitorOperations.class,
-            UseSMT.class,
-            SortMembership.class,
-            FreshOperations.class);
+    private final String hookPropertiesFileName = "hooks.properties";
```
yilongli: make it static?

kernel/src/main/java/org/kframework/main/AbstractKModule.java
```diff
@@ -11,9 +13,12 @@
 import org.kframework.krun.tools.LtlModelChecker;
 import org.kframework.krun.tools.Prover;
 import org.kframework.main.KModule;
+import org.kframework.utils.inject.Builtins;
 import org.kframework.utils.inject.Options;
 
+import com.beust.jcommander.internal.Lists;
```
yilongli: Use guava's `Lists`.

andreistefanescu:
I am unclear on how hooks are detected to be impure.

kernel/src/main/java/org/kframework/main/AbstractKModule.java
dwightguth: oops, thanks

java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
dwightguth, line 342: this is actually the best way because it localizes disabling the check to the specific check and location in the code that you care about. Other checks in the same file are not affected.

java-backend/src/main/java/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
dwightguth: this is actually needed for functions that reference the configuration. I am open to other suggestions for how to get this information, but this seemed like the most obviousc choice to me.

@cos, you need this for what you want to do with strategies, can you provide your opinion here?

dwightguth:
@andreistefanescu , hooks are impure if their implementation when you execute them in kompile throws the ImpureFunctionException. A pure function is otherwise identical to an impure function except that impure functions are not subject to partial evaluation.

The problem with the approach we used before is that if you needed a particular dependency in order to construct a builtin hook implementation, it would still try to resolve that dependency during kompile. I think that this is a simpler and cleaner solution than trying to disable all rules that use a particular functions symbol during kompile only. It also has the advantage of hypothetically allowing special forms of a function to be used during partial evaluation.

andreistefanescu:
You still have not answered my question: how do you choose one implementation over the other (the one which throws the exception vs the one that does not).

I'm not sure I would say that redefining a large number of functions to have bodies that just throw an exception is clean. Maybe the hooks should just be label with [impure] at declaration?

java-backend/src/main/java/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
cos: I'm not that worried about this detail.

My main concern is being able to take snapshots of the entire state. It would be very hard to achieve with this rewriter anyhow, as it does destructive updates. It may work if we have a good cloning mechanism for everything. Do we?

dwightguth:
@andreistefanescu I went ahead and figured out a way to implement your suggestion. Please take a look and re-review.

dwightguth:
@yilongli please re-review

yilongli:
lgtm

andreistefanescu:
Thanks Dwight, looks good (and I would say better).

PR #1215, opened 2014-11-28
fix Andrei's bug with read stdin

dwightguth:
@andreistefanescu please review and let me know if this fixes your issue.


andreistefanescu:
lgtm

PR #1216, opened 2014-11-28
remove maude backend and make java the default

dwightguth:
so many test failures to go through... but the output is easier to read on Jenkins.

**do not review yet**


dwightguth:
Unfortunately I am rapidly coming to the conclusion that the java rewrite engine is simply not stable enough to use as a default yet. I will try to salvage as much of this diff as I can, but I'm giving up on the project of getting rid of maude for now. It requires too much cleanup and bugfixes first.

grosu:
That does not sound encouraging.  Let's discuss this tomorrow.

Grigore

PR #1217, opened 2014-11-30
Feature/generic collection interfaces

cos:
**check this out**


cos:
`CollectionsTest` is the interesting part (you can also look under the hood if you like). It shows how to use immutable collections as Java 8 streams (bottom line, it doesn't matter they are Scala collections, they are black boxes from the user's perspective). Basically, our API is just a `stream(...)` method taking an immutable collection into a `Stream`, and a set of `to*` collectors which recreate the immutable collections.

Also, a bit of assoc magic :-)

Next, using KORE collection in the same way.

cos:
added an example of manipulating a KList in the same way.

cos:
and KSeq

k-distribution/src/test/java/org/kframework/kore/convertors/TstKOREtoKILIT.java
```diff
@@ -71,7 +71,7 @@ public void syntaxWithOR() throws IOException {
         outerKILtoKOREtoKILTest();
     }
 
-    @Test
+    @Test @Ignore
```
dwightguth, line 74: can you explain why this unit test was dropped?

kore/src/main/scala/org/kframework/kore/Sorts.scala
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
+
+package org.kframework.kore
+
+import org.kframework.kore
+
+object Sorts {
+  def KString = kore.KString
+}
```
dwightguth, line 9: note this

dwightguth:
Just to clarify: is this diff changing the existing API that you had previously decided upon in any way?

cos:
no, it only adds a nice way to dead with the classes from Java

k-distribution/src/test/java/org/kframework/kore/convertors/TstKOREtoKILIT.java
cos, line 74: The unit test passes when run by itself, and fails when run in a suite. I guess that SDF somehow gets polluted. It would probably require separate class loaders (similar to what you did for nailgun) or some other trick to make it work, but it is not worth the effort.

dwightguth, line 74: Is there a way to specify that a test should not be run in parallel with
other tests? that would also probably solve the issue. But I agree it's not
worth it if the fix isn't that simple.

dwightguth:
lgtm

k-distribution/src/test/java/org/kframework/kore/convertors/TstKOREtoKILIT.java
cos, line 74: They're not run in parallel, actually. But somehow the state gets tainted. I don't think it is worth the effort to track down.

cos:
will merge it once the core-convertors branch gets merged.

dwightguth:
Jenkins: test this please

dwightguth:
Merge if tests pass

PR #1218, opened 2014-12-01
Added Lower-Case and Upper-Case Identifiers

omarzd:
Just like the builtin Id is defined, I made LCId and UCId.


dwightguth:
I'm not inclined to accept this diff as it stands right now. This is adding a lot of weight to the tutorial, and I'm not sure it is going in the right places. If you want to do this, I suggest you create a smaller diff that modifies just id.k to add the following lines:

```
syntax CamelCaseId
syntax PascalCaseId
rule isPascalCaseId(X:Id) => isUppercase(substrString(#tokenToString(X), 0, 1))
rule isCamelCaseId(X:Id) => isLowercase(substrString(#tokenToString(X), 0, 1))
```

You will also have to define isUppercase and isLowercase in string.k.

radumereuta:
But do we want to support this by default?
I can see cases where this would produce ambiguities, since they are included automatically in every definition. There is the corner case when you don't specify a start symbol for the program parser (then it tries all possible sorts), and you give as input an Id. Then the parser will return an ambiguity with one of the new Ids, and the old one...

dwightguth:
@radumereuta hence my suggestion of declaring their syntax indirectly.

omarzd:
@dwightguth Just to be clear, I didn't really understand your suggestion. Do you mean to abandon this whole diff and make the smaller one you suggested, or should I add to it?

dwightguth:
I mean to abandon the diff in favor of what I suggested.

omarzd:
Oh I see. So instead of LowerCaseId and UpperCaseId, they should be CamelCaseId and PascalCaseId. And they should be subsorts of Id. Then whenever I want a rule to match only one of these two subsorts, say PascalCaseId, I will specify it using a "when isPascalCaseId(TheIdInMyRule)". Is that right?
This is a very good suggestion. Thank you.

dwightguth:
Well actually, given that they are declared as sorts, you can actually say X:PascalCaseId, and have it automatically infer that you want to call the isPascalCaseId function. But otherwise, yeah.

omarzd:
Of course!! I don't know why I tend to overcomplicate things. Thank you. I will close this now.

dwightguth:
Jenkins: ok to test

omarzd:
@dwightguth I'm sorry, I did this

```
syntax Id ::= CamelCaseId
syntax Id ::= PascalCaseId
```

but you suggested

```
syntax CamelCaseId
syntax PascalCaseId
```

dwightguth:
The problem with this approach is that it declares that if something is a CamelCaseId, then it is an Id. This means that isId has to call out to isCamelCaseId, which in turn is defined using isId. This creates a loop which does not terminate. The solution that I suggested is what gets around this issue by not declaring it as a proper subsort, but simply as a predicate which identifies a subset of Ids.

omarzd:
OK, got it. I corrected it.

omarzd:
Hey sorry, I'm new to github. I was trying to create another pull request, but this one was standing in the way. It was neither merged nor rejected. I don't know what to do in this case.

dwightguth:
If you want to create a new pull request, either put your code on a different branch, or else close this pull request, push your changes, and then reopen the pull request.

omarzd:
It should pass the test this time

dwightguth:
Jenkins: ok to test

omarzd:
OK. I did the following sample K definition:

```
module TEST
    syntax Exp ::= PascalCaseId
    configuration <k> $PGM:Exp </k>
endmodule
```

And when I krun the following test program:

```
Asd
```

I get this:

```
[Error] Critical: Parse error: Desired start symbol not found: Exp
```

Why is this happening? This is why I wanted PascalCaseId and CamelCaseId to be subsorts of Id or define them as tokens like I did initially with LowercaseId and UppercaseId like so:

```
syntax CamelCaseId ::= Token{[a-z][\_a-zA-Z0-9]*}       [notInRules, regex("(?<![A-Za-z0-9\\_])[a-z\\_][A-Za-z0-9\\_]*")]
syntax PascalCaseId ::= Token{[A-Z][\_a-zA-Z0-9]*}       [notInRules, regex("(?<![A-Za-z0-9\\_])[A-Z\\_][A-Za-z0-9\\_]*")]
```

Should I do it like this?

omarzd:
I just tried defining them as tokens and it didn't work.

dwightguth:
Are you trying to parse programs by limiting them to only parsing
identifiers with a particular case?

omarzd:
Yes, that's exactly what I'm trying to do.

dwightguth:
Okay. Unfortunately, you will probably not be able to do this in such a way
that the change can be made part of the prelude. I suggest you take a look
at fun-untyped in the tutorial and see how it elects to specify its own
syntax of identifiers by means of declaring the token sort directly in
your  definition.

omarzd:
OK. Well that's what I did initially and it worked. Then I wanted to add that here so I started this pull request.

PR #1221, opened 2014-12-02
add logik back to tutorial in java rewrite engine

dwightguth:
**please review**


dwightguth:
@grosu @yilongli please review

grosu:
LGTM.  Thanks for changing the .out files to not use the regexes anymore, they are so much easier to read now!  Also, we may want eventually to rename the "builtin" labels uniformly to all start with `#`, but no rush with that right now.

yilongli:
lgtm

PR #1225, opened 2014-12-02
fix superheating and supercooling for java backend

dwightguth:
@traiansf @yilongli please review


dwightguth:
Seems to have segfaulted. Not entirely sure why. For some reason it didn't archive the jvm dump like it was supposed to. I'm going to rerun it. If it recurs, I'll track down the issue with the error log and go from there.

dwightguth:
Jenkins: test this please

traiansf:
I don't understand what happened with the superheat rules in the Maude backend.  This pull request seems to remove the code for generating them without leaving anything instead.

I thing that until the Java backend can replace the Maude backend, no feature should be removed.  You might of course move it entirely in the Maude backend, but removing it seems like giving up important features (which were used in the POPL'12 C semantics paper, for example).

Please help me understand.

dwightguth:
Jenkins: test this please

dwightguth:
@grosu :

Apparently the behavior that causes you to get 2 instead of 3 in lambda++ happens when a heating rule is marked as a transition, because it occurs only in the case that a heating rule and a computational rule form a critical pair. When I modified the way superheating works to not be hacky, this then surfaced, as it did if I marked "--transition computational".

The undesirable behavior goes away if you force the argument to cc to be evaluated before restoring the environment.

Do you approve of this change? You can create a separate pull request to modify the readme if necessary, assuming you believe this is the correct solution.

dwightguth:
@traiansf : superheating is not removed. It's just that superheating is now a synonym for declaring a heating rule to be a transition. This is actually better because before, we did not get the nondeterminism arising from a critical pair between a rule and a heating application.

dwightguth:
@traiansf @yilongli @grosu this is ready for review now.

yilongli:
lgtm

traiansf:
I agree that superheating using transition is much simpler.   But I think it would impact a lot the performance of search.   Anyway,  lgtm for now.

dwightguth:
I think it's probably better to deal with the performance issue by cleverly
optimizing the search mode of the java rewrite engine than by trying to
repeat what superheating was doing before. Maybe we can tell the rewrite
engine to collapse all successive pairs of rules tagged with superheating
into a single transition. This way we get conceptual purity and performance
both. In any case, this is something to deal with when issues arise, not
now.

k-distribution/include/k-prelude.k
```diff
@@ -80,16 +80,8 @@ endmodule
...
 
-  rule heat(heated(K:K)) => K
-  rule heat(heated(_,,K:K,,_)) => K
-     [heat-choice]
-
-  rule heat((redex(_:K,, K:K,, KL:KList) => K,,KL),,_)
-    [owise]
-  rule heat((redex(K:K) => .KList),,_,,heated(_,,(.KList => K)))
-    [owise]
+  syntax K ::= cool(K) [function, klabel(cool)]
```
grosu, line 84: No big deal, but I believe that klabel is not needed in this case, because that is the default klabel anyway.

k-distribution/tutorial/1_k/3_lambda++/lesson_5/lambda.k
```diff
@@ -59,5 +59,5 @@ module LAMBDA
   syntax Exp ::= "callcc" Exp  [strict]
   syntax Val ::= cc(Map,K)
   rule <k> (callcc V:Val => V cc(Rho,K)) ~> K </k> <env> Rho </env>
-  rule <k> cc(Rho,K) V ~> _ =>  V ~> K </k> <env> _ => Rho </env>
+  rule <k> cc(Rho,K) V:Val ~> _ =>  V ~> K </k> <env> _ => Rho </env>
```
grosu, line 62: Thanks for fixing this.  I wonder how we didn't notice it before.

k-distribution/tutorial/2_languages/3_fun/1_untyped/1_environment/fun-untyped.k
```diff
@@ -686,7 +686,7 @@ above. */
   syntax KItem ::= assignTo(Names,Exps)  [strict(2)]
 
   rule <k> assignTo(.Names,.Vals) => . ...</k>            [structural]
-  rule <k> assignTo((X:Name,Xs => Xs),(V:Val,Vs => Vs)) ...</k>
+  rule <k> assignTo((X:Name,Xs => Xs),(V:Val,Vs:Vals => Vs)) ...</k>
```
grosu, line 689: I guess this is to avoid non-deterministic behaviors.  OK.  I'll deal with the readme, if needed, later when I prepare cs422 for Spring.

k-distribution/include/k-prelude.k
dwightguth, line 84: technically the default klabel is 'cool, with an apostrophe. I elected to leave the klabel the same rather than changing the references to it. Hence the attribute.

k-distribution/tutorial/2_languages/3_fun/1_untyped/1_environment/fun-untyped.k
dwightguth, line 689: without this, stuck-letrec does not get stuck, but instead can evaluate to 2. This seemed incorrect to me, so I changed it.

grosu, line 689: Interesting.  Yes, I approve the change.

Grigore

PR #1227, opened 2014-12-02
remove "remove preincluded rules" for rules in k definitions

dwightguth:
**please review**

PR #1228, opened 2014-12-02
fix hybrid in java backend

dwightguth:
**do not review yet**


dwightguth:
Turns out this is inseparable from #1227. Closing as duplicate and combining the two.

PR #1234, opened 2014-12-03
Adding #time function

omarzd:
# time should return the number of milliseconds since the epoch. However,

the return type is long, and long isn't supported in K as far as I know.
I made this commit, even though it's a mess, to get feedback from the
developers on how to properly do this. I was following [this pull
request](https://github.com/kframework/k/pull/967/files) which is why I
put it under IO but then I changed my mind in hooks.properties and made
time.k!


dwightguth:
Just make the return type of #time be an Int. Ints are arbitrary precision in K, so you don't have to worry about it being out of range.

k-distribution/include/builtins/time.k
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2012-2014 K Team. All Rights Reserved.
+require "long.k"
+
+module TIME-HOOKS
+  imports #TIME-INTERFACE
+  imports LONG-HOOKS
+
+  syntax Long ::= "time" "(" ")" [function, klabel(randomRandom), hook(#TIME:time)]
```
dwightguth: fix this klabel

k-distribution/include/io/util.k
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) 2013-2014 K Team. All Rights Reserved.
```
dwightguth: should this file be deleted?

k-distribution/include/builtins/time.k
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2012-2014 K Team. All Rights Reserved.
```
dwightguth: this is a new file so the copyright should just be 2014

dwightguth:
There is one more component of this that you need to handle in order to make it work with maude, which is the actual invocation of the IO server. Take a look at some code in tcp.k for examples of how the IO server is called.

dwightguth:
One thing I'd like to note is that we actually plan to deprecate the Maude backend moving forward and provide only minimal support for it. So you may consider it in your interest to provide an implementation of this hook in the java backend as well as the maude backend. But I won't insist on it.

omarzd:
If I delete util.k, where do I put

```
syntax KItem ::= "#time" "(" ")" [function, klabel(#time), hook(#TIME:#time)]
```

dwightguth:
I'm not sure I understand. It seems to be duplicated from the file time.k.

omarzd:
Yeah you're right. I will delete util.k
I don't see how the IO server is called in tcp.k.

dwightguth:
Take a look at the rules which include "#tcpRequest" in them.

omarzd:
Regarding the backend, How do I implement a java hook? I just imitated code from the #external pull request that I linked in the first comment.

dwightguth:
If you look at that pull request, you will see that it actually modifies code in the java backend as well. You can just add the function as a method on the BuiltinIOOperations class.

k-distribution/include/builtins/time.k
dwightguth: can you please add the attribute "impure" to this function to indicate to the java backend that this function accesses state outside its inputs?

PR #1220, opened 2014-12-02
improved handling of bottom

andreistefanescu:
not ready for review yet

PR #1235, opened 2014-12-03
fix bug in #1220

dwightguth:


dwightguth:
I believe this failed because of a race condition in the Jenkins scripts associated with the C semantics and nailgun. I'm not entirely sure what the correct approach is here... I'll think about it and deal with it if it recurs.

dwightguth:
Jenkins: test this please

andreistefanescu:
lgtm

PR #1237, opened 2014-12-04
Fixed substitution.

dwightguth:
Creating pull request for @traiansf 's code, but rebased onto correct history. I will review this myself since it's his code.


java-backend/src/main/java/org/kframework/backend/java/kil/Sort.java
```diff
@@ -39,6 +39,8 @@
     public static final Sort BIT_VECTOR     =   Sort.of("MInt");
     public static final Sort META_VARIABLE  =   Sort.of("MetaVariable");
 
+    public static final Sort VARIABLE = Sort.of("Variable");
```
dwightguth: can you please align this to the right with the others?

dwightguth:
looks okay, but as you said, you were going to add comments to the algorithm to make it easier to read.

dwightguth:
I'm going to close this and sublimate it into the review for #1238. That makes the merge much cleaner and then we don't duplicate work that has already been given a more permanent solution in that branch. so... lgtm.

PR #1238, opened 2014-12-04
Convert many examples to java backend

dwightguth:
**please review**


dwightguth:
Something is in an inconsistent state because jenkins is reporting errors that I can't reproduce. I am going to rerun this and then try to inspect the working directory on the box.

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
@grosu this is almost ready to be reviewed, but I still have to deal with the issue with kool. Can you please either give me an ETA on when you will make a pull request for it, or else tell me what the solution you would like to have is so I can unblock myself?

dwightguth:
Jenkins: test this please

dwightguth:
segfault...

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
This is ready for review. I rebased the entire history so that it's pretty conceptually clean. Moving forward, I'll leave the history as is. But I took the liberty because I wasn't done with the diff at all before.

@grosu please review changes to the tutorial. They are all highly trivial, with the exception that Id is no longer subsorted to Variable. I did this because I wanted to avoid the case where a developer doesn't want Ids to be variables and ends up with undesired behavior they can't do anything about.

@andreistefanescu please review changes to the symbolic execution. You'll note I didn't actually make a prelude by default, simply modified the existing preludes so that smt queries can be run even when no prelude is provided. This is because I ran into lots of unpredictable behaviors I couldn't figure out when I tried to do things the way you suggested, and when I changed back to not having a default prelude, they went away. If you want to take care of this issue, you will have to address it yourself.

@yilongli please review the changes to Java code. Nothing too major, just bug fixes here and there, plus the substitution fix that Traian wrote.

Note that there is still one test failing, but this is a test in kool that is nondeterministic and should only require @grosu 's say-so to fix. I just didn't fix it because I didn't know if he wanted to modify the semantics of kool to remove the nondeterminism, or modify the test output so that it does a search. @grosu , please either fix this test yourself in your own pull request, or clarify which of these two outcomes you want so I can do it myself.

This diff removes very nearly all of the remaining maude tests in the tutorial. Still left is agent and a few regression tests. I will address these in the next pull request which removes the maude backend entirely.

k-distribution/tutorial/1_k/5_types/lesson_4/lambda.k
```diff
@@ -20,31 +18,30 @@ module LAMBDA
                | "mu" Id "." Exp                  [binder]
 
   syntax Type ::= "int" | "bool"
-                | Type "->" Type
+                | Type "->" Type                  [smtlib(funcType)]
```
grosu: Why is smtlib needed here?  It will make the tutorial harder to present.

k-distribution/tutorial/1_k/5_types/lesson_4/lambda.k
```diff
@@ -20,31 +18,30 @@ module LAMBDA
...
-  rule T1:Type  * T2:Type => T1 = int ~> T2 = int ~> int
-  rule T1:Type  / T2:Type => T1 = int ~> T2 = int ~> int
-  rule T1:Type  + T2:Type => T1 = int ~> T2 = int ~> int
-  rule T1:Type <= T2:Type => T1 = int ~> T2 = int ~> bool
-  syntax Exp ::= Exp "->" Exp  [strict]
-
-  rule lambda X . E:Exp => ?T:Type -> E[?T/X]
-  rule T1:Type T2:Type => T1 = (T2 -> ?T:Type) ~> ?T
-  rule if T:Type then T1:Type else T2:Type => T = bool ~> T1 = T2 ~> T1
+  rule int * int => int
```
grosu: Eventually I replaced the rules you removed with the ones you added, but in some later lessons.  The idea of this lesson was to do things in a more ugly but computationally cleaner way.  But if this way does not work anymore, then fine, I'll deal with changing the tutorial when I get there next semester.

k-distribution/tutorial/1_k/5_types/lesson_4/lambda.k
```diff
@@ -20,31 +18,30 @@ module LAMBDA
...
+  rule int / int => int
+  rule int + int => int
+  rule int <= int => bool
+
+  rule lambda X . E:Exp => E[?T/X] ~> ?T:Type -> HOLE
```
grosu, line 39: Can't you make the `->` construct strict or at least strict(2), so we can write the RHS in the rule above as it used to be?

k-distribution/tutorial/1_k/5_types/lesson_4/tests/composition.lambda.out
```diff
@@ -2,5 +2,5 @@ Search results:
 
 Solution 1:
 V:K -->
-(#symType(3) -> #symType(4)) -> ((#symType(4) -> #symType(5)) -> (#symType(3) 
-  -> #symType(5)))
+(GeneratedAnonVar0:Type -> GeneratedAnonVar1:Type) -> ((GeneratedAnonVar1 ->
```
grosu: These variable names are quite large and ugly.  I think @andreistefanescu used names like `_172` or `_v172` or `_Var171` at some moment, like in Prolog.

k-distribution/tutorial/1_k/5_types/lesson_6/lambda.k
```diff
@@ -67,11 +64,10 @@ module LAMBDA
   rule <k> mu X . E => ?T:Type ...</k>  <tenv> TEnv </tenv>
        (. => <task> <k> E = ?T </k> <tenv> TEnv[X <- ?T] </tenv> </task>)
 
+  rule T1 -> T2 =/=K T1 => true [smt-lemma]
```
grosu: Why is this needed?  And in general, why do we need to get the SMT involved here?  It looks like all we need is unification.  Also, the rule above is incorrect in some type systems, for example ones with infinite types.  @andreistefanescu can you please step in here and help us simplify things out?

k-distribution/tutorial/1_k/5_types/lesson_6/lambda.k
```diff
@@ -67,11 +64,10 @@ module LAMBDA
...
+
   syntax KItem ::= Exp "=" Exp  [strict]
-  rule <k> T:Type = T':Type => . ...</k>
-       <mgu> Theta:Mgu => updateMgu(Theta,T,T') </mgu>
-  rule <tasks> <task> <k> T:Type => applyMgu(Theta,T) </k> ...</task> </tasks>
-       <mgu> Theta:Mgu => . </mgu>
```
grosu, line 74: I still think that it would be nice and instructive to have a version of a type system where the designer makes calls to the unifier explicitly.  This way people will understand it better and understand the magic underlying the narrowing-based method.

k-distribution/tutorial/2_languages/3_fun/1_untyped/1_environment/tests/config.xml
```diff
@@ -17,4 +18,23 @@
...
+  <test
+      definition="fun-untyped.k"
+      programs="../../../programs"
+      extension="fun"
+      results="." >
+     <kompile-option name="--superheat" value="strict" />
+     <kompile-option name="--backend" value="java" />
+     <all-programs>
+        <krun-option name="--pattern-matching" />
+        <krun-option name="--pattern" value="&lt;k&gt; V:K &lt;/k&gt;" />
```
grosu, line 30: This is orthogonal to the issues discussed, but I always wondered if there is anything we can do about this ugly notation, which I never remember:

```
value="&lt;k&gt; V:K &lt;/k&gt;"
```

I'd like to write it as:

```
value="<k> V:K </k>"
```

Isn't there any smart XML parser to realize that the k "tags" occur in a string?  Don't XML people care about such issues?

k-distribution/tutorial/2_languages/3_fun/1_untyped/2_substitution/tests/polymorphism-6.fun.out
```diff
@@ -2,12 +2,12 @@ Search results:
...
-fun (x -> (fun (x -> (fun (x -> (fun (x -> (fun (x -> fun (y -> x))) ((fun (x 
-  -> fun (y -> x))) x))) ((fun (x -> (fun (x -> fun (y -> x))) ((fun (x -> fun
-   (y -> x))) x))) x))) ((fun (x -> (fun (x -> (fun (x -> fun (y -> x))) ((fun
-   (x -> fun (y -> x))) x))) ((fun (x -> (fun (x -> fun (y -> x))) ((fun (x ->
-   fun (y -> x))) x))) x))) x))) ((fun (x -> (fun (x -> (fun (x -> (fun (x -> 
-  fun (y -> x))) ((fun (x -> fun (y -> x))) x))) ((fun (x -> (fun (x -> fun (y
-   -> x))) ((fun (x -> fun (y -> x))) x))) x))) ((fun (x -> (fun (x -> (fun (x
-   -> fun (y -> x))) ((fun (x -> fun (y -> x))) x))) ((fun (x -> (fun (x -> 
-  fun (y -> x))) ((fun (x -> fun (y -> x))) x))) x))) x))) x))
+fun (_1 -> (fun (_2 -> (fun (_3 -> (fun (x -> (fun (x -> fun (y -> x))) ((fun
```
grosu: What are the `_1`, `_2`, ... variables?  Are they generated?  If yes, then why do they look different from the previously generated variables, `GeneratedAnnonVar1`, etc?  I actually prefer it this way, `_1`, etc.

Also, how comes that before this expression had no anonymous variables, while now it has?

k-distribution/tutorial/1_k/5_types/lesson_4/lambda.k
dwightguth: If you do not include the smtlib lemma that I include here, the unification rewriter does not know that it can not unify T1 -> T2 = T1, and it keeps going past the point it should get stuck, outputting a single variable as the typed result of typing polymorphic-identity-2-fails.

k-distribution/tutorial/1_k/5_types/lesson_4/lambda.k
dwightguth: Actually, both work, I believe. I can change it if you want. Can you identify for me which lessons you would like reverted back to the old way?

grosu: I think it would be ideal to keep everything as much as possible as it was, because all the documentation was written for that old code.  So if that is not too much, I'd prefer to keep it things that way.  There are also latter definitions that use the nicer approach, so nothing is lost.  Thanks,

k-distribution/tutorial/1_k/5_types/lesson_4/lambda.k
dwightguth, line 39: If -> is strict, then the unification rewriter will attempt to heat type variables, and never terminate.unless you specify some kind of strategy. This was the solution that I found. I agree strictness is nicer, but I don't know how to make that work with the current form of the tool.

k-distribution/tutorial/1_k/5_types/lesson_4/tests/composition.lambda.out
dwightguth: This is easy enough to fix. I will take care of it.

k-distribution/tutorial/1_k/5_types/lesson_6/lambda.k
dwightguth: see above. This relates to polymorphic-identity-2-fails. The problem is that the smtlib does not know that it can not unify T1 -> T2 = T1, and so it adds this as a constraint and continues to rewrite, and so eventually, instead of getting stuck, it outputs that the type of the program is a variable, and does not realize that the constraint is not actually satisfied.

k-distribution/tutorial/1_k/5_types/lesson_4/lambda.k
grosu: The unification procedure should be able to detect two problems all by itself, without any smt involvement:
1) equalities of the form X = T[X]; and
2) equalities of the form f(...) = g(...), with f different from g
In both cases it should fail, smoothly.  If that is not the case, then we have a problem with our unification procedure.

I think @andreistefanescu should also say something about this.

grosu: It is OK with me to let it pass in order to speed up the removal of Maude, but at least we should add a task for this to be fixed asap after the removal of Maude.  It is quite unnatural to get the SMT involved here.

k-distribution/tutorial/1_k/5_types/lesson_6/lambda.k
dwightguth, line 74: That functionality does not currently work in the java backend. We can probably fix that, but I elected instead to make the tutorial work the way the java backend is currently designed. I'm not sure I have the time to try to make all of that work and still get done with maude in time to work on the C semantics though, so if that's what you want, getting rid of Maude will probably have to wait further.

k-distribution/tutorial/1_k/5_types/lesson_4/lambda.k
grosu, line 39: But I think it worked before, didn't it?  @andreistefanescu ?  Again, I'm OK with letting it pass in the interest of moving on with our plan, but in that case:
1) Add a code comment the first time it appears that we will fix it using strictness to look more syntactic
2) Add a github issue about it, to make sure we do not forget it

k-distribution/tutorial/2_languages/3_fun/1_untyped/2_substitution/tests/polymorphism-6.fun.out
dwightguth: these are not variables, they are fresh constants. That is why they look different. I will change the Generated\* variables though.

k-distribution/tutorial/1_k/5_types/lesson_6/lambda.k
grosu: Now I understand the problem better, but it still very strongly seems to be an artifact of a mis-implementation of our unification procedure, and not an SMT related issue.  The two checks I mentioned before (circularity and different constructs) should be checked by Java when it generates the equations for the SMT, and not by the SMT.

Or is this related to the fact that we want to have "unification modulo domain reasoning"?  @andreistefanescu ?

k-distribution/tutorial/2_languages/3_fun/1_untyped/2_substitution/tests/polymorphism-6.fun.out
grosu: ooops, right.  Hm, tricky.  We definitely need different notations for them.  `_1` looks like a variable and not like a constant.  I like this notation for variables.  For constants we should want something else.  Maybe `#1`, to indicate that it is a "builtin" thing?

k-distribution/tutorial/1_k/5_types/lesson_6/lambda.k
grosu, line 74: OK, let's not do this then, it is more important to clean up the framework.  However, please add an issue to github to keep track of it.  We should basically export the functionality of the unifier in the K API, and then be able to just hook it to KLabels like update MGU, applyMGU, etc.

k-distribution/tutorial/2_languages/3_fun/1_untyped/1_environment/tests/config.xml
dwightguth, line 30: most likely in order to fix this we will have to abandon xml in favor of a different markup language. I think that could be worthwhile, but not inside the scope of this diff.

k-distribution/tutorial/1_k/5_types/lesson_4/lambda.k
dwightguth: done

k-distribution/tutorial/1_k/5_types/lesson_4/tests/composition.lambda.out
dwightguth: done

k-distribution/tutorial/2_languages/3_fun/1_untyped/2_substitution/tests/polymorphism-6.fun.out
dwightguth: done

java-backend/src/main/java/org/kframework/backend/java/kil/JavaSymbolicObject.java
```diff
@@ -121,6 +122,35 @@ public void visit(Variable variable) {
         return Collections.unmodifiableSet(variableSet);
     }
 
+     /**
+     * Returns a {@code Set} view of the user variables in this
+     * {@code JavaSymbolicObject}.
+     * <p>
+     * When the set of user variables has not been computed, this method will do the
+     * computation instead of simply returning {@code null}
+     * {@link JavaSymbolicObject#getVariableSet()}.
```
yilongli: are you try to say `@see {@link JavaSymbolicObject#getUserVariableSet()}` here? BTW, what is user variable?

yilongli:
The changes to Java code looks fine to me.

k-distribution/tutorial/1_k/5_types/lesson_4/lambda.k
andreistefanescu: @grosu is right, there should be an occurs check here that makes `X = T[X]` false. If this does not happen, it must be a bug. I will look into it.

k-distribution/tutorial/1_k/5_types/lesson_4/lambda.k
andreistefanescu, line 39: @grosu this used to work before, mainly because `Type` is subsorted to `KResult`, and thus no heating should occur for `Type` variables.

k-distribution/tutorial/1_k/5_types/lesson_4/lambda.k
dwightguth: I took care of it in this diff.

k-distribution/tutorial/1_k/5_types/lesson_6/lambda.k
dwightguth: same as above: done

k-distribution/tutorial/1_k/5_types/lesson_4/lambda.k
dwightguth, line 39: @grosu I added an issue and assigned it to Andrei, and created a TODO in lesson 4. Andrei says he will get to it after he finishes his current project.

dwightguth:
@grosu please re-review

@andreistefanescu please review changes to symbolic execution.

k-distribution/tutorial/1_k/5_types/lesson_5/lambda.k
```diff
@@ -35,9 +33,10 @@ module LAMBDA
   rule T1:Type  / T2:Type => T1 = int ~> T2 = int ~> int
   rule T1:Type  + T2:Type => T1 = int ~> T2 = int ~> int
   rule T1:Type <= T2:Type => T1 = int ~> T2 = int ~> bool
-  syntax Exp ::= Exp "->" Exp  [strict]
-  rule <k> lambda X:Id . E:Exp => ?T:Type -> E ~> tenv(TEnv) ...</k>
+
+  rule <k> lambda X:Id . E:Exp => E ~> ?T:Type -> HOLE ~> tenv(TEnv) ...</k>
```
grosu, line 37: A TODO comment for Andrei here as well, to do it with strictness once fixed?

dwightguth, line 37: I mean, I can put it in every lesson if you want me to, it's just that they're all essentially the same change. It's just to a bunch of different lessons.

k-distribution/tutorial/1_k/5_types/lesson_5/tests/polymorphic-identity-1-fails.lambda.out
```diff
@@ -2,5 +2,5 @@ Search results:
 
 Solution 1:
 V:K -->
-#symType(5) = (bool) ~> #symType(4) = #symType(3) ~> #symType(4) ~> (#symType(
-  2) -> HOLE) ~> tenv ( .Map ) ~> (HOLE (#symType(1) -> #symType(1)))
+bool -> _0:Type = (int -> _1:Type) ~> _1 ~> if _0 then HOLE else id 2 ~> (bool
```
grosu, line 5: Not a big deal, but I'm just curious why the top equality has not been split into bool = int and _0 = _1.

k-distribution/tutorial/2_languages/3_fun/1_untyped/2_substitution/tests/polymorphism-6.fun.out
```diff
@@ -2,12 +2,12 @@ Search results:
...
-  fun (y -> x))) ((fun (x -> fun (y -> x))) x))) x))) x))) x))
+fun (#1 -> (fun (#2 -> (fun (#3 -> (fun (x -> (fun (x -> fun (y -> x))) ((fun 
+  (x -> fun (y -> x))) x))) ((fun (x -> (fun (x -> fun (y -> x))) ((fun (x -> 
+  fun (y -> x))) x))) #3))) ((fun (#3 -> (fun (x -> (fun (x -> fun (y -> x))) 
+  ((fun (x -> fun (y -> x))) x))) ((fun (x -> (fun (x -> fun (y -> x))) ((fun 
+  (x -> fun (y -> x))) x))) #3))) #2))) ((fun (#2 -> (fun (#3 -> (fun (x -> (
+  fun (x -> fun (y -> x))) ((fun (x -> fun (y -> x))) x))) ((fun (x -> (fun (x
+   -> fun (y -> x))) ((fun (x -> fun (y -> x))) x))) #3))) ((fun (#3 -> (fun (
+  x -> (fun (x -> fun (y -> x))) ((fun (x -> fun (y -> x))) x))) ((fun (x -> (
+  fun (x -> fun (y -> x))) ((fun (x -> fun (y -> x))) x))) #3))) #2))) #1))
```
grosu, line 13: Nice!  Can you please, @dwightguth , add a short sentence in the manual-to-be, to discuss the difference between _7 and #7?  Thanks!

grosu:
LGTM.  Thanks, Dwight, it is very nice, a pleasure to read definitions now.  I see there are some nice improvements also on the pretty-printer / unparser.

k-distribution/tutorial/1_k/5_types/lesson_5/lambda.k
grosu, line 37: It is OK if you have one comment per situation in the first K definitions in which it occurs.  I thought the other one you commented was a different one, but if the same then fine, ignore this comment.

Grigore

k-distribution/tutorial/1_k/5_types/lesson_5/tests/polymorphic-identity-1-fails.lambda.out
dwightguth, line 5: because we aren't actually relying on equalities down at that level. Once we go that deep, it becomes the built in unification.

k-distribution/tutorial/2_languages/3_fun/1_untyped/2_substitution/tests/polymorphism-6.fun.out
dwightguth, line 13: done

k-distribution/tutorial/1_k/5_types/lesson_5/tests/polymorphic-identity-1-fails.lambda.out
grosu, line 5: Right, I expected that to be the reason.  It would be nice though to have a hook to the builtin equality that the internal unification algorithm uses, for those who want to implement things really efficiently.  But this can wait.  It is good to merge.

andreistefanescu:
lgtm

dwightguth:
An error in the Jenkins configuration caused a race condition. I am pretty sure that is why this failed. I am retesting, and fixed the Jenkins issue.

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

PR #1239, opened 2014-12-04
Fix rule cache. Taking into consideration changes to the attributes of rules.

radumereuta:
@dwightguth please review.


radumereuta:
Nevermind. Got a better solution.

PR #1240, opened 2014-12-04
Better fix for the rule attributes cache.

radumereuta:
@dwightguth please review.
I'm copying the attributes from the StringSentence now, instead of taking them from the previous cached rule.


dwightguth:
merge if tests pass.

PR #1241, opened 2014-12-04
Rule parser

radumereuta:
Let's see if tests pass.

### DIFF
--- a/k-distribution/tests/config.xml
+++ b/k-distribution/tests/config.xml
@@ -5,10 +5,10 @@
   <!-- Tutorial Languages and Programs. -->
   <include file="../tutorial/tests/config.xml" directory="tutorial" />
 
-  <!-- Include the tests for the new parser. -->
+  <!-- Include the tests for the new parser. 
   <include file="newparser/config.xml"
            directory="tests/newparser" skip="pdf" />
-
+-->
   <include file="../samples/tests/config.xml" directory="samples" />
 
   <include file="regression/config.xml" directory="tests/regression" skip="pdf" />
--- a/k-distribution/tests/config.xml
+++ b/k-distribution/tests/config.xml
@@ -5,10 +5,10 @@
   <!-- Tutorial Languages and Programs. -->
   <include file="../tutorial/tests/config.xml" directory="tutorial" />
 
-  <!-- Include the tests for the new parser. 
+  <!-- Include the tests for the new parser. -->
   <include file="newparser/config.xml"
            directory="tests/newparser" skip="pdf" />
--->
+
   <include file="../samples/tests/config.xml" directory="samples" />
 
   <include file="regression/config.xml" directory="tests/regression" skip="pdf" />
--- a/kernel/src/main/java/org/kframework/kil/StringSentence.java
+++ b/kernel/src/main/java/org/kframework/kil/StringSentence.java
@@ -11,13 +11,15 @@ import org.kframework.kil.visitors.Visitor;
  */
 public class StringSentence extends ModuleItem {
     private String content;
-    private String contentLocation;
+    private int contentStartLine;
+    private int contentStartColumn;
     private String label;
     private String type;
 
-    public StringSentence(String content, String contentLocation, String type, String label) {
+    public StringSentence(String content, int contentStartLine, int contentStartColumn, String type, String label) {
         this.content = content;
-        this.contentLocation = contentLocation;
+        this.contentStartLine = contentStartLine;
+        this.contentStartColumn = contentStartColumn;
         this.type = type;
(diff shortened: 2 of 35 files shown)


radumereuta:
@dwightguth could you have a look?
I had to fix other things than the parser, and I wouldn't like to fall behind with the merge conflicts.

k-distribution/tests/config.xml
```diff
@@ -5,10 +5,10 @@
   <!-- Tutorial Languages and Programs. -->
   <include file="../tutorial/tests/config.xml" directory="tutorial" />
 
-  <!-- Include the tests for the new parser. -->
+  <!-- Include the tests for the new parser.
```
dwightguth: why are these tests commented out?

kernel/src/main/java/org/kframework/kil/StringSentence.java
```diff
@@ -11,13 +11,15 @@
...
 public class StringSentence extends ModuleItem {
     private String content;
-    private String contentLocation;
+    private int contentStartLine;
+    private int contentStartColumn;
     private String label;
     private String type;
 
-    public StringSentence(String content, String contentLocation, String type, String label) {
+    public StringSentence(String content, int contentStartLine, int contentStartColumn, String type, String label) {
```
dwightguth, line 19: why is this only half of a Location?

kernel/src/main/java/org/kframework/parser/concrete/disambiguate/CollectVariablesVisitor.java
```diff
@@ -69,17 +69,19 @@ public Void visit(TermCons node, Void _void) {
...
-                } catch (ParseFailedException e) {
-                    e.printStackTrace();
+                if (!node.isListTerminator()) {
+                    UserList ul = (UserList) node.getProduction().getItems().get(i);
+                    Term t1 = node.getContents().get(0);
+                    Term t2 = node.getContents().get(1);
+                    try {
+                        new CollectVariablesVisitor2(context, ul.getSort()).visitNode(t1);
+                        new CollectVariablesVisitor2(context, node.getProduction().getSort()).visitNode(t2);
+                    } catch (ParseFailedException e) {
```
dwightguth: please remove this catch block.

kernel/src/main/java/org/kframework/parser/generator/ParseRulesFilter.java
```diff
@@ -56,40 +65,64 @@ public ASTNode visit(StringSentence ss, Void _void) throws ParseFailedException
...
+
+                st = (Sentence) new JavaClassesFactory(context).getTerm((Element) xmlTerm);
+                assert st.getLabel().equals(""); // labels should have been parsed in Outer Parsing
+                st.setLabel(ss.getLabel());
+                st.setAttributes(ss.getAttributes());
+                st.setLocation(ss.getLocation());
+                st.setSource(ss.getSource());
+            } else  {
+                // parse with the new parser for rules
+                Grammar ruleGrammar = getCurrentModule().getRuleGrammar(kem);
```
dwightguth, line 104: I don't like that this is duplicated with ParseConfigsFilter. You need to find a way to share the code path.

dwightguth:
Looks  reasonable but please see my comments that I'd like  addressed before this merges.

k-distribution/tests/config.xml
radumereuta: These are the newparser tests. Since I'm rebuilding how I generate parsers, they fail right now. I need to work a bit more on parsing rules before I can make these tests work properly again.

kernel/src/main/java/org/kframework/kil/StringSentence.java
radumereuta, line 19: I only need the offset in the file so I can correct the location information in the second stage of parsing.
If you want I could store that in a `Source` object, but I'm not sure how to properly determine the end location of the sentence. I would just duplicate the end position to match the start since I only need the beginning of the string.

dwightguth, line 19: no, that's fine.

radumereuta:
@dwightguth please review.

dwightguth:
As I said, you need to find a way to merge the common code path between ParseRulesFilter and ParseConfigsFilter.

kernel/src/main/java/org/kframework/parser/generator/ParseRulesFilter.java
radumereuta, line 104: I did for the new parser code, but I don't want to go in the SDF part. That part should just be removed.

PR #1242, opened 2014-12-05
debugger substitution feature

msaxena2:
Update to the current debugger to show rules applied, and the substitution at every transition. Do not review yet.


java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaSymbolicExecutor.java
```diff
@@ -171,4 +175,48 @@ private ConstrainedTerm javaKILRun(org.kframework.kil.Term cfg, int bound) {
     public SymbolicRewriter getSymbolicRewriter() {
         return symbolicRewriter.get();
     }
+
+    private ConstrainedDebugResult javaKILDebugStep(org.kframework.kil.Term originalState) throws KRunExecutionException {
```
dwightguth, line 179: this doesn't look right to me. Can you explain to me why you were unable to use javaKILRun?

java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaSymbolicExecutor.java
```diff
@@ -171,4 +175,48 @@ private ConstrainedTerm javaKILRun(org.kframework.kil.Term cfg, int bound) {
...
+            return getSymbolicRewriter().debugStep(constrainedTerm);
+        }
+
+    }
+
+
+    private KRunDebuggerResult transformDebugResult(ConstrainedDebugResult debugResult) {
+        org.kframework.kil.Term kilTerm = (org.kframework.kil.Term) debugResult.getSteppedState().term().accept(
+                new BackendJavaKILtoKILTransformer(context));
+        System.out.println("term transform successful");
```
dwightguth, line 197: you should get in the habit of making sure that code like this is removed before you request code review.

kernel/src/main/java/org/kframework/krun/api/DebugTransition.java
```diff
@@ -0,0 +1,42 @@
...
+import org.kframework.kil.Term;
+import org.kframework.kil.Variable;
+
+import java.util.Map;
+
+/**
+ * Represents a transition in the debugger, contains extra information
+ * about substitutions, apart from the rules.
+ */
+public class DebugTransition {
```
dwightguth, line 14: this is wrong. If a transition needs a substitution, add it to Transition, don't create a wrapper class.

java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedDebugResult.java
```diff
@@ -0,0 +1,46 @@
...
+
+package org.kframework.backend.java.kil;
+
+import java.util.Map;
+
+/**
+ *  Debugger Result object to be used in the symbolic rewriter.
+ *  Internal Representation to be used in the java rewrite engine.
+ */
+public class ConstrainedDebugResult {
```
dwightguth, line 11: I would call this RewriteStep

java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaSymbolicExecutor.java
```diff
@@ -171,4 +175,48 @@ private ConstrainedTerm javaKILRun(org.kframework.kil.Term cfg, int bound) {
...
+        KRunState kilState = new KRunState(kilTerm, counter);
+
+        org.kframework.kil.Rule kilRule = (org.kframework.kil.Rule) debugResult.getRule().accept(
+                new BackendJavaKILtoKILTransformer(context));
+        System.out.println("rule transfrom successful");
```
dwightguth, line 202: same here as above

kernel/src/main/java/org/kframework/krun/api/ExecutorDebugger.java
```diff
@@ -29,12 +32,14 @@
...
 import edu.uci.ics.jung.graph.util.Pair;
 
+import java.util.Map;
 import java.util.Map.Entry;
 
 public class ExecutorDebugger implements Debugger {
     private Integer currentState;
     private KRunGraph graph;
     private BidiMap<Integer, KRunState> states;
+    private Graph<KRunState, Map<Variable, Term>> substGraph;
```
dwightguth, line 42: note here what happens as a result of your design decision. By not adding the substitution to Transition, you end up having to propagate the change forward instead of localizing it. This field is unnecessay; if you added the substitution to transition, you would get it for free from KRunGraph.

kernel/src/main/java/org/kframework/krun/api/KRunDebugGraph.java
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
+
```
dwightguth, line 2: this class should also be unnecessary.

kernel/src/main/java/org/kframework/krun/api/KRunDebuggerResult.java
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
```
dwightguth, line 1: something feels funny to me here. We have the exact same functionality on both sides of the KIL vs JavaKIL divide. I think what you need is an opaque interface which defines getters for the Transition and KRunState, and an implementation of that interface returned by the java backend which lazily evaluates back to frontend KIL only those states for which it was explicitly requested.

kernel/src/main/java/org/kframework/krun/tools/Debugger.java
```diff
@@ -49,8 +49,8 @@
     public abstract KRunGraph getGraph();
 
     /**
-     * Reset the debugger's configuration to the graph specified by the parameter
-     */
+    * Reset the debugger's configuration to the graph specified by the parameter
```
dwightguth, line 52: the alignment was right before, you should revert this file.

kernel/src/main/java/org/kframework/krun/tools/Executor.java
```diff
@@ -76,6 +77,17 @@ the resulting term after executing the specified number of steps (or fewer if no
...
 
+    /**
+    Execute a term in debugger mode. Execution only happens for a single step.
+    @param originalState The initial configuration to step from
+    @exception KRunExecutionException Thrown if the backend fails to successfully execute the
+    term
+    @return An object containing information about the step operation, mainly the stepped state,
+    the transition (rule applied) and the substitution
+    */
+    public abstract KRunResult<KRunDebuggerResult> debugStep(Term originalState) throws KRunExecutionException;
```
dwightguth, line 88: I like the idea of having a step function that returns a RewriteStep object, but you should refactor it to be the interface for the existing step function. We don't actually need an extra method in the API.

dwightguth:
I note that even when I check out your branch and use the debugger, it throws exceptions when I run commands. I get the feeling you never tested this, because some of the errors I get are immediately apparent consequences of the code that you wrote, which testing should have detected.

In general, you should always test the code yourself manually in addition to relying on the test suite to tell you what does or does not work. For example, the debugger is not covered by tests, which is why the tests pass even though your actual change is broken. I won't merge this until at least the code you are running has been tested and operates successfully.

dwightguth:
This looks like a step in the right direction, in the sense that it is a much more useful diff which gives us something valuable. But there are still issues with ExectorDebugger crashing on the java backend that need to be addressed.

That said, a couple changes to the API are desirable.
1. No debugStep method
2. signature of step becomes `KRunResult<RewriteStep> step(Term cfg, int steps)`
3. KRunDebuggerResult becomes the RewriteStep interface as follows:

```
public interface RewriteStep {
  KRunState resultState();
  Transition stepTaken();
}
```
1. Transition gains new field for substitution, and replaces DebugTransition
2. ConstrainedDebugResult becomes JavaRewriteStep class which implements RewriteStep and lazily calls BackendJavaKILToKILTransformer

PR #1244, opened 2014-12-08
fixed kool example

grosu:


grosu:
Alright, I fixed the print issue the simplest possible way: I modified the example to be deterministic.  I checked it is deterministic using --search and --transition computational, both with the Maude and with the Java backend.  speaking of which, the Java backend search is SO MUCH faster than Maude's, about 10 times faster!  Good work, guys, I am very impressed!  Let us remove the Maude backend from the basic K distribution asap, so that we can focus on the new engine.

PR #1252, opened 2014-12-11
Keep location information of KItem in Java backend

daejunpark:
for the Jenkins test

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinVisitorOperations.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinVisitorOperations.java
@@ -82,7 +82,8 @@ public class BuiltinVisitorOperations {
             term = KItem.of(
                     visitLabel,
                     KList.concatenate(visitParams),
-                    context);
+                    context,
+                    term.getSource(), term.getLocation());
             return ((KItem) term).evaluateFunction(true/*to be safe*/, context);
         }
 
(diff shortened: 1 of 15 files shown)


dwightguth:
Can you explain what in particular you want to use this information for?

andreistefanescu:
lgtm

daejunpark:
I need this information for measuring semantics test coverage in the term-level granularity.

Suppose that we have the following K definition:

```
...
PseudoCode ::= "If" K "then" K "else" K
...
rule a1(x) => If x then b(x) else c1(x)
rule a2(x) => If x then b(x) else c2(x)
...
```

Currently, what we can measure is whether the rule `a1(x)` is applied or not, but we cannot measure which branch (i.e., `b(x)` or `c1(x)`) is taken. Also, if we have the same terms in multiple places (i.e., `b(x)`), we need to distinguish it based on where it came from.

So, my solution for the problem is to maintain location information for term-level, which allows to measure location information of all terms that have been in the top of the K cell. For now, I think `KItem` is sufficient for my purpose.

daejunpark:
@yilongli could you please review this?

grosu:
This is needed to evaluate the semantic coverage of a test suite, i.e, how well a set of programs covers the semantics.  Daejun crafted this by hand for his JavaScript semantics, and he found several behaviors which were undefined by the Ecmascript test-suite.  Interestingly, those turned out to be mis-implemented by browsers, so he found bugs in Chrome, IE, Firefox, Safari.  So we really want this feature in the K framework.

Grigore

yilongli:
Lgtm

PR #1255, opened 2014-12-12
remove maude backend and make java the default

dwightguth:
**please review**


dwightguth:
**DO NOT MERGE**. This pull request needs to be merged as part of the release process for K 3.5, so that all changes up to this diff are part of K 3.5 and this diff is part of K 3.6.

@yilongli please review changes to java backend and java code

@grosu please review changes to documentation files and tutorial

@radumereuta please review changes to parser and prelude

@traiansf please review changes to agent

@yilongli @radumereuta please review everything else.

@cos you're welcome to comment as well if you like. You might be interested in seeing what I did to deal with productions that declare KLabel constants.

radumereuta:
I'm about to board my flight to Europe, so the best I can do is Thursday.

yilongli:
The changes to the java backend look fine to me.

traiansf:
changes to agent look good to me.  Does quote/unquote work now with the java backend?

kernel/src/main/java/org/kframework/kil/Production.java
```diff
@@ -215,6 +216,21 @@ public int getArity() {
         return arity;
     }
 
+    public int getArityOfKItem() {
```
cos, line 238: could you add some javadoc explaining the distinction between this method and `getArity`?

kernel/src/main/java/org/kframework/kil/Production.java
```diff
@@ -391,4 +407,14 @@ public void setBinderMap(Multimap<Integer, Integer> binderMap) {
     public void setChildren(List<ProductionItem> children, Enum<?> _void) {
         this.items = children;
     }
+
+    public String getKLabelOfKItem() {
```
cos, line 437: and between this method and the simple `getKLabel`

grosu:
Looks excellent to me!  I am so glad to see this pull request.

k-distribution/include/builtins/model-checker.k
```diff
@@ -2,7 +2,6 @@
 require "bool.k"
 
 module LTL-SYNTAX-HOOKS
```
radumereuta, line 4: Is this file still needed? I think this worked only with the maude backend.

k-distribution/include/io/uris.k
```diff
@@ -1,94 +0,0 @@
...
-  rule #string2klist(K:K) => K when isString(K) =/=K true
-
-  syntax KList ::= #statTypes(KList) [function]
-  rule #statTypes(S1:String,,S2:String,,S3:String,,S4:String,,S5:String,,S6:String,,S7:String,,S8:String,,S9:String,,S10:String,,S11:String,,S12:String) => String2Int(S1),,String2Int(S2),,String2Int(S3),,String2Bool(S4),,String2Bool(S5),,String2Bool(S6),,String2Int(S7),,String2Int(S8),,String2Int(S9),,String2Int(S10),,String2Int(S11),,String2Int(S12)
-  rule #statTypes(K:K) => K
-
-// ------------------------------
-// Dynamic parsing
-
-  rule #parse(StringToParse:String, Sort:String) => 'String2K(#tcpRequest("parse\x01" +String Sort +String "\x01" +String StringToParse +String "\x01"))::K
```
radumereuta, line 78: I haven't gotten to the end of the diff yet, but what happens to this?

k-distribution/include/builtins/model-checker.k
dwightguth, line 4: I would like to keep it; it defines the LTL API that Traian's work on model checking will have to support. We can make adjustments to that API of course, but it needs to stick around as a base, otherwise we'll have to redo everything from scratch.

k-distribution/include/io/uris.k
dwightguth, line 78: It still works; they are implemented as hooks in the java engine. It is only the janky tcp mechanism for communicating with maude that has been affected. There are a few hooks that were implemented in maude only, but these can be reimplemented as needed.

kernel/src/main/java/org/kframework/kil/Production.java
```diff
@@ -169,13 +170,19 @@ public String getLabel() {
...
+     * KItem (ie, is a function or a constructor).
+     * @return
+     */
     public String getKLabel() {
         String klabel = getAttribute("klabel");
         if (klabel == null && isSyntacticSubsort()) {
             return null;
         } else if (klabel == null) {
             if (sort.equals(Sort.KLABEL) && getArity() == 0)
-                klabel = getPrefixLabel();
```
radumereuta, line 178: I don't think `getPrefixLabel` should be integrated into `getKLabel`. That function was created separately only because of maude.

kernel/src/main/java/org/kframework/kil/Syntax.java
```diff
@@ -90,8 +90,9 @@ public String toString() {
             return lbls;
         for (PriorityBlock pb : priorityBlocks) {
             for (Production prod : pb.getProductions()) {
-                if (prod.getSort().isComputationSort() || prod.getSort().equals(Sort.KLABEL) && prod.isConstant())
-                    lbls.add(prod.getKLabel());
+                if (prod.getSort().isComputationSort() || prod.getSort().equals(Sort.KLABEL) && prod.isConstant()) {
```
radumereuta, line 93: `isConstant` in the future is going to be true iff it has the attribute `token`.
Is this condition still true with this information?

radumereuta:
lgtm

kernel/src/main/java/org/kframework/kil/Production.java
dwightguth, line 178: Note I am actually removing a call to getPrefixLabel here. The issue is that `syntax KLabel ::= "foo"` doesn't declare a construct whose KLabel is "foo", it's declaring a KLabel /itself/, so returning "foo" from this method leads to incorrect assumptions in the code later on, such as that the arity of a production is always equal to the arity of KItems whose KLabel comes from the KLabel of that production. In any event, this is all temporary hackiness in the implementation. I believe that in kore, there should never be any productions whose return sort is "KLabel". KLabel declarations like the one above in full K should be syntactic sugar for " `syntax KItem ::= foo(KList)` or `syntax KItem ::= foo(K, K)` depending on whether an arity has been declared. And only KLabels with an arity can heat/cool.

kernel/src/main/java/org/kframework/kil/Syntax.java
dwightguth, line 93: See above: I believe this method should eventually go away.

kernel/src/main/java/org/kframework/kil/Production.java
dwightguth, line 178: @cos @grosu you are the kore gatekeepers, does this sound like an acceptable solution? In particular it's the invariant mentioned in the above post that I would like to guarantee for all KItems.

grosu, line 178: LGTM (and yes, this code will change soon, hopefully, because this is just a deconstructor for a KORE syntactic constructor; the plan is that the KORE library will provide implementations for all the constructors and the associated deconstructors).

Syntactically, KORE does not restrict you to declare `syntax KLabel ::= "foo"`.  But we are going to have all sorts of additional restrictions that will be checked and enforced after parsing.  Not sure yet whether this will be one of them or not, but we do not have to decide it now.

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

PR #1257, opened 2014-12-12
some fixes to the way klabels are heated

dwightguth:
**do not review yet**


dwightguth:
Never mind, this is actually inteleaved more than I thought with the kill maude diff, I'm just going to put them together.

PR #1264, opened 2014-12-19
feature/new parser on kore

cos:
**please review** @dwightguth 

A bit of context:
take textual full K outer definition ( with KAST inner)
->
KIL outer + bubbles
-> via KIL-to-KORE
KORE outer + bubble
->
parse KAST bubbles with new parser using a KORE outer definition of KAST (from kore.k)
->
complete KORE :-)

@radumereuta, this is the work we did last week. I fixed some tests and cleaned up a bit. have a look and adjust as you see fit.


cos:
Jenkins, test this please.

cos:
this seems like a failure of the ktest system

kernel/src/main/java/org/kframework/parser/concrete2kore/TreeCleanerVisitor.java
```diff
@@ -0,0 +1,39 @@
...
+import org.kframework.parser.TermCons;
+import org.kframework.parser.Transformer;
+import org.kframework.utils.errorsystem.ParseFailedException;
+
+import scala.util.Either;
+
+/**
+ * Remove parsing artifacts such as single element ambiguities.
+ */
+// TODO: simplify by removing leftover "null" checks and recheck this code
```
radumereuta, line 17: You can remove this comment.

radumereuta:
Well... I was sitting behind you when you wrote this code... I can't tell you more than that.
You should ask someone else to have a look now.

cos:
I will ask for review from someone else, but this is also your pull request, so you may want to make sure you agree with everything before we show it off.

radumereuta:
All I can say right now is lgtm.
We'll see in the future what else is needed.

k-distribution/src/test/java/org/kframework/kore/convertors/TestParserOnKORE.java
```diff
@@ -0,0 +1,95 @@
...
+            Definition kilDefinitionOfKORE = foo.definition;
+            kilDefinitionOfKORE.setMainModule("K");
+            org.kframework.kore.outer.Definition definitionOfKORE = kilToKore1.apply(kilDefinitionOfKORE);
+            Module kastModule = definitionOfKORE.getModule("K").get();
+
+            kastGrammar = KSyntax2GrammarStatesFilter.getGrammar(kastModule);
+        }
+
+        protected Module parseBubbles(Module m) {
+            Set<Module> newImports = stream(m.imports()).map(this::parseBubbles).collect(Collectors.toSet());
```
dwightguth: where is this stream() coming from?

kernel/src/main/java/org/kframework/parser/concrete2kore/Grammar.java
```diff
@@ -0,0 +1,576 @@
+// Copyright (c) 2014 K Team. All Rights Reserved.
```
dwightguth: This seems to be duplicated code. That's a recipe for disaster given that both paths are likely to need updating. We need to come up with a way to consolidate these two code paths without checking in this duplicated code.

kore/src/main/scala/org/kframework/POSet.scala
```diff
@@ -0,0 +1,76 @@
+package org.kframework
```
grosu, line 1: Why is this in Scala?  @cos, let me remind you our deal, which was already a stretch: we can only use Scala for the kore data-structures, and for getters and setters for them.  Maybe also the base classes that transformers and visitors implement, but basically nothing that normal developers need to see.  Please understand that you work in a team here, and the team had its rules and language set well before you joined the team.  And that language is Java and _not_ Scala.  Also, if you need to rewrite existing code that is not related to the above, do it by actually rewriting that existing code, not by duplicating it or creating another version of it.

kore/src/main/scala/org/kframework/parser/TreeNodesToKORE.scala
```diff
@@ -0,0 +1,18 @@
+package org.kframework.parser
```
grosu, line 1: Same here.

k-distribution/src/test/java/org/kframework/kore/convertors/TestParserOnKORE.java
dwightguth: oh never mind I see, it's coming from the Collections scala static class. Is it at all possible for us to use the Java "default" keyword on the interfaces instead? That would make for a nicer syntax.

cos: not really for these classes as we cannot extend them. Remember our discussion regarding implementation choices for the collection data structures. I went with the simple solution of leaving the pure Scala classes underneath and simply gave them a `stream()` and `iterable()` through `Collections` (there is also an `immutable()` method taking a mutable Java class and returning a Scala immutable class, and stream specialized stream collectors). This covers all basic needs and keeps things immutable and safe. I could implement our own Set and List classes so that we have better control over the interface, but it wasn't a priority. The only change would be changing the calling syntax you noticed, `.stream()` vs `stream(...)`, which is simple and mechanical and can be done at any later point in time.

kernel/src/main/java/org/kframework/parser/concrete2kore/Grammar.java
cos: Yes, we choose to duplicate the `concrete2` package. I don't think trying to solve this by abstraction is worth it. The `concrete2` package working over KIL is meant to be removed once we have a good implementation based on Radu's small data structures. Furthermore, by doing this we were able to but out large swats of code which were no longer relevant.

If we are to merge these, we'll basically have to either clean up the KIL classes to conform to a nicer smaller interface, or muddy up KORE and Radu's classes to conform to KIL interfaces.

kore/src/main/scala/org/kframework/POSet.scala
cos, line 1: I follow exactly this rule, don't write in Scala things that other developers need to see/tweak. This is a POSet implementation. Once implemented, nobody really needs to touch it. They can simply use it by instantiating the class with their relations. Furthermore, it only has a constructor and handful of public-facing methods (less then, greater then, in a relation, least upper bound). I implemented it in about two hours, including tests. An alternate implementation in Java would not be that hard to do and could be substituted without breaking any code.

Regarding duplication, I've addressed it above. I don't see a good solution for it.

grosu, line 1: @cos, you are pushing the envelope here.  Nothing from what you said is visible there, there is no javadoc or useful comments for users.  Just adhoc Scala code, which could have just as well be Java.  I don't know about others, but I am personally tired of having this discussion over and over again.  This slows us down and we've got lots of things to do.  You got the maximum you can get wrt Scala for now.  We may re-evaluate the situation after we have a clean, KORE based implementation that works well.  Maybe K 5.0.

cos, line 1: I agree I should add more javadoc and comments here. 

> Just adhoc Scala code

I take pride in my code. I'm not claiming perfection but, aside from the lack of documentation, could you please explain what is ad hoc about this code? 

> which could have just as well be Java

Actually, not easily. We needed the Poset for the subsorts exposed by a `Module`. We initially tried to use the existing poset implementation, but it is not that easy. The problem is the kore classes are separated from the rest of the code base. The kernel module depends on the kore module. This means that we cannot use any existing Java code from within the KORE classes. They can be used by other classes but not the other way around. So, the options were to either implement a POSet, or create an interface and add a poset factory as a parameter to the Module class. The latter would basically means adding a lot of abstraction machinery for a very simple thing. So I chose to simply implement the POSet by the KORE class. Other classes in KORE will likely need it (we'll need also a labels POSet soon) so this is keeping complexity down overall.

> I don't know about others, but I am personally tired of having this discussion over and over again. This slows us down and we've got lots of things to do. You got the maximum you can get wrt Scala for now. 

I am not asking for anything more regarding Scala. I'm just trying to implement things in the fastest cleanest manner possible, while making sure that anything I do will be very usable for developers trying to use the KORE classes from Java. I'm not taking here about names, which can be easily changed (e.g., for the POSet we'll need aliases for the <, >, and ~ methods as Java doesn't allow method names like that), but higher-level decisions regarding how things are put together.

So far this means I've implemented in Scala everything related to KORE classes and their supporting infrastructure (visitors, underlying data structures, etc.). I use Java for everything that uses the KORE classes for a specific purpose, e.g., compiler passes.

Is this different from our understanding? I'm really doing my best here and these discussions slow me down also. 

> We may re-evaluate the situation after we have a clean, KORE based implementation that works well. Maybe K 5.0.

I was hoping K 5.0 would be implemented mostly in K.

In any case, if you prefer a specific different solution for this specific issue, let me know and I'll implement it.

dwightguth, line 1: Here's my suggested solution: move the preexisting Poset class into the
kore package.

cos, line 1: I thought about this but didn't really want to create a precedent. Initially I didn't like that the KORE classes cannot depend on existing code, but it may be for the best in order to not muddy the waters. 

The existing POSet implementation is pretty clean but it is mutable. As I need the poset as a field of an immutable KORE class, I wouldn't plug it in directly. One way would be to use it to compute the transitive closure, then wrap it in an immutable data structure. The the Java POSet object will only live during the computation. I didn't think it is worth it considering the simplicity of doing a transitive closure. Still, yes, it is workable, and I can take this path if it pleases people. Let me know.

dwightguth, line 1: Okay, yeah, I guess I can follow this as having the desire to create an
immutable poset. @grosu, I am fine with this use of Scala as explained to
me.

cos:
@radumereuta, I just talked to @dwightguth about getting rid of the duplication. Do you think it would be possible to use the new code path exclusively sooner rather then later? We could even convert from the new frontend data structures to KIL so you could use the current post-processing steps when playing with the concrete bubble parser. The issue mostly concerns the new parser, so it shouldn't have any impact on the main pipeline, right?

cos:
@dwightguth, we've removed the _concrete2_. could you have a look?

cos:
also, I piggybacked some small changes to convertors -- they've been implemented over time in parallel and it would be quite hard to rip them apart

kore/src/main/scala/org/kframework/kore/Attributes.scala
```diff
@@ -49,6 +49,17 @@ case class Attributes(att: Set[K] = Set()) extends Collection[K] with Indexed[St
...
 object Attributes {
   def apply(ks: K*) = new Attributes(ks.toSet)
+
+  val classFromUp = "classType"
+}
+
+object Location {
+  import KORE._
+
+  def apply(startLine: Int, startColumn: Int, endLine: Int, endColumn: Int) = {
```
dwightguth, line 59: if we're going to be redoing this from scratch, the location object should, in addition to line and column numbers, store string indices.

cos, line 59: what do you mean by string index?

dwightguth, line 59: I mean offset in the string that it was parsed at. ie:

parse("foo")
"foo".charAt(0) == 'f'
"foo".charAt(1) == 'o'

This isn't a conversion we need to make right away, it's just info I think the new location class should probably have

PR #1265, opened 2014-12-19
update changelog and poms for release version 3.5

dwightguth:
I ended up doing all three sections of the changelog diffs because I just wanted to get through it and I got in early today. @yilongli @traiansf if you have completed any portion of your versions, please check to make sure that I am not missing anything that you put in.

Once this is merged I will start the release process.


yilongli:
Thanks, Dwight! I can confirm that you haven't missed anything in the merged pull questions I have gone through. Only better.

PR #1266, opened 2014-12-23
Performance improvements to krun

dwightguth:
- stream outputs directly to and from files in many cases to save memory usage
- remove old --output raw so that the java backend does not invoke the unparser unnecessarily.
- other minor performance enhancements

**please review**


dwightguth:
@yilongli please review

yilongli:
lgtm

grosu:
@dwightguth if there are any visible performance improvements for users of the K framework, can you please briefly let us know what kind of improvements and how much?

dwightguth:
It's not that noticeable for most users of the semantics. Mosly the case I am improving is the case where users have very very very large configurations, in which case some of the changes here are meant to speed up the semantics and/or prevent crashes.

PR #1267, opened 2014-12-23
add rule auditing feature

dwightguth:
Feature shows a trace for a particular rewrite step showing why rules or rule did not apply.

@yilongli please review


yilongli:
lgtm

PR #1273, opened 2014-12-31
add new isInfinite for floats

dwightguth:
@yilongli please review


k-distribution/include/builtins/float.k
```diff
@@ -75,4 +76,6 @@ module FLOAT-HOOKS
 
   rule sqrtFloat(F:Float) => rootFloat(F, 2)
 
+  rule isInfinite(F:Float) => F >Float maxValueFloat(precisionFloat(F), exponentBitsFloat(F)) orBool F <Float --Float maxValueFloat(precisionFloat(F), exponentBitsFloat(F))
+
```
yilongli, line 80: If you are sure that `+Inf >Float maxValueFloat` indeed returns true then I am fine with this pull request.

dwightguth, line 80: cf the test I wrote, which passes.

PR #1274, opened 2014-12-31
various minor bug fixes and optimizxations to the rewriter

dwightguth:
Comprises:
- add auditing for anywhere rules
- fix race conditions associated with get/put critical pairs in static caches
- disable user time (apparently JMX is really slow)
- fix for bug with anywhere rules not applying because the eval instruction was not generated (due to Definition.anywhereRules() not being set yet at compilation time)
- fix to use Reader instead of String for parts of the parser in order to avoid creating large arrays in memory

@yilongli please review


yilongli:
Looks much better

PR #1275, opened 2015-01-02
Refactoring for Debugger's Substitution Feature

msaxena2:
Refactoring now allows KRun to return the graph of execution, if needed. The graph will be needed in the debugger, and can also be used for other purposes. Changes introduced:
1) The original step and resume functions have now been modified to support KRun graphs.
2) KRun will now return a RewriteRelation object, instead of the original KRun State, which has the graph as an optional parameter.

 Please do not merge yet.


dwightguth:
Jenkins: test this please

java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedRewriteRelation.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.util.Optional;
+
+/**
+ * Backend Representation of the Rewrite Relation represented in the
+ * Frontend by the rewrite relation class.
+ * Used by the Symbolic Rewriter, and then converted to the generic notation by
+ * an implementation of the executor interface.
+ */
+public class ConstrainedRewriteRelation {
```
dwightguth, line 11: so, the idea as I understand it is that instead of doing a conversion of everything in the backend form to everything in KIL, you make KRunState an abstract class in which getRawResult is abstract. And then the implementation would be a java backend class that lazily converts to the frontend only the values that it actually needs.

java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaSymbolicExecutor.java
```diff
@@ -153,12 +138,87 @@ public SearchResults search(
         return retval;
     }
 
+
+    private KRunGraph genericGraphTransformer(ConstrainedExecutionGraph constrainedGraph) {
```
dwightguth, line 142: as per above, this method should not be necessary.

java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaSymbolicExecutor.java
```diff
@@ -153,12 +138,87 @@ public SearchResults search(
...
+            returnGraph.addVertex(state1);
+            returnGraph.addVertex(state2);
+            returnGraph.addEdge(genericTransition, state1, state2);
+
+        }
+        return returnGraph;
+    }
+
+    private RewriteRelation genericKilTransformer(ConstrainedRewriteRelation constrainedRewriteRelation) {
+        /* processing the final Term */
```
dwightguth, line 183: same here

java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaSymbolicExecutor.java
```diff
@@ -153,12 +138,87 @@ public SearchResults search(
...
+        return returnRelation;
+    }
+
+    private ConstrainedRewriteRelation internalTraceStep(org.kframework.kil.Term cfg, int steps, boolean computeGraph)
+            throws KRunExecutionException{
+        Term term = kilTransformer.transformAndEval(cfg);
+        TermContext termContext = TermContext.of(globalContext);
+        if (javaOptions.patternMatching) {
+            if (computeGraph) {
+                throw new KRunExecutionException("compute Graph with pattern matching not yet implemented!");
```
dwightguth, line 204: please use the KExceptionManager here; KRunExecutionException is old deprecated code.

java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedRewriteRelation.java
grosu, line 11: What @dwightguth just said is also consistent with my understanding, as described in the PPT slides that I send on the k-list yesterday.  It is this sort of things that I had in mind when I recommended that Manasvi stops implementing anything new for a few days and instead spend some time on designing the API interfaces mentioned in the PPT slides.  I would actually like to see and actual (Java) interface fully defined first, and then everything Manasvi does from there on to be to implement that interface.  And I think that we should all participate in the design of the interface.

PR #1295, opened 2015-01-09
Refactoring - Lazy conversion of Backend Specific Kil to Generic Kil and Graph option in step/resume

msaxena2:
#### Major Highlights:
- Wrapper Classes which allow for lazy evaluation from Backend Kil to Generic Kil.
- Step/Resume now have an option to compute execution graphs.
- Executor now returns a Rewrite Relation object, containing an optional graph parameter, if compute Graph was specified as true in step/resume.
- Other supporting structures around lazy evaluation/ execution graphs.

Please Do not merge yet.


dwightguth:
This is not at all the design that I had in mind. The idea was to make KRunState and Transition be abstract classes in which the getters which returned the KIL terms were abstract. Then the java backend would be able to exist largely unchanged except for the very minor alteration which accumulates the graph as you rewrite, and a quite small change to the way KRunStates and Transitions are constructed in JavaSymbolicExecutor. The only new classes this diff should have to introduce are RewriteRelation, JavaBackendKRunState, and JavaBackendTransition. The later two classes should extend KRunState and Transition and implement only one method, which is respectively getRawResult and getRule. This should be able to be done in a fairly small diff, rather than having to introduce a number of new classes and constructs, not to mention significantly altering the interface of the rewriter.

dwightguth:
Please find the time to talk to me to make sure you understand the class design before we proceed any further. I'd like to get past this circling we are doing, and to do that it's necessary that you have the correct mental model for how the pieces of the system should interact.

grosu:
I insist that we need to nail down the two interfaces ASAP.  One is the rewrite engine interface and the other is the debugger interface.  Then different backends can be connected with the generic K debugger provided they implement the rewrite engine interface (e.g., Maude, or OCAML, or the forthcoming LLVM one), and various observers/shells for the debugger (e.g., textual, graphical, IDE, etc.) can co-exist provided that they implement the interface of the debugger.  Probably Dwight is saying the same thing, but I just want to re-iterate the importance of nailing down these interfaces.  It is like making a plan before a complex operation.  It is a must, to avoid mistakes and wasted time.

Grigore

dwightguth:
So, what I'm talking about is not actually the conceptual design; the
conceptual design in this diff is pretty much about what I am aiming for,
and it achieves all the types of separation you are talking about (in fact,
the lazy interpretation of backend terms is the last piec; we already have
a good rewrite engine and debugging api that supports the potential for
multiple backends and frontends).

The problem with this diff is at a lower level than that: it has bad _class
design_. Essentially, instead of being a simple, non-intrusive change which
provides a minor extension of capabilities beyond what is already there, it
upends the entire interface for rewriting in the java backend and uses half
a dozen new classes and hundreds of lines of poorly designed code to do the
same work that is already being done by the system we have now. It's not
that the design itself is wrong; the design of the implementation is just
all wrong.

I am working with Manasvi to make sure we get it right this time before any
further coding is done. He is going to submit a class design for feedback
to me tomorrow and we will iterate feedback until it matches what I have in
mind. I agree, enough time wasting.

grosu:
Alright, thanks for the clarification, Dwight. Let us get this thing merged first, and then we will re-discuss the big picture.  To me, the incorporation of KORE everywhere and the debugger interface are two of the most important features of K that I'd like to see incorporated as soon as possible.  However, considering that I start teaching CS422 in 10 days or so, making the most of our existing debugging capability is the best we can do.

Thanks,
Grigore

Issue #1256, opened 2014-12-12
Incremental kompile provides incorrect location information


daejunpark:
The incremental kompile provides incorrect location information. It seems that each line is inconsistently shifted down, while the column keeps correct.

For example, for the following K definition, `c.k`:

```
  1 module C
  2 
  3 configuration <T>
  4   <k> $PGM:Cmd </k>
  5 </T>
  6 
  7 syntax Cmd ::= "x" "(" ")"
  8              | "y" "(" ")"
  9 
 10 rule x() => y()
 11 
 12 endmodule
```

when we run the following two consecutive kompilations (i.e., the second is the incremental kompilation, because nothing is changed.):

```
$ kompile --no-prelude --backend java c.k
$ kompile --no-prelude --backend java c.k
```

The result of the second run provides incorrect location information of the rule at Line 10, while the first run provides the correct one. In `org.kframework.backend.FirstStep#compile`, the argument `def` has `location` field of the rule that refers to Line 14 (non-exsiting line).

andreistefanescu:
Amarin also encountered the same issue.

PR #1300, opened 2015-01-14
Fixing issue #1256, line number gets moved during recompilation because ...

radumereuta:
...of cache. Just had to switch some parameters into the correct position.

@dwightguth please review.


dwightguth:
lgtm. Merge if tests pass.

PR #1301, opened 2015-01-15
Significantly speed up --output pretty, move some functionality to new --output sound

dwightguth:
**please review**


dwightguth:
Jenkins: test this please

dwightguth:
@yilongli @radumereuta please review

yilongli:
Lgtm, though I am not very familiar with this part of the code.

kernel/src/test/java/org/kframework/backend/kore/ToKAppTransformerTest.java
```diff
@@ -37,7 +40,13 @@
...
+        transformer = new ToKAppTransformer(context, new Comparator<Term>() {
+
+            @Override
+            public int compare(Term o1, Term o2) {
+                return o1.toString().compareTo(o2.toString());
+            }
+        });
         mapSort = new DataStructureSort("Map", Sort.MAP, "'_Map_", "'_|->_", "'.Map", Maps.<String, String>newHashMap());
         listSort = new DataStructureSort("List", Sort.LIST, "'_List_", "'ListItem", "'.List", Maps.<String, String>newHashMap());
         setSort = new DataStructureSort("Set", Sort.SET, "'_Set_", "'SetItem", "'.Set", Maps.<String, String>newHashMap());
```
radumereuta, line 52: I think these klabel strings should be constants declared somewhere.
If we suddenly switch to the new KORE format, they have to change.
Is this the only place where we use them?

kernel/src/test/java/org/kframework/backend/unparser/UnparserFilterNewTest.java
```diff
@@ -27,10 +26,9 @@ public void setup() {
      */
     @Test
     public void testTrailingUnderscores() {
-        UnparserFilter v = new UnparserFilter(context);
+        Unparser v = new Unparser(context);
         KApp t = KApp.of(KLabelConstant.of("'__"), Variable.getAnonVar(Sort.K), Variable.getAnonVar(Sort.K));
```
radumereuta, line 30: See above.

radumereuta:
lgtm, although I didn't look into much detail.

dwightguth:
I'm inclined to leave your comments as is because they are both in unit tests so it will be very easy to fix them when the time comes because the test will break.

Since Yilong and Radu both feel that they don't know this code very well, I think it might be good to get more eyes on it, just in case. @cos is there any chance you can take a moment to review this as well?

radumereuta:
@dwightguth I could live with that.

kernel/src/main/java/com/davekoelle/AlphanumComparator.java
```diff
@@ -130,4 +129,95 @@ public int compare(String o1, String o2)
 
         return s1Length - s2Length;
     }
+
+    private final int getChunkLength(String s, int marker, int end)
+    {
```
cos: why is this file part of the k codebase? legacy? also, this method doesn't follow style conventions. as well as the ones after.

kernel/src/main/java/org/kframework/backend/unparser/AddBracketsFilter.java
```diff
@@ -250,27 +254,8 @@ private boolean isAtom(Term inner) {
...
-        public boolean getFound() {
-            return found;
-        }
-    }
-
-    private static boolean contains(Term outer, Term inner, org.kframework.kil.loader.Context context) {
-        ContainsVisitor visit = new ContainsVisitor(inner, context);
-        visit.visitNode(outer);
-        return visit.getFound();
+    private static boolean contains(Term outer, Term inner) {
```
cos: why is `contains` an equality now?

kernel/src/main/java/org/kframework/backend/unparser/SortCollections.java
```diff
@@ -0,0 +1,71 @@
...
+import org.kframework.kil.ASTNode;
+import org.kframework.kil.Bag;
+import org.kframework.kil.Cell;
+import org.kframework.kil.ListBuiltin;
+import org.kframework.kil.MapBuiltin;
+import org.kframework.kil.SetBuiltin;
+import org.kframework.kil.Term;
+import org.kframework.kil.loader.Context;
+import org.kframework.kil.visitors.CopyOnWriteTransformer;
+
```
cos, line 16: javadoc

kernel/src/main/java/org/kframework/backend/unparser/Unparser.java
```diff
@@ -0,0 +1,805 @@
...
+import org.kframework.kil.Token;
+import org.kframework.kil.UserList;
+import org.kframework.kil.Variable;
+import org.kframework.kil.loader.Context;
+import org.kframework.kil.visitors.NonCachingVisitor;
+import org.kframework.krun.ColorSetting;
+import org.kframework.utils.ColorUtil;
+
+import com.davekoelle.AlphanumComparator;
+
```
cos, line 67: it's clear what the class does, but the control-flow seems quite involved -- there are several inner classes. documentation would be useful.

kernel/src/main/java/org/kframework/backend/unparser/AddBracketsFilter.java
dwightguth: This was being called in only one place, and that place was causing an increase in asymptotic running time because it was searching the entire AST for a term that we knew was at the top layer. I will inline this method so it's not so confusing.

dwightguth:
Seems to have triggered known transient test failures. Running again.

dwightguth:
Jenkins: test this please

cos:
looks good.

PR #1302, opened 2015-01-16
Refactoring and Option to compute Graph in KRun's Executor Interface

msaxena2:
Please don't not merge yet.
Highlights:
- Changed Executor interface, with option to compute execution Graph.


java-backend/src/main/java/org/kframework/backend/java/util/ConstrainedTransition.java
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2015 K Team. All Rights Reserved.
```
dwightguth: I don't think you need this class. You should be able to just create JavaTransitions and put them directly in the backend's graph.

java-backend/src/main/java/org/kframework/backend/java/util/ConstrainedExecutionGraph.java
```diff
@@ -0,0 +1,13 @@
...
+
+import edu.uci.ics.jung.graph.DirectedOrderedSparseMultigraph;
+import org.kframework.backend.java.kil.ConstrainedTerm;
+
+
+/**
+ * Backend specific execution graph.
+ */
+public class ConstrainedExecutionGraph extends DirectedOrderedSparseMultigraph<ConstrainedTerm,
+        ConstrainedTransition> {
```
dwightguth: this should probably be a graph of JavaKRunState and JavaTransition, and be called JavaKRunGraph

kernel/src/main/java/org/kframework/krun/GenericTransition.java
```diff
@@ -0,0 +1,55 @@
...
+        return rule;
+    }
+
+
+    @Override
+    public Map<Variable, Term> getSubstitution() {
+        return substitution;
+    }
+
+    public static GenericTransition rule(ASTNode rule) {
```
dwightguth: unused method, delete it

kernel/src/main/java/org/kframework/krun/GenericTransition.java
```diff
@@ -0,0 +1,55 @@
...
+        return new GenericTransition(TransitionType.RULE, null, rule, null, null);
+    }
+
+    public static GenericTransition label(String label) {
```
dwightguth: same here

kernel/src/main/java/org/kframework/krun/GenericTransition.java
```diff
@@ -0,0 +1,55 @@
...
+        return new GenericTransition(TransitionType.LABEL, label, null, null, null);
+    }
+
+    public static GenericTransition unlabelled() {
+        return new GenericTransition(TransitionType.UNLABELLED, null, null, null, null);
+    }
+
+    public static GenericTransition deadlock() {
```
dwightguth: and here

dwightguth:
please set your ide to delete unused imports, and go over this code again to make sure they're removed. I see several of them in this diff.

java-backend/src/main/java/org/kframework/backend/java/util/ConstrainedTransition.java
msaxena2: I still think having this class should be considered. The SymbolicRewriter did not deal with KRunStates/Transitions. It returned terms, which where then put into states. Having this class allows us to keep it that way, and have KRunStates/Transitions in the SymbolicExecutor instead of the rewriter, making for a cleaner implementation.

kernel/src/main/java/org/kframework/krun/api/Transition.java
```diff
@@ -29,41 +39,16 @@
```
dwightguth, line 31: please move this field (readString) to GenericTransition, and then delete label (it will be unused after you remove GenericTransition.label()

kernel/src/main/java/org/kframework/krun/GenericTransition.java
```diff
@@ -0,0 +1,55 @@
...
+import org.kframework.krun.api.Transition;
+
+import java.util.Map;
+
+/**
+ * Generic Transition for Rules that have already been converted to generic Kil
+ */
+public class GenericTransition extends Transition {
+
+    public GenericTransition(TransitionType type, String label, ASTNode rule,
```
dwightguth: this should be private probably because it's only called by the static factories in this class.

java-backend/src/main/java/org/kframework/backend/java/util/ConstrainedExecutionGraph.java
msaxena2: Again the same argument applies, I think keeping the States/Transitions in the executor is a better idea.

java-backend/src/main/java/org/kframework/backend/java/util/JavaTransition.java
```diff
@@ -0,0 +1,80 @@
...
+    public JavaTransition(TransitionType type, String label, Rule rule,
+                          Map<org.kframework.backend.java.kil.Variable, org.kframework.backend.java.kil.Term> javaSubs,
+                          String readString, Context context) {
+        super(type, label, null, null, readString);
+        javaRule = rule;
+        this.javaSubs = javaSubs;
+        this.context = context;
+    }
+
+    public JavaTransition(TransitionType type, String label, ASTNode rule, Map<Variable, Term> subs,
```
dwightguth: this should only take two parameters: the rule and the substitution

kernel/src/main/java/org/kframework/krun/GenericTransition.java
msaxena2: All these methods have been copied from the previous Transitions class. Are you sure you want the unused ones deleted?

kernel/src/main/java/org/kframework/krun/api/Transition.java
```diff
@@ -77,6 +62,12 @@ public void setLabel(String label) {
...
     }
 
+    public abstract Map<Variable, Term> getSubstitution();
+
+    public void setSubstitution(Map<Variable, Term> substitution) {
+        this.substitution = substitution;
+    }
+
     public enum TransitionType {
         /**
```
dwightguth, line 58: you should delete label and deadlock from this list.

dwightguth:
This is pretty close now to what I had in mind, there's just a lot of small details. I suggest you simply make the new commits on the same branch and push them: that will add them to this pull request without requiring that you create a completely new pull request.

java-backend/src/main/java/org/kframework/backend/java/util/ConstrainedTransition.java
dwightguth: So, there's arguably not anything cleaner about having an extra layer of glue. Ideally I would like to get rid of the JavaSymbolicExecutor class entirely and make PatternMatchRewriter and SymbolicRewriter implement Executor. Doing things the way you are describing adds to the glue rather than thinning it, which is what the solution I am proposing would be a step towards. And if we follow this to the logical solution that I just described, then the inconsistency in abstraction levels you just described goes away, because we no longer maintain a distinction between the implementation of the kernel interfaces and the implementation of the rewriting logic itself.

java-backend/src/main/java/org/kframework/backend/java/util/ConstrainedExecutionGraph.java
dwightguth: see above. Also note that this is problematic for another reason, which is that the pattern match rewriter doesn't have ConstrainedTerms, nor does your so-called ConstrainedTransition actually have a constraint.  Same with the name of the class.

msaxena2:
Jenkins: test this please

msaxena2:
Jenkins: test this please

msaxena2:
@dwightguth Please review.

java-backend/src/main/java/org/kframework/backend/java/symbolic/JavaSymbolicExecutor.java
```diff
@@ -155,9 +157,9 @@ public SearchResults search(
     }
 
     @Override
-    public KRunState step(org.kframework.kil.Term cfg, int steps)
+    public RewriteRelation step(org.kframework.kil.Term cfg, int steps, boolean computeGraph)
             throws KRunExecutionException {
-        return internalRun(cfg, steps);
+        return javaKILRun(cfg, steps, computeGraph);
```
grosu: I only took a very brief look at the code to see how it looks overall, and I find the method names sometimes confusing, in that their name does not make it clear what they do.  For example, what is `javaKILrun`?

msaxena2: For this particular pull request, I didn't have to add new functions to already existing classes/interfaces, and most of my work revolved around making graphs work with the current interface. For example in this case, the function javaKILRun exists in the current codebase as well, and acts as a private helper function that calls the rewriter. However, I agree that names might be a bit ambiguous and changing them is be a good idea, and an easy fix. My only question is whether you'd like me to change names of functions that I've not written, and are being used in the current code base (given that they've already been peer reviewed)?

grosu: I think there is nothing bad in proposing a new name to an existing function once in a while.  The names, like many other things, may also get obsolete as the system evolves.  Also, some of them may have lurked inthere from ancient times, when we were not that picky about names.  But now we are.  Since people are generally reluctant to writing documentation, the names of functions end up being their main documentation.  Don't worry about long names.  They are in fact good.  What is important is for them to well capture the meaning of the function.  So unless Dwight strongly disagrees, I would suggest that you should feel free to change function names, too, but please make it in separate pull requests, which Dwight can approve or reject quickly.

dwightguth: Yeah, that makes sense. I would suggest that @msaxena2 runs the proposed names by me first though, because the fact that I am more familiar with the project as a whole gives me a  better grasp for what is or is not an appropriate name for something.

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -49,27 +54,48 @@
...
-    public ConstrainedTerm rewrite(ConstrainedTerm constrainedTerm, int bound) {
-        stopwatch.start();
+    public KRunGraph getExecutionGraph() {
+        return executionGraph;
+    }
 
+    public KRunState rewrite(ConstrainedTerm constrainedTerm, int bound, boolean computeGraph) {
+        stopwatch.start();
+        Context context = definition.context();
+        KRunState initialState = new JavaKRunState(constrainedTerm.term(),
```
dwightguth: to make this run as fast as possible, you should declare this outside the if, initialize it to null, and initialize its value inside the if.

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -49,27 +54,48 @@
...
+        if (computeGraph) {
+            executionGraph = new KRunGraph();
+            executionGraph.addVertex(initialState);
+        }
         for (step = 0; step != bound; ++step) {
             /* get the first solution */
             computeRewriteStep(constrainedTerm, 1);
             ConstrainedTerm result = getTransition(0);
             if (result != null) {
+                KRunState finalState = new JavaKRunState(result.term(), context, counter);
```
dwightguth: same here, put as much inside the if as possible so that it runs as fast as possible in the default case.

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -80,7 +106,7 @@ public ConstrainedTerm rewrite(ConstrainedTerm constrainedTerm, int bound) {
             System.err.println("[" + step + ", " + stopwatch + "]");
         }
 
-        return constrainedTerm;
+        return initialState;
```
dwightguth, line 115: in order to make this work, you will probably have to check for null here and create a new KRunState if it is null.

dwightguth:
This looks so good! Thank you! Just one quick thing I am suggesting here that will hopefully make the common case of rewriting marginally faster by avoiding memory overhead. When you make that change, it will be ready to merge.

Awesome!

msaxena2:
Jenkins: test this please

msaxena2:
@dwightguth @grosu  Please review. 

I've also changed the name of the function mentioned in one of the earlier comments. I've tried to come up with a good name, but still feel that we can come up with a better one. If you feel the name is inappropriate, or that a better name is possible, please let me know, and I'll implement the changes as quickly as possible. I'll also make sure to change other confusing  names as I modify more files in the future. If there are other functions that need renaming, please let me know, and I'll be happy to make the changes.

dwightguth:
I'm honestly not super concerned about the name of this function, all things considered. Good names are valuable, but as I said before, I would like to see the SymbolicRewriter and PatternMatchRewriter classes implement the Executor interface directly, so the entire class should go away sooner or later, at which point the name of this function is moot. That said, you have room for improvement in the way you pick names for things, so it'd be good if we sync up on the names for things _before_ the actual code is changed in the future.

I am merging this now. Good work! Let's see if we can get the next diff through faster.

PR #1304, opened 2015-01-22
check hashcode before calling equals on two terms

dwightguth:

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicUnifier.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicUnifier.java
@@ -111,7 +111,7 @@ public class SymbolicUnifier extends AbstractUnifier {
     private void unify(Term term, Term otherTerm) {
         if (term.isGround() && otherTerm.isGround()
                 && term.isNormal() && otherTerm.isNormal()) {
-            if (!term.equals(otherTerm)) {
+            if (term.hashCode() != otherTerm.hashCode() || !term.equals(otherTerm)) {
                 fail(term, otherTerm);
             }
             return;


java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicUnifier.java
```diff
@@ -111,7 +111,7 @@ public boolean symbolicUnify(Term term, Term otherTerm) {
     private void unify(Term term, Term otherTerm) {
         if (term.isGround() && otherTerm.isGround()
                 && term.isNormal() && otherTerm.isNormal()) {
-            if (!term.equals(otherTerm)) {
+            if (term.hashCode() != otherTerm.hashCode() || !term.equals(otherTerm)) {
```
grosu, line 114: Don't you need `&&` instead of `||` ?  And wouldn't it be more effective to move the hashcode test inside the term.equals method?

andreistefanescu, line 114: I also asked @dwightguth to se if he can move the hashcode test inside `equals`.

@grosu `||` is correct.

dwightguth:
After talking to Yilong, I am going to hold off on making this change until I get a chance to do more thorough profiling information on whether this is likely to be a uniform optimization, or if it will slow some cases down.

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicUnifier.java
grosu, line 114: @andreistefanescu Wasn't the intent here to skip the expensive term equality check when the two terms have the same hashcode?  Well, the above still calls the expensive term equals in this case.  Also, couldn't it be possible to have equal terms that have different hashcodes?  In that case the first condition is true and fail is returned, which seems wrong to me.

andreistefanescu, line 114: In java, if two terms are equal, then they must have the same hashcode, but if they have the same hashcode, they don't need to be equal. So if the hashcodes are different you can conclude the terms are different, but if they have the same hashcode, you must make the equals check.

grosu, line 114: > if two terms are equal, then they must have the same hashcode, 

Are you 100% sure?  This seems highly improbable to work well in general, because term equality can be quite tricky, for example equality modulo domain properties.  Take for example, `lambda x . x` and `lambda y . y`.  It looks like the hashcode and equals methods need to be synchronized somehow, and it seems hard, if not impossible, for Java to know how to do that automatically.  Even if we make abstraction of domain properties and only think of how objects are laid out in memory, it should be able to recognize when two graphs (the heap representations of the two objects) are isomorphic, shouldn't it?

andreistefanescu, line 114: Java assumes that if `o1.equals(o2)` then `o1.hashCode() == o2.hashCode()`. It is the developer's responsibility to make sure this assumption holds when overriding `equals` and `hashCode`.

grosu, line 114: And you have done that for Term.equals and Term.hashcode, I presume?  How about the terms lambda x . x and lambda y . y for LAMBDA with binders?  I guess I should look over the code, but I don't have time, so I just want to make sure that we are not doing anything inconsistent here.  If you guys are confident we are fine, then I'm OK with it.

PR #1306, opened 2015-01-23
Add profiling information about functions + fix an OOM for large configurations

dwightguth:
@yilongli please review


dwightguth:
Jenkins: test this please

dwightguth:
I am struggling to get this working because of a piece of this diff that is not really on my critical path anymore now that I have decided I need to bring the size of C semantics configurations down. So I am going to close this and reopen just the piece that provides useful new functionality without bugs. Will get to this later if it's still an issue.

PR #1307, opened 2015-01-23
Improve error message when List is not declared

bmmoore:
A NullPointerException was the only error when using --no-prelude on a definition that didn't include a List sorts. This pull requests changes all places that assume context.dataStructureListSortOf(DataStructureSort.DEFAULT_LIST_SORT) succeeds with calls to a new method that will raise a nice exception if List isn't a recognized sort.


dwightguth:
lgtm. Merge if tests pass

PR #1310, opened 2015-01-26
Move Coq backend to separate maven module

bmmoore:
This moves the Coq backend to a separate maven module, without changing it's behavior.
Some test code in kernel is modified because it used CoqBackend as basically a dummy instance of Backend for setting up injection.


dwightguth:
Awesome, thanks!

PR #1315, opened 2015-01-28
fix bug with latex attributes

dwightguth:
@grosu please review


grosu:
I think it is OK to merge, although I was not able to test it.  Besides the windows problems that I reported yesterday, which prevent me from running `mvn package`, the instructions on how to check out someone else's branch at https://github.com/kframework/k/wiki/Git-FAQ didn't work.  I think they assume that the branch is under the k repo and not in someone's fork, like it is the case here.  Can you @dwightguth please update those instructions, too?

PR #1316, opened 2015-01-28
Small bug fix for graph in rewriter.

msaxena2:
Small fix to set final state in case of finite steps in symbolic rewriter.


msaxena2:
@dwightguth please review

traiansf:
@msaxena2 @dwightguth I have some questions about these pull requests: Can I get the graph of all possible transitions in one step with the code that you're adding?

I need that for my work on the model checker, but I guess it would be useful for a debugger, too, so I was wondering if you implemented that or you plan on implementing that anytime soon.

dwightguth:
When it's ready, you should be able to get a graph from any KRun execution. The code is currently untested though and likely contains bugs.

dlucanu:
This is a good news. I know that it is not easy to estimate about the time it is done, but a clue will be very helpful.

traiansf:
@dwightguth Just to make sure: execution includes search?  And we would not mind being alpha-testers

dwightguth:
search is currently unimplemented. Patches are welcome, or Manasvi will likely get to it when he reaches that point.

PR #1317, opened 2015-01-28
Lazy conversion from backend kil to generic kil for faster step and resume. 

msaxena2:
Faster debugger step and resume due to lazy conversion from backend kil to generic kil.

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/util/JavaKRunState.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/util/JavaKRunState.java
@@ -40,7 +40,8 @@ public class JavaKRunState extends KRunState{
     @Override
     public boolean equals(Object o) {
         if (!(o instanceof JavaKRunState)) {
-            return false;
+            KRunState genericObj = (KRunState) o;
+            genericObj.getRawResult().equals(this.getRawResult());
         }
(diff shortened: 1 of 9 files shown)


java-backend/src/main/java/org/kframework/backend/java/util/JavaKRunState.java
```diff
@@ -40,7 +40,8 @@ public Term getRawResult() {
     @Override
     public boolean equals(Object o) {
         if (!(o instanceof JavaKRunState)) {
-            return false;
+            KRunState genericObj = (KRunState) o;
```
dwightguth: this is not a valid implementation of equals. For example, it is not symmetric. I refer you to the javadoc for Object.equals: http://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#equals%28java.lang.Object%29

Another point to note is that this function now violates the contract of hashCode. In particular, hashCode guarantees that if two objects compare equal with equals(), they must have the same hash code.

I think probably the simplest solution to this problem is to get rid of the GenericKRunState class and replace instances of its constructor with `executor.step(term, 0, false).getFinalState()`

msaxena2:
Jenkins: test this please!

msaxena2:
Jenkins: test this please

msaxena2:
Jenkins: test this please

PR #1322, opened 2015-02-02
add #configuration builtin

dwightguth:
@yilongli please review


yilongli:
lgtm

grosu:
Really?  Is that all the code needed for this super-feature?  :-)

dwightguth:
Well no, not really. In particular I chose to write my one rule that required this feature without the configuration abstraction, rather than fix the configuration abstraction which was broken for function rules. I figure since we are rewriting that pass anyway, we can address it then rather than trying to rewrite the code in its current form.

PR #1324, opened 2015-02-03
break circular dependency on BuiltinFunction so that we can inject Defin...

dwightguth:
...ition


andreistefanescu:
lgtm

PR #1325, opened 2015-02-03
backend on kore

andreistefanescu:
do not review


andreistefanescu:
Jenkins: test this please

andreistefanescu:
@dwightguth @yilongli please review

andreistefanescu:
This pull request contains the modification needed to support KORE, but it does not contain that code that builds a definition from KORE. That code will be in the next pull request.

yilongli:
lgtm

java-backend/pom.xml
```diff
@@ -203,6 +203,28 @@
   <build>
     <plugins>
       <plugin>
+        <groupId>net.alchim31.maven</groupId>
```
dwightguth: not doing this.  The whole idea was for the scala portion of the codebase to be something that is a very fixed dependency with a well defined interface, which could be easily swapped out for java code in less than a week. We can't do that /and/ have a scala dependency in the rewrite engine itself.

java-backend/src/main/java/org/kframework/backend/java/kil/Collection.java
```diff
@@ -21,18 +21,16 @@
      */
     protected final Variable frame;
 
+    protected final TermContext context;
```
dwightguth, line 24: you can't store a termcontext object like this inside a term, unless you intend to use it strictly as a way to lazily compute some information which is fixed at creation time, in which case it needs to be transient. TermContext changes over the course of rewriting so in particular the value stored here will become out of date and thus lead to incorrect results. If you really need more information than you currently have it should be safe to store Definition however.

java-backend/src/main/java/org/kframework/backend/java/kil/GlobalContext.java
```diff
@@ -11,17 +11,19 @@
...
 
+import java.io.Serializable;
+
 @Singleton
-public class GlobalContext {
+public class GlobalContext implements Serializable {
     private Definition def;
-    public final FileSystem fs;
-    public final Tool tool;
+    public final transient FileSystem fs;
```
dwightguth, line 19: this is not safe either. When you deserialize this class you're going to get an object with invalid state. I have no idea why it seems to be working now but at some point it's likely to break down. This class should remain non-serializable.

dwightguth:
main concern with this pull request seems to be several unexpected departures from the current design of the backend. We are going to need to have a much longer discussion if you want to try to persuade me the changes I commented on are a good idea. But other than these specific points it looks tolerable.

java-backend/src/main/java/org/kframework/backend/java/kil/Collection.java
andreistefanescu, line 24: This is needed for translating a list/map/set in kore format, and `KItem` depends on it. I agree with you that this is not ideal, but we will change it later when we refactor `TermContext/GlobalContext`.

java-backend/src/main/java/org/kframework/backend/java/kil/GlobalContext.java
andreistefanescu, line 19: The packaging of the individual fields in `TermContext/GlobalContext` makes this refactoring difficult. We are going to change that in a future refactoring.

dwightguth:
Seems to have introduced nontermination in our large semantics. Reverting for now while @andreistefanescu looks into it.

PR #1330, opened 2015-02-06
Note on debugging in developer readme

bmmoore:
Mention main class for debugging kompile/krun


bmmoore:
@radumereuta, should this also mention anything in particular to make sure the SDF parser tables have been generated?

radumereuta:
@dwightguth knows more about this, but I had to add this to the PATH environment variable:
`$PATH;d:\work\k\k-distribution\target\release\k\lib\native\windows;d:\work\k\k-distribution\target\release\k\lib\native\windows64;d:\work\k\k-distribution\target\release\k\lib\native\64`

There is also the `-kast` option for calling the parser only.

PR #1340, opened 2015-02-10
Don't do strict type checking on semantic casts.

radumereuta:
@dwightguth please review.
This is related to the issue reported yesterday by @laurayuwen.


dwightguth:
Can we get a test in tests/regression/regression.k?

radumereuta:
Done.

PR #1344, opened 2015-02-11
Debugger Transitions

msaxena2:
Enhanced support for viewing rules in transitions. Unparsed rules can now be viewed when viewing transitions. Update to K wiki with more details on usage to follow.


kernel/src/main/java/org/kframework/backend/unparser/PrintTransition.java
```diff
@@ -28,6 +43,22 @@ public PrintTransition(
...
+     * Populates a map containing the rules, for pretty printing.
+     */
+    private void populateRuleStore() {
+        this.ruleStore = new HashMap<>();
+        definition.getItems().stream().filter((item) -> item instanceof Module).forEach((item) -> {
+            ((Module) item).getItems().stream().forEach((moduleItem) -> {
+                Source src = moduleItem.getSource();
+                Location location = moduleItem.getLocation();
+                if (src != null && location != null) {
+                    ruleStore.put(new MutablePair<>(src, location), moduleItem);
```
dwightguth: as I said, I think you should use Pair.of

kernel/src/main/java/org/kframework/backend/unparser/PrintTransition.java
```diff
@@ -35,14 +66,38 @@ public String run(Transition trans, Attributes a) {
...
+            definition = a.typeSafeGet(Attribute.Key.get(Definition.class));
+            populateRuleStore();
+        }
+        if (unparser == null) {
+            Context context = a.typeSafeGet(Context.class);
+            unparser = new Unparser(context);
+        }
+        Location location = trans.getLocation();
+        Source source = trans.getSource();
+        Pair keyPair = new MutablePair<>(source, location);
```
dwightguth: same: Pair.of

dwightguth:
Looks awesome! just that one tiny thing to fix and then you can merge.

dwightguth:
Hmm. Looks like somehow your changes broke the pdf backend. Can I get you to take a look, see if you can figure out what the problem is? If you can't reproduce it, let me know and I'll try to think how to proceed.

msaxena2:
Sure.. I'll have a look.

msaxena2:
Jenkins: test this please

PR #1345, opened 2015-02-13
Declarative disambiguation filters for KORE new parser.

radumereuta:
- Added PreferAvoid, Priority and Associativity filters
- implemented the visitor using the try/catch mechanism
- various fixes to scala reported by the IDE.
  @dwightguth and @cos please review.


kernel/src/main/java/org/kframework/parser/concrete2kore/CatchTransformer.java
```diff
@@ -0,0 +1,103 @@
...
+import org.kframework.utils.errorsystem.KException.ExceptionType;
+import org.kframework.utils.errorsystem.KException.KExceptionGroup;
+import org.kframework.utils.errorsystem.ParseFailedException;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+public class CatchTransformer {
```
dwightguth: javadoc?

kernel/src/main/java/org/kframework/parser/concrete2kore/ParserUtils.java
```diff
@@ -0,0 +1,66 @@
...
+        def.setMainSyntaxModule(mainModule);
+
+        KILtoKORE kilToKore = new KILtoKORE(null);
+        org.kframework.kore.outer.Definition koreDef = kilToKore.apply(def);
+        return parseWithDef(theTextToParse, mainModule, startSymbol, koreDef);
+    }
+
+    public static K parseWithDef(CharSequence theTextToParse, String mainModule, String startSymbol, org.kframework.kore.outer.Definition definition) {
+
+        Module kastModule = definition.getModule(mainModule).get();
```
dwightguth, line 48: can you please factor this one step further and make a parseWithGrammar function that takes a CharSequence, a Grammar, and a start symbol? And then factor out the code which generates a grammar from a Definition. In particular, for performance reasons it is going to be necessary that the grammar only be computed once for multiple executions of krun.

kernel/src/main/java/org/kframework/parser/concrete2kore/Rule.java
```diff
@@ -208,6 +208,12 @@ public DeleteRule(int length, boolean reject) {
         abstract Set<KList> apply(KList context, Set<KList> set, MetaData metaData);
     }
 
+    private static  List<Term> arrayList(Term term) {
```
dwightguth: use Lists.newArrayList instead.

kernel/src/test/java/org/kframework/parser/concrete2kore/TreeCleanerVisitorTest.java
```diff
@@ -57,19 +57,15 @@ public void testKList() throws Exception {
 //        amb([amb([NOKLABEL(amb([#emptyKRequireList()]),amb([#KModuleList(amb([#KModule(amb([#token(KModuleName,"FOO ")]),amb([#emptyKImportList()]),amb([#emptyKSentenceList()]))]),amb([#emptyKModuleList()]))]))])])
```
dwightguth: can you please break this line?

dwightguth:
I'll give this another look once you fix my comments, but it seems reasonable. After this diff is merged, how close would we be to being able to parse programs is kast using the new parser?

kernel/src/main/java/org/kframework/parser/concrete2kore/Rule.java
radumereuta: Actually Lists gives you a fixed size ArrayList, meaning you can't add elements to it. I had to use the java.utils.ArrayList because of that.

radumereuta:
After this pr, I should have pretty much all the components of the new parser required to parse programs, implemented in KORE. So it is my immediate task.

kernel/src/main/java/org/kframework/parser/concrete2kore/ParserUtils.java
radumereuta, line 48: Done. I'm using these functions more for testing, so I will refine them further as I will find more use cases.

kernel/src/main/java/org/kframework/parser/concrete2kore/Rule.java
dwightguth: no, not Lists.newArrayList. It gives you, as it says, an ArrayList.

dwightguth:
okay. If you fix the thing with Lists.newArrayList, you can merge this.

radumereuta:
Done with @dwightguth comments.
@cos since I wrote scala, please have a look.

kernel/src/main/java/org/kframework/parser/concrete2kore/CatchTransformer.java
```diff
@@ -0,0 +1,113 @@
...
+ * Base class Transformer used by the front-end to filter out ambiguities in
+ * the DAG returned by the parser. It matches on every node based on the class hierarchy
+ * from the most general to the most specialized.
+ * The Ambiguity node is special since every call to visit a child is surrounded by a
+ * try/catch statement, which captures errors thrown by the child visitor. If such an error
+ * occurs, then the child is eliminated from the tree.
+ *
+ * This a lightweight wiring mechanism, and error handling is isolated greatly.
+ */
+public class CatchTransformer {
```
cos: I see that you don't want to use the previous exception gathering mechanism. Does this report all exceptions, or just one?

radumereuta: At the moment I want to reproduce the code we have in kompile, so just one, but it can be made to throw all of them.
I will add a TODO

kore/src/main/scala/org/kframework/kore/outer/outer.scala
```diff
@@ -27,12 +27,37 @@ case class Module(name: String, imports: Set[Module], localSentences: Set[Senten
...
+
+          (newPairs | all, current)
+      }
+      x._1
+    }.flatten
+  lazy val priorities = POSet(expressedPriorities)
+  lazy val leftAssoc  = BuildAssoc(Associativity.Left)
+  lazy val rightAssoc = BuildAssoc(Associativity.Right)
+
+  private def BuildAssoc(side:Associativity.Value): Set[(Tag, Tag)] = {
```
cos: the method shouldn't be capitalized.

kore/src/main/scala/org/kframework/kore/outer/outer.scala
```diff
@@ -27,12 +27,37 @@ case class Module(name: String, imports: Set[Module], localSentences: Set[Senten
...
   }
 
+  private lazy val expressedPriorities: Set[(Tag, Tag)] =
+    sentences
+      .collect({ case SyntaxPriority(ps, _) => ps })
+      .map { ps: Seq[Set[Tag]] =>
+      val x = ps.foldLeft((Set[(Tag, Tag)](), Set[Tag]())) {
+        case ((all: Set[(Tag, Tag)], prev: Set[Tag]), current: Set[Tag]) =>
+          val newPairs: Set[(Tag, Tag)] =
+            for (a <- prev; b <- current) yield {(a, b) }
```
cos: this code seems familiar but it is ugly. did I help write it? :-)

radumereuta: History says you did.

radumereuta:
I've done multiple exception throwing. Can I merge now?

cos:
@radumereuta, do you have any tests confirming that the code actually catches multiple exceptions? Wiring this with exceptions is tricky, and it is not clear that no exceptions are missed.

radumereuta:
Test added.

cos:
@radumereuta, but you are not selecting the exceptions at the top. You are simply getting the first exception appearing on a child term, which may be much deeper then the exception appearing on a sibling. 

This seems like an instance of "it's not a bug, it's a feature". If you really think we should only report the exceptions appearing higher in the AST, then make an implementation for exactly that. You're fitting the requirements to the implementation.

kore/src/main/scala/org/kframework/kore/outer/outer.scala
cos: Yes, I implemented this. We were pair programming. I'll try to figure out a better solution.

cos: I cleaned up the fold a bit, but I cannot push to your account. I pushed to `k/feature/new-outer-parser` (it may be good to always use the main repo when more then one person works on a pull request). Could you merge to your account? Thanks!

radumereuta:
I've been talking to @cos, and we seem to disagree on the implementation of how I do disambiguation.
I prefer a lightweight implementation where I don't mix wiring with error handling. Whereas Cosmin wants to return objects that can either be a set of errors, or the correct term. His approach may help bring more errors to the users in a single run, but my approach avoids visiting parts of terms that have already been decided to be wrong, therefore making it more efficient.

Well need someone else to break the tie here.

cos:
@radumereuta, your solution does not avoiding the mix of of wiring with error handling. It simply avoids error handling by only reporting the first encountered error in many cases.

If you decide to implement reporting of all errors (or of a relevant subset, not the first encountered one), your solution will be at least as complex as the one based on `Transformer`. Actually, the solution based on `Transformer` is just as simple even now (see the diff for `TreeCleanerVisitor`).

Normally I wouldn't make such a big deal out of this, but you replaced the code we implemented together (which I know you didn't like as it introduced the concept of `Either`), with an inferior solution. I call it inferior because it doesn't report all errors. Furthermore, even if you make it report all errors, it will be a solution based on complex exception handling, not clean recursion. And it will lack the type safety of the previous solution which checked (at compile time) that  both branches of the `Either`. Now you can easily miss error cases (like you did after you added multiple error handling) without getting any warning either at compile or runtime.

As for efficiency, yes, if you fail to report all errors, you will avoid visiting some branches. But anything you gain here will be offset by probably at least an order of magnitude by the fact you're using exception handling extensively. Afaik, exceptional control flow is much slower then normal control-flow. 

Finally, in a nutshell, what you're doing, i.e., using exceptional control flow for encoding regular control flow, is very bad design (if you like fancy terms, it is an anti-pattern). If you don't believe me, believe others: https://www.google.com/search?client=safari&rls=en&q=using+exceptions+for+control+flow&ie=UTF-8&oe=UTF-8

grosu:
Let's discuss this in today's K meeting.

Grigore

dwightguth:
For what it's worth, I am strongly in favor of multiple error reporting if we can manage it.

radumereuta:
@dwightguth I think you should have another look over this.
I've reworked the visitors that @cos proposed to include warning reporting.
Now it makes a lot more sense to use this type of visitors. I removed the try/catch mechanism.

kore/src/main/scala/org/kframework/parser/Transformer.scala
```diff
@@ -5,60 +5,154 @@ package org.kframework.parser
...
+  def apply(a: Ambiguity): Either[E, Term] = mapChildren(a)._1
+  def apply(tc: TermCons): Either[E, Term] = mapChildrenStrict(tc)._1
+  def apply(kl: KList): Either[E, Term] = mapChildrenStrict(kl)._1
+  def apply(c: Constant): Either[E, Term] = Right(c)
+
+  override def mergeWarnings(a: Ignore, b: Ignore) = Ignore
+  override val warningUnit = Ignore
+}
+
+abstract class SafeTransformer extends ChildrenMapping[Ignore, Ignore] {
```
dwightguth, line 144: javadoc

kore/src/main/scala/org/kframework/parser/Transformer.scala
```diff
@@ -5,60 +5,154 @@ package org.kframework.parser
...
+    case c: Constant => apply(c)
+  }
+
+  def apply(a: Ambiguity): (Either[E, Term], W) = mapChildren(a)
+  def apply(tc: TermCons): (Either[E, Term], W) = mapChildrenStrict(tc)
+  def apply(kl: KList): (Either[E, Term], W) = mapChildrenStrict(kl)
+  def apply(c: Constant): (Either[E, Term], W) = { (Right(c), warningUnit) }
+}
+
+abstract class TransformerWithErrors[E] extends ChildrenMapping[E, Ignore] {
```
dwightguth, line 110: javadoc

kore/src/main/scala/org/kframework/parser/Transformer.scala
```diff
@@ -5,60 +5,154 @@ package org.kframework.parser
...
+  val warningUnit: W
+
+  val errorUnit: E
+  /**
+   * Merges the set of problematic (i.e., Left) results.
+   */
+  def mergeWarnings(a: W, b: W): W
+}
+
+abstract class GeneralTransfomer[E, W] extends ChildrenMapping[E, W] {
```
dwightguth: javadoc

kore/src/main/scala/org/kframework/parser/Transformer.scala
```diff
@@ -5,60 +5,170 @@ package org.kframework.parser
...
+}
+
+/**
+ * Visitor pattern for the front end classes.
+ * Applies the visitor transformation on each node, and returns a tuple of either a term, or a set of errors, and
+ * a set of possible warnings.
+ * @tparam E container for error.
+ * @tparam W container for warning.
+ */
+abstract class GeneralTransfomer[E, W] extends ChildrenMapping[E, W] {
```
dwightguth: s/fomer/former/

dwightguth:
That's my last remaining comment.

kore/src/main/scala/org/kframework/parser/Transformer.scala
radumereuta: Huh? I don't understand what you mean by that.

dwightguth: oh, it's a sed thing. You wrote GeneralTransfomer and I wanted it replaced with GeneralTransformer. With an r.

cos:
Looks good. I'm ok with merging now.

dwightguth:
lgtm

PR #1347, opened 2015-02-17
fix Chucky's bug

dwightguth:
@bmmoore please review


bmmoore:
We might consider making Linux an explicit case rather than the default.
Consulting http://en.wikipedia.org/wiki/Uname it looks like every Linux distribution actually prints "Linux", and the remaining possible outputs are BSD or other unix variants.
Continuing after a warning might be a better option, but this change is acceptable.

bmmoore:
Looks good.

PR #1357, opened 2015-02-24
revert DefinitionScope of java kil Definition because of race in JS sema...

dwightguth:
...ntics

@yilongli please review

If the bulk of the ~7min performance optimization on the C semantics test suite is lost by this diff, I will track down and deal with the remaining mutable state in the Definition object. But I want to get this revert in today in order to unbreak the build.


yilongli:
looks ok to me

PR #1361, opened 2015-02-25
KORE

cos:
**please review**
- various fixes and small changes
- a simple `Rewriter` interface proposal -- we'll have to agree on it, but not necessarily before merging
- stump for E-KORE to KORE meta-translator
- tiny, my rewrite engine, is able to execute a simple IMP program, Chucky's `while` loop counting to 1000 , in 120s. This is exactly one order of magnitude slower then the Java backend (when executed in the current pipeline; I think the Java backend running on KORE lacks many compiler optimizations and may be significantly slower then the one integrated in the pipeline), but should be more then enough for writing compiler passes in K.
- update Scala to 2.11.5 and activate the `-Xperimental` flag which significantly improves Java interop.


cos:
Jenkins, test this please.

cos:
Jenkins, test this please.

cos:
@dwightguth, @andreistefanescu, please review

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
```diff
@@ -149,14 +209,14 @@ public Rule apply(org.kframework.kil.Rule r) {
...
         switch (assocOrig) {
-        case "left":
-            return Associativity.Left();
-        case "right":
-            return Associativity.Right();
-        case "non-assoc":
-            return Associativity.NonAssoc();
-        default:
-            throw new AssertionError("Incorrect assoc string: " + assocOrig);
+            case "left":
```
dwightguth: this was better the way it was before. Please revert.

kore/src/main/scala/org/kframework/Rewriter.scala
```diff
@@ -0,0 +1,21 @@
...
+  def normalize(k: K): K
+  def substitute(k: K, s: KVariable => K): K
+
+  /**
+   * Takes one rewriting step.
+   *  - for regular execution, it returns the next K or False (i.e. an empty Or)
+   *  - for symbolic execution, it can return any formula with symbolic constraints
+   *  - for search, it returns an Or with multiple ground terms as children
+   */
+  def step(k: K): K
```
dwightguth, line 18: are True, False, And, Or, Not, etc privileged values under this proposal or are they equivalent to the true, false, andBool, etc in K right now?

dwightguth:
looks okay although I agree we'll have to iterate over the rewriter interface. I'd like to have an answer to my question about it before we merge though.

kore/src/main/scala/org/kframework/Rewriter.scala
cos, line 18: I haven't put this down yet, but I think they should be privileged. In the sense that we should have "marker" interfaces for `KApply` for `And` and `Or`. `True` and `False` should simply be empty `And` and `Or`.

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
cos: I know the Java code convention is to align `case` with the `switch` but it doesn't really make sense. I think the Google code convention has it right. In any case, not worth a discussion.

kore/src/main/scala/org/kframework/Rewriter.scala
dwightguth, line 18: I agree; trying to make them the same introduces a confusion between program syntax and the meta level. I just wanted to get your opinion. This seems like a workable API if we start from that assumption.

dwightguth:
I am okay with this being merged now. Should we still wait for @andreistefanescu to review, or try to find someone else to substitute as the second reviewer?

cos:
I added @andreistefanescu to get his opinion on the rewriter API, but we'll discuss and change it quite a bit anyhow. So I think we should merge for now.

PR #1367, opened 2015-03-02
Rule grammar2

radumereuta:


radumereuta:
Rule Parser:
@dwightguth @cos please review.
I've added code which takes a module as input, and another separate module representing the K grammar, and create a parser for rule: generate the sort lattice to connect concrete user syntax with K syntax, and casts (for now). Also I reimplemented and adapted the disambiguation filters.

I added junits for now, but after this merge I will try larger examples from `k-distribution`.

k-distribution/src/test/java/org/kframework/kore/convertors/BaseTest.java
```diff
@@ -49,7 +49,7 @@ public DefinitionWithContext(Definition d, Context c) {
 
     private File testResource(String baseName) {
         return new File(new File("k-distribution/src/test/resources" + baseName)
-                .getAbsoluteFile().toString().replace("k-distribution/k-distribution", "k-distribution"));
+                .getAbsoluteFile().toString().replaceAll("k-distribution.k-distribution", "k-distribution"));
```
dwightguth: if you want to support windows, use File.separator. You don't need a regex.

kernel/src/main/java/org/kframework/parser/concrete2kore/ParseInModule.java
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Parse as input the given string and start symbol using the module stored in the object.
+     * @param input          the string to parse.
+     * @param startSymbol    the start symbol from which to parse.
+     * @return the Term representation of the parsed input.
+     */
+    // TODO: require source location to this call
+    // TODO: figure out how to handle parsing errors
+    public Tuple2<Either<Set<ParseFailedException>, Term>, Set<ParseFailedException>> parseString(CharSequence input, String startSymbol) {
```
dwightguth, line 47: I would recommend Apache Commons Lang3's Pair class for consistency with the rest of the code base.

kernel/src/main/java/org/kframework/parser/concrete2kore/ParserUtils.java
```diff
@@ -33,37 +30,30 @@ public static K parseWithFile(CharSequence theTextToParse, String mainModule, St
     }
 
     public static K parseWithString(CharSequence theTextToParse, String mainModule, String startSymbol, String definitionText) {
```
dwightguth: break this line

kernel/src/main/java/org/kframework/parser/concrete2kore/ParserUtils.java
```diff
@@ -33,37 +30,30 @@ public static K parseWithFile(CharSequence theTextToParse, String mainModule, St
...
+        Module kastModule = parseMainModuleOuterSyntax(definitionText, mainModule);
+        return parseWithModule(theTextToParse, startSymbol, kastModule);
+    }
+
+    public static K parseWithModule(CharSequence theTextToParse, String startSymbol, org.kframework.definition.Module kastModule) {
```
dwightguth: and this one

k-distribution/src/test/java/org/kframework/kore/convertors/BaseTest.java
radumereuta: This is something that @cos told me to do. And it works on windows.
And for future reference, evaluating paths with slashes in Java works both in unix and in windows. Java does all the magic. So I don't think this is a problem.

kernel/src/main/java/org/kframework/parser/concrete2kore/ParserUtils.java
```diff
@@ -33,37 +30,30 @@ public static K parseWithFile(CharSequence theTextToParse, String mainModule, St
...
+        ParseInModule parser = new ParseInModule(kastModule);
```
dwightguth, line 46: this code needs to be factored to support the fact that we will eventually need to compute Grammar objects during kompile and serialize them.

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/AmbFilter.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.Set;
+
+/**
+ * Eliminate remaining ambiguities by choosing one of them.
+ */
+public class AmbFilter extends SetsGeneralTransformer<ParseFailedException, ParseFailedException> {
+
+    @Override
+    public Tuple2<Either<Set<ParseFailedException>, Term>, Set<ParseFailedException>> apply(Ambiguity amb) {
```
dwightguth, line 24: same here: Pair

k-distribution/src/test/java/org/kframework/kore/convertors/BaseTest.java
dwightguth: okay then @cos why was this change needed?

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/CorrectCastPriorityVisitor.java
```diff
@@ -0,0 +1,52 @@
...
+                tc.items().set(0, rez.right().get());
+        }
+        return super.apply(tc);
+    }
+
+    private static class PriorityVisitor2 extends SetsTransformerWithErrors<ParseFailedException> {
+        public Either<java.util.Set<ParseFailedException>, Term> apply(TermCons tc) {
+            // TODO: add location information
+            if (tc.production().items().last() instanceof NonTerminal) {
+                String msg = "Due to typing errors, Casting is too greedy. Use parentheses to set proper scope.";
```
dwightguth: this is a bad error that we wanted to get rid of in the new parser, remember? so why is it still here?

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/CorrectKSeqPriorityVisitor.java
```diff
@@ -0,0 +1,59 @@
...
+            }
+        }
+        return super.apply(tc);
+    }
+
+    private static class PriorityVisitor2 extends SetsTransformerWithErrors<ParseFailedException> {
+        public Either<java.util.Set<ParseFailedException>, Term> apply(TermCons tc) {
+            // TODO: add location information
+            if (tc.production().klabel().isDefined() && tc.production().klabel().get().name().equals("#KSequence")) {
+                String msg = "Due to typing errors, => is not greedy. Use parentheses to set proper scope.";
```
dwightguth: same here

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/CorrectRewritePriorityVisitor.java
```diff
@@ -0,0 +1,57 @@
...
+            }
+        }
+        return super.apply(tc);
+    }
+
+    private static class PriorityVisitor2 extends SetsTransformerWithErrors<ParseFailedException> {
+        public Either<java.util.Set<ParseFailedException>, Term> apply(TermCons tc) {
+            // TODO: add location information
+            if (tc.production().klabel().isDefined() && tc.production().klabel().get().name().equals("#KRewrite")) {
+                String msg = "Due to typing errors, => is not greedy. Use parentheses to set proper scope.";
```
dwightguth: and here

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/VariableTypeInferenceFilter.java
```diff
@@ -0,0 +1,461 @@
...
+                    }
+                    // The above loop looks for variables that can have multiple sorts, collected from multiple solutions.
+                    // In rare cases (couldn't think of one right now) it may be that the
+                    // solution may be different because of different variable names
+
+                    // Ok, I found one example now. In C with unified-builtins, the follow restriction for ==Set doesn't work
+                    // and it creates multiple parses with different amounts of variables
+                    // This makes it that I can't disambiguate properly
+                    // I can't think of a quick fix... actually any fix. I will delay it for the new parser.
+                    String msg = "Unsolvable ambiguities. Please write the rule in labeled form.\n (Generally because of __ productions mixing with variables).";
```
dwightguth: and here

kernel/src/main/java/org/kframework/parser/concrete2kore/ParseInModule.java
radumereuta, line 47: The basic visitors are defined in scala, so I used the `Tuple2` from there.
Using it makes the scala code look much better. I don't think the pair you are suggesting can be used like that.

dwightguth, line 47: oh I see okay. That makes sense then. Never mind, ignore this one.

k-distribution/src/test/java/org/kframework/kore/convertors/BaseTest.java
cos: because `toString` prints out the file-system-specific separator so we need to replace it in a file-system-specific manner. So in this case the localization actually hurts. 

Yes, it can be done with File.separator, it's just a bit more verbose. Feel free to change it if you like.

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/CorrectCastPriorityVisitor.java
radumereuta: The message here is kind of irrelevant since if I managed to generate the grammar correctly, there should be a branch in the parse forest which should be available after this step.
I will change the message when we have cases where this is a problem.

k-distribution/src/test/java/org/kframework/kore/convertors/BaseTest.java
dwightguth: does toPath also print the file-specific separator though?

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/CorrectCastPriorityVisitor.java
dwightguth: If this code is supposed to be unreachable then you need to modify this code to indicate that it is supposed to be unreachable (ie, throw an uncaught exception so the user is informed that it's a bug in the framework). It is not at all clear from this error message that this state is supposed to be unreachable, which is a bad user experience.

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/CorrectRewritePriorityVisitor.java
radumereuta: Now this is different. Take for example something like this:

```
rule foo bar => Y
syntax A ::= foo A
syntax  B ::= "bar"
```

In this case, the only possible parse would be: `'foo_(=>('bar(), Y))`.
What I'm able to detect here is that the rewrite's left child has a right most nonterminal. In normal circumstances there will be another ambiguity where rewrite will be at the top, but in this case it can't be because `foo bar` doesn't parse.

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/CorrectCastPriorityVisitor.java
radumereuta: It is reachable, and it is part of the disambiguation algorithm. Each time I see a case where I need to eliminate the branch, I return an error which needs to be propagated until the last `Ambiguity` node visited where we can do disambiguation.

At this stage I would like to worry more about functionality and less about user experience.
Let's have the new parser and kore integrated into kompile and krun first before we spend the time to polish error messages.

I will feel more comfortable working on this when I have more examples.

k-distribution/src/test/java/org/kframework/kore/convertors/BaseTest.java
radumereuta: Can I ignore this in the current pull request? This is @cos code.

kernel/src/main/java/org/kframework/parser/concrete2kore/ParserUtils.java
radumereuta, line 46: The grammar itself is actually only half the parser. You also need the module where you find the priority and associativity rules. I created a wrapper which keeps these two objects together in `ParseInModule`. I also updated these helper functions. (will commit soon)

k-distribution/src/test/java/org/kframework/kore/convertors/BaseTest.java
cos: I will fix it, you team player, you!

radumereuta: I'm just trying to keep things separated.

cos:
> Fixing some of @dwightguth _commits_.

Did @dwightguth have any commits in this pull request?

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/CorrectCastPriorityVisitor.java
dwightguth: No, do it now. If you need to adjust it later, you can. But in general plans to clean code up later do not lead to code actually being cleaned up later, so I want to deal with this now because otherwise it will likely never get fixed.

radumereuta:
_Fixing some of @dwightguth comments_

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/CorrectRewritePriorityVisitor.java
dwightguth: Same here. Needs a better error message to indicate what might actually cause this error to manifest.

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/CorrectCastPriorityVisitor.java
grosu: My 2 philosophical cents: It is a good idea in general, in everything you do in your life, to achieve local maxima when it gets to minor things.  There is never a better moment to fix something simple which can be fixed quickly than when you just noticed it.  I strongly recommend this book:

http://www.amazon.com/The-Habits-Highly-Effective-People/dp/1455892823

Grigore

radumereuta: Talked to @dwightguth  and error messages improved.

radumereuta:
I've addressed all comments. Please have another look.

dwightguth:
Can you please construct example programs that generate each of the error messages you changed, and post both the program and the error message they print on the command line? it's hard for me to judge the UI experience of these errors in the abstract without that info.

dwightguth:
Sorry to clarify obviously except the ones that you marked in the error message as being a bug to be reported.

k-distribution/src/test/java/org/kframework/kore/convertors/BaseTest.java
cos: @radumereuta, you may need to decide at some point between collective or individual code ownership. 

Scenarios:
1. We work together to extend the visitors, I code, you sit by my side. At the end, you summarize our activity as:
   
   > I've reworked the visitors that @cos proposed to include warning reporting.
   > (https://github.com/kframework/k/pull/1345#issuecomment-74777567)
2. You work on something, ask me how to fix an issue, I advise you on a working method -- good or bad is irrelevant. You summarize the activity as:
   
   > I'm just trying to keep things separated.

I don't mind either 1. or 2., but I do mind both. If you decide that the code is ours, and we don't care to be precise with attribution, I am perfectly ok with this (actually like it). But if you decide you want to keep things separated, please try to make sure you're precise about it all the time.

radumereuta:
I've made the tests print out more information and had to modify the error reporting system a bit.
My test right now would print this:

```
rule = #KRewrite('Plus(#token(Exp,"1"),#token(Exp,"2")),#KSequence(#SemanticCast(#token(KVariable,"A")),#OuterCast(#token(KVariable,"B"))))
[Warning] Compiler: Variable 'B' was not declared. Assuming sort K.
    Location: (1,1,1,19)
rule = #KRewrite('Decl(#token(KVariable,"X"),#SemanticCast(#token(KVariable,"S"))),'tuple(#token(KVariable,"X"),#token(KVariable,"S")))
[Warning] Compiler: Variable 'X' was not declared. Assuming sort Id.
    Location: (1,1,1,25)
rule = #KRewrite('Decl(#token(KVariable,"X"),#token(KVariable,"S")),'tuple(#token(KVariable,"X"),#token(KVariable,"S")))
[Warning] Compiler: Variable 'S' was not declared. Assuming sort Stmt.
    Location: (1,1,1,20)
[Warning] Compiler: Variable 'X' was not declared. Assuming sort Id.
    Location: (1,1,1,20)
[Error] Critical: Associativity filter exception. Cannot use 'Plus as a left
child of 'Plus
    Location: (1,1,1,4)
[Error] Critical: Associativity filter exception. Cannot use 'Plus as a right
child of 'Plus
    Location: (1,3,1,6)
rule = 'Plus(#token(Exp,"1"),'Plus(#token(Exp,"2"),#token(Exp,"3")))
1: Exp ::= syntax Exp ::= Exp "+" Exp [#klabel(#token(AttributeValue,"'Plus"))
org.kframework.attributes.Location(#token(Int,"1"),#token(Int,"28"),#token(Int,"1"),#token(Int,"54"))]
    'Plus(#token(Exp,"1"),'Plus(#token(Exp,"2"),#token(Exp,"3")))
2: Exp ::= syntax Exp ::= Exp "+" Exp [#klabel(#token(AttributeValue,"'Plus"))
org.kframework.attributes.Location(#token(Int,"1"),#token(Int,"28"),#token(Int,"1"),#token(Int,"54"))]
    'Plus('Plus(#token(Exp,"1"),#token(Exp,"2")),#token(Exp,"3"))
    Location: (1,1,1,6)
[Error] Critical: Could not parse rewrite as a child of syntax A ::= "foo" A
[#klabel(#token(AttributeValue,"'foo"))
org.kframework.attributes.Location(#token(Int,"1"),#token(Int,"26"),#token(Int,"1"),#token(Int,"47"))]
   Use parentheses to set the proper scope of the rewrite operation.
    Location: (1,5,1,13)
```

I've also added a flag to switch between printing or not, since we don't want to clutter the console.

dwightguth:
Okay, that's what I thought it would say, but that's not really what we discussed. The problem isn't anything to do with this parse; the problem is with the /other/ parse, the one we already rejected as not obeying the grammar. We need to indicate that /that/ parse was unsuccessful.

radumereuta:
Well... that parse doesn't exist. The kernel of the parser automatically discarded it because `bar` can't be a child of `foo_`. I'm not sure exactly what you want me to say about it.
Inferring that it might exist could be dangerous, for example `foo foo foo bar => X` needs to go up a few levels. Maintaining such information is kind of tricky.

dwightguth:
well but you should be able to infer the location that the parse exists over because it goes from the beginning of the term that actually /was/ parsed to the =>. And that's all you really need, you don't need the parse itself.

radumereuta:
Sure, I can change the location information to be reported for the parent instead of the child rewrite, but it doesn't work that well for my previous example. And even so. There is another problem.
When I was trying to explain what happens there, and I told people to put parenthesis to set the proper scope of the rewrite, their first instinct was to write something like this: `(foo bar => X)` instead of
`(foo bar) => X`.

radumereuta:
I managed to solve all the comments that @dwightguth had.
@cos please have another look.

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/AmbFilter.java
```diff
@@ -0,0 +1,48 @@
...
+import scala.Tuple2;
+import scala.util.Either;
+import scala.util.Right;
+
+import java.util.Set;
+
+/**
+ * Eliminate remaining ambiguities by choosing one of them.
+ */
+public class AmbFilter extends SetsGeneralTransformer<ParseFailedException, ParseFailedException> {
```
cos, line 21: should the right-hand side be of a different type representing a warning, not an exception?

cos, line 21: I'm talking about the type parameters.

kernel/src/test/java/org/kframework/parser/concrete2kore/RuleGrammarTest.java
```diff
@@ -0,0 +1,179 @@
...
+                "syntax Exp ::= Exp \"+\" Exp [klabel('Plus)] " +
+                "| r\"[0-9]+\" [token] " +
+                "endmodule";
+        Module test = ParserUtils.parseMainModuleOuterSyntax(def, "TEST");
+        ParseInModule parser = gen.getRuleGrammar(test);
+        Tuple2<Either<Set<ParseFailedException>, Term>, Set<ParseFailedException>> rule =
+                parser.parseString("1+2+3", startSymbol);
+        Assert.assertEquals("Expected 1 warning: ", 1, rule._2().size());
+        Assert.assertTrue("Expected no errors here: ", rule._1().isRight());
+        printout(rule);
```
cos, line 152: there's a lot of code duplication in this file. the code for each test has been copy-pasted and could be extracted instead in a separate method.

kore/src/main/scala/org/kframework/POSet.scala
```diff
@@ -60,14 +60,17 @@ class POSet[T](directRelations: Set[(T, T)]) {
    * Returns true if x < y
    */
   def lessThen(x: T, y: T): Boolean = <(x, y)
+  def lessThenEq(x: T, y: T): Boolean = <(x, y) | x == y
```
cos, line 63: not necessarily for this pull request, but I'm wondering... Wouldn't it be better to have distinct strict and non-strict `POSet`s?

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/AmbFilter.java
radumereuta, line 21: `ParseFailedException` is just a name used to wrap up exceptions originating in the parser. The actual information is in the KException class.
If you want I can choose a different name which fits in both locations.
We may want at one point to look over the entire error reporting system. I had to do a few patches to make it work right now.

kore/src/main/scala/org/kframework/POSet.scala
radumereuta, line 63: It could be a parameter to this class. At the moment I only need the strict version.

kernel/src/test/java/org/kframework/parser/concrete2kore/RuleGrammarTest.java
radumereuta, line 152: Refactored.

radumereuta:
@cos, please re-review.

cos:
ok. looks good to merge.

PR #1371, opened 2015-03-04
fix failing test in c semantics due to warning

dwightguth:
@radumereuta accidentally broke the -w flag which was surpressing the warning from  the exception thrown by this hook. It's rather time-sensitive, so I'm fixing the warning while he fixes the bug. @yilongli please review


yilongli:
lgtm

PR #1376, opened 2015-03-05
fix #1375: add better error message for unsupported set pattern

### LINKED ISSUE #1375: Error message "Unsupported map pattern in the rule left-hand side" reported for definitions not using maps
The compilation of the module 

```
module TST

  syntax Edge ::= Int "-->" Int
  syntax SetOfEdges ::= "{" Set "}"

// the set of vertices of a graph

  syntax Set ::= vertices(SetOfEdges) [function]

  rule vertices({ SetItem(I:Int --> J:Int) G:Set }) => (SetItem(I) SetItem(J) vertices({ G }))

endmodule
```

reports the following error:

```
$ kompile tst.k 
[Error] Critical: Unsupported map pattern in the rule left-hand side
    Source: File:
/Users/dlucanu/Documents/cursuri/pa/2015/seminar-special/03mar2015/clasa/xxx/././tst.k
    Location: (10,3,10,94)
```

but the definition does not use maps.

Version of the kompile tool:

```
$ kompile --version
K framework version 3.6-SNAPSHOT
Git revision: 4023860
Git branch: master
Build date: Tue Mar 03 16:11:32 EET 2015
```

dwightguth:
@yilongli please review


yilongli:
lgtm

PR #1377, opened 2015-03-05
fix #1372 : improve error for "cannot compute least sort"

### LINKED ISSUE #1372: strange "Cannot compute least sort" error message
The compilation of the module 

```
module TST

  syntax Typename ::= "int"

  syntax Decl ::= Typename IvarList

  syntax IvarList ::= List{Ivar, ","}

  syntax Ivar ::=
       Id
     | Id "=" Expr [ivar]

  syntax Varref ::= 
       Id
     | Id "[" Expr "]"

 syntax Const ::= Int 

  syntax Expr ::=
       Varref
     | Const

  syntax Assign ::=
       Varref "=" Expr [assign]

  rule <k> T:Typename X:Id = V:Const => .K ...</k>

endmodule
```

reports the following error:

```
$ kompile tst.k 
[Error] Critical: Cannot compute least sort of term: '_=_(_8:Id,, _9:Const)
Possible least sorts are: [Assign, Ivar]
```
1. No line & column numbers are displayed.
2. The error is given by the only rule in the module. From my point of view, "T:Typename X:Id = V:Const " is not ambiguous, it should be infered that the least sort of "X:Id = V:Const" is Ivar. Moreover, neither of (X:Id = V:Const)::Ivar , (X:Id = V:Const):>Ivar, (X:Id = V:Const):Ivar, (X:Id = V:Const)<:Ivar solves the problem.

dwightguth:
@yilongli please review

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
@@ -236,7 +236,9 @@ public class KItem extends Term implements KItemRepresentation {
         Sort sort = sorts.isEmpty() ? kind.asSort() : subsorts.getGLBSort(sorts);
         if (sort == null) {
             throw KExceptionManager.criticalError("Cannot compute least sort of term: " +
-                            this.toString() + "\nPossible least sorts are: " + sorts);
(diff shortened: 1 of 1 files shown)


yilongli:
lgtm

PR #1378, opened 2015-03-05
fix issue with hashcode introduced by previous diff. :/

### LINKED ISSUE #1369: Strange error when parsing maps with the key zero
Consider the following simple definition:

```
module ARR
  syntax Value ::=   "{" Map "}"

  configuration
    <k> .K </k>
    <state> $ST:Map </state> 

endmodule
```

The following test is ok:

```
$ krun -cST="a |-> { 1 |-> 2 }"
<k>
    .K
</k>
<state>
    a |-> { 1 |-> 2 }
</state>
```

but the following one returns a very strange error:

```
$ krun -cST="a |-> { 0 |-> 2 }"
[Error] Internal: Uncaught exception thrown of type NullPointerException.
```

Note that the only thing changed is the key of the map. For any key =/= 0 it works fine (inlcuding negatives).

dwightguth:
fixes #1369 

@yilongli please review


yilongli:
lgtm

PR #1380, opened 2015-03-06
KORE

cos:
**please review**
- work on consolidating builtins
- fixing issues with meta up/down, and reactivating part of the tests
- moving syntactic sugar from tiny to kore -- so that it is available when using other rewrite engines
- improvements to tiny -- unique labels and parallelization. it is now 3x slower then the Java rewrite engine


cos:
@dwightguth, please review

dwightguth:
When you say 3x slower than the Java rewrite engine, are you comparing it against the --pattern-matching flag run with Nailgun? because If you can get it faster than that for the C semantics, I will cheerfully start using your backend to execute programs.

cos:
I think my backend is very far from executing any C program. At the very least, we'll need a KORE definition of C. The 3x is a very rough comparison: Brandon said he executed Chucky's "sum to 1000" example in about 12s on his notebook. I've executed the same IMP program in about 40s on mine. It's a ballpark comparison, it may be 2x or 10x slower in a comprehensive comparison.

My engine would actually fare worse with nailgun because I've already parallelized the matching. So there would be fewer processors available for the nailgun's courser-grained parallelization.

I haven't spent much time optimizing so I think there are still several low-hanging fruits for making it faster. So far I've only done some indexing, caching, and parallelization. There are likely fine-grained optimizations, like eliminating some cases for matchers, which can have large effects. 

Finally, my purpose is not to beat the Java engine in terms of speed, but have an implementation that has everything I need for program transformation (i.e., assoc (and some AC) matching for any term, matches in side conditions, strategies).

grosu:
I continue to believe that it is a good idea to experiment with a new engine for the purpose of nailing down the API of the rewrite engine.  And it may also evolve into a fast one, maybe some pushout of it with the existing one.

In the meanwhile, we still have Maude as a state-of-the-art competitor.  My daughter told me today that half of her class spent 30 min trying to solve the following problem: you are given numbers 5,5,7,4 and you can apply any arithmetic operation on them; can you obtain 24?  Of course, I was not able to solve while driving home, but as soon as I got home I wrote the following Maude program:

```
mod T is
  including RAT .
  sort MSet .
  subsort Rat < MSet .
  op _,_ : MSet MSet -> MSet [assoc comm] .
  vars A B : Rat .
  rl A,B => A * B .
  rl A,B => A - B .
  rl A,B => A + B .
  rl A,B => A / B .
endm

search 5,5,7,4 =>* 24 .
show path 634 .
```

The output is:

```
Solution 1 (state 634)
states: 635  rewrites: 4805 in 4ms cpu (3ms real) (1201250
    rewrites/second)
empty substitution

No more solutions.
states: 676  rewrites: 5306 in 4ms cpu (4ms real) (1326500
    rewrites/second)
state 0, MSet: 4,5,5,7
===[ rl A,B => A / B . ]===>
state 19, MSet: 1,4,7
===[ rl A,B => A - B . ]===>
state 197, MSet: 4,6
===[ rl A,B => A * B . ]===>
state 634, NzNat: 24
```

So 4 miliseconds!  We have a long way to go.

grosu:
I wish some day to write K programs like this:

```
  import BAG{RAT}
  rule A,B => A*B \/ A-B \/ A+B \/ A/B
```

The RHS of the rule above is a correct ML pattern.  Also, if there is only one module, no need to have `module` and `endmodule`, so the above is the complete program.  Then "prove" CTL formulae like

```
prove 5,5,7,4 -> E F 24
```

(that is, 5,5,7,4 implies there exists (E) a path on which in the future (F) we see 24).

cos:
Regarding the Maude example, it is a bit tricky... unless paying the huge heating/cooling price, it requires anywhere AC matching. My engine doesn't support that yet as it is tricky to implement. I will think about it. 

The `\/` rhs looks cool.

cos:
Just for fun, I implemented what was necessary and tried Grigore's challenge with tiny (my prototype engine). It is crazy slow, about 10s, but I haven't implemented any optimizations. 

You can have a look at the code here:
https://github.com/kframework/k/blob/kore/tiny/src/test/scala/org/kframework/tiny/GrigoreChallange.scala

There are two versions of the module there: one with normal rules (and extra variables for matching) and one with anywhere rules.
Also, there are two tests for each of them, one doing rewriting and showing all possible final results, and another searching for final pattern (in this case 24).

The commit basically adds full AC (modulo any bugs I might have planted) and anywhere (so anywhere-AC as well) matching.

Regarding speed, I think I can easily cut about 10x of the time, and another 10x with reasonable effort, but that would bring us to 100ms, still very far from Maude's 4ms.

grosu:
This is very cool!  So the `Rewriter` class is your tiny rewriter?  How does it know that `~` is infix?  Are you assuming that for all AC operations?  Note that this is not required in Maude, for example, you can write things like `f(a,f(b,c))` where f is an AC operation.  Thanks @cos .

cos:
> This is very cool! So the Rewriter class is your tiny rewriter? 

Yes.

> How does it know that `~` is infix? 

That's just Scala syntactic sugar. It's the same as writing `KApply(KLabel("~"), a, b)`. Basically, I made a few common operators applicable to `K` as infix (https://github.com/kframework/k/blob/kore/kore/src/main/scala/org/kframework/kore/ScalaSugar.scala#L25-L36). There's also more sugar transforming Java/Scala primitives to `KToken`s, and allowing easy instantiation of `KApply` like this `'foo('bar('foo()) + 5, "I'm a string")`.

> Are you assuming that for all AC operations? Note that this is not required in Maude, for example, you can write things like f(a,f(b,c)) where f is an AC operation. 

Only the sugar is infix -- the rewriter itself doesn't care about syntactical issues like `~` being infix. I used `~` because I wanted to use infix syntax for clarity, and the `,` cannot be used as an operator in Scala.

PR #1381, opened 2015-03-06
Add ability to index on more than one cell if desired

dwightguth:
**please review**


dwightguth:
Jenkins: test this please

dwightguth:
I'm not entirely certain why, but apparently the problem I was having, that unindexed rules were a performance bottleneck, goes away with this diff even if I don't actually index on more than the K cell. I'm inclined to try to commit this though, since it seems to have a roughly 40% speed increase on some programs in the C semantics. If we commit it and run into bugs or failed tests or loss in performance, I'll investigate further.

dwightguth:
@yilongli please review

yilongli:
lgtm

dwightguth:
The problem is that the indexingPair method is still used by stream cells, and I'm not entirely sure how to get the cell associated with those rules, which would be needed in order to store it in the map.

yilongli:
OK, then I am fine with it.

dwightguth:
Jenkins: test this please

java-backend/src/main/java/org/kframework/backend/java/indexing/IndexingTable.java
```diff
@@ -29,16 +40,16 @@
 /**
  * The indexing scheme currently used in the Java Backend
  */
-public class IndexingTable implements Serializable, RuleIndex {
+public class IndexingTable implements Serializable {
 
-    private Map<Index, ImmutableSet<Rule>> ruleTable;
+    private transient Table<CellLabel, Index, ImmutableSet<Rule>> ruleTable;
```
andreistefanescu, line 45: why is this transient?

dwightguth, line 45: not sure... seems like a bug. sorry

PR #1384, opened 2015-03-06
Kast new parser

radumereuta:
Testing new parser for programs, do not review.


radumereuta:
@dwightguth please review, but do not merge. For the final merge I have to revert the default parser to be SDF.
I merged master and warnings2errors, so this has to be merged after we fix the issues with ktest.

kernel/src/main/java/org/kframework/parser/ProgramLoader.java
```diff
@@ -131,10 +137,25 @@ public Term processPgm(Reader content, Source source, Sort startSymbol,
             } catch (IOException e) {
                 throw KExceptionManager.internalError("Error reading from binary file", e);
             }
-        } else {
+        } else /*if (whatParser == ParserType.NEWPROGRAM)*/ {
+            Definition def = loader.loadOrDie(Definition.class, context.files.resolveKompiled("definition-concrete.bin"));
```
dwightguth, line 141: you're going to revert this part, right?

radumereuta, line 141: Yes, this is the part I was talking about.

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/KSyntax2GrammarStatesFilter.java
```diff
@@ -97,7 +134,17 @@ public static void processProduction(Production prd, Grammar grammar) {
                         + prdItem.getClass().getName();
             }
         }
-        RuleState labelRule = new RuleState("AddLabelRS", nt, new WrapLabelRule(prd));
+        Pattern pattern = null;
+        if (prd.att().contains("token")) {
+            // TODO: calculate reject list
+            if (prd.att().contains("autoReject") && prd.att().contains("rejectt"))
```
dwightguth: some typos here in the names of the attributes? maybe you should make constants?

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Rule.java
```diff
@@ -72,9 +73,16 @@ public MetaData(Location start, Location end, CharSequence input) {
      */
     public static class WrapLabelRule extends KListRule {
         private final Production label;
+        public final Pattern pattern;
```
dwightguth: should this be named reject?

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Rule.java
```diff
@@ -82,6 +90,9 @@ protected KList apply(KList klist, MetaData metaData) {
             if (label.att().contains("token")) {
                 // TODO: radum, figure out how to reject constants from here.
```
dwightguth: is this todo still relevant?

dwightguth:
I have comments to make about the code you wrote to integrate this code with kast, because it has multiple undesirable inefficiencies... but we can deal with that later, since it's going to get reverted for now. The rest looks okay, but I didn't look at it too hugely carefully. You should get Cosmin to review also.

radumereuta:
Did the renamings and extracted constants out of those strings. FYI, that wasn't a typo. `reject` right now is being used differently with SDF, so I was trying to avoid the name conflict, since the two features are incompatible. I renamed it to `reject2` since will make a bit more sense. Once we get rid of  SDF, we will rename it back to simply `reject`.

@cos please have a look also.

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinStringOperations.java
```diff
@@ -60,15 +60,23 @@ public static IntToken ord(StringToken term, TermContext context) {
...
         int codePoint = term.intValue();
-        StringUtil.throwIfSurrogatePair(codePoint);
-        char[] chars = Character.toChars(codePoint);
-        return StringToken.of(new String(chars));
+        try {
+            StringUtil.throwIfSurrogatePair(codePoint);
+            char[] chars = Character.toChars(codePoint);
+            return StringToken.of(new String(chars));
+        } catch (IllegalArgumentException e) {
+            return null;
```
cos, line 68: what does returning `null` mean?

k-distribution/src/test/resources/convertor-tests/syntaxPriorities-expected.k
```diff
@@ -4,10 +4,10 @@ module TEST [org.kframework.attributes.Location(#token(Int,"3"),#token(Int,"1"),
...
-  syntax Foo ::= "m" [org.kframework.attributes.Location(#token(Int,"9"),#token(Int,"18"),#token(Int,"9"),#token(Int,"20"))]
+  syntax Foo ::= "z" [#klabel(#token(KString,"'z")) org.kframework.attributes.Location(#token(Int,"6"),#token(Int,"18"),#token(Int,"6"),#token(Int,"20"))]
+  syntax Foo ::= "y" [#klabel(#token(KString,"'y")) org.kframework.attributes.Location(#token(Int,"5"),#token(Int,"18"),#token(Int,"5"),#token(Int,"40")) someattribute(#token(AttributeValue,""))]
+  syntax Foo ::= "x" [#klabel(#token(KString,"'x")) org.kframework.attributes.Location(#token(Int,"4"),#token(Int,"18"),#token(Int,"4"),#token(Int,"20"))]
+  syntax Foo ::= "u" [#klabel(#token(KString,"'u")) off(#token(AttributeValue,"")) org.kframework.attributes.Location(#token(Int,"11"),#token(Int,"18"),#token(Int,"11"),#token(Int,"29"))]
+  syntax Foo ::= "t" [#klabel(#token(KString,"'t")) andanohter(#token(AttributeValue,"")) org.kframework.attributes.Location(#token(Int,"8"),#token(Int,"18"),#token(Int,"8"),#token(Int,"36"))]
+  syntax Foo ::= "m" [#klabel(#token(KString,"'m")) org.kframework.attributes.Location(#token(Int,"9"),#token(Int,"18"),#token(Int,"9"),#token(Int,"20"))]
```
cos, line 12: I see why these failed on your machine now. You changed it such that productions which don't explicitly say `#token` are `KApply`s, so they're are assigned a label.

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
```diff
@@ -287,19 +314,83 @@ public Value applyAssoc(String assocOrig) {
...
+        res.add(prod2);
+        res.add(prod3);
+        res.add(prod4);
+        res.add(SyntaxSort(Sort(sort.name() + "Terminator")));
+        res.add(SyntaxSort(Sort("Ne" + sort.name())));
+        if (!nonEmpty) {
+            res.add(prod5);
+        }
+    }
+
```
cos: I know we don't care deeply about these convertors (as we're gonna get rid of them), but would it be easy to not duplicate the entire `applyUserList`?

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/TreeCleanerVisitor.java
```diff
@@ -24,14 +24,14 @@
     @Override
     public Either<Set<ParseFailedException>, Term> apply(TermCons tc) {
         Either<Set<ParseFailedException>, Term> vis;
-        if (tc.production().klabel().isEmpty()) {
+        if (tc.production().isSyntacticSubsort() && tc.production().klabel().isEmpty()) {
```
cos, line 27: I know `isSyntacticSubsort` sneaked into the `Module` some time ago, but we should do something about it before it spreads too much. Any ideas?

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinStringOperations.java
dwightguth, line 68: it indicates an undefined value for a partial function.

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/TreeCleanerVisitor.java
dwightguth, line 27: It shouldn't be too hard to create a class that takes a concrete-syntax definition and computes this table...

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinStringOperations.java
cos, line 68: Could we use `Optional` instead?

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/TreeCleanerVisitor.java
cos, line 27: It seems to me we're putting in the same bucket, and using the same notation, for two different very concepts: subsorts and productions. One way to solve this elegantly would be to add a special keyword for sort matters. This would not only solve this issue, but would allow all remaining productions to always have a klabel, a desirable property we're now missing.
So, instead of this:

```
syntax Exp
syntax Int < Exp
syntax Exp ::= Exp "+" Exp   [klabel("_+_")]
syntax Foo ::= Exp                [klabel("Foo")]
```

we woud have this:

```
sort Exp
sort Int < Exp
syntax Exp ::= Exp "+" Exp   [klabel("_+_")]
syntax Foo ::= Exp                [klabel("Foo")]
```

This would make it crystal-clear which are subsorts, and which are productions. Futhermore, in full-K, we would simply have:

```
sort Exp
sort Int < Exp
syntax Exp ::= Exp "+" Exp
syntax Foo ::= Exp
```

with the klabels inferred in both cases, and without worrying that the user might have forgot to put a kabel.

Again, regarding the desirable property, now all productions have a klabel, and all subsorting (`sort Exp` is essentially `sort Exp < K`) does not.

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
radumereuta: It's a completely different pattern, and I would like it to be in two different functions because one will be used for definitions, and the other will be used for programs. Having them enclosed in separate entities will make it easier to read it.

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/TreeCleanerVisitor.java
radumereuta, line 27: But from a parsing perspective, they are actually the same. The parsing community doesn't even call them subsorts, but chain productions.
We like BNF, for the fact that it includes nicely subsorts, without needing us to add new keywords and notation.

Regarding the syntacticSubsorts, you said that the parser should only take in KORE. Since I need that information and you don't want it in KORE, does it mean my parser doesn't take in KORE? (I'm just trying to get to your understanding by stretching the argument a bit).

radumereuta, line 27: P.S. not all productions have KLabels, remember `bracket`s.

cos, line 27: I understand the parser kernel doesn't know about subsorts -- and, considering this, we should not have the concept of "syntactic subsort". Still, looking at the entire parser, it does knew about subsorts -- it sees them as productions when feeding them to the kernel, and it filters them out with the post-parsing filters.

`brackets` are entirely a parsing concept.

Probably we should go with @dwightguth's suggestion and move some of this functionality to a module wrapper.

Still, my suggestion regarding the `sort` keyword stands. We should make a decision regarding this -- either we embrace sorts or we dive into doing everything with injections. I, at least, am not comfortable with  "in the middle".

radumereuta, line 27: This looks like a discussion that we need to have in the K meeting.
So I guess we'll delay this PR for tomorrow.

grosu, line 27: No, we had this discussion already several times and I am literally tired of it.  I do not want it to slow down any second anymore.  We've decided that we use productions for everything.  If there is something that I've done research on and I  never want to do it again, that is order-sorted algebra.  You guys have no idea what complications you can run into with subsorts.  Look at Maude.  So please, do not bring this up anymore.  The issue of whether we want to have a klabel for everything in KORE is orthogonal and yes, I think we should have such a label for everything.  We then need an "empty" label, that we can associate to "subsorts", brackets, and maybe other productions, too.  How about something like this?

```
syntax Exp ::= Int  [klabel()]
                     | "(" Exp ")"  [klabel()]
                     | "begin" Exp "end"  [klabel()]
```

And now we do not even need `bracket` anymore :-)

grosu, line 27: Not to mention that we actually do not even need to have such a special empty label, because we can always do this in KORE:

```
syntax Exp ::= Int  [klabel(Int2Exp)]
              | "(" Exp ")"  [klabel(parentheses)]
              | "begin" Exp "end"  [klabel(begin-end)]
rule Int2Exp(I) => I  [macro]
rule parentheses(E) => E [macro]
rule begin E end => E  [macro]

```

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
dwightguth: Cosmin is right, this is significantly the same function. You should refactor out the common parts.

kernel/src/main/java/org/kframework/parser/ProgramLoader.java
```diff
@@ -131,6 +137,21 @@ public Term processPgm(Reader content, Source source, Sort startSymbol,
             } catch (IOException e) {
                 throw KExceptionManager.internalError("Error reading from binary file", e);
             }
+        } else if (whatParser == ParserType.NEWPROGRAM) {
+            Definition def = loader.loadOrDie(Definition.class, context.files.resolveKompiled("definition-concrete.bin"));
```
dwightguth, line 141: you should get this object from dependency injection so that the K server can cache the load.

kernel/src/main/java/org/kframework/parser/ProgramLoader.java
```diff
@@ -131,6 +137,21 @@ public Term processPgm(Reader content, Source source, Sort startSymbol,
...
+            Module synMod = new KILtoKORE(context, true).apply(def).getModule(def.getMainSyntaxModule()).get();
+            ParseInModule parser = RuleGrammarGenerator.getProgramsGrammar(synMod);
```
dwightguth, line 143: I would be interested to see a performance metric on how long this function takes to execute for medium-to-large-size configurations.

kernel/src/main/java/org/kframework/parser/ProgramLoader.java
radumereuta, line 141: Can you point me to an example?

dwightguth, line 141: hmm, never mind. This is slightly nontrivial to deal with, I'd prefer not to get you bogged down in it. I can address this later when need be.

kernel/src/main/java/org/kframework/parser/ProgramLoader.java
radumereuta, line 143: You mean parsing in total? Or only generating the grammar for programs?

dwightguth, line 143: I mean generating the grammar.

radumereuta, line 143: I'm only parsing programs right now, so are some quick tests I made:

```
lambda.k, identity.lambda
Gen   = 43
Parse = 50
```

```
imp.k, sum.imp
Gen   = 17
Parse = 71
```

```
simple-untyped.k, factorial.simple
Gen   = 20
Parse = 118
```

```
kool-untyped, matrix.kool (2.2KB)
Gen   = 21
Parse = 849
```

Measuring time to generate the parser and to parse a program (in ms). I did this only once so I don't know how much this may vary. One explanation of why matrix.kool takes so long is because I haven't implemented the efficient way of parsing lists at this moment (but I'm not sure). Usually, calling the same parser the second time goes much faster since everything is cached in memory.

grosu, line 143: This times are OK for now, I would not get stuck with trying to improve them.  Let us have the new parser incorporated in the framework asap and we will think about improving it later, when its performance becomes a bottleneck.

dwightguth, line 143: okay. Yeah, 20 milliseconds isn't too bad. If it's higher for C, we may do something, but this is fine for now.

dwightguth:
lgtm

Issue #1382, opened 2015-03-06
Non confluent rules in kool-typed-static


dwightguth:
in kool-typed-static, if the rule at line 709 applies before the extendsAll cell has finished populating, you can get a stuck state even though the classes actually are compatible. I ran into this when I was modifying the order in which rules are applied, so it would be good if you could fix this ASAP to get me unblocked.

PR #1388, opened 2015-03-09
Fixing non confluence bug #1382

grosu:
@dwightguth Here is a naive fixing of the non-confluence bug.  Please see if it works, because I didn't have your test.  I also pushed a quick fixed MD typo in a README, please let it go through, too.


dwightguth:
I remember I tested this exact fix and it seemed to work, because the extendsAll rules are structural and therefore apply before any transition. I agree it seems like strategies are the most desirable solution for this problem. In any event, I will merge this now.

PR #1393, opened 2015-03-10
Extract graph functionality from ExecutorDebugger.

traiansf:
I need a way to keep graphs compact for the model checker.
In my view, this should be the default behavior for KRunGraph, but I decided to keep it as a separate class for now.

@dwightguth please review


dwightguth:
You've tested this with the debugger and things look okay at a glance? if so, go ahead and merge this.

traiansf:
@dwightguth I've fixed the debugger-related bug I had introduced and now the debugger seems to be working as well as it did before (See Issue #1418).  Do you want to re-review?

PR #1396, opened 2015-03-11
fix assertion error caused by incorrect rule with multiple cells of the ...

dwightguth:
...same label

@yilongli please review


yilongli:
lgtm

PR #1408, opened 2015-03-15
Refactoring and work on tiny

cos:
- aside from tiny stuff, the main change is that the KIL-to-KORE convertors removes the quote, i.e., `'` from all productions and labels. As we now have a different escaping mechanism, there is no point in leaving it there.

### DIFF
--- a/k-distribution/src/test/java/org/kframework/kore/convertors/TstTinyOnKORE.java
+++ b/k-distribution/src/test/java/org/kframework/kore/convertors/TstTinyOnKORE.java
@@ -60,7 +60,7 @@ public class TstTinyOnKORE extends BaseTest {
                                 ))),
                 cons.KLabel("'<state>").apply(
                         cons.KLabel("'_Map_").apply(
-                                cons.KLabel("'_|->_").apply((K) cons.stringToId("n"), (K) cons.intToToken(10)),
+                                cons.KLabel("'_|->_").apply((K) cons.stringToId("n"), (K) cons.intToToken(200)),
                                 cons.KLabel("'_|->_").apply((K) cons.stringToId("s"), (K) cons.intToToken(0)))
                 ),
                 cons.KLabel("'<s>").apply()
--- a/k-distribution/src/test/java/org/kframework/kore/convertors/TstTinyOnKORE.java
+++ b/k-distribution/src/test/java/org/kframework/kore/convertors/TstTinyOnKORE.java
@@ -60,7 +60,7 @@ public class TstTinyOnKORE extends BaseTest {
                                 ))),
                 cons.KLabel("'<state>").apply(
                         cons.KLabel("'_Map_").apply(
-                                cons.KLabel("'_|->_").apply((K) cons.stringToId("n"), (K) cons.intToToken(100)),
+                                cons.KLabel("'_|->_").apply((K) cons.stringToId("n"), (K) cons.intToToken(10)),
                                 cons.KLabel("'_|->_").apply((K) cons.stringToId("s"), (K) cons.intToToken(0)))
                 ),
                 cons.KLabel("'<s>").apply()
--- a/k-distribution/src/test/java/org/kframework/kore/convertors/TstTinyOnKORE.java
+++ b/k-distribution/src/test/java/org/kframework/kore/convertors/TstTinyOnKORE.java
@@ -12,6 +12,7 @@ import org.kframework.tiny.Constructors;
 import org.kframework.tiny.FreeTheory;
 import org.kframework.tiny.K;
 import org.kframework.tiny.KApp;
+import org.kframework.tiny.KIndex$;
 import org.kframework.tiny.Rewriter;
 import org.kframework.tiny.Rule;
 import org.kframework.tiny.package$;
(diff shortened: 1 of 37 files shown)


cos:
@dwightguth, I need a project published to the RV maven repo -- the developer didn't put the publish the latest version to Maven Central.

The project is here:
https://github.com/nicolasstucki/multisets

It's a Scala project so it needs to be published using SBT. You can do that by changing these lines:
https://github.com/nicolasstucki/multisets/blob/2.11.x/build.sbt#L19-L25

I changed the selection to:

```
publishTo := Some(Resolver.file("file",  new File(Path.userHome.absolutePath+"/.m2/repository")))
```

for my local Maven repo. You may need to do something slightly different for RV.

dwightguth:
Assuming that I don't know SBT and don't really want to learn at this time, I am fine with running a set of commands with my privileges if you give me an exact set of files and commands to run on them; but I don't really want to spend significant amounts of effort trying to figure out what those commands should be.

The KORE stuff looks fine; I haven't looked over the rewriter changes very closely. I'm fine with you handling that however you want.

cos:
Are you running the publish command directly on the server which has the maven repo? If so, the commands are:

```
git clone git@github.com:nicolasstucki/multisets.git
cd multisets
```

In `build.sbt`, replace these lines: https://github.com/nicolasstucki/multisets/blob/2.11.x/build.sbt#L19-L25
with

```
publishTo := Some(Resolver.file("file",  new File(Path.userHome.absolutePath+"/.m2/repository")))
```

if the maven repo is in its standard directory.

Then:

```
sbt publish
```

dwightguth:
Jenkins: test this please

radumereuta:
There are issues in this diff.
First of all, I don't agree with how @cos uses `dropQuote` and `addQuote`. If a production has a klabel on it, then that is the klabel. No processing should be done after that.

Secondly there are print messages that were forgotten, and when I run something I get all sorts of strange printouts.

I think this should be reverted, fixed and then made another pull request.

cos:
> First of all, I don't agree with how @cos uses dropQuote and addQuote. If a production has a klabel on it, then that is the klabel. No processing should be done after that.

@radumereuta, the fact klabels have quotes is an artifact of the SDF parser. Generally, we don't want definitions with quotes and, with the new parser, this is possible. If we see the quote as the correspondent of the back quote in the new parser, they should be removed when converting to KORE.

> Secondly there are print messages that were forgotten, and when I run something I get all sorts of strange printouts.

Very likely I missed something, put what you see in a comment and I'll make a pull request with the fix.

> I think this should be reverted, fixed and then made another pull request.

I disagree.

dwightguth:
No, Radu is right about the quote. It is possible to have no quotes in the new parser, but nonetheless, the quote does not correspond to the backtick, because not all klabels have quotes, and a klabel with and without a quote is not the same klabel. I missed this when I originally reviewed this diff, but I think we should try to find another migration pattern than to put this in the converter. The correct solution, I believe, is for the K frontend to not generate quotes as part of the default (ie, non-user-specified) klabels for productions.

radumereuta:
But I need the klabels to do disambiguation. This is how I apply the priority filters and associativity.
If a user says `syntax priority '_*_ > '_+_` I have to be able to know which productions he/she is referring to.

Adding klabels to a production is in fact a step separate from the transformers. It should be done immediately after outer parsing and it is there where all the decisions about how the default labels look, should be done. The rest of the tool should take them as they are.

grosu:
Guys, if in doubt, always consult the wiki https://github.com/kframework/k/wiki/KAST-and-KORE and the files that it links to: https://github.com/kframework/k/blob/master/kernel/src/main/resources/kore.k

We've spent dozens of hours revising these documents, and arguing and then arguing again about them.  We have not done that for nothing.  So please do not just take adhoc decisions after thinking 5 minutes about the issue.  If you are not happy with something in the wiki, then complain about that and propose something better and get that approved, or otherwise understand the big picture and the motivations behind it better.

The short story is that the quote `'`, which is an artifact of our previous use of Maude's QIDs, will play no role whatsoever in KORE.  If users want to use it in their explicitly given KLabels, fine, it is their choice.  But we do not force them.  Also, we do have backticks now for grouping, including for grouping the characters in a KLabel.  So one can write ``_+_`` for now on a KLabel position if `_+_` is ambiguous with the current parser; future parsers, even if they will be able to handle `_+_`, will still work when writing ``_+_``.  This will also nicely allow us to have whitespaces in KLabels, as needed for some languages.

cos:
The issue is that we don't want the quote in KORE, and SDF does not allow us to have klabels without the quote. My solution was to see the quote as a convention (like the backquote), and have the KILtoKORE convertor remove them. This way we have clean KORE definitions, and we do not need to touch or affect the current KIL-based pipeline in any way. Remember that KILtoKORE is just our testing mechanism. It is not part of the legacy pipeline and it will not be part of the new pipeline. If you see any other solution (aside from using the new parser also for bubbles, which will be great) that both gives us quote-less klabels and does not touch the current pipeline, I'll gladly take it.

radumereuta:
1. We do want to use quote in KORE. We should not force users not to have something like that.
2. SDF allows you that, see predicates
3. quote is a convention only when generating default labels, and that's it.
4. depending on how issue #1411 goes, KILtoKORE will be part of the pipeline (at least for a while).

cos:
1. The change affects KILtoKORE. KILtoKORE is our testing mechanism. No users will ever get to use it. They will use the SDF-based parser until the new parser is ready. And they will use the new parser afterwards. There will be no intermediate step. As for us, I think we can live with not having labels starting with quote until the new parser is ready.
2. I tried it and it fails. 
3. I tried manually (i.e. by specifying it) using a quote-less label and I'm getting parsing errors. 
4. I disagree. #1411 has nothing to do with it as it regards the program parser. We're talking here about the new parser for the new pipeline. KILtoKORE should not be part of that. We should either bootstrap the new parser, or, harder, write a javacc version of outer. As for the current pipeline, I don't see any intermediate solution. If you want to make an intermediate solution that uses the new parser for the current pipeline, you'll also have to write a KOREtoKIL (or parser data structures to KIL) convertor, which is probably hard, very time-consuming, and error prone.

cos:
I just saw #1410. Unless you're offering an alternate solution for generating clean KORE, please re-merge.

dwightguth:
As I understand it, it broke the new parser. I can say with confidence you are incorrect about your #3 though. It is definitely possible to have productions with a klabel without a quote, and they are distinct from the same klabel with a quote. This is already done numerous times in existing definitions. The quote is not an indicator that what follows is a klabel: it is part of the klabel itself, and a convention for generating klabels from productions that do not specify them. As such, as I suggested, I feel the correct solution is to modify the outer parser to change its default klabel algorithm in the new implementation.

cos:
Regarding #3, I tried what seems natural, using `klabel(foo)` as an attribute. This doesn't work. As I understand from Radu, this is not way it is done because of yet another quirk. Anyway.

Regarding "breaking" the new parser, it just breaks using the new parser as a program parser for the legacy pipeline -- Radu's pending pull request. I acknowledge this as an issue, and I think it should have been raised like this, not by going into philosophical discussions about the correctness of the approach in general. The fix is actually very easy, just add an option to to KILtoKORE to determine whether to drop the quote or not. This is growing way out of proportion for a temporary convertor.

cos:
Also, I don't it is worth going into the trouble of adapting the current outer parser just for this. That is code that hasn't been touched in some time and my guess is that we don't want to mess with it. We want to rewrite the outer parser for the generating KORE modules anyway.

cos:
the fix, about 10 lines of code (not counting merging issues partially introduced by the revert): #1412

PR #1412, opened 2015-03-17
Feature/refactoring and work on tiny

cos:
again


radumereuta:
lgtm, @dwightguth please review.
We decided to add a switch which will allow both me and @cos to use the same implementation of the transformers.

PR #1417, opened 2015-03-20
minor optimizations to java backend

dwightguth:
**please review**


dwightguth:
@yilongli please review

The second commit here is some wizardry intended to hack together a solution to the excessive conversion in the old K implementation. We won't need this trick with KORE, which is good, but since it was really fast to write code for and makes linking C programs **substantially** faster, I want to get this merged now.

yilongli:
lgtm

PR #1421, opened 2015-03-25
more specific exclude pattern

daejunpark:
Since `exclude` matches the given string on FULL file paths, the exclude pattern `io` is too general, which may lead to exclude unexpected ones. For example, `k-distribution` matches the exclude pattern `io`, so that it excludes all files under the path `k-distribution/tutorial/`.

Well, the suggested fix `io.imp` may be still problematic. So, `ktest` would be better to report explicitly which files are excluded.

### DIFF
--- a/k-distribution/tutorial/1_k/4_imp++/lesson_7/exercises/abort/tests/config.xml
+++ b/k-distribution/tutorial/1_k/4_imp++/lesson_7/exercises/abort/tests/config.xml
@@ -5,7 +5,7 @@
   <include file="../../../tests/config.xml"
            more-programs="."
            more-results="."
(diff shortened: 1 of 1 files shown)


daejunpark:
@grosu Please review this.

grosu:
ok with me for now, although we need to redesign ktest, as discussed.

cos:
I'm closing and deleting the branch. The change is tiny so it can be recreated easily if needed. 

Pasting it here:
k-distribution/tutorial/1_k/4_imp++/lesson_7/exercises/abort/tests/config.xml

```
-           exclude="io" >  <!-- Because of bug #159-->
+           exclude="io.imp" >  <!-- Because of bug #159-->
```

PR #1426, opened 2015-03-31
assoc attribute for function rule -> AC matcher.

traiansf:
ready for review


traiansf:
This PR adds support for using the AC matcher to match functional patterns if the **assoc** attribute is explicitly specified in the rule.

@dwightguth please review

java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
```diff
@@ -444,11 +437,31 @@ public Term evaluateFunction(KItem kItem, boolean copyOnShareSubstAndEval, TermC
...
                                 System.err.println("\nAuditing " + rule + "...\n");
                             }
-                            /* function rules should be applied by pattern match rather than unification */
-                            Map<Variable, Term> solution = NonACPatternMatcher.match(kItem, rule, context);
-                            if (solution == null) {
-                                continue;
-                            }
+
+
+                                Map<Variable, Term> solution;
```
andreistefanescu: wrong indentation

kernel/src/main/java/org/kframework/kil/Attribute.java
```diff
@@ -21,6 +21,7 @@
 
     public static final String BUILTIN_KEY = "builtin";
     public static final String FUNCTION_KEY = "function";
+    public static final String ASSOCIATIVE_KEY = "assoc";
```
andreistefanescu, line 24: maybe `assoc-comm`, or both `assoc` and `comm`?

dwightguth, line 24: I'm inclined to agree with the latter, actually. Probably make either one have the same effect.

java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
```diff
@@ -444,11 +437,31 @@ public Term evaluateFunction(KItem kItem, boolean copyOnShareSubstAndEval, TermC
...
+                                        continue;
+                                    }
+                                } else {
+                                    List<Substitution<Variable, Term>> matches = PatternMatcher.match(kItem, rule, context);
+                                    if (matches.isEmpty()) {
+                                        continue;
+                                    } else {
+                                        if (matches.size() > 1) {
+                                            if (javaOptions.deterministicFunctions) {
+                                               throw KExceptionManager.criticalError("More than one possible match. " +
```
dwightguth: I would say make this a warning...

java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
traiansf: I my defense I would say it was indented badly to begin with (a try block was added without indenting its body).

java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
traiansf: It is a warning when the deterministic functions option is not used.  I thought it should be stronger when checking that functions are deterministic.  If you want a warning, should I collapse them, or there should be in different warning levels?

kernel/src/main/java/org/kframework/kil/Attribute.java
traiansf, line 24: so, should it have both assoc and comm, or assoc-comm, or any of the three would work?

dwightguth, line 24: I think probably since KORE has the assoc and comm attributes to indicate associativity and commutivity, and since the AC matching algorithm is used for both, we should accept either assoc or comm.

java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
dwightguth: hmm. I guess since that is what the option is already doing, an error is okay here. But you should probably change the assert statement into a KException also just for consistency.

kernel/src/main/java/org/kframework/kil/Attribute.java
traiansf, line 24: Ok, then I would argue against the additional assoc-comm rule, since we
would have "assoc comm", "assoc", and "comm" to cover all options.

2015-03-31 18:12 GMT+03:00 dwightguth notifications@github.com:

java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
traiansf: Ok, I've changed the assert (I believe you mentioned) into a critical error.

2015-03-31 18:15 GMT+03:00 dwightguth notifications@github.com:

traiansf:
I think I've addressed all comments.  @dwightguth please re-review.

dwightguth:
It looks like your test is failing with a parse error, can you fix it before I do the final review? I believe the issue is that attributes are separated by commas.

traiansf:
@dwightguth: fixed,  ready for final review

PR #1431, opened 2015-04-01
modify error message on uncaught exception to provide better instruction...

dwightguth:
...s to users reporting bugs

@yilongli please review


yilongli:
lgtm

PR #1434, opened 2015-04-04
change Unparser.java to output new format kast

laurayuwen:
do not review.


dwightguth:
I don't think this is the right way to go about this. What we really want is for there to be a transformation that converts KORE into terms referencing productions. Then you create a new implementation of the unparser following the same paradigm as this one, but operating over the new classes in treeNodes.scala.

laurayuwen:
So are things other than the unparser you mentioned ready to use now?
And if this file does not lead to the right new KAST syntax, why keep this one?

bmmoore:
I'm planning to implement a printer in the new syntax for the Kore classes soon, and we should be able to use that plus the KILToKORE transformer to get the output. Even if we wanted to do more work on KIL classes, Dwight says there's a KoreFilter class that would be a better place than Unparser to make these changes.

PR #1438, opened 2015-04-08
Removing imp-hoare, imp-reachability, and imp-symbolic from samples.

andreiarusoaie:
As part of the cleaning up samples I removed imp-hoare, imp-reachability and imp-symbolic. All these were supposed to be tests for the old symbolic execution engine which is not present anymore in the tool.

Do not review yet.


andreiarusoaie:
@dwightguth please review

PR #1445, opened 2015-04-15
add code to generate  configuration syntax from declaration + refactor K...

dwightguth:
...ompile

**please review**


dwightguth:
@cos @bmmoore please review. Note that I did not add the code to generate rules for the initializers yet. It only generates the productions at the moment, because that was all I needed for parsing. We can do the rest later.

kore/src/main/scala/org/kframework/attributes/Att.scala
```diff
@@ -64,7 +64,7 @@ case class Att(att: Set[K]) extends AttributesToString {
   def stream = att.asJava.stream
   def addAll(that: Att) = this ++ that
 
-  def remove(k: String): Att = new Att(att filter { case KApply(KLabel(k), _) => false; case _ => true })
+  def remove(k: String): Att = new Att(att filter { case KApply(KLabel(`k`), _) => false; case _ => true })
```
cos, line 67: ups. good catch.

cos:
is `GenerateSentencesFromConfigDecl` tested in any way? There are no unit tests and `kore_imp_tiny.k` doesn't seem to be touched.

dwightguth:
It actually was tested, but I have pushed a change to kore_imp_tiny just to double-check.

kernel/src/main/java/org/kframework/kore/compile/GenerateSentencesFromConfigDecl.java
```diff
@@ -0,0 +1,195 @@
...
+            KApply kapp = (KApply) body;
+            if (kapp.klabel().name().equals("#configCell")) {
+                if (kapp.klist().size() == 4) {
+                    K startLabel = kapp.klist().items().get(0);
+                    K endLabel = kapp.klist().items().get(3);
+                    if (startLabel.equals(endLabel)) {
+                        if (startLabel instanceof KToken) {
+                            KToken label = (KToken) startLabel;
+                            if (label.sort().equals(Sort("#CellName"))) {
+                                String cellName = label.s();
```
bmmoore, line 86: It's harder to read this complicated processing inside the deep nesting, maybe extract a method taking minimal parameters.

kernel/src/main/java/org/kframework/kore/compile/GenerateSentencesFromConfigDecl.java
```diff
@@ -0,0 +1,215 @@
...
+                                String sortName = getSortOfCell(cellName);
+                                Sort sort = Sort(sortName);
+                                Att cellProperties = getCellPropertiesAsAtt(kapp.klist().items().get(1), cellName, ensures);
+                                ConfigurationInfo.Multiplicity multiplicity = convertStringMultiplicity(
+                                        cellProperties.<String>get("multiplicity"), body);
+                                Tuple2<Set<Sentence>, List<Sort>> childResult = genInternal(
```
bmmoore: I don't understand this recursion. Maybe commenting genInternal would help.

bmmoore:
Looks fine to me now. You might still clean up that `computeSentencesOfWellFormedCell` method a bit more.

cos:
lgtm.

cos:
my main concern was the lack of any testing, I see you changed kore_imp_tiny.k. lgtm now.

PR #1446, opened 2015-04-15
try to make variables include _

dwightguth:
**please review**

### DIFF
--- a/k-distribution/include/builtin/kast.k
+++ b/k-distribution/include/builtin/kast.k
@@ -44,7 +44,7 @@ endmodule
 // To be used when parsing/pretty-printing symbolic configurations
 module KSEQ-SYMBOLIC
   imports KSEQ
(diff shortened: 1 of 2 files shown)


dwightguth:
@radumereuta seems to work just fine. Are you okay with this being merged? it will significantly benefit my ability to verify C programs.

radumereuta:
lgtm, but this is a K design decision, so @grosu do you approve that variables can contain underscore in their name?

grosu:
But they already do, don't they?  _1, _2, etc., are variables generated by unification in combination with the ?X, ?Y, etc., variables.  Actually, I think people should even be allowed to use _1, _2, etc., in their rules the same way as they use _ alone.

Grigore

radumereuta:
@andreistefanescu can you clarify what _1, _2 are? Because I can tell you they weren't parsed as variables.

dwightguth:
I believe they are variables, they are just not parsed by the semantics (being generated by the rewriter) and are therefore not subject to the syntax of variables.

PR #1449, opened 2015-04-21
Adding back the ability to parse programs with the new parser.

radumereuta:
- fixed examples and regular expressions.

PR #1450, opened 2015-04-21
Add primitive new kompile frontend + substantial improvements to incremental performance

dwightguth:
**please review**


kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Grammar.java
```diff
@@ -148,16 +149,16 @@ public void addWhiteSpace() {
      * @param start NextableState to which to attach the whitespaces
      * @return the remove whitespace state
      */
-    private NextableState addWhitespace(NextableState start) {
+    private NextableState addWhitespace(NextableState start, NonTerminal nt) {
```
radumereuta: Why?

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Grammar.java
```diff
@@ -232,33 +233,8 @@ public void compile() {
...
-                                nts.child, new HashSet<NonTerminalState>(Arrays.asList(nts)));
-                        }
-                        reachableNonTerminals.get(nts.child).add(nts);
-                        collectNTCallers(((NonTerminalState) st).child.entryState,
-                            visited, reachableNonTerminals);
-                    }
-                    collectNTCallers(st, visited, reachableNonTerminals);
-                }
-            }
-        }
```
radumereuta, line 261: I'm pretty sure this was useful, please explain.

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Parser.java
```diff
@@ -712,15 +712,16 @@ boolean addRule(Function that, final Rule rule, final StateReturn stateReturn, f
     ////////////////
 
     private final ParseState s;
+    private final Grammar g;
```
radumereuta: Why?

dwightguth:
@cos @bmmoore @radumereuta  please review. This diff takes the first-time compilation time for the C semantics down from 3:00 to 2:20, and the incremental compilation time from 2:20 to 0:20. Remaining performance issues seem to have to do with serialization and with the compiler pipeline. I would like to try serilizing K to KAST strings to see if the performance is better, but this will do for now. @cos, you might want to put some effort into thinking of a way to make the compilation infrastructure faster because as we add more compiler passes it will get slower and slower, and it seems to be a significant performance overhead.

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Grammar.java
dwightguth: I was getting significant performance penalties from the process by which java serialization handles cycles. I solved this by removing cycles from the Grammar and using unshared serialization. One disadvantage of this approach is that it cannot support NFAs with infinite languages, but we do not currently support any syntax that would cause this to occur in production.

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Grammar.java
dwightguth, line 261: same as above; since all nonterminals are now indexed in the grammar object, this method is no longer necessary.

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Parser.java
dwightguth: see above

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Grammar.java
radumereuta: What? Don't serialize `Grammar` objects. To generate the parser from a K definition takes 15ms. There's no way you can get such speeds from serialization.
You should serialize the K definition instead, which you have to do anyway. This is also comparable for cache purposes. I am not convinced of this change.

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -121,11 +127,9 @@ public CompiledDefinition run(File definitionFile, String mainModuleName, String
     }
 
     public Definition parseDefinition(File definitionFile, String mainModuleName, String mainProgramsModule, boolean dropQuote) {
-        String definitionString = files.loadFromWorkingDirectory(definitionFile.getPath());
-
         Definition definition = parser.loadDefinition(
                 mainModuleName,
-                mainProgramsModule, REQUIRE_KAST_K + definitionString,
+                mainProgramsModule, REQUIRE_KAST_K + "require " + StringUtil.enquoteCString(definitionFile.getPath()),
```
radumereuta, line 132: This feels very wrong. You shouldn't connect modules by adding text to a string that has to be parsed. What if the user decides to change the syntax of K? This is something that we want to support in the future.

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Grammar.java
grosu: Sorry for jumping into the middle of this discussion without really knowing what's going on, but my eye caught the "serialization of the K definition" substring.  I really hope that you are not thinking of doing anything else here but printing textual KORE as "serializations" of K definitions.  That should be very very fast to parse (if it is not then we have a problem to fix) and it gives us the benefit that we can also inspect it manually, pass it around to other tools in other languages, etc.

Grigore

radumereuta:
I disapprove with half of this commit. I don't think the changes to the `ParserState` and `Grammar` objects are necessary. It makes the code more coupled and harder to read or maintain. You are imposing restrictions on grammar states to not have cycles, for serialization performance reasons. I have examples where parsing a list of objects with a cycle in the grammar performs much, much better than SDF. This is something that we would like to have in the tool at one point. It's not implemented currently because we didn't get to the point where this problem is clearly visible, but we will get there soon.

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Grammar.java
cos: I agree with @radumereuta. Considering we can generate grammars almost instantly, I see no point in serializing Grammar objects. As @grosu says, we should serialize K definitions to their KORE form. We can discuss crazy fast alternatives later on (there's this library I've been eyeing: https://github.com/scala/pickling), once the textual KORE version works. 

@radumereuta, does the KORE definition parser (kore.k file with the new parser) still work? If it does, we should just make sure the outer KORE pretty printing can be parsed -- it probably needs a few trivial fixes.

cos:
regarding performance, how much of the time (both direct and incremental) is spent in the compiler passes?

kernel/src/main/java/org/kframework/kompile/Kompile.java
cos, line 132: @radumereuta, let's leave this for now. We have to think carefully about and refactor Kompile anyhow.

radumereuta:
I can answer this. From my tests, about 70% of the time is spent doing only parsing stuff (generating grammars, parsing, disambiguating). I'm not sure exactly where the rest of 30% of the time comes from. I did these tests on the C semantics, where it takes more than 2 minutes to execute everything.

kernel/src/main/java/org/kframework/kompile/Kompile.java
radumereuta, line 132: I actually found an issue with doing this. The location information was off by one line because of adding require in the beginning. Not sure exactly if this fixed it.

cos:
How about out of the 20s incremental compilation?

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Grammar.java
radumereuta: Last time I checked it worked. I Think I had a small unit test for it.

radumereuta:
Well I guess 0, since there's no parsing involved.

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Grammar.java
dwightguth: When we switched to dk.brics.automaton, one of the reasons that the regular
expressions were so much faster is that unlike java.util.regex, the new
regex engine determinizes and minimizes the nfa into a minimal dfa. This
task is a performance bottleneck, solving of which was responsible for
roughly 20-second reduction in incremental parse time. For this reason, I
chose to serialize the grammar and what followed from that. Since this
seems to be unpopular, I will instead explore the possibility of
serializing the dfa as part of the production items. That said, it will
always be prohibitively slow to serialize regular expressions as text, and
as such, given the performance constraints involved, I suspect it will
never be a satisfactory default to serialize definitions to text. We need
to make decisions that scale rather than simply naively assuming that
performance can be dealt with later. Guess what? It's later, and I'm
dealing with it. One of the natures of optimization is that it inevitably
makes code at least somewhat less nice. I can try to avoid serializing the
grammar, see how that goes, but some sacrifices do have to be made; the
decision not to serialize the parser state to kast is one of them.

@cos: In response to your question, the 4 compiler passes that use
ModuleTransformer total about 6 seconds. Roughly 12 seconds is spent in
serialization/deserialization, and the remaining time is spent in
miscellaneous locations.

cos: Ok. 6s is not that bad but we could look into improving it a bit. One simple option would be to externalize (to disk) `ModuleTransformer`'s cache/memoization. Of course, it is simple assuming we do as you suggest and figure out a fast serialization/deserialization solution now instead of later. A more involved solution would be to figure out which parts of the compiler pipeline could benefit from incrementality -- but the simple caching solution might actually subsume most of this.

I'll be at RV tomorrow. We can brainstorm and maybe try out some solutions then.

grosu: Interesting.  Dwight makes good points regarding the NFA2DFA translation and the state of the parser.  We do not want to regenerate those.  But can we store them, serialized, as meta-data in the KORE definition?  For example, the DFA corresponding to a terminal can be store as metadata to the terminal or the production, and the parser state as meta-date to the module.  That way, we have both a human readable KORE definition (that the user can output at different stages of the kompilation to understand what's going on or to debug their definition), and efficient caching (for work done in previous kompilation stages).  Would that work?

Grigore

cos: @grosu, sure, it can work.

Still, considering the current speed of the parser, if we go from serializing the parser to serializing-to-K the module, it will have a very significant speed impact. I would guess that we will even slow it down compared to non-incremental as parsing full-K-outer is done with javacc, while our current KORE parser is the new parser.

If we want speed, I think the best solution is to: 
1. represent the regex in `SyntaxProduction` directly as a regex, not as a String.
2. Serialize the Modules using pickle (or some other fast serialization/deserialization solution). In theory, the serialized versions are still KORE (because the Module is KORE). If a user ever wants to inspect a serialized Module, we can easily make a convertor from serialized to `kore.k` syntax. Furthermore, we could have loading a definition work in a cascade, trying first `.k`, then `.kore`, then `.kled` (pickled). When we'll have a crazy-fast KORE parser, we'll simply not use `.kled` anymore.

kernel/src/main/java/org/kframework/kompile/Kompile.java
dwightguth, line 132: This fixes that.

radumereuta:
This should not be merged before #1449. My pull request adds more test for the new parser, and it seems there are some weird random failures.

kernel/src/main/java/org/kframework/parser/concrete2kore/generator/RuleGrammarGenerator.java
```diff
@@ -184,7 +186,7 @@ public ParseInModule getCombinedGrammar(Module mod) {
         return prods;
     }
 
-    public ParseInModule getProgramsGrammar(Module mod) {
+    public Module getProgramsGrammar(Module mod) {
```
radumereuta, line 191: This should be static, and not call `getCombinedGrammar` at the end.

radumereuta:
Ok, I guess you fixed all the tests. The diff looks much better now.

kernel/src/main/java/org/kframework/parser/concrete2kore/generator/RuleGrammarGenerator.java
dwightguth, line 191: it has to; otherwise if a user creates a syntax module that imports one of the "magic" modules, it will not behave as expected.

radumereuta, line 191: Then move the code from this method into a 'magic module' and import that module. Then we trully have only one function to create modules.

dwightguth, line 191: added a TODO

dwightguth:
@cos @bmmoore can you re-review now that I have cleaned this diff up to serialize regexes as part of the Terminal instead of serializing the entire grammar?

radumereuta:
lgtm

k-distribution/src/test/java/org/kframework/kore/convertors/TstBackendOnKORE.java
```diff
@@ -25,17 +29,22 @@
...
 import scala.collection.JavaConversions;
 
-import com.google.inject.Guice;
-import com.google.inject.Injector;
+import java.io.File;
+import java.io.IOException;
+import java.net.URISyntaxException;
 
 
 public class TstBackendOnKORE extends BaseTest {
```
cos, line 50: does the Java backend work on KORE again now? The test is not followed by _IT so it doesn't get automatically executed.

kernel/src/main/java/org/kframework/kore/compile/BooleanUtils.java
```diff
@@ -0,0 +1,24 @@
...
+
+import static org.kframework.kore.KORE.KApply;
+import static org.kframework.kore.KORE.KLabel;
+import static org.kframework.kore.KORE.KToken;
+import static org.kframework.kore.KORE.Sort;
+
+/**
+ * Created by dwightguth on 4/17/15.
+ */
+public class BooleanUtils {
```
cos: this sort of thing should probably live in org.kframework.builtin in the kore project

kernel/src/main/java/org/kframework/parser/concrete2kore/ParseCache.java
```diff
@@ -0,0 +1,60 @@
...
+import org.kframework.utils.errorsystem.ParseFailedException;
+
+import java.io.Serializable;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Created by dwightguth on 4/20/15.
+ */
+public class ParseCache implements Serializable {
```
cos, line 15: is this class still needed? At most, it seems we need a `ModuleCache` class.

cos:
`</comments>`

k-distribution/src/test/java/org/kframework/kore/convertors/TstBackendOnKORE.java
radumereuta, line 50: This is intended. It is a quick way to test the C semantics. I asked for this to be pushed to master, since it makes it so much easier to set up the C semantics until we have complete integration into kompile and krun.

cos, line 50: @radumereuta, you're answering a "does" question with "this is intended"...

As for using this test for C, if the Java backend has been fixed to work on KORE again, I think we should test it for IMP with jenkins. This was the original purpose of this test class.

radumereuta, line 50: Ah, I misread. Sorry.

dwightguth, line 50: I am inclined to agree, this test should be in jenkins. We can just add an @Ignore flag to the c semantics test once we fix that. It's outside the scope of this diff, however

kernel/src/main/java/org/kframework/parser/concrete2kore/ParseCache.java
dwightguth, line 15: If we intend to create a generic cache in common between parsing and compilation, we will have to unify the transformer framework, because currently, parser transformers can return warnings and compiler transformers can't. So I'm inclined to leave this as is, and allow you to refactor it later as necessary.

dwightguth:
@cos please re-review

k-distribution/src/test/java/org/kframework/kore/convertors/TstBackendOnKORE.java
cos, line 50: Again, does the Java backend work on KORE again now? Should Andrei and I look into it or have you already fixed it?

kernel/src/main/java/org/kframework/parser/concrete2kore/ParseCache.java
cos, line 15: ok. we'll postpone this.

k-distribution/src/test/java/org/kframework/kore/convertors/TstBackendOnKORE.java
dwightguth, line 50: I am working off of the C semantics, which does not yet pass successfully through the configuration concretization pass. So I have not yet started looking into the process of compatibility with the Java Rewrite Engine. I will probably get to that sometime next week.

cos, line 50: Ok. Just trying to figure out where we are.

PR #1453, opened 2015-04-23
Float2String(Float, String)

### LINKED ISSUE #1451: float format
Currently the builtin Float2String method prints float in e+ format:  instead of 0.0, I got 0e+00. 
This made some of my tests fail (not the same format as Java prints). Shall we get the traditional one back?
e.g.  instead of
      return value.toString();
we can have 
       return String.valueOf(value.doubleValue());
in method  printKFloat.

dwightguth:
fixes #1451 
@laurayuwen please test and verify that it works in a satisfactory manner. You may still have to special-case NaN and Infinity, but there's not much I can do about that.

### DIFF
--- a/k-distribution/include/builtins/string.k
+++ b/k-distribution/include/builtins/string.k
@@ -28,31 +28,31 @@ module STRING-HOOKS
   syntax Bool ::= String "==String" String [function, left, hook(#STRING:_==String_)]
 
   syntax Int ::= "lengthString" "(" String ")" [function, klabel(lengthString), hook(#STRING:lengthString)]
-  syntax String ::= "chrChar" "(" Int ")"   [function, klabel(chrChar), hook(#STRING:chrChar)]
-  syntax Int ::= "ordChar" "(" String ")"  [function, klabel(ordChar), hook(#STRING:ordChar)]
+  syntax String ::= "chrChar" "(" Int ")"      [function, klabel(chrChar), hook(#STRING:chrChar)]
+  syntax Int ::= "ordChar" "(" String ")"      [function, klabel(ordChar), hook(#STRING:ordChar)]
 
   syntax String ::= "substrString" "(" String "," Int "," Int ")" [function, klabel(substrString), hook(#STRING:substrString)]
-  syntax Int ::= "findString" "(" String "," String "," Int ")" [function, klabel(findString), hook(#STRING:findString)]
-  syntax Int ::= "rfindString" "(" String "," String "," Int ")" [function, klabel(rfindString), hook(#STRING:rfindString)]
-  syntax Int ::= "findChar" "(" String "," String "," Int ")" [function, klabel(findChar), hook(#STRING:findChar)]
-  syntax Int ::= "rfindChar" "(" String "," String "," Int ")" [function, klabel(rfindChar), hook(#STRING:rfindChar)]
+  syntax Int ::= "findString" "(" String "," String "," Int ")"   [function, klabel(findString), hook(#STRING:findString)]
+  syntax Int ::= "rfindString" "(" String "," String "," Int ")"  [function, klabel(rfindString), hook(#STRING:rfindString)]
+  syntax Int ::= "findChar" "(" String "," String "," Int ")"     [function, klabel(findChar), hook(#STRING:findChar)]
+  syntax Int ::= "rfindChar" "(" String "," String "," Int ")"    [function, klabel(rfindChar), hook(#STRING:rfindChar)]
 
-  syntax String ::= "Float2String" "(" Float ")" [function, klabel(Float2String), hook(#CONVERSION:float2string)]
+  syntax String ::= "Float2String" "(" Float ")"            [function, klabel(Float2String), hook(#CONVERSION:float2string)]
   syntax String ::= "Float2String" "(" Float "," String ")" [function, klabel(FloatFormat), hook(#CONVERSION:floatFormat)]
-  syntax Float ::= "String2Float" "(" String ")" [function, klabel(String2Float), hook(#CONVERSION:string2float)]
-  syntax Int ::= "String2Int" "(" String ")"     [function, klabel(String2Int), hook(#CONVERSION:string2int)]
-  syntax String ::= "Int2String" "(" Int ")"     [function, klabel(Int2String), hook(#CONVERSION:int2string)]
-  syntax String ::= "Base2String" "(" Int "," Int ")" [function, klabel(Base2String), hook(#CONVERSION:base2string)]
-  syntax Int ::= "String2Base" "(" String "," Int ")" [function, klabel(String2Base), hook(#CONVERSION:string2base)]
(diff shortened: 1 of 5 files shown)


dwightguth:

laurayuwen:
Hi, I checked out your branch and run a trival example (due to unparsing issue, I didn't try on java semantics), I get the satisfying String2Float result. 

module IMP-SYNTAX
  syntax AExp  ::= Int | Id | Float | String
  syntax Pgm ::= AExp
endmodule

module IMP
  imports IMP-SYNTAX
  rule S:Float => Float2String(S)
endmodule

But I also receive a warning like when I krun a program, say 0.0
[Warning] Internal: The Maude backend does not officially support floating
point numbers. The results of this semantics may be undefined or, in some
cases, incorrect.

And when I type krun --version, I get such an error:
[Error] Internal: Uncaught exception thrown of type NumberFormatException.

dwightguth:
Have you been building K 3.5 recently? this is a error that is specific to the Maude backend and should not occur in the revision of this branch. Make sure you are running the following commands to check out my branch:

```
git checkout dwightguth/floatformat
mvn clean package
```

k-distribution/include/builtins/string.k
```diff
@@ -38,6 +38,7 @@ module STRING-HOOKS
   syntax Int ::= "rfindChar" "(" String "," String "," Int ")" [function, klabel(rfindChar), hook(#STRING:rfindChar)]
 
   syntax String ::= "Float2String" "(" Float ")" [function, klabel(Float2String), hook(#CONVERSION:float2string)]
+  syntax String ::= "Float2String" "(" Float "," String ")" [function, klabel(FloatFormat), hook(#CONVERSION:floatFormat)]
```
andreistefanescu, line 41: can you preserve indentation?

andreistefanescu:
lgtm

k-distribution/include/builtins/string.k
dwightguth, line 41: what do you mean? it is preserved...

andreistefanescu, line 41: all the `[function, ...` attributes are aligned except the one you introduced.

dwightguth, line 41: oh I see, okay. That indentation. Never mind.

dwightguth:
Jenkins: test this please

laurayuwen:
I have been using k3.6 lately and I checked out your branch as a seperate project and added to path.

dwightguth:
I am merging this. You must have gotten your paths mixed up. The errors you reported are part of the Maude Backend.

laurayuwen:
I am using a later version of k (May 5), and found that float changed again. The same test above is yeilding different result, I get e+00 again when calling Float2String. So do I need to specify an option or anything?

dwightguth:
As I said, you need to call the version of FloatToString that takes 2 arguments, and pass it an MPFR format string for the format you want the float in.

laurayuwen:
I searched and tried, but could not figure out what exact string to put here as the second argument, could you provide an explicit reference? I also asked two other people and they don't know either, so I think it woulc benefit to users if some comments could be added there.

dwightguth:
http://www.mpfr.org/mpfr-current/mpfr.html#Formatted-Output-Functions

laurayuwen:
@dwightguth  I looked into the codebase of mpfr-java and tried to find the syntax of 'usable' format string. Seems that only "%Re" and "%Rf" etc are usable, if I added the precision, then the precision is ok (the number of digits is equal to the number of precision); BUT the printed string is simply a sequence of 0 in the form of "0.0000...." 
I am using format string like this: "%.128Rf"
// this is what I found in the mpfr's documentation.
//     mpfr_printf ("%.128Rf", x);

If the precision is contronable, then is there any special syntax w.r.t. the format string?

dwightguth:
I'm not really sure what you're asking. The semantics of this function is
precisely equivalent to a call to mpfr_asprintf with the specified format
string. You can do exactly as much or as little as that function supports.
If you want me to try to determine what your format string should look
like, you're out of luck, because I have no idea the outcome you are trying
to achieve. But you have all the information you need to make a decision
for how to proceed.

laurayuwen:
After many tests, I think the second argument along is not enough to give me the exact output as Java.

For this java program, 

public class float_22_numeric_ops {
  public static void main(String[] args) {
    System.out.println("+1.23 = " + (+1.23));
    System.out.println("-1.23 = " + (-1.23));
    System.out.println("7 \* 1.2 = " + (7.0 \* 1.2));
    System.out.println("7 / 1.2 = " + (7.0 / 1.2));
    System.out.println("7 % 1.2 = " + (7.0 % 1.2));
    System.out.println("7 + 1.2 = " + (7.0 + 1.2));
    System.out.println("7 - 1.2 = " + (7.0 - 1.2));

```
double f = 1.23;
System.out.println("++1.23  = " + (++f));

f = 1.23;
System.out.println("--1.23  = " + (--f));

f = 1.23;
System.out.println("(f = 1.23)++ = " + (f++));
System.out.println("(f =           " + f);

f = 1.23;
System.out.println("(f = 1.23)-- = " + (f--));
System.out.println("(f =           " + f);

System.out.println("Done!");
```

  }
}

The Java output 

+1.23 = 1.23
-1.23 = -1.23
7 \* 1.2 = 8.4
7 / 1.2 = 5.833333333333334
7 % 1.2 = 1.0000000000000002
7 + 1.2 = 8.2
7 - 1.2 = 5.8
++1.23  = 2.23
--1.23  = 0.22999999999999998
(f = 1.23)++ = 1.23
(f =           2.23
(f = 1.23)-- = 1.23
(f =           0.22999999999999998
Done!

sometimes has precision 17, sometimes 16, sometimes 15.

Having the second argument of "%.17RNg" is right for some cases, whereas "%.16RNg" or "%.15RNg" is right for some other cases. I also tried other rounding modes which doesn't help.

My output when using precision 17. 

+1.23 = 1.23
-1.23 = -1.23
7 \* 1.2 = 8.4000000000000004
7 / 1.2 = 5.8333333333333339
7 % 1.2 = 1.0000000000000002
7 + 1.2 = 8.1999999999999993
7 - 1.2 = 5.7999999999999998
++1.23  = 2.23
--1.23  = 0.22999999999999998
(f = 1.23)++ = 1.23
(f =           2.23
(f = 1.23)-- = 1.23
(f =           0.22999999999999998
Done!

I think I need another process to decide the precision depending on the difference. 
@dwightguth Am I missing something?

dwightguth:
Honestly, I'm not sure what to recommend. I think the best next step for you would be to try to figure out exactly how Java's toString method works by examining the source code of Java. It could be that Java is performing some kind of automatic rounding or doing something else special before printing the numbers. You might also try printing the numbers at an artificially enhanced precision to see if they are actually being computed as slightly different values.

bmmoore:
The documented behavior of Java is to print just enough digits to exactly reconstruct the number, plus to switch between decimal notation and an explicit exponent at certain ranges. Finding that minimum precision is pretty tricky, and the only mprf specifier that offers that functionality is e/E.

PR #1455, opened 2015-04-23
indexing on multiple cells

andreistefanescu:


cos:
@andreistefanescu, this is no longer applicable to the new matcher. Can we delete the pull request?

PR #1460, opened 2015-04-27
Fix not in x

radumereuta:
Adding support for notInRules, notInGround, and notInPrograms.
Improved list desuggaring.

@dwightguth please review. These changes are needed in order to parse the JavaScript semantics, but it is only partial work. I still need to fix some ambiguities issues before we will be able to parse the entire semantics.

### DIFF
--- a/k-distribution/include/builtin/domains.k
+++ b/k-distribution/include/builtin/domains.k
@@ -356,7 +356,7 @@ module STRING-SYNTAX
   //              | [\\][u] 4*Hex // "\uFFFF"     Backslash 'u' followed by four  hexadecimal characters
   //              | [\\][U] 8*Hex // "\UFFffFFff" Backslash 'U' followed by eight hexadecimal characters
   //                              // the code must not be in the range [0xdfff, 0xd800] or exceed 0x10ffff
-  syntax String ::= r"[\\\"](([^\\\"\n\r\\\\])|([\\\\][nrtf\\\"\\\\])|([\\\\][x][0-9a-fa-f]{2})|([\\\\][u][0-9a-fa-f]{4})|([\\\\][u][0-9a-fa-f]{8}))*[\\\"]"      [token, hook(org.kframework.kore.KString)]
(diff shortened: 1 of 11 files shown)


dwightguth:
I do not like this. notInRules/notInGround/etc is essentially a hack in order to get around the limitations of SDF, and we should not have to introduce it back at this time. The pattern that I would like to see us pursue going forward is to parse every string inside a module, and parse every module with exactly the syntax of that module. You will have to make changes to the JavaScript semantics in order to bring it into alignment with the new syntax, but that is needed anyway because the two syntaxes are not actually compatible (cf klabels).

radumereuta:
Well... in a way I am generating different modules from the original module. The productions annotated with those attributes are filtered out in a preprocessing step. I'm trying to maintain some backwards compatibility so we have an easier time converting the old definitions to the new implementation.
I know for sure that I can't refactor the module system in the JavaScript semantics. It's way too big.

Also, we're gonna have a hard time desuggaring user lists, since the program parser is different from the rule parser.

dwightguth:
I am quite confident that the module system in the JavaScript semantics can, in fact, be refactored.  Moreover, in so doing, it generates a much clearer and less muddy picture of how parsing works. Since we are already breaking all the definitions anyway, I think we should pursue things to be in exactly the form that we want them, which is to say, each module is parsed with exactly the productions in that module. We should have as few exceptions to this rule as possible and this one, I believe, is not necessary.

As for user lists, I believe the simple solution is to do the same thing we did with RULE-CELLS, ie, create a module, let's call it "EXPLICIT-USER-LISTS" ,which, when imported, removes epsilon productions tagged with the attribute userList and replaces them with .Ids, .Exps, etc.

radumereuta:
I will propose this topic in tomorrow's K meeting. We can either decide to keep the `noInX` attributes, or more likely, I will inform everybody about this change.

dwightguth:
You're welcome to discuss this in tomorrow's K meeting if you want, but
bear in mind that if you decide to keep the attributes, then you'll have to
have another discussion to persuade me of that fact. Which might be quite
hard.

dwightguth:
Closing this since we decided not to go this route.

radumereuta:
agreed

PR #1463, opened 2015-04-28
fix #1435 : latex attribute for list/map/set unit

### LINKED ISSUE #1435: .List .Map .Set incorrect latex output
`.List`, `.Map` and `.Set` are considered terminals in `LatexFilter` and so they appear in the tex output as `\terminal{.List}` instead of `\dotCt{List}`. I don't think this is intended because `.K` appears as `\dotCt{K}`.

dwightguth:
@grosu please review


grosu:
lgtm

PR #1464, opened 2015-04-28
fix #1351 : add error message for duplicate cell

### LINKED ISSUE #1351: Inappropriate error message
The following module includes an error (a duplicate cell in the rule), but the error message is total inappropriate. In a big definition it is not easy to see such an error. I think that it can be easily detected.

```
module TEST

  configuration 
    <k> 1 </k>
    <env> .Map </env>

  rule  <k> 1 => .K ...</k>
        <env> E:Map </env>
        <env> .Map </env>

endmodule
```

Here is the error mesage:

```
$ kompile test.k 
[Error] Internal: Uncaught exception thrown of type AssertionError.
```

dwightguth:
@bmmoore please review. Note that I wouldn't bother with this were it not so extremely quick to fix. The new KORE implementation already catches this error, after all. But it's an easy way to close out the issue, so I took it.


dwightguth:
Jenkins: test this please

bmmoore:
lgtm

PR #1465, opened 2015-04-28
add build instructions for reference manual

dwightguth:
@grosu please review.

Note that once we include documentation for the prelude in the reference manual, the build instructions will look much nicer. But that's not a huge priority since we haven't created a manual yet, and likely will not until K 4.0.


grosu:
nice, thanks.  lgtm

PR #1466, opened 2015-04-28
fix #1436 : fix float tendencies of \ksentence in latex

### LINKED ISSUE #1436: latex: configuration centers but rules do not
If I take the latex output of a configuration block and do:

```
\begin{figure}
\centering
\kconfig{
  ...
}
\end{figure}
```

then it comes out centered. However, if i do the same for a `\krule`, it does not.

dwightguth:
@grosu please review to make sure that I didn't break the appearance of anything in the latex that you care about.

@traiansf were you the one who wrote k.sty originally? if so, can you take a look? if not, who was?

### DIFF
--- a/k-distribution/include/latex/k.sty
+++ b/k-distribution/include/latex/k.sty
@@ -916,8 +916,7 @@ rectangle,rectangle
 \newcommand{\ruleAttributes}[1]{\hfill[#1]}
 
 \newcommand{\ksentence}[7][]{%
-\wrapRules{%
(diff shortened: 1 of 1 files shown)


dwightguth:
Jenkins: test this please

grosu:
It seems hard to test, so let's not slow down the release because of this.  If there is any problem we'll notice later and will get back to it.  We have to integrate @denis-bogdanas 's latex render anyway.

PR #1467, opened 2015-04-28
fix race condition

dwightguth:
@cos please review

This fixes a crash I saw that was the result of an unsynchronized race on this method.


cos:
I think this doesn't preclude races. I would make `exceptions` into a thread-safe/concurrent collection instead.

dwightguth:
@cos please re-review

PR #1470, opened 2015-04-28
delete unused defintions in samples

dwightguth:
@traiansf @dlucanu @ciobaca @grosu these definitions belong to one of you. I am removing them because they are not tested and therefore likely broken. If you want to add them back, please create a pull request doing so which adds ktest coverage for the programs in the definition.


grosu:
It is OK with me to remove these for now.  I would like to add the funcons back, eventually, though.  @traiansf , @dlucanu , @ciobaca, you should also add your examples back when you get a chance.

PR #1473, opened 2015-04-29
fix #1437 : align latex cells by cell label instead of cell content

### LINKED ISSUE #1437: latex: cells not horizontally aligned
In the drawing of rules or configurations, cells that contain different types of content do not align horizontally, e.g., cells that contain a `\dotCt{something}` are a little below other cells containing `\terminal{}` and `\constant{}`.
I tried playing with the `k.sty` file, but nothing affected that.

dwightguth:
@grosu please review. If you could compare what the code looks like before vs after, and let me know if you think it's an improvement, that would be great. Basically, this code makes all the cells align based on their cell label instead of based on the baseline of their content.


grosu:
lgtm.  it actually looks nicer aligned to the top.

PR #1478, opened 2015-04-29
equality for KORE terms

cos:


cos:
@dwightguth, the Att-less equals -- it also adds hashCode and equals for all KORE terms (this superseeds any case class generated equals and hashCode)

dwightguth:
Looks good, but can you make hashCode be a lazy val? Everything is immutable, so that should be safe, and it will have a significant performance impact on large terms.

PR #1480, opened 2015-04-30
finish most of kompile frontend + add hook for krun frontend

dwightguth:
@cos please review

### DIFF
--- a/kernel/src/main/java/org/kframework/krun/KRun.java
+++ b/kernel/src/main/java/org/kframework/krun/KRun.java
@@ -1,3 +1,4 @@
+// Copyright (c) 2015 K Team. All Rights Reserved.
 package org.kframework.krun;
 
 import org.kframework.kil.Attributes;
--- a/kernel/src/main/java/org/kframework/krun/KRun.java
+++ b/kernel/src/main/java/org/kframework/krun/KRun.java
@@ -1,0 +1,36 @@
+package org.kframework.krun;
+
+import org.kframework.kil.Attributes;
+import org.kframework.kompile.CompiledDefinition;
+import org.kframework.transformation.Transformation;
+import org.kframework.utils.errorsystem.KExceptionManager;
+import org.kframework.utils.file.FileUtil;
+
+/**
+ * Created by dwightguth on 4/30/15.
+ */
+public class KRun implements Transformation<Void, Void> {
+
+    private final KExceptionManager kem;
+    private final FileUtil files;
+
+    public KRun(KExceptionManager kem, FileUtil files) {
+        this.kem = kem;
+        this.files = files;
+    }
+
+    public int run(CompiledDefinition compiledDef, KRunOptions options) {
+        System.err.println("TODO(cos): krun");
+        return 0;
+    }
+
+    @Override
+    public Void run(Void aVoid, Attributes attrs) {
+        return null;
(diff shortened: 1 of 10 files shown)


kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.kframework.kil.Attributes;
+import org.kframework.kompile.CompiledDefinition;
+import org.kframework.transformation.Transformation;
+import org.kframework.utils.errorsystem.KExceptionManager;
+import org.kframework.utils.file.FileUtil;
+
+/**
+ * Created by dwightguth on 4/30/15.
+ */
```
cos, line 12: the created-bys are redundant having blame, is this a general interface for anything `KRun`-ish?

cos, line 12: I see, it's the actual KRun for KORE

dwightguth, line 12: this is intended to be the KORE-based KRun frontend, yes. Eventually it will have methods in it for various types of activities like running programs, proving, pretty-printing, etc. But I just made a shell for now.

kernel/src/main/java/org/kframework/krun/KRunFrontEnd.java
```diff
@@ -78,14 +88,18 @@ public int run() {
...
             scope.enter(kompiledDir.get());
             try {
-                Transformation<Void, Void> tool = toolProvider.get();
-                Attributes a = new Attributes();
-                tool.run(null, a);
-                Integer exitCode = a.typeSafeGet(Integer.class, Executor.Tool.EXIT_CODE);
-                if (exitCode == null) {
-                    exitCode = 0;
+                if (kRunOptions.experimental.kore) {
+                    return new KRun(kem, files).run(loader.loadOrDie(CompiledDefinition.class, files.resolveKompiled("compiled.bin")), kRunOptions);
```
cos, line 92: does this flow allow me to add command-line parameters at will, without adding them to same configuration files, etc?

cos:
if I understand this correctly, the double-layer Guice injection is still active in this flow, right?

dwightguth:
double-layer? most of the guice injection does not fire in this case because we never call `get` on the TransformationProvider. A little bit does fire to initialize everything, but what that is doing is parsing options and determining the location of the compiled definition from options. The rest does not fire.

kernel/src/main/java/org/kframework/krun/KRunFrontEnd.java
dwightguth, line 92: well, you can add whatever you want to KRunOptions... but you really shouldn't have to for what we're doing since we're not adding new features... you just need to implement the functionality of the existing options.

cos:
I remember injection was somehow used before parsing options, and then another layer of injection was started for the rest. In any case, I played a bit with this branch and it seems lightweight enough. I can build upon it. lgtm.

kernel/src/main/java/org/kframework/krun/KRunFrontEnd.java
cos, line 92: I need a `--console` or `--debugger` that would leave an open console.

PR #1481, opened 2015-04-30
hashCode caching for KORE AST classes

cos:


dwightguth:
This seems overkill. Why not just make the hashCode method a lazy val?

cos:
that would force any (Scala) subclasses of K to have "stable" (i.e., final) hashCode values, so we wouldn't be able to have mutable implementations, which require hashCode updating, anymore.

PR #1483, opened 2015-04-30
basic KORE-based execution in krun

cos:
**please review**


cos:
ok, so let's take this in bite-size pieces. @dwightguth, please review

the `Rewriter` will very likely change once we add logic And/Or to KORE, but I want to make sure any interaction is done though the abstract interface, not the tiny `Rewriter`.

tiny/src/main/scala/org/kframework/tiny/Rewriter.scala
```diff
@@ -89,7 +89,7 @@ class Rewriter(module: definition.Module, index: K => Option[String] = KIndex) {
   var indexFailures = 0
 
   def executeStep(k: K): Option[K] = {
-//    println("\n\n MATCHING ON: " + k)
+    //    println("\n\n MATCHING ON: " + k)
```
radumereuta, line 92: Try to keep diffs to a minimum by avoiding auto-formatting.

radumereuta:
Why do you want to add boolean logic to KORE? I thought KORE should be something really simple.

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -21,8 +29,25 @@ public KRun(KExceptionManager kem, FileUtil files) {
...
     public int run(CompiledDefinition compiledDef, KRunOptions options) {
-        System.err.println("TODO(cos): krun");
-        return 0;
+        try {
+            String pgmFileName = options.configurationCreation.pgm();
+            if (!options.configurationCreation.term()) {
+                KExceptionManager.criticalError("Unsupported options: term=false");
+            }
+
+            String pgm = Files.toString(new File(pgmFileName), Charset.defaultCharset());
```
dwightguth: files.loadFromWorkingDirectory

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -21,8 +29,25 @@ public KRun(KExceptionManager kem, FileUtil files) {
...
+            K program = compiledDef.getProgramParser().apply(pgm, Source.apply(pgm));
```
dwightguth: needs support for external parsers

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -21,8 +29,25 @@ public KRun(KExceptionManager kem, FileUtil files) {
...
+
+            org.kframework.Rewriter rewriter = new Rewriter(compiledDef.executionModule());
+
+            K result = rewriter.execute(program);
+
+            System.err.println("result:" + result);
```
dwightguth: should call the unparsing pipeline we have created so far

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -21,8 +29,25 @@ public KRun(KExceptionManager kem, FileUtil files) {
...
+        try {
+            String pgmFileName = options.configurationCreation.pgm();
+            if (!options.configurationCreation.term()) {
+                KExceptionManager.criticalError("Unsupported options: term=false");
+            }
+
+            String pgm = Files.toString(new File(pgmFileName), Charset.defaultCharset());
+            K program = compiledDef.getProgramParser().apply(pgm, Source.apply(pgm));
+
+            org.kframework.Rewriter rewriter = new Rewriter(compiledDef.executionModule());
```
dwightguth: something about this feels odd to me... We need support for different backends, some of which may not be part of the kernel, so you need some kind of indirect way of instantiating the rewriter. I think that this method should take a `Function<Module, Rewriter>` instance and apply the function...

kernel/src/main/java/org/kframework/krun/KRun.java
cos: again, why do we have special functions for things that are one-liners with vanilla JDK?

kernel/src/main/java/org/kframework/krun/KRun.java
cos: **comment here instead**
again, why do we have special functions for things that are one-liners with vanilla JDK?

kernel/src/main/java/org/kframework/krun/KRun.java
cos: I know, we're far from there.

dwightguth: I'm not sure what you're asking here... The purpose of external parsers has nothing to do with the JDK, it derives from the fact that we may want to parse things using something other than K.

dwightguth: oh I see never mind

kernel/src/main/java/org/kframework/krun/KRun.java
dwightguth: well, in particular Files.toString likely will not  pick up the thread-local working directory, and so will actually in some cases when running a server output the wrong value.

kernel/src/main/java/org/kframework/krun/KRun.java
cos: yes, that's a way -- eventually we'll have to figure out the mechanism for choosing a rewriter (or even multiple rewriters). I kept it simple for now. Having a `Function<Module, Rewriter>` as an argument to `run` sounds good if you want that now.

kernel/src/main/java/org/kframework/krun/KRun.java
dwightguth: we aren't far from this. It would take me less than two hours to implement this functionality, which is required for interacting with most large semantics. I think it's reasonable to get this over with, especially because the code you wrote cannot, actually, be used for any definitions with syntax modules.

kernel/src/main/java/org/kframework/krun/KRun.java
dwightguth: yes, I would like this now. You can just pass in Rewriter::new to the function for now though...

kernel/src/main/java/org/kframework/krun/KRun.java
cos: which one is the right unparser?

tiny/src/main/scala/org/kframework/tiny/Rewriter.scala
cos, line 92: thank you @radumereuta, this is highly useful

dwightguth, line 92: please don't be sarcastic about this. It actually has been making your diffs quite hard to read for quite some time now. I've given up persuading you to change your mind, but it is kind of a serious annoyance when you do this. So don't joke.

kernel/src/main/java/org/kframework/krun/KRun.java
cos: somehow this doesn't feel right, but I'll play along

kernel/src/main/java/org/kframework/krun/KRun.java
dwightguth: This is the one-liner I have at the moment... Eventually we can make the interface more uniform.

```
private String unparseTerm(K input, Module test) {
+        return KOREToTreeNodes.toString(KOREToTreeNodes.apply(KOREToTreeNodes.up(input), test));
+    }
```

kernel/src/main/java/org/kframework/krun/KRun.java
cos: I have no idea how to hook an external parsers, so you're more then welcome to add that

the code I wrote, for now, requires definitions with syntax modules

tiny/src/main/scala/org/kframework/tiny/Rewriter.scala
cos, line 92: I agree it can be annoying but I'm not going to spend time cherry picking through diffs -- it is much harder then disregarding them when reviewing.

The right solution is for all of us to have the same style file. Somehow, it seems we don't. Is there a common style file for the K project?

kernel/src/main/java/org/kframework/krun/KRun.java
dwightguth: well, no, it requires definitions whose syntax module is the same as their main module, because it requires definitions whose syntax module contains the cell productions, which most syntax modules do not. Essentially, what you want to do is take the value of the --parser flag and invoke it as a process with pgm() as its argument. Then you take everything from stdout and parse it as KAST. We can add some tweaks to remove redundant process creations fairly easily later.

kernel/src/main/java/org/kframework/krun/KRun.java
cos: it's one line so I'll copy it for now.

cos:
> Why do you want to add boolean logic to KORE? I thought KORE should be something really simple.

it's not boolean logic, it's matching logic.

radumereuta:
So have you discussed this in the K meeting? I would like to understand why this is necessary. It's making me question what kore is.

kernel/src/main/java/org/kframework/krun/KRun.java
cos: it seems there are two relatively distinct issues here:
1. supporting definitions with pure syntax modules (no cell declarations). We probably have to copy the cell productions into the compiled -SYNTAX. Radu, you did the wiring for -SYNTAX (or, anyway, -PROGRAMS), could you have a look?
2. supporting an external parser. for this, at the very least, we need a macro module with rules for initializing the configuration -- @dwightguth, if I remember correctly, you never got to this when implementing the explicit-configuraiton-to-syntax-productions pass.

cos:
we've touched upon it, but didn't make a decision yet

kernel/src/main/java/org/kframework/krun/KRun.java
radumereuta: I'm not sure I understand what is the issue here, but this is how I view the parser:
1. there is a function `parse(String input, String startSymbol, Source source)`. Source is only for annotating terms of their origin, startSymbol may have a default, and input is mandatory. The return is a KAST term.
2. the function above is attached to an object which knows how to do the parsing. In this case, it can either be a newparser object, which has been initialized with the appropriate grammar (see ParseInModule.java), or some sort of wrapper that know show to call an external parser and transform the output into KAST. (I think this is what @dwightguth is asking for).

Note that this may get a bit more complex once we add into consideration error reporting.

cos: @radumereuta, I'm glad the `Module` is not part of the signature anymore :-)

the problem is with where the productions for configuration lay (this is something you could address), and how the initial configuration is constructed when not given by the user (i.e. when `--term=false`).

tiny/src/main/scala/org/kframework/tiny/Rewriter.scala
grosu, line 92: I think we do have a style guidelines file for the K project:

```
https://github.com/kframework/k/wiki/Style-Guidelines
```

Those of you who created their style configuration files for their favorite IDE shuld feel free to make it public as links on the style guidelines page.

But I agree that minor things, especially, become very annoying when they do not conform with the policies.  No need to make a big deal out of it now, but it would be good in general to avoid mixing things.  Maybe add an item on the policies page specifically about that, and then, in the future, we can reject code and refer the authors to it?

grosu:
@radumereuta , you should read the matching logic paper that I sent around.  I agree that it makes sense to have them in KORE earlier than later, because in the long terms they will be added anyway.  However, since they are not crucial right now, we should not rush to add them if that slows us down with anything.

kernel/src/main/java/org/kframework/krun/KRun.java
dwightguth: I do not believe we should copy the productions of the configuration into any modules. I think instead, we should allow users of krun to specify which module they are parsing in on the command line somehow. Whether we do this using the external parser interface or some other flag, I do not feel strongly about at this time. However, we should, I feel, address this issue one way or another before committing this.

kernel/src/main/java/org/kframework/krun/KRun.java
dwightguth: Can you now add the call to AddBrackets between KOREToTreeNodes.apply and KOREToTreeNodes.toString, now that it's committed?

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -20,11 +31,27 @@ public KRun(KExceptionManager kem, FileUtil files) {
...
+        }
+
+        String pgm = files.loadFromWorkingDirectory(pgmFileName);
+        K program = compiledDef.getProgramParser().apply(pgm, Source.apply(pgm));
+
+        Rewriter rewriter = rewriterGenerator.apply(compiledDef.executionModule());
+
+        K result = rewriter.execute(program);
+
+        System.err.println("result:" + unparseTerm(result, compiledDef.kompiledDefinition.mainSyntaxModule()));
```
dwightguth: this needs to be the main module, again, because it needs to contain all the syntax that is possible to exist in an output configuration.

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -20,11 +31,27 @@ public KRun(KExceptionManager kem, FileUtil files) {
         this.files = files;
     }
 
-    public int run(CompiledDefinition compiledDef, KRunOptions options) {
-        System.err.println("TODO(cos): krun");
+    public int run(CompiledDefinition compiledDef, KRunOptions options, Function<Module, Rewriter> rewriterGenerator) {
+        String pgmFileName = options.configurationCreation.pgm();
+        if (!options.configurationCreation.term()) {
+            KExceptionManager.criticalError("Unsupported options: term=false");
```
dwightguth: needs the throws keyword

tiny/src/main/scala/org/kframework/tiny/Rewriter.scala
cos, line 92: @grosu, the style guidelines are very general. They do not touch upon issues like whether to put the `//` characters at the beginning of the line or align them with the code. There are very many such small decisions (several pages of configuration in Intellij). I personally, do not care about most of those things one way or another, but I do think we should all use the exact same style file. Unless we have the same style file, we have to either be very careful when writing code and keep all of these rules in mind (instead of simply auto-formatting), or be very careful committing and adjust such things when necessary. Being this careful about formatting when there are automatic solutions is a waste of time in my opinion.

@dwightguth merged his Intelliij formatting file. I will simply start using that. If we all do, our changes will be confluent from now on.

kernel/src/main/java/org/kframework/krun/KRun.java
cos: regardless of how we link it to the user interface, we still need need the [`initializer` rules](https://github.com/kframework/k/wiki/KAST-and-KORE) in order to construct the configuration around a given program.

kernel/src/main/java/org/kframework/krun/KRun.java
cos: hm... the main module is not the right one to use either because it doesn't have `Id`s. Considering that we will want to allow printing of symbolic terms, it seems we need a special mix that has configuration, k stuff (with a special representation for variables such that they don't clash with Ids), and Ids. 

In the meantime, I will stick with mainSyntaxModule.

cos:
@dwightguth, I'm getting this error when unparsing the result of the counter imp test program:

```
java.lang.IndexOutOfBoundsException
    at org.pcollections.ConsPStack.listIterator(ConsPStack.java:85)
    at java.util.AbstractSequentialList.get(AbstractSequentialList.java:88)
    at org.kframework.parser.TermCons.get(treeNodes.scala:36)
    at org.kframework.unparser.KOREToTreeNodes$$anonfun$1.apply(KOREToTreeNodes.scala:42)
    at org.kframework.unparser.KOREToTreeNodes$$anonfun$1.apply(KOREToTreeNodes.scala:38)
    at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
    at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
    at scala.collection.immutable.List.foreach(List.scala:381)
    at scala.collection.TraversableLike$class.map(TraversableLike.scala:245)
    at scala.collection.immutable.List.map(List.scala:285)
    at org.kframework.unparser.KOREToTreeNodes$.toString(KOREToTreeNodes.scala:38)
    at org.kframework.unparser.KOREToTreeNodes$$anonfun$1.apply(KOREToTreeNodes.scala:42)
    at org.kframework.unparser.KOREToTreeNodes$$anonfun$1.apply(KOREToTreeNodes.scala:38)
    at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
    at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
    at scala.collection.immutable.List.foreach(List.scala:381)
    at scala.collection.TraversableLike$class.map(TraversableLike.scala:245)
    at scala.collection.immutable.List.map(List.scala:285)
    at org.kframework.unparser.KOREToTreeNodes$.toString(KOREToTreeNodes.scala:38)
    at org.kframework.unparser.KOREToTreeNodes$$anonfun$1.apply(KOREToTreeNodes.scala:42)
    at org.kframework.unparser.KOREToTreeNodes$$anonfun$1.apply(KOREToTreeNodes.scala:38)
    at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
    at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
    at scala.collection.immutable.List.foreach(List.scala:381)
    at scala.collection.TraversableLike$class.map(TraversableLike.scala:245)
    at scala.collection.immutable.List.map(List.scala:285)
    at org.kframework.unparser.KOREToTreeNodes$.toString(KOREToTreeNodes.scala:38)
    at org.kframework.unparser.KOREToTreeNodes.toString(KOREToTreeNodes.scala)
    at org.kframework.krun.KRun.unparseTerm(KRun.java:54)
    at org.kframework.krun.KRun.run(KRun.java:49)
    at org.kframework.krun.KRunFrontEnd.run(KRunFrontEnd.java:93)
    at org.kframework.main.FrontEnd.main(FrontEnd.java:51)
    at org.kframework.main.Main.runApplication(Main.java:110)
    at org.kframework.main.Main.runApplication(Main.java:100)
    at org.kframework.main.Main.main(Main.java:52)
```

dwightguth:
Can you show me the KORE term you are unparsing? The only way I can see
that you might have this issue is if you generate a term where the arity of
the production does not match the arity of the term. Since this should not
happen, I have to assume you are generating wrong terms somehow.

tiny/src/main/scala/org/kframework/tiny/Rewriter.scala
grosu, line 92: I thought we agreed some time ago that we should not commit these IDE configuration files in the repo, because different people may have different IDEs and different settings, which this way they may end up pushing by mistake and screwing up others settings, too.  I would feel more comfortable having some special folder under src, say "IDE styles", or something similar, where people can share their configurations if they want to.  In short, I do not like having a top .idea folder in the master.  And I think it was not only me, but many others, when we agreed that we should not do this.

Grigore

dwightguth, line 92: Grigore, the purpose of not committing all the .idea files is that they
vary from user to user, yes. However, it's worth noting that in some sense
Cosmin is right: having a consistent style file across all developers
should in principle remove the kinds of issues that we are dealing with in
Cosmin's diffs. So making this exception is actually completely in keeping
with the methodology according to which we made the original decision:
namely, it is material we expect should /not/ change across developers.

cos:
basically, the test under `TstTinyOnKORE_IT`. The term to unparse is:

```
<top>(<k>(#KSequence()),<state>(_Map_(_|->_(n:Id,-1),_|->_(s:Id,55))))
```

dwightguth:
Yes, I figured out what the problem here is. It was a bug in KOREToTreeNodes.up. You need to replace the case of KSequence with the following block: 

```
case s: KSequence =>
      if (s.items.size() == 0)
        KApply(KLabel("#EmptyK"), KList(), s.att)
      else
        KApply(KLabel("#KSequence"), upList(s.items.asScala), s.att)
```

cos:
you have access to the pull request branch, could you do that? thanks!

cos:
ok, I replaced it and it works.

cos:
I think I've addressed everything easily addressable at this stage. I'll try ktest now, and improve things on the way. Is everyone ok with the merge?

radumereuta:
modulo whitespace changes, this lgtm

kernel/src/main/java/org/kframework/krun/KRun.java
dwightguth: can you remove the "result:" to make it consistent with the behavior krun used to have?

dwightguth:
lgtm if you fix my minor issue.

PR #1485, opened 2015-04-30
add code style file for intellij

dwightguth:
@cos please review. You asked for a style file, here is one. Feel free to compare it against your own and suggest differences, but I am unlikely to want to change the way I decided to indent labels.


cos:
I have no strong preferences here, just don't want to worry about formatting.

PR #1488, opened 2015-05-05
KAST-syntax unparsing of KORE terms

bmmoore:


kore/src/test/scala/org/kframework/kore/UnparseTest.scala
```diff
@@ -0,0 +1,62 @@
...
+  @Test def Tokens() {
+    Assert.assertEquals("""#token("Int","9")~>#token("String","Test")""",ToKast(intToToken(9)~>"Test"))
+  }
+
+  @Test def Variables() {
+    Assert.assertEquals("A=>B",ToKast(KRewrite(KVariable("A"),KVariable("B"))))
+  }
+
+  @Test def Precedence1() {
+    Assert.assertEquals("a()=>b()~>c()",ToKast(KRewrite('a(),'b())~>'c()))
```
dwightguth: this doesn't look right. shouldn't ~> bind tighter than =>?

kore/src/test/scala/org/kframework/kore/UnparseTest.scala
```diff
@@ -0,0 +1,62 @@
+package org.kframework.kore
+
+import org.junit.{Assert, Test}
+import org.kframework.attributes.Att
+import org.kframework.kore.KORE._
+
+class UnparseTest {
+  @Test def Token() {
+    Assert.assertEquals("""#token("Int","12")""",ToKast(KToken(Sort("Int"), "12")))
```
dwightguth, line 9: isn't sort supposed to be second?

kore/src/test/scala/org/kframework/kore/UnparseTest.scala
```diff
@@ -0,0 +1,62 @@
...
+  @Test def EscapeLabel() {
+    Assert.assertEquals("`_+_`",ToKast(KLabel("_+_")))
+  }
+
+  @Test def WrappedKLabel() {
+    Assert.assertEquals("#klabel(foo)",ToKast(InjectedKLabel('foo, Att())))
+  }
+
+  @Test def EmptyApp() {
+    Assert.assertEquals("foo()",ToKast('foo()))
```
dwightguth: pretty sure you need the .::KList for it to parse back.

kore/src/test/scala/org/kframework/kore/UnparseTest.scala
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  @Test def NestedApp() {
+    Assert.assertEquals("foo(a(),b())",ToKast('foo('a(), 'b())))
+  }
+
+  @Test def SequenceEmpty() {
+    Assert.assertEquals(".K",ToKast(ADT.KSequence(List(),Att())))
```
dwightguth, line 32: .::K

kore/src/test/scala/org/kframework/kore/UnparseTest.scala
```diff
@@ -0,0 +1,62 @@
...
+  @Test def Variables() {
+    Assert.assertEquals("A=>B",ToKast(KRewrite(KVariable("A"),KVariable("B"))))
+  }
+
+  @Test def Precedence1() {
+    Assert.assertEquals("a()=>b()~>c()",ToKast(KRewrite('a(),'b())~>'c()))
+  }
+
+  @Test def Precedence2() {
+    Assert.assertEquals("a()=>``b()~>c()``",ToKast(KRewrite('a(),'b()~>'c())))
```
dwightguth: you only need one ` here.

bmmoore: Yes, but it's a bit complicated to figure that out. Using two should be correct.

kore/src/test/scala/org/kframework/kore/UnparseTest.scala
bmmoore: Right, I'm reversing that. It's unfortunately not described in kast.k

kore/src/test/scala/org/kframework/kore/UnparseTest.scala
bmmoore, line 32: .K is equivalent to .::K in kast.k - and is what's used on the wiki.

kore/src/test/scala/org/kframework/kore/UnparseTest.scala
bmmoore, line 9: Hmm, that is what's on the wiki, but the Kore types and the previous unparser agree on putting Sort first. I suppose I do prefer putting the sort second.

kore/src/test/scala/org/kframework/kore/UnparseTest.scala
bmmoore: .KList is what's used on the wiki, and kast.k accepts both.

bmmoore:
Ok, I've added the explicit .KList. I think swapping the argument order on token should be left for a separate request, that will touch a lot more files if we do the KToken constructor too (and the parser/Down works by reflection so that's almost required).

dwightguth:
lgtm

PR #1490, opened 2015-05-06
fix java backend for imp + add back java backend as default rewriter for kore

dwightguth:
**please review**


dwightguth:
@radumereuta @andreistefanescu @cos please review

Note that I have made the java backend the default, but you can still get tiny by using kompile --backend tiny. @cos , it might be nice if you moved the code which initializes the tiny rewriter into a K module though.

java-backend/src/main/java/org/kframework/backend/java/compile/KOREtoBackendKIL.java
```diff
@@ -99,8 +98,8 @@ else if (k instanceof org.kframework.kore.KSequence)
             return KSequence(((org.kframework.kore.KSequence) k).items(), k.att());
         else if (k instanceof org.kframework.kore.KVariable)
             return KVariable(((org.kframework.kore.KVariable) k).name(), k.att());
-//        else if (k instanceof org.kframework.kore.InjectedKLabel)
-//            return InjectedKLabel(((org.kframework.kore.InjectedKLabel) k).klabel(), k.att());
+        else if (k instanceof org.kframework.kore.InjectedKLabel)
+            return InjectedKLabel(((org.kframework.kore.InjectedKLabel) k).klabel(), k.att());
         else
             throw new AssertionError("BUM!");
```
radumereuta, line 105: Funny, but really?

dwightguth, line 105: predates me. Given this is a case that is supposed to never occur, I think it's probably fine.

java-backend/src/main/java/org/kframework/backend/java/kil/Definition.java
```diff
@@ -217,28 +216,69 @@ public Definition(org.kframework.definition.Module module, KExceptionManager kem
...
     public void addKoreRules(Module module, TermContext termContext) {
         KOREtoBackendKIL transformer = new KOREtoBackendKIL(termContext);
         JavaConversions.setAsJavaSet(module.sentences()).stream().forEach(s -> {
             if (s instanceof org.kframework.definition.Rule) {
                 org.kframework.definition.Rule rule = (org.kframework.definition.Rule) s;
+                K leftHandSide = RewriteToTop.toLeft(rule.body());
                 org.kframework.kil.Rule oldRule = new org.kframework.kil.Rule();
                 oldRule.setAttributes(new KOREtoKIL().convertAttributes(rule.att()));
+                Location loc = rule.att().getOptional(Location.class).orElse(null);
+                Source source = rule.att().getOptional(Source.class).orElse(null);
```
radumereuta, line 274: Can you explain how `Location` and `Source` work? This looks counter intuitive at a first glance.

dwightguth, line 274: I'm not really sure what you're asking... This is the way we want to get typed attributes going forward. Location is being requested via the getOptional method, which looks up the attribute with the klabel equal to the class name of the Location class.

kore/src/main/scala/org/kframework/Rewriter.scala
```diff
@@ -19,8 +19,5 @@ trait Rewriter {
    * - for search, it returns an Or with multiple ground terms as children
    */
 
-
-  def rewrite(k: K)(implicit sofar: Set[K] = Set()): Set[K]
```
cos, line 23: I know these are in flux, but we do need an interface method for search even now.

kore/src/main/scala/org/kframework/builtin/Sorts.scala
```diff
@@ -16,7 +16,9 @@ object Sorts {
   val File = Sort("File")
 
   val K = Sort("K")
+  val KBott = Sort("KBott")
   val KVariable = Sort("KVariable")
+  val KItem = Sort("KItem")
```
cos, line 21: not sure KItem is the same as KBott in all contexts. @radumereuta ?

kore/src/main/scala/org/kframework/collections.scala
```diff
@@ -25,7 +25,6 @@ object Collections {
     x.addAll(s.asJava)
     x
   }
-  def mutable[T](s: Iterable[T]): java.lang.Iterable[T] = s.asJava
```
cos, line 28: this is not redundant, it could call `iterable` if you like but it's nice to have a uniform convention for converting from immutable Scala data structure to Java ones

kernel/src/main/java/org/kframework/krun/KRunModule.java
```diff
@@ -194,12 +196,19 @@ protected void configure() {
...
+            });
+
             MapBinder.newMapBinder(
                     binder(), String.class, LtlModelChecker.class);
 
             MapBinder.newMapBinder(
                     binder(), String.class, Prover.class);
 
+            //TODO(cos): move to tiny module
+            rewriterBinder.addBinding("tiny").toInstance(m -> new org.kframework.tiny.Rewriter(m));
```
cos, line 210: I don't understand. What do you want moved to tiny?

kore/src/main/scala/org/kframework/Rewriter.scala
dwightguth, line 23: is that what that's supposed to be? I removed it because the implicits were making it particularly not nice to deal with from Java. I will add it back, but without the implicit.

kore/src/main/scala/org/kframework/collections.scala
dwightguth, line 28: I figure one method is enough. It was only being used at one call site, so I just changed it.

kore/src/main/scala/org/kframework/Rewriter.scala
cos, line 23: the implicit is simply a second parameter in Java -- in any case, we can go without the implicit for now.

kernel/src/main/java/org/kframework/krun/KRunModule.java
dwightguth, line 210: the code which adds "tiny" to the list of backends

kore/src/main/scala/org/kframework/collections.scala
cos, line 28: ok.

kore/src/main/scala/org/kframework/Rewriter.scala
dwightguth, line 23: maybe it wasn't the implicit, maybe it was the default value, but in any event, it was an interface that did not look nice to implement because it had some nasty mangled method name with $s in it.

kernel/src/main/java/org/kframework/krun/KRunModule.java
cos, line 210: sorry, but I do not know how the entire injection system works. I know it is documented, and I've read and understand it once, but I don't want to get back to it considering we'll likely refactor it in the future. Could you make the necessary additions?

kore/src/main/scala/org/kframework/Rewriter.scala
cos, line 23: oh, yes, it's because of the default value -- let's not deal with the implicit parameter for now then.

kernel/src/main/java/org/kframework/krun/KRunModule.java
grosu, line 210: But Cosmin, the overall agreement was that we will add backends and tools to the framework by means of dependency injection, because that's the cleanest way to do it.  The agreement was that we will not overuse dependency injection, but I remember quite clearly that we all agreed that such modules will still be added with dependency injection.  In fact, we want people to add such modules themselves, without even being incorporated in the K framework.  If you cannot do it then it means that other people have to chance to do it.  However, other people actually did it (Traian and Andrei A).  So the documentation must be good enough.  If it is not, then complain about that.

So if I understand the context well, then your answer to Dwight is inadequate and all that discussion about people being responsible for the entire code base was useless.

Grigore

cos, line 210: @grosu, I think the idea is to also simplify the modularization mechanism (we're just not there yet). If we are, then there is no point in me working an hour for something that Dwight can do in 5 min. It would be useful to do to refresh my memory, but I would rather do it when we discuss modularization, instead of now when I don't really have an hour for this (until Friday) and we'll likely not work on modularization soon anyhow.

java-backend/src/main/java/org/kframework/backend/java/util/Z3Wrapper.java
```diff
@@ -81,7 +82,7 @@ public boolean checkQueryWithExternalProcess(String query, int timeout) {
         String result = "";
         try {
             for (int i = 0; i < Z3_RESTART_LIMIT; i++) {
-                ProcessBuilder pb = this.pb.get().command(
+                ProcessBuilder pb = files.getProcessBuilder().command(
```
radumereuta, line 85: Why is the process builder inside `FileUtil`?

k-distribution/include/builtin/domains.k
```diff
@@ -9,7 +9,7 @@ module MAP
 
   /*@ Construct a new Map consisting of key/value pairs of two Maps (the keys of
    the two Maps are assumed disjoint) */
-  syntax Map ::= Map Map                        [left, function, hook(Map:__), klabel('_Map_), assoc, comm, unit(.Map)]
+  syntax Map ::= Map Map                        [left, function, hook(Map:__), klabel('_Map_), assoc, comm, unit(.Map), element(_|->_)]
```
radumereuta, line 12: This is new. Who uses it? Also, it would be nice to have a place where we explain the meaning of all attributes.

kore/src/main/scala/org/kframework/Rewriter.scala
cos, line 23: conclusion of offline discussion: we'll remove the method for now and add finer-grained method moving forward

cos:
lgtm.

k-distribution/src/test/resources/convertor-tests/kore_imp.k
```diff
@@ -1,299 +1,68 @@
...
+  syntax Block ::= "{" "}"
+                 | "{" Stmt "}"
+  syntax Stmt  ::= Block
+                 | Id "=" AExp ";"            [strict(2)]
+                 | "if" "(" BExp ")"
+                   Block "else" Block         [strict(1)]
+                 | "while" "(" BExp ")" Block
+                 > Stmt Stmt                  [left]
+  syntax Pgm ::= "int" Ids ";" Stmt
+  syntax Ids ::= List{Id,","}
```
radumereuta, line 27: I haven't solved the list issue yet. Is this tested?

k-distribution/src/test/resources/convertor-tests/kore_imp.k
```diff
@@ -1,299 +1,68 @@
...
+  rule <k> X = I:Int; => .K ...</k> <state>... X |-> `_ => I` ...</state>
+  rule S1:Stmt S2:Stmt => S1 ~> S2  [structural]
+  rule if (true)  S else _ => S
+  rule if (false) _ else S => S
+  rule while (B) S => if (B) {S while (B) S} else {}  [structural]
+// Pgm
+  rule <k> int `X,Xs => Xs`;_ </k> <state> Rho:Map `.Map => X|->0` </state>
+    when notBool `X in keys(Rho)`
+  rule int .Ids; S => S  [structural]
+endmodule
```
radumereuta, line 68: For consistency, new line please.

kernel/src/main/java/org/kframework/krun/KRunModule.java
dwightguth, line 210: There is a point to it. The point is that I don't have to spend 15 minutes
on something that adds no value for the company. Time adds up and mine is
not infinite. If you want to start working on tasks that benefit the
company, I will be more than happy to trade with you and work on tasks that
benefit your research, but somehow I'm pretty sure you don't. So in the
meantime, here's the choice I have for you. You can add this to your list
of tasks to work on in the next couple weeks, or I can remove tiny from the
krun frontend entirely.

kernel/src/main/java/org/kframework/kore/compile/CloseCells.java
```diff
@@ -93,10 +93,14 @@ void resetVars() {
         rhsOf = null;
     }
 
-    KVariable newDotVariable() {
+    KVariable newDotVariable(Sort s) {
```
radumereuta, line 97: I would like to see some JavaDoc here. You are required to have a sort for this function, but you may also have `null`. That can be dangerous if not properly documented.

kernel/src/main/java/org/kframework/kore/compile/GenerateSentencesFromConfigDecl.java
```diff
@@ -158,7 +158,7 @@
             Sentence bagSubsort = Production(bagSort, Seq(NonTerminal(sort)), Att());
             Sentence bagUnit = Production("." + bagSort.name(), bagSort, Seq(Terminal("." + bagSort.name())));
             Sentence bag = Production("_" + bagSort + "_", bagSort, Seq(NonTerminal(bagSort), NonTerminal(bagSort)),
-                    Att().add("assoc").add("unit", "." + bagSort.name()));
+                    Att().add("assoc", "").add("comm", "").add("unit", "." + bagSort.name()));
```
radumereuta: Extract constant from string.

radumereuta: Actually, @cos was really into having typed attributes. This looks untyped.

kernel/src/main/java/org/kframework/kore/compile/ResolveSemanticCasts.java
```diff
@@ -24,10 +23,11 @@
 public class ResolveSemanticCasts {
 
     private Set<KApply> casts = new HashSet<>();
-    private Map<String, String> varSorts = new HashMap<>();
+    private Map<KVariable, KVariable> varSorts = new HashMap<>();
```
radumereuta: Where is the sort here? You have a mapping from KVariable to KVariable. Something is wrong, since the name of this variable doesn't match it's type.

kernel/src/main/java/org/kframework/kore/compile/SortCells.java
```diff
@@ -144,10 +145,9 @@ void addOccurances(KLabel cell, KVariable var, List<K> items) {
 
         K replacementTerm() {
             if (remainingCells.size() == 1) {
-                return var;
+                return KVariable(var.name(), var.att().remove("sort"));
```
radumereuta: There are a lot of places where you use strings instead of constants. Please fix them all.

kernel/src/main/java/org/kframework/krun/KRunModule.java
cos, line 210: @dwightguth, I'm tempted to reply but this is not the right place for this discussion.
Feel free to do whatever you like.

kernel/src/main/java/org/kframework/kore/compile/GenerateSentencesFromConfigDecl.java
cos: @radumereuta, actually @dwightguth was really into having typed attributes and he convinced me. Still, here it is not that relevant.

still, @dwightguth, why require the `,""`?

kore/src/main/scala/org/kframework/builtin/Sorts.scala
radumereuta, line 21: I'm not sure I understand the question. Yes, KBott is different than KItem when I'm parsing. What does that have to do with this instance?

java-backend/src/main/java/org/kframework/backend/java/util/Z3Wrapper.java
dwightguth, line 85: it's a way of limiting the amount of dependency injection we are using. We need to be able to construct ProcessBuilderss specially because they need to be configured with the working directory and environment of the calling process, even if it is being invoked from the k server.

k-distribution/src/test/resources/convertor-tests/kore_imp.k
dwightguth, line 27: yes, I have tested this on the sum program.

java-backend/src/main/java/org/kframework/backend/java/compile/KOREtoBackendKIL.java
```diff
@@ -61,7 +60,7 @@ public KApply KApply(KLabel klabel, org.kframework.kore.KList klist, Att att) {
     }
 
     public Term KApply1(org.kframework.kore.KLabel klabel, org.kframework.kore.KList klist, Att att) {
-        return KItem.of(KLabel(klabel.name()), KList(klist.items()), context).evaluate(context);
```
andreistefanescu, line 64: Why remove this? I believe we added it to convert builtin collection from klabel/klist to internal representation.

kernel/src/main/java/org/kframework/kore/compile/GenerateSentencesFromConfigDecl.java
dwightguth: it crashes otherwise. I forget why exactly, but you can look into it if you want. I believe it was an error in Down which incorrectly handled an attribute with a KList of length 0.

java-backend/src/main/java/org/kframework/backend/java/compile/KOREtoBackendKIL.java
dwightguth, line 64: the code as it was evaluates the left hand side of rules with themselves, which causes significant problems for rewriting. So I replaced it with a call to KILtoBackendJavaKILTransformer.evaluateDefinition; it seems to work.

k-distribution/include/builtin/domains.k
dwightguth, line 12: This attribute is being used by the Java backend to initialize information about the data structures in the definition.

java-backend/src/main/java/org/kframework/backend/java/util/Subsorts.java
```diff
@@ -61,6 +61,7 @@ public Subsorts(Module module) {
                 .collect(Collectors.toSet());
 
         this.subsort = ArrayTable.create(sorts, sorts);
+        Set<org.kframework.kore.Sort> kSorts = Sets.newHashSet(Sorts.K(), Sorts.KList(), Sorts.KItem(), Sorts.KLabel(), Sorts.KBott(), Sorts.KVariable(), Sorts.KString());
```
andreistefanescu: where is this used?

java-backend/src/test/java/org/kframework/backend/java/builtins/BuiltinIOOperationsTest.java
```diff
@@ -1,36 +0,0 @@
-// Copyright (c) 2014-2015 K Team. All Rights Reserved.
```
andreistefanescu, line 1: why remove the test?

andreistefanescu:
lgtm

java-backend/src/test/java/org/kframework/backend/java/builtins/BuiltinIOOperationsTest.java
dwightguth, line 1: it's not really doing anything particularly valuable and I didn't really want to fix it.

dwightguth:
I have tried to address most of everyone's comments. Assuming the tests still pass, can you all take another look and let me know if you think this is ready to merge or if you want me to do anything more in particular? @cos @radumereuta @andreistefanescu

andreistefanescu:
you can merge.

dwightguth:
Jenkins: test this please

radumereuta:
lgtm

dwightguth:
Jenkins: test this please

Issue #1489, opened 2015-05-06
Can't Install K


lucaspena:
When running both "mvn install" and "mvn verify", the installation process hangs after "main:". There were also some errors when installing, such as "java.io.IOException: Cannot run program "kast" (in directory ...)" but this did not stop the installation process.

radumereuta:
Could you provide a few more information please?
OS, Java version, make sure you have a recent maven version.

If these things don't work, I could also have a look directly at your computer. It's working fine for most of us, so such problems are interesting.

dwightguth:
@radumereuta from the error he mentions I suspect that K is not in his PATH. You might test mvn verify that way and see if it reproduces the issue.

radumereuta:
Hmm... true. Also, running the examples could fail if you run the tests on windows. The tests are set up to call `kast`, but on windows you have to call `kast.bat`. It's just some weird thing in how java executes commands.

lucaspena:
It seems as though @dwightguth was right, I just added K to my PATH. I'm running mvn verify now to make sure but it looks like it's already gotten farther than it had before. Thanks for your help.

traiansf:
For my part, I never run more than mvn package in general, and on  Windows in special.

If K is added to the path, then mvn package seems enough for most practical purposes (unless one wants to stress their own computer by running the mighty K test suite).

radumereuta:
I usually run `mvn package -DskipTests`. Even faster. It just builds the tool in order for me to run definitions, skipping all unit tests.

grosu:
Guys, can we do something to avoid such silly errors?  I know you will say that this is not urgent and that we have something else which is more important or more intellectually interesting to do, but in the end these minor things matter a lot.  @lucaspena was really motivated, so he bothered to report the issue and try to make it work, but others will simply give up and never look back.  They would rightfully say that if they have problems installing the thing, likely they will also have lots of problems running it, so why bother.  So _please_, let us stop for awhile from what we do and try to find a clean solution to this (which hopefully does not involve asking users to read in detail some README in order to do something that is typically not done when installing a Java app from sources).

andreistefanescu:
I think that `ktest` and `mvn package | verify | ...` should not use `PATH` for any of the K related binaries/libraries. @dwightguth is that hard to achieve?

dwightguth:
[...]

PR #1491, opened 2015-05-06
Improve README for #1489.

radumereuta:
@dwightguth please review.
I'm also waiting for suggestions on how to improve the error message here.


grosu:
I don't like this solution.  It is a patch to a symptom and not a clean fix.  We just tell people what to do if things go wrong, but the way the README is written things _will_ go wrong:

First, we say that updating the PATH is optional.  Then we tell people "To completely test the current version of the K framework, run mvn verify".  But this is going to fail if they do not update the PATH.  So we either require updating the PATH or alternatively we fix ktest to not fail if the PATH is not updated.

Grigore

dwightguth:
Frankly I think the correct solution is to set the PATH variable in all processes spawned by ktest. it's not like we don't already have the path in question at that point.

radumereuta:
Should it be inside `ktest`, or `krun`? `krun` is the one that calls `kast`.

PR #1492, opened 2015-05-07
Adding /bin to the PATH in ktest

radumereuta:
so test work even if the user didn't update his environment.

@dwightguth please review.

### DIFF
--- a/kernel/src/main/java/org/kframework/ktest/Proc.java
+++ b/kernel/src/main/java/org/kframework/ktest/Proc.java
@@ -151,7 +151,6 @@ public class Proc<T> implements Runnable {
         String path = this.env.keySet().stream().filter(key -> key.toUpperCase().equals("PATH")).findFirst().orElseGet(() -> "PATH");
         if (this.env.containsKey(path)) {
             String allPaths = this.env.get(path);
-            System.out.println("allPaths = " + allPaths);
             if (!allPaths.contains(binDir))
                 this.env.put(path, allPaths + ";" + binDir);
         } else {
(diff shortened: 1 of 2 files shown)


andreistefanescu:
did you try a fresh installation on a machine without a changed `PATH`?

radumereuta:
No.

radumereuta:
Tested without the path, and it works as expected. Can we merge this?

andreistefanescu:
Thanks! You can merge.

kernel/src/main/java/org/kframework/ktest/Proc.java
```diff
@@ -141,9 +142,20 @@ public Proc(T obj, String[] args, File inputFile, String procInput,
...
         success = options.dry;
         this.warnings2errors = warnings2errors;
+        // add the /bin directory to the PATH if not there already
+        String binDir = ExecNames.getBinDirectory();
+        // windows is case insensitive w.r.t environment variables, so first find the exact match
+        String path = this.env.keySet().stream().filter(key -> key.toUpperCase().equals("PATH")).findFirst().orElseGet(() -> "PATH");
+        if (this.env.containsKey(path)) {
+            String allPaths = this.env.get(path);
+            if (!allPaths.contains(binDir))
+                this.env.put(path, allPaths + ";" + binDir);
```
dwightguth, line 155: you need to use File.pathSeparator instead of ";" since it is separated by ":" on linux.

radumereuta:
Ah, I will create a new pull request in just a sec.

PR #1493, opened 2015-05-07
Making :: and <: strict for the inner sort.

radumereuta:
@dwightguth please review.
This should fix your problem with not being able to chose certain productions.


dwightguth:
I don't understand how this can have fixed the issue since you don't seem to be using the strict variable anywhere.

radumereuta:
@dwightguth please re-review.

PR #1495, opened 2015-05-07
kast with new pipeline + external parsers + resolve anonymous vars + …

dwightguth:
…misc

**please review**


dwightguth:
@cos @andreistefanescu @radumereuta please review

radumereuta:
lgtm

java-backend/src/main/java/org/kframework/backend/java/kil/Rule.java
```diff
@@ -132,9 +131,9 @@ public Rule(
         setLocation(oldRule.getLocation());
         setSource(oldRule.getSource());
 
-        if (oldRule.containsAttribute(Constants.STDIN)
+        if ((oldRule.containsAttribute(Constants.STDIN)
                 || oldRule.containsAttribute(Constants.STDOUT)
-                || oldRule.containsAttribute(Constants.STDERR)) {
+                || oldRule.containsAttribute(Constants.STDERR)) && lhsOfReadCells != null) {
```
andreistefanescu: is this a way os saying the rule comes from the new pipeline and not from the old one?

andreistefanescu:
lgtm

java-backend/src/main/java/org/kframework/backend/java/kil/Rule.java
dwightguth: well that's effectively what it's doing, since we haven't added support for IO yet. It wasn't needed before because we were deleting all rule attributes, but now it is.

andreistefanescu: I think I would prefer to see a flag for that rather than this ad-hoc check.

cos:
lgtm.

PR #1496, opened 2015-05-08
rule initializers

dwightguth:
**please review**


dwightguth:
@bmmoore @cos please review

dwightguth:
Jenkins: test this please

k-distribution/include/builtin/kast.k
```diff
@@ -40,15 +40,18 @@ module KSEQ
...
 
 
 // To be used when parsing/pretty-printing symbolic configurations
 module KSEQ-SYMBOLIC
   imports KSEQ
-  syntax KVariable ::= r"(?<![A-Za-z0-9_\\$!\\?])(\\$|\\!|\\?)?([A-Z][A-Za-z0-9'_]*|_)"   [token, autoReject, hook(org.kframework.kore.KVariable)]
-  syntax KBott     ::= KVariable
-  syntax KLabel    ::= KVariable
+  syntax KVariable  ::= r"(?<![A-Za-z0-9_\\$!\\?])(\\!|\\?)?([A-Z][A-Za-z0-9'_]*|_)"   [token, autoReject, hook(org.kframework.kore.KVariable)]
+  syntax KConfigVar ::= r"(?<![A-Za-z0-9_\\$!\\?])(\\$)([A-Z][A-Za-z0-9'_]*)"          [token, autoReject]
```
cos, line 51: instead of extending the KORE syntax with a special variable, it may be better to implement the variables through hooks. the `$PGM` would become something like `syntax Input ::= "$" String [ hook(org.kframework.builtin.Input ]`. From the KORE perspective, it would be just another `KApply`.

dwightguth, line 51: note that I am not giving this particular token a hook, nor does it `down` into anything unique. It is simply another KToken of a specific sort.

cos, line 51: we've been usually very careful about adding anything to the KORE syntax, but this will require a bit of a clean-up soon anyhow.

cos:
lgtm.

k-distribution/include/builtin/kast.k
cos, line 51: Actually, I don't think we should have KConfigVar as a special construct that is linked by magic. It needs to follow the same conventions as everything else, including being hooked in order to work.

cos:
the branch doesn't merge cleanly now. @radumereuta helped fix things: https://github.com/kframework/k/commits/feature/ktest-tutorial2

k-distribution/include/builtin/kast.k
dwightguth, line 51: Again, I am not actually changing the syntax; merely changing the parse for variables that begin with a $ so that they are no longer variables, but KTokens. I think it's fine. But if you really want to change it, I don't have any particular objection.

bmmoore:
Could you include some comments and unit tests?

k-distribution/include/builtin/kast.k
cos, line 51: What I'm saying is that $PGM should not be a token but a KApply hooked (via the standard hooking mechanism) to a stream which happens to be the program. But having things linked in any way is still a step forward, so go ahead with the merge from this perspective. It will be clearer what I mean once I get to this and make a pull request.

cos:
ok, I'll retract the "lgtm". The configuration initializers are not the same as described on the wiki:
https://github.com/kframework/k/wiki/KAST-and-KORE

Instead of cleanly initializing the configuration, an `Init` Map (which contains the `$PGM`) is propagated through all initializers. I see why you went on this route. If you consider the `$PGM` as a special magic variable you need to somehow resolve it to the program at the end of initialization. Still, this is unnecessary if we consider the `$PGM` as a hooked function. 

```
  rule =>(initTopCell(Init),<top>(initKCell(Init),initStateCell(Init))) requires #token(Bool,"true") ensures #token(Bool,"true") []
  rule =>(initStateCell(Init),<state>(_Map_())) requires #token(Bool,"true") ensures #token(Bool,"true") []
  rule =>(initKCell(Init),<k>(Map:lookup(Init,#token(KConfigVar,"$PGM")))) requires _andBool_(isK(Map:lookup(Init,#token(KConfigVar,"$PGM"))),isPgm(Map:lookup(Init,#token(KConfigVar,"$PGM")))) ensures #token(Bool,"true") []
```

Also, there's a bug in `plugConfigVars`. The empty map is initialized with `.Map()`, while the klabel for an empty map is `_Map_`. This breaks at least tiny. We should have some tests for this part of the pipeline.

dwightguth:
`plugConfigVars` is correct. _Map_(.KList) is a badly typed term. You should fix your rewriter to obey the written syntax of maps, in which .Map is the unit. A production with two NonTerminals should always correspond to a KApply with two children in its klist. I have no idea why you are so interested in breaking this property, but it's an extremely valuable one.

As far as the initializers go, I like the idea you propose in theory but I'm not sure it works out in practice in a way that is particularly nice. In order to achieve the semantics you describe, it becomes necessary to decompose a hook string in some mysterious fashion into pieces that tell you that the hook corresponds to a configuration variable, and the name of the configuration variable. I would strongly prefer hook strings to remain opaque and statically fixed in each backend. As far as I can tell there isn't a way to do both of these, and I feel like the solution I went with is completely feasible and easy to understand. But I am willing to have discussion about how to proceed here; I just don't see a better solution yet based on the information I currently have.

cos:
Regarding .Map, _Map_ is the current klabel in misc.k

The $PGM can be:

syntax ProgramVariable ::= "$" String [hook=....]

(sent from mobile phone)

dwightguth:
misc.k is deprecated. I left it in place because I didn't want to spend time trying to fix your rewriter, but you should fix your code to be compatible with domains.k, which contains a much more accurate and complete prelude.

dwightguth:
@bmmoore I added some comments, let me know if you want comments anywhere more specific as well. As far as testing goes, the functionality I am adding is tested by TstBackendOnKORE_IT... do you want more specific tests?

grosu:
When I wrote https://github.com/kframework/k/wiki/KAST-and-KORE, what I had in mind for configuration variables was to regard them as constants with desugaring macros:

```
syntax KItem ::= "$PGM" | "$STATE"
rule $PGM => "int x; x =7;"
rule $STATE => "x |-> 0"
...
rule initKCell => <k> #parse($PGM, IMP-SYNTAX) </k>
...
```

I still believe that this is the ultimate, most general solution.  Recall that macros are meant to be like the anywhere rules, but they apply statically to the entire definition (and program).

However, I agree that this general solution would be inconvenient right now, because it will require to re-kompile the definition each time, which with our current kompile times is a problem.  Even if it took 10 seconds to kompile the entire C semantics, which is not even the case, it is still too much to pay when executing each program.  I believe Cosmin's solution is somewhat equivalent, passing the macros job to functions.  So I am OK with Dwight's solution for now.

bmmoore:
For comments, I am concerned about documenting the translation strategy, especially now that it doesn't match what's documented on the wiki. The comments are pretty good, but a general description of the translation somewhere is what's really missing. Maybe it's best to put that on the wiki and refer to it from there?

As for unit tests, I mean an example that tests just this pass, so we have at least a handful of test cases showing expected input and output of just this configuration desugaring pass.

I'm fine with the general implementation strategy. This approach may even be a bit better than macros, because I think we should continue enforcing that configuration variables are used only in the initial configuration, rather than letting them be used in any rule.

However, one detail needs to be improved - the treatment of cell initializers added in completion. It currently supplies an empty map which will lead to a runtime error if the cell references confiugration variables, when we could easily produce a static compilation error.
I suggest that cells not containing (transitively) configuration variables should get an initializer label that does not take any arguments (while continuing to pass a map into the ones that do need the variables). Then cell completion can revert to generate a simple no-argument label application for legal code, and more easily generate an appropriate error message otherwise. (I'd rather see the information explicitly represented, maybe modifying the existing ConfigurationInfo lookup methods, but if you insist on working by checking the arity, comments around the creation and check should document that you expect initializers to have those two arities, from which you infer additional semantics).

cos:
@bmmoore, what is the advantage of having configuration variables to be used only in the initial configuration? Furthermore, why do we need the concept of "configuration variable" in the first place? They can simply input streams to the definition:

```
syntax ConfigVar ::= "$" String [ hook(ReadNamedStreamInput) ]
```

Then:

```
rule initKCell => <k> #parse($PGM, IMP-SYNTAX) </k>
```

becomes, in (semi)labeled form:

```
rule initKCell => <k> #parse(`$`("PGM"), IMP-SYNTAX) </k>
```

where `$` is just a function reading the entire stream named "PGM". 

This approach gets rid of the concept of "configuration variable" and give us more flexibility. E.g., it allows us to easily write definitions that behave like continuous interpreters of input streams (e.g., a console).

bmmoore:
The advantage of allowing configuration variables only in the initial state is that that makes them purely part of the initialization process, rather than the definition of the semantics.

Generalizing from simple named values to arbitrarily input streams is just unnecessary complexity. Do we really want to think about what happens if the contents of $PGM change halfway through parsing?!

dwightguth:
Given that we already have a solution that meets our needs (which Brandon and I agree on and which is mostly implemented in this pull request with some minor alterations to follow soon), I would prefer that we close this discussion and move on, only re-evaluating the implementation if we can come up with some concrete use case justifying the need for things to behave differently than they already do. Is this going to be a problem?

grosu:
Oh, is this still not closed?  I thought that it was over.  So yes, we do have a solution which works fine and does everything we were able to do before, so we are good for now.  Let us please not waste any more time with it.

Any re-design or change of this from here on should only happen after we discuss it in person.  All of us directly interested in this.  We have to look at all possible uses of the concept before proposing (or worse, implementing) any new design.  For example, we want to allow programs to contain variables, or even be just variables, and then do semantics-based program generation, like @yilongli used to do and like Tom (a new intern) will do soon.  In that light, $PGM _is_ a variable, one which will be unified as rewriting takes place via narrowing.  So your solution, @cos, is not going to be general enough.  And I believe mine was not general enough either.

But let's not get into this, it is really not important now.  The bottom line: we have something which works with the new KORE and can do everything we wanted it to do so far, so we are fine.  _Please_ merge asap and any other change of it should be bug fixes, not re-designs, unless we discuss and agree on a new design.  Same with configuration abstraction (which we also know that it is not as general as it can/should eventually be), please, no changes there either, except for bug fixes!  Let us first have everything work with KORE and the worry about making things as general as possible.

bmmoore:
Ok, looks good now.

Issue #1402, opened 2015-03-13
Variable => KVariable


grosu:
For uniformity with other K-related sorts, we should rename the sort `Variable` (telling K which terms can be used as variables in bindings) to `KVariable`.

I'd like to use this opportunity to see if we are on the same wavelength regarding the play between `bind` and `KVariable`.  Suppose that I define a let binder which can bind variables through constructor patterns, something like this:

```
...
syntax Exp ::= "let" Pattern "=" Exp "in" Exp  [binder(1 -> 3)]
syntax KVariable ::= Id
```

Then what I expect to happen within a term like say `let p = e1 in e2`, is that all the KVariables that occur in p are bound within e2 (but not within e1).  So, for example, substitution

```
(let [h|t] = [1,2,3] in if h<5 then x else t)[h/x]
```

reduces to something like this:

```
let [_72|t] = [1,2,3] in if _72<5 then h else t
```

Is that correct?  Does this already work?

radumereuta:
Something to mind for the future, is that I'm already using the `KVariable` sort in the definition of full K.
If you plan to use it like above, that will break the parser for rules. I think we should have a convention to avoid name collision with parsing only sorts. Maybe take over `#`? In the past it wasn't very clear what was the meaning.

dwightguth:
Assigning to Radu so we don't lose track of this. I think that convention is good. #Variable for the parser sort of K variables, and KVariable for the user-defined sort used by the substitution and pattern-matching modules.

radumereuta:
Can we close this, or do we need to assign someone else?

dwightguth:
I am going to close this. We still need to deal with the Variable user sort, but that is specific to substitution and pattern-matching, which we intend to redo in the new KORE regardless.

grosu:
Let's just make sure we don't forget to rename the user Variable into KVariable, which was the original issue here.

bmmoore:
I think we were pretty close to forgetting that.

PR #1498, opened 2015-05-08
Rename KVariable to #KVariable to allow bidings (see #1402).

radumereuta:
@dwightguth please review.


dwightguth:
You should also remove the KVariable case from RuleGrammarGenerator now that it begins with a hash.

grosu:
Would this be a good moment to also rename Variable to KVariable in what the user sees in K definitions, to be consistent with KResult, KItem, etc.?

dwightguth:
Jenkins: test this please

dwightguth:
@radumereuta it looks like you need to push a new commit to github to make Jenkins pick up your build.

radumereuta:
@dwightguth please re-review. Tests passed.

Issue #1500, opened 2015-05-11
strange behavior when in --search mode and read from istream


kheradmand:
Hi,

I am trying to do one of the exercises in the tutorials: 1_k/4_imp++/lesson_7/exercises/non-determinism

When I run 
`krun tests/nondet-0.imp --search  --pattern '<k> K:K </k> <out> ListItem(#buffer(S:String)) </out>' < tests/nondet-0.imp.in`

I get the folowing output:

```
Search results:

Solution 1:
K:K -->
read ( ) ~> HOLE + String2Int ( "10 20 30" ) ~> HOLE + read ( ) ~> HOLE, 
  .AExps ~> print ( HOLE ) ; ~> .Stmts
S:String -->
""

Solution 2:
K:K -->
read ( ) ~> HOLE + read ( ) ~> HOLE + String2Int ( "10 20 30" ) ~> HOLE, 
  .AExps ~> print ( HOLE ) ; ~> .Stmts
S:String -->
""

Solution 3:
K:K -->
read ( ) ~> String2Int ( "10 20 30" ) + HOLE ~> HOLE + read ( ) ~> HOLE, 
  .AExps ~> print ( HOLE ) ; ~> .Stmts
S:String -->
""

Solution 4:
K:K -->
read ( ) ~> String2Int ( "10 20 30" ) + read ( ) + HOLE ~> HOLE, .AExps ~> 
  print ( HOLE ) ; ~> .Stmts
S:String -->
""

Solution 5:
K:K -->
read ( ) ~> read ( ) + HOLE ~> HOLE + String2Int ( "10 20 30" ) ~> HOLE, 
  .AExps ~> print ( HOLE ) ; ~> .Stmts
S:String -->
""

Solution 6:
K:K -->
read ( ) ~> read ( ) + String2Int ( "10 20 30" ) + HOLE ~> HOLE, .AExps ~> 
  print ( HOLE ) ; ~> .Stmts
S:String -->
""
```

It seems that when in --search mode, it reads the whole input stream in the first read operation (it does not behave like this if not in --search mode).

I think it is K's problem, or maybe I'm missing something.

daejunpark:
Fixed in #1501

PR #1501, opened 2015-05-11
Reapply commit 5a9a69f that was missed during migration uris.k to io.k. Fix Issue #1500.

daejunpark:


dwightguth:
please apply this fix also to domains.k in k-distribution/include/builtin

PR #1502, opened 2015-05-11
change unparser.k to print in correct kast format

laurayuwen:
please review.

### DIFF
--- a/k-distribution/tests/regression/tests/test20.issue.out
+++ b/k-distribution/tests/regression/tests/test20.issue.out
@@ -2,4 +2,5 @@ Search results:
 
 Solution 1:
 Result:K -->
-foo ( true , '`(`) , x , .Exps )
+foo ( true , `'`(`) , x , .Exps )
+
--- a/k-distribution/tests/regression/tests/test20.issue.out
+++ b/k-distribution/tests/regression/tests/test20.issue.out
@@ -2,5 +2,5 @@ Search results:
 
 Solution 1:
 Result:K -->
-foo ( true , `'`(`) , x , .Exps )
+foo ( true , `'()` , x , .Exps )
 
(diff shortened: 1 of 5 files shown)


laurayuwen:
There are three tests whose expected output are in the old kast format, it is not a surprise they failed. Maybe we can change the expected output, they should be in the new kast format.

laurayuwen:
@dwightguth Please review.

k-distribution/tests/regression/tests/test20.issue.out
```diff
@@ -2,4 +2,5 @@ Search results:
 
 Solution 1:
 Result:K -->
-foo ( true , '`(`) , x , .Exps )
```
dwightguth, line 5: this does not appear to be correct. It should be ``'()``

laurayuwen, line 5: The plain format is foo(true, '(), x, )
new kast adds a backstick to both ' and ( 
I don't know why there is a ` inside the brackets, but it was there before, I don't think I introduced it

dwightguth, line 5: it's not adding a backtick to (, it's delimiting the beginning and end of the klabel with a backtick. The klabel's string is "'()". Previously we were escaping klabels with a backtick before ( and ), but now we do not. So ``'()`` is correct.

laurayuwen, line 5: Oh, yes, thanks for the explanation, let me remove the backstick I added to the beginning of bracets and instead append it to the end of klabel.

laurayuwen, line 5: There was four replace here: 
replaceAll("`", "``").replaceAll("\\(", "`(").replaceAll("\)", "`)").replaceAll(",", "`,")
you mentioned the two ( ), are the other two also not needed, or I just remove the two related to ( ).

laurayuwen, line 5: I removed all replacements for klabel. Please review.

dwightguth:
good enough for government work. :P we are going to prune this code eventually anyway so the last remaining detail (escaping backticks in klabels) isn't that crucial.

PR #1508, opened 2015-05-18
beginning of new rewriter interface + primitive new debugger

dwightguth:
@andreistefanescu @cos please review


dwightguth:
Note that I am not going to merge this yet because I am waiting for Andrei to finish his change first. At this stage I just want feedback on the interfaces proposed in this diff.

kore/src/main/scala/org/kframework/Rewriter.scala
```diff
@@ -19,5 +19,36 @@ trait Rewriter {
    * - for search, it returns an Or with multiple ground terms as children
    */
 
-  def execute(k: kore.K): kore.K
+  /**
+   * Converts a term into a K which has been optimized for execution in the backend.
+   * @param k A K term to convert
+   * @return A K term which, when passed to other methods in this interface, does not require further conversion.
+   */
+  def convert(k: K): K = k
```
cos, line 27: I think this method should be on the `Constructors` instance. Or, if we play a bit with type tags, we could even make it static.

kore/src/main/scala/org/kframework/Rewriter.scala
```diff
@@ -19,5 +19,36 @@ trait Rewriter {
...
+
+  /**
+   * Matches a particular subject against a pattern
+   * @param k The subject to match against.
+   * @param trace true if trace information about the matching should be printed to standard output.
+   * @param rule The rule to match and check the side condition of.
+   * @throws UnsupportedOperationException if the class does not support rule traces, and trace is true.
+   * @return A list of substitutions satisfying the rule.
+   */
+  def `match`(k: K, trace: Boolean, rule: Rule): java.util.List[_ <: java.util.Map[_ <: KVariable, _ <: K]]
```
cos, line 37: We may want to generalize this to unification between two `K`s, with a boolean optional attribute (extra method in Java) telling it to force matching instead of unification. 
The result can be a Matching Logic formula. For simple matches can be a disjunction of conjunctions of equalities, with the lhs of each equality being a variable.

So the method would look something like:
`def unify(pattern: K, subject: K, forceMatching: boolean): K`

We could have two other helper methods:
`def unify(pattern: K, subject: K) : K`
`def match(pattern: K, subject: K) : K`

As for the disjunction, conjunction, and equality I think we probably want to add them to KORE as interfaces extending KApply. We could also have a special `Binding` interface extending `Equality`. 

Having the unification/matching available independent of rules can be very helpful for debugging. When something doesn't match you could, for example, give it a subpattern and a subexpression and check whether it matches.

kore/src/main/scala/org/kframework/Rewriter.scala
```diff
@@ -19,5 +19,36 @@ trait Rewriter {
...
+
+  /**
+   * Substitutes a particular term using the right hand side of a rule.
+   * @param substitution A mapping of substitutions for variables in the rule.
+   * @param rule The rule whose right hand side should be used for the substitution
+   * @return the right hand side of rule substituted using the substitution.
+   */
+  def substitute(substitution: java.util.Map[_ <: KVariable, _ <: K], rule: Rule): K
```
cos, line 45: could substitution be made independent of the rewrite engine?

kore/src/main/scala/org/kframework/Rewriter.scala
```diff
@@ -19,5 +19,36 @@ trait Rewriter {
...
+
+  def execute(k: K): K
+
+  /**
+   * Returns the list of rules in the definition being rewritten over, in the order they should be tried.
+   * These rules can be passed to other methods in this interface without further conversion.
+   */
+  def rules : java.util.List[_ <: Rule]
```
cos, line 53: I'm starting to wonder whether it is a good idea to build a `Rewriter` based on a module.... we're already compiling a module for a particular `Rewriter`, and we want to ability to make rewriters on the fly anyhow. So, I'm thinking, how about we have:
`Definition x Backend => KompiledDefinition`
With the Kompiled definition holding the compiled `Rule`s. Rules would extend `K => K`.

So, in order to get an _execution_ `Rewriter` for the semantics module `IMP` with rules executed in order you would do:
`SequentiallyComposedRule(list of rules)`. 

Basically, I'm proposing to break the `Rewriter` completely apart. I know it is radical but please think about it for a moment.

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -161,7 +161,13 @@ trait Sentence {
...
 case class Context(body: K, requires: K, att: Att = Att()) extends Sentence with OuterKORE with ContextToString
 
-case class Rule(body: K, requires: K, ensures: K, att: Att = Att()) extends Sentence with RuleToString with OuterKORE
+trait Rule extends Sentence {
+  val body: K
+  val requires: K
+  val ensures: K
+}
+
+case class RuleImpl(body: K, requires: K, ensures: K, att: Att = Att()) extends Rule with RuleToString with OuterKORE
```
cos, line 170: @dwightguth, I think this `RuleImpl` goes towards what I was saying above. Still, I think we should keep implementation related stuff out of the definition.

kore/src/main/scala/org/kframework/Rewriter.scala
dwightguth, line 27: it can't be static, it behaves differently on each backend. For example, in the Java backend, it converts K's into org.kframework.backend.java.kil.Term.

kore/src/main/scala/org/kframework/Rewriter.scala
dwightguth, line 45: no, because it has to be able to do things like resolve functions.

kore/src/main/scala/org/kframework/Rewriter.scala
dwightguth, line 37: I would propose instead to have the following interface for unification:

```
def unify(k: k, constraint: K, rule: Rule) : List<K> //constraints of sort #Formula
def substitution(constraint: K) : java.util.List[_ <: java.util.Map[_ <: KVariable, _ <: K]]
```

@andreistefanescu would this give you enough info to do symbolic execution? We would still have to revisit the issue of verification.

kore/src/main/scala/org/kframework/Rewriter.scala
dwightguth, line 53: This is completely meaningless in any kind of non-Java context. A KompiledDefinition in which Rules extend a function type is utterly incompatible with e.g. an OCAML backend. I propose the following semantics for backends:

```
public interface Backend {
  void compile(CompiledDefinition def);
  Rewriter load(Module mod);
}
```

The first method is called after the main compilation pipeline finishes, and the second is called by krun to obtain a rewriter.

kore/src/main/scala/org/kframework/definition/outer.scala
dwightguth, line 170: You're welcome to refactor it if you want, but bottom line is that Rule needs to be an interface. I did this cheaply and simply since I wasn't entirely sure how I wanted it. If you want it a different way, feel free to adjust it after I finish this.

kore/src/main/scala/org/kframework/Rewriter.scala
cos, line 27: the `Constructors` instance specific to each backend already knows about the backend's classes. So the convert is just a generic recursive function (see `tiny.Constructors`). The only backend specific piece is the `instanceof` check.

kore/src/main/scala/org/kframework/Rewriter.scala
cos, line 37: @dwightguth, please read my comment above again. A few notes:
- there is no `#Formula` sort. Everything is a formula. in effect, there is no need of a separate `constraint` parameter.
- the result of unification is not `List<K>`, but a formula, i.e., a `K`

I know all of this is very different from the way we were doing things, but I believe we should do it right this time, not reimplement what we had.

kore/src/main/scala/org/kframework/Rewriter.scala
cos, line 45: The fact functions are evaluated during substitution is just an artifact of the Java rewrite engine. There is nothing fundamental about it. Substitution and function evaluation are essentially  separate operations. The only reason to merge them is performance, and I don't think we actually gain much there anyhow.

kore/src/main/scala/org/kframework/Rewriter.scala
cos, line 53: Everything not-Java is incompatible with anything Java. If we ever want to integrate OCaml or any other backend with the Java execution engine, we will need to have an adaptor to give it a Java interface. In that case, the Java function type is just as hard to as any other interface you may like to have.

The main compilation pipeline is already backend specific. having `compile(CompiledDefinition` sounds unnatural. Also, we may want to construct a `Rewriter` from an entire definition to give it meta-level access to non-imported modules.

kore/src/main/scala/org/kframework/definition/outer.scala
cos, line 170: I don't think we should mix the definition `Module` `Rule` with the rules of a particular backend. The way I designed them, the `.definition.*` classes are not meant to ever include the internal representation of a particular backend.

cos:
I think we have reached the point where we can seriously discuss this interface. I will put together  a short proposal when I will find a bit of time in the next week or so.

dwightguth:
Closing. Not going to bother with this, just move on to ocaml. We can resurrect this code later if need be.

PR #1511, opened 2015-05-19
Clean up traces of Context sensitive rules from the parser kernel.

radumereuta:
Removed Primitive state adding a constant to the AST, so for now DeleteRule is not used (will be when we have whitespace non-terminals).

@dwightguth please review.
The performance improvement is minimal, but the kernel is a bit cleaner now. Managed to remove over 200loc.


kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Parser.java
```diff
@@ -675,37 +528,7 @@ boolean addNTCall(Function call, final Function exit) {
          * @return 'true' iff the mapping in this function changed
          */
         boolean addRule(Function that, final Rule rule, final StateReturn stateReturn, final Rule.MetaData metaData) {
-            if (rule instanceof ContextFreeRule) {
```
dwightguth, line 678: can you remove the ContextFreeRule class and make everything just be over the Rule interface?

dwightguth:
lgtm. merge if tests pass

PR #1512, opened 2015-05-19
Change #klabel attribute back to klabel

bmmoore:
Attempting to change the attribute label used for giving a klabel back to klabel as it has been, instead of the "#klabel" as used in the new code. There are still test failures.


bmmoore:
@radumereuta, could you look at the failure? The first error is this:
    Failed tests: 
    ParserTest.testListAmbiguity:575 The error:  expected:<NOKLABEL(amb(#token(K,"-1"),NOKLABEL(#token(K,"1"))))> but was:<#token(K,"-1")>

bmmoore:
I see I missed some files, which seem to fix this. I'm recreating this with a cleaner branch.

PR #1514, opened 2015-05-20
Put value first in KToken and #token syntax.

bmmoore:
Changes the argument order in KToken and the
expected order in parsing and printing #token,
to put the value before the sort,
as documented and shown on the Wiki.


radumereuta:
lgtm

PR #1516, opened 2015-05-21
Add hook function4 printing java double

xiaohe27:
@laurayuwen @dwightguth This pull request aims to provide a function for java-semantics so that the floating point (specifically the type 'double') data type in java can be printed by k the same way as java does.

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinStringOperations.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinStringOperations.java
@@ -1,14 +1,14 @@
 // Copyright (c) 2013-2015 K Team. All Rights Reserved.
 package org.kframework.backend.java.builtins;
 
-import java.math.BigInteger;
-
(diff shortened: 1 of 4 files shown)


dwightguth:
Haha, no. This belongs in the Java semantics, not the K standard library.

andreistefanescu:
Actually, this is a reasonable format. What is the standard format for other languages? I think C/C++ and Python have a similar format. Can you check if it is the same?

dwightguth:
Neither C, Python, nor Java yield the same result that I observe:

Java, String.valueOf: `2.2250738585072014E-308`
Python, float.**repr**: `2.2250738585072014e-308`
C, printf("%f", ...): `0.000000`

This is quite aside from the fact that all three of these languages implement completely disparate string formatting languages that a semantics of these languages logically would have to deal with regardless, for which this method is entirely ill-suited. I maintain that this is simply not suitable material for the K standard library.

xiaohe27:
Yes, you are right, I agree. I'll try to implement this using K instead :)

Issue #1521, opened 2015-05-27
~> cannot be the immediate child of #KList


dwightguth:
try to parse the following in the KSEQ module:

```
`foo`(`bar`(.KList) ~> .K , .KList)
```

This should work, but it declares the priorities to be wrong. I suggest that you generate priority declarations for ~> and => and manually inspect them rather than trying to rely on an ad-hoc compiler pass. You will have to do this in order for unparsing to work anyway.

PR #1522, opened 2015-05-28
Fixing #1521.

radumereuta:
@dwightguth please review.


dwightguth:
Can you add a test?

dwightguth:
This looks okay, but I still feel like having a  custom pass for this isn't really what we want... This will eventually lead to unsoundness in the unparser because you are not representing the priorities as actual priorities but rather handling them ad-hoc. I'm not going to oppose merging this because it gets the job done, but I strongly encourage you to consider automatically generating the relevant "syntax priority" statements rather than relying on a custom pass.

radumereuta:
I will keep this in mind for the future. I would like to have the POSet performance fixed at least before I do this.

dwightguth:
okay. lgtm.

PR #1523, opened 2015-05-29
Fix pgm lists

radumereuta:
@dwightguth please review.
I've cleaned up a bit the list generation mechanism.
From the user specification of `Es ::= List{E,","}` I first generate three productions which can be used directly in rules:

```
Es ::= Es "," Es [right, userList, klabel(...)]
Es ::= ".Es"     [userList, klabel(...)]
Es ::= E         [userList]
```

For programs though these don't work so well, so I'm detecting these productions and generate a different pattern:

```
Es#Terminator ::= "" [klabel('.Es)]
Ne#Es ::= E "," Ne#Es [klabel('_,_)]
Ne#Es ::= E Es#Terminator [klabel('_,_)]
Es ::= Ne#Es
Es ::= Es#Terminator // if the list is *
```

I'm sure there are ways to improve this, but it makes sense for the tests we have right now. I'm expecting this to evolve once the implementation gets more complete.


dwightguth:
I haven't looked at this too closely, but I propose you move the production-modifying code into any module that imports a module in kast.k called "LIST-EPSILON"

radumereuta:
@dwightguth please review.
This ended up being a lot more involved than I anticipated.
I had to change a bit how we create parsers and how we handle them when we consider caching.
Right now I'm using the `seed module`, which imports all the modules which will generate the desuggared productions. Each time I want to recreate this module, I use the `RuleGrammarGenerator` class to regenerate the final grammars, since is pretty cheap to do so.

k-distribution/include/builtin/kast.k
```diff
@@ -1,8 +1,12 @@
 // Copyright (c) 2015 K Team. All Rights Reserved.
 
 // Module defining only the sorts K and KString, useful for modularity
-module BASIC-K
+module SORT-K
```
dwightguth, line 4: I would prefer you copy the module structure of this part of kast.k from my pull request, that way we don't have to worry about whether the merge will trigger latent bugs. If you run into problems because my BASIC-K includes a few other sorts, then expand further from there based on your own needs. Right now this exists in kind of a state in which whichever one of us merges first, we may suddenly find that the other person, when they merge afterwards, suddenly has bugs to clean up that have nothing to do with their own code. And that's not great, hence my wanting to align these closer together before the first merge.

k-distribution/include/builtin/kast.k
```diff
@@ -141,6 +145,21 @@ module AUTO-FOLLOW
   // because it would turn the terminal into the terminal "ab".
 endmodule
 
+module PROGRAM-LISTS
```
dwightguth, line 150: LIST-EPSILON

k-distribution/include/builtin/kast.k
```diff
@@ -141,6 +145,21 @@ module AUTO-FOLLOW
...
+  imports SORT-K
```
dwightguth, line 151: I don't actually see what this import is used for.

kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
```diff
@@ -115,7 +115,7 @@ public int run() {
                     }
                     mod = mod2.get();
                 }
-                K parsed = def.getParser(mod, sort, kem).apply(FileUtil.read(stringToParse), source);
+                K parsed = def.getParser(def.getParserModule(mod), sort, kem).apply(FileUtil.read(stringToParse), source);
```
dwightguth: this shouldn't be necessary because getParser should call getParserModule.

kernel/src/main/java/org/kframework/kompile/CompiledDefinition.java
```diff
@@ -2,6 +2,7 @@
 package org.kframework.kompile;
```
dwightguth, line 2: I disagree with every change in this file.

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -139,7 +139,7 @@ public Definition addSemanticsModule(Definition d) {
     }
 
     public Definition addProgramModule(Definition d) {
-        Module programsModule = gen.getProgramsGrammar(d.mainSyntaxModule());
+        Module programsModule = gen.getProgramsGrammar(d.mainSyntaxModule()).module();
```
dwightguth: again, none of this is necessary as far as I can tell.

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
```diff
@@ -379,72 +379,31 @@ public static RegexTerminal getRegexTerminal(String regex) {
...
-            prod3 = Production(sort, Seq(Terminal("." + sort.toString())),
-                    attrsWithKilProductionId.remove("strict").add("klabel", dropQuote(p.getTerminatorKLabel())));
-
-            res.add(prod1);
-            res.add(prod2);
-            if (!nonEmpty) {
-                res.add(prod3);
-            }
-        }
+        // lst ::= lst sep lst
```
dwightguth: this comment does not seem to be correct

kernel/src/main/java/org/kframework/parser/concrete2kore/ParseCache.java
```diff
@@ -29,12 +29,12 @@ public Module getModule() {
         return module;
     }
 
-    private transient ParseInModule parser;
+    private transient Module parser;
 
-    public ParseInModule getParser() {
```
dwightguth, line 40: again I disagree with the changes in this file

radumereuta:
@bmmoore please review.
I've been discussing this pull request with @dwightguth, and we reached the conclusion that it's not the perfect solution, but is something that can work for now. Please have a look at this and maybe suggest some improvements.

The problem was that I had to add specific productions to the grammar in order to parse different inputs. For example parsing lists in programs requires something different than rules and configurations. To fix this, I created the concept of a `seed` module, which contains the original module, and some marker imports. These marker imports, tell the parser how and how to generate productions. Another problem is that those productions may not be to any use to the user. A good example here is rewriting the grammar to support priorities. This will add new sorts, which we don't want to interfere with variable type inference for example. This is why I create two different modules: one for disambiguation, and one for parsing. I use the parsing module to generate the AST and eliminate some ambiguities related to priorities/associativity, and I use the disambiguation module for the subsort lattice and other information that are visible to the user.

@dwightguth please add other things that I forgot in my comment above.

bmmoore:
I reviewed the design of this with Radu, and I think we need only a little change.
The "parsing module" is already completely encapsulated inside the parser implementation, so no code outside of the parser has to know or care that the information is kept in kore data structures, or ever deal with that object as if it was a module that was part of the definition. It might be ugly as a K module, but as far as I'm concerned it's just an internal data structure, and bonus points to Radu for re-using exiting code to represent it.
The "disambiguation module" is currently exposed through a public acessor, but we saw it is used in only two places and seems to be used incorrectly (radu may need dwight's help to fix this). So, it can be changed to also be completely encapsulated, and then I will care very little about the contents of the module. The only remaining concern is that because it's used for various disambiguation calculations, we might get confusing bugs if the generated disambiguation module isn't faithful to the user's syntax declarations - I think the best we can do there is check that the code copies most productions over unchanged, which I'll check when I read the code after further changes from Radu.

A little more detail on the contents of the disambiguation module:
there are only two things in the generated productions that later compiler passes might need to deal with, which are the productions for the loose versions of cells, and for semantic casts.
The cells are already represented in parse results by using a constant number of reserved labels, including `#cells` and `#dots`. Existing passes handle these just by looking at the strict versions of the cell productions. The semantic casts are currently represented by a family of labels like `#semanticCastToInt`, but I think the current code uses an ugly `String.startsWith()` test to recognize them instead of properly lookup up productions in the disambiguation module, and there are tons of alternative representations we could use.

grosu:
I obviously don't know what you guys discussed in your recent meetings, due to my (overly-extended) travel, but I just want to remind ourselves what we discussed in our meetings before I left, which should still be valid: `List{Sort,Separator}` is just syntactic sugar in Full K, which the frontend desugres differently in different generated KORE modules.  The generated KORE modules will have no `List{Sort,Separator}` anymore, but just normal productions, and they are used for generating different parsers for different tasks.  What @radumereuta said in his original message is a reasonable start.  It should be easy to change that if needed.  What I did not see there were rules in the semantics making use `#rewrite(Module,Term)` to translate the K terms generated by the program parser into K terms recognizable by the semantics; for example, you may have different KLabels for the program lists vs the semantics lists.

radumereuta:
I just commited the modifications we've discussed earlier today, so @bmmoore @dwightguth please review. 
@grosu what I ended up with, is something similar. I start from a module that the user specified, and then from that one I generate modules for different purposes.
For parsing rules, I take all the included syntax, and generate a module specialized for that purpose. To parse programs, I may take the same module and build a different module which is more appropriate for parsing programs.
After all this, I have a few constants:
- each module contains a single parser associated with it. Note that to parse rules, programs and configurations I may need something else, derived from what the users specified, so I generate internally these modules.
- the implementation is clean, since all the modifications are done using only KORE. The parser generator takes as input a module and returns a parser object.

Regarding lists and labels. After a few days discussing with Dwight and Brandon we come up with a solution where I'm actually using two modules. One is the parser module, and I'm using it to generate the internal states of the parser (and that's it). The other one, called the disambiguation module, is used to disambiguate the AST. The parser will produce nodes only in this second module. The first one can have extra sorts, and extra productions that may come from rewriting the grammar to account for priorities/associativity or other transformations, but the disambiguation module is a simple extension that can be used to easily filter and analyze the AST.

I hope I'm making sense since this is a recent result, and we're still working on it.

k-distribution/include/builtin/kast.k
dwightguth, line 4: still relevant

k-distribution/include/builtin/kast.k
dwightguth, line 151: still relevant

kernel/src/main/java/org/kframework/kast/KastFrontEnd.java
dwightguth: still relevant

kernel/src/main/java/org/kframework/kompile/CompiledDefinition.java
```diff
@@ -44,7 +45,20 @@ public CompiledDefinition(KompileOptions kompileOptions, Definition parsedDefini
      * A function that takes a string and the source of that string and parses it as a program into KAST.
      */
     public BiFunction<String, Source, K> getProgramParser(KExceptionManager kem) {
-        return getParser(kompiledDefinition.mainSyntaxModule(), programStartSymbol, kem);
```
dwightguth, line 49: if you keep the getParser method's signature the same, then getGroundParser is no longer a method that needs to exist.

kernel/src/main/java/org/kframework/kompile/CompiledDefinition.java
```diff
@@ -83,7 +96,7 @@ public Module executionModule() {
         };
     }
 
-    public Module getParserModule(Module module) {
```
dwightguth, line 90: something feels wrong about the accessibility of this method now... It should be possible to make this method return the expanded module without having to change all the various call sites for this class.

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
dwightguth: still relevant

kernel/src/main/java/org/kframework/parser/concrete2kore/ParseInModule.java
```diff
@@ -36,15 +36,54 @@
...
+     * necessarily representable in KORE (sorts like Ne#Ids, to avoid name collisions).
+     * In this case the modified production will be annotated with the information from the
+     * original production, so disambiguation can be done safely.
+     */
+    private final Module parsingModule;
     private final Grammar grammar;
-    public ParseInModule(Module module) {
-        this.module = module;
-        this.grammar = KSyntax2GrammarStatesFilter.getGrammar(module);
+    public ParseInModule(Module seedModule) {
```
dwightguth: this constructor feels wrong now... I think it should be a Module and a Definition and then the constructor calls getCombinedGrammar.

dwightguth: either that or make all constructors for this class package-private

dwightguth:
okay, I am done reviewing for now, but I'm going to have to do another pass once some of the crufty changes in this diff are reverted.

kernel/src/main/java/org/kframework/parser/concrete2kore/ParseCache.java
radumereuta, line 40: `ParseInModule` now takes multiple arguments, so the only way to properly create the parsing object, is to call 
`RuleGrammarGenerator.getCombinedGrammar`. I made the cache store only the seed module, and build the parser when it's needed.

dwightguth, line 40: I am pretty sure that this isn't good enough. One of the reasons why we stored the parser module was that it was a huge performance bottleneck to recompute the regex automata. Can you parse the C semantics twice and let me know the initial and incremental parse times? It should be roughly on the order of 3 minutes and 30 seconds.

kernel/src/main/java/org/kframework/parser/concrete2kore/ParseInModule.java
radumereuta: At the moment, this is used only in `ParserUtils` which from what I remember is used only in unit tests.
I will make them package-private.

kernel/src/main/java/org/kframework/kompile/CompiledDefinition.java
radumereuta, line 90: Renamed the method, and now returns only the extension module which will be used for unparsing.

k-distribution/include/builtin/kast.k
radumereuta, line 151: K is the default start symbol if the user doesn't specify one.
I'm subsorting all sorts to K, so every user defined sort is considered when parsing programs.

k-distribution/include/builtin/kast.k
radumereuta, line 4: Merged with master and your changes, but I still need K to be in a separate module, since I'm including it in the program parser, so I can use it as a default start symbol when the user doesn't specify one.

kernel/src/main/java/org/kframework/kompile/CompiledDefinition.java
```diff
@@ -50,6 +51,15 @@ public CompiledDefinition(KompileOptions kompileOptions, Definition parsedDefini
     }
 
     /**
+     * A function that takes a string and the source of that string and parses it as a ground term
+     * in the syntax module into KAST.
+     */
+    public BiFunction<String, Source, K> getGroundParser(KExceptionManager kem, String moduleName) {
+        RuleGrammarGenerator gen = new RuleGrammarGenerator(parsedDefinition);
+        return getParser(gen.getConfigGrammar(kompiledDefinition.getModule(moduleName).get()), Sorts.K(), kem);
```
dwightguth: this doesn't seem right. Why are we calling getConfigGrammar if we aren't parsing a configuration?

k-distribution/src/test/java/org/kframework/kore/convertors/TstBackendOnKORE_IT.java
```diff
@@ -63,7 +63,7 @@ public void kore_imp() throws IOException, URISyntaxException {
         try {
             CompiledDefinition compiledDef = new Kompile(new KompileOptions(), FileUtil.testFileUtil(), kem, false).run(definitionFile, "IMP", "IMP-SYNTAX", Sorts.K());
 
-            BiFunction<String, Source, K> programParser = compiledDef.getProgramParser(kem);
+            BiFunction<String, Source, K> programParser = compiledDef.getGroundParser(kem, "IMP-SYNTAX");
 
             K program = programParser.apply(
                     "int s, n, .Ids; n = 10; while(0<=n) { s = s + n; n = n + -1; }", Source.apply("generated by " + getClass().getSimpleName()));
```
dwightguth: why is this test passing? The program grammar for this language should no longer contain the .Ids production.

dwightguth:
Looks much better but there is still some small stuff I'd like to see fixed.

radumereuta:
That test worked because I was using the configuration parser, which had the `.Ids` token.
I changed it to program parser, and renamed `groundParser` to `configParser`.

dwightguth:
Just delete the getConfigParser method you created. It doesn't seem to be used anywhere except TestTiny which is commented out anyway because it doesn't correctly handle the new strictness.

radumereuta:
Done.

dwightguth:
Merge if tests pass.

PR #1526, opened 2015-06-04
add lemma attribute as workaround for smtlib translation issue

daejunpark:


daejunpark:
I close this, since it hasn't been active. We can bring it back later, if needed.

PR #1528, opened 2015-06-08
first revision of ocaml backend

dwightguth:
@andreistefanescu @cos @bmmoore please review
- adds a number of new compiler passes
- fixes a few minor issues in the Java Backend
- a few other minor improvements to the KORE pipeline
- prototype OCAML rewriter


java-backend/src/main/java/org/kframework/backend/java/kil/Collection.java
```diff
@@ -113,4 +91,8 @@ public ASTNode accept(Transformer transformer) {
         return transformer.transform(this);
     }
 
+    public TermContext context() {
```
andreistefanescu, line 94: is this used?

andreistefanescu:
lgtm

java-backend/src/main/java/org/kframework/backend/java/kil/Collection.java
dwightguth, line 94: yes it's called by the default methods in KItemCollection.

dwightguth:
@radumereuta please review sections related to parsing.

bmmoore:
I made a pull request into your source branch with some further suggestions.
https://github.com/dwightguth/k/pull/11

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -264,7 +280,7 @@ private Module resolveBubbles(Module module) {
                 .parallel()
                 .filter(s -> s instanceof Bubble)
                 .map(b -> (Bubble) b)
-                .filter(b -> !b.sentenceType().equals("config"))
+                .filter(b -> b.sentenceType().equals("rule"))
```
radumereuta, line 288: How about contexts?

kernel/src/main/java/org/kframework/kore/compile/ConvertDataStructureToLookup.java
```diff
@@ -0,0 +1,372 @@
...
+    public synchronized Sentence convert(Sentence s) {
+        if (s instanceof Rule) {
+            return convert((Rule) s);
+        } else if (s instanceof Context) {
+            return convert((Context) s);
+        } else {
+            return s;
+        }
+    }
+}
```
radumereuta, line 456: This file feels weird. It is very big, and it's hard to follow structures in it.
If this can't be refactored into smaller pieces, then more comments are necessary.
Inline stuff, as well as an overall description in the beginning.

kernel/src/main/java/org/kframework/kompile/Kompile.java
dwightguth, line 288: handled below, see line 307

kernel/src/main/java/org/kframework/parser/concrete2kore/ParseInModule.java
```diff
@@ -37,9 +37,11 @@
  */
 public class ParseInModule implements Serializable {
     private final Module module;
+    private final boolean strict;
     private final Grammar grammar;
-    public ParseInModule(Module module) {
+    public ParseInModule(Module module, boolean strict) {
         this.module = module;
+        this.strict = strict;
```
radumereuta, line 44: Please also add a constructor which has only one argument and sets strict to true by default.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
```diff
@@ -0,0 +1,1008 @@
...
+            "           | InjectedKLabel of klabel\n" +
+            "           | Map of t m\n" +
+            "           | List of t list\n" +
+            "           | Set of s\n" +
+            "           | Int of Z.t\n" +
+            "           | String of string\n" +
+            "           | Bool of bool\n" +
+            "           | Bottom\n";
+
+    public static final String prelude = "module type S =\n" +
```
radumereuta, line 90: Make external files and load them into memory when you need them.
Do not merge like this.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
```diff
@@ -0,0 +1,1008 @@
...
+            } else {
+                encodeStringToIdentifier(sb, klabel);
+            }
+        }
+    }
+
+    private boolean isLookupKLabel(KApply k) {
+        return k.klabel().name().equals("#match") || k.klabel().name().equals("#mapChoice") || k.klabel().name().equals("#setChoice");
+    }
+}
```
radumereuta, line 1005: I have a feeling that this file can be split into multiple smaller steps which can be easier to manage.

kernel/src/main/java/org/kframework/kompile/Kompile.java
radumereuta, line 288: Ah, ok.

kernel/src/main/java/org/kframework/parser/concrete2kore/ParseInModule.java
dwightguth, line 44: I don't see the need, that literally never arises outside of unit tests.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
dwightguth, line 1005: So, one of the planned next steps is to set things up so that there is a representation of ocaml code inside Java. I am trying to leave that for our interns to work on if I can though because I have enough other stuff I can focus on instead right now. But it is going to get somewhat cleaner eventually.

kernel/src/main/java/org/kframework/parser/concrete2kore/ParseInModule.java
radumereuta, line 44: I will add it when I need it.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
dwightguth, line 90: I honestly don't know the best way to go about this. There are advantages to all three approaches (structured ASTs, file imports, string splicing). Overall I think what we're likely to do is some combination of all three, where part of the prelude is in its own .ml file, much of the tiny code snippets will be represented as spliced strings, and the structure of code that is primarily generated by an algorithm represented as a structured AST. That said, I am trying to focus on implementing as much of K as possible as fast as possible, and I don't always entirely know what shape things will take eventually.  Things might also change as I determine new optimizations that are needed as well. I would prefer to wait until I have an interpreter that can execute C before trying to refactor this whole thing.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
radumereuta, line 1005: Hmm... others agree?

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
radumereuta, line 90: I don't buy the argument of `implementing as much of K as possible as fast as possible`. We set up to rewrite K using KORE because we got into a state where the code was unmanageable. This approach will only lead to the same outcome.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
bmmoore, line 1005: I think it's reasonable to merge this request with the pass in one big file. The interns have been making a representation for O'Caml and other languages, and plan to change Dwight's code to use it.

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -136,6 +171,27 @@ case class Module(name: String, imports: Set[Module], localSentences: Set[Senten
 
   lazy val subsorts: POSet[Sort] = POSet(subsortRelations)
 
+  lazy val freshFunctionFor: Map[Sort, KLabel] =
```
cos, line 174: could this be moved outside `Module`?

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -136,6 +171,27 @@ case class Module(name: String, imports: Set[Module], localSentences: Set[Senten
...
+    productions.groupBy(_.sort).mapValues(_.filter(_.att.contains("freshGenerator")) )
+      .filter(_._2.nonEmpty).mapValues(_.map(p => p.klabel.get)).mapValues { set => {
+      if (set.size > 1)
+        throw KEMException.compilerError("Found more than one fresh generator for sort " + sortFor(set.head)
+          + ". Found: " + set)
+      else
+        set.head
+    }}
+
+  lazy val collectionFor: Map[KLabel, KLabel] = productions.filter(_.att.contains("assoc")).flatMap { p => {
```
cos: same; is this a KORE concept?

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -101,6 +124,18 @@ case class Module(name: String, imports: Set[Module], localSentences: Set[Senten
 
   val sortDeclarations: Set[SyntaxSort] = sentences.collect({ case s: SyntaxSort => s })
 
+  lazy val sortDeclarationsFor: Map[Sort, Set[SyntaxSort]] =
+    sortDeclarations
```
cos, line 128: there is some rather ugly reformatting of code in this file. did something change in IntelliJ's Scala plugin?

kore/src/main/scala/org/kframework/definition/outer.scala
dwightguth, line 174: I'm not sure I understand why this is a problem. it represents an index that is used by one of the core compilation passes of the K framework. I think this is the right place to put it, and since it is a lazy val, it causes no overhead whatsoever if it is not used. We could definitely move it outside, but I don't really see how that gains us anything given that it contains information relevant to the core K language.

kore/src/main/scala/org/kframework/definition/outer.scala
dwightguth, line 128: I'm not sure what happened, but you persuaded me not to care about formatting changes, so I didn't. If you want me to care again, you will have to care yourself in the future.

kore/src/main/scala/org/kframework/definition/outer.scala
dwightguth: Again, I'm not sure I see the harm here. Adding a new class just for this doesn't really seem warranted to me given that it is used by one of the core compiler passes for backends that intend to perform rapid evaluation of matching involving collections.

bmmoore:
GenerateSortPredicateRules/Syntax still don't seem to entirely ignoring user-declared parsing only sorts, but they are at least closer to depending entirely on queries that might be adjusted in the module.

dwightguth:
GenerateSortPredicateRules now generates exactly the syntax present in the module that it receives. In the normal course of events, this should contain no parsing sorts or productions, so it should be fine.

bmmoore:
I suppose it's fine now - I see there are no more calls to isParserSort, so at least there's no danger of assumptions that this pass is responsible for excluding parser-only sorts.

lgtm.

kore/src/main/scala/org/kframework/definition/outer.scala
cos, line 174: could it be moved into the compilation phase in this case?

kore/src/main/scala/org/kframework/definition/outer.scala
cos: it could be moved in the relevant compilation phase then.

kore/src/main/scala/org/kframework/definition/outer.scala
cos, line 128: if we're all using the same auto-formatting settings (the ones you committed) then we may want to  be careful if something is broken with them. is this auto-format?

kore/src/main/scala/org/kframework/definition/outer.scala
dwightguth, line 174: I'm sorry, I misspoke, this is actually used by all backends in order to implement the fresh variable hook. This seems like a reasonable place to access that information to me.

kore/src/main/scala/org/kframework/definition/outer.scala
dwightguth, line 128: the only autoformatting in this file is what was done to  importedModules and sentences, and there's really not much that can be done to control the autoformatting of blocks such that it can sensibly decide whether to add a newline or not.

dwightguth:
@cos @radumereuta please review

Note that I am probably not going to engage in a large-scale refactoring of the OCAML code at this point. I intend to wait until we have a better representation of OCAML in Java first, and that is work that the interns are doing.

radumereuta:
lgtm, but I think this will conflict in some areas with the changes I made in my pull request.
Let me know if you decide to merge this one first.

kore/src/main/scala/org/kframework/definition/outer.scala
cos, line 174: Backends can choose to use a different way to implement freshness -- I don't think we consider this approach fundamental to K. I see how this functionality can live in `Module` but, on the principle of keeping the KORE classes as light as possible, I would prefer it is moved someplace else. In any case, not worth the energy to discuss it too much. If you don't want to move it, you can merge the PL as it is.

dwightguth:
@cos @radumereuta @andreistefanescu @bmmoore any last comments before I merge?

radumereuta:
lgtm

cos:
lgtm

PR #1530, opened 2015-06-09
improved isNormal and isGround computations

andreistefanescu:
do not review

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/kil/JavaSymbolicObject.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/kil/JavaSymbolicObject.java
@@ -129,10 +129,10 @@ public abstract class JavaSymbolicObject extends ASTNode
      */
     public boolean isGround() {
         if (isGround == null) {
-            if (variableSet == null || !variableSet.isEmpty()) {
+            if (variableSet == null) {
                 new IsGroundFieldInitializer().visitNode(this);
             } else {
-                isGround = true;
+                isGround = variableSet.isEmpty();
             }
         }
         return isGround;
--- a/java-backend/src/main/java/org/kframework/backend/java/kil/JavaSymbolicObject.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/kil/JavaSymbolicObject.java
@@ -20,6 +20,7 @@ import java.util.IdentityHashMap;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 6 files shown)


andreistefanescu:
@dwightguth please review

dwightguth:
looks good but there's a couple minor improvements I'd like to see with respect to the isGround and variableSet fields. Namely, if isGround is true, we should be able to deduce that variableSet is empty without traversing the entire term, and similarly, if we have already computed the variableSet, we should be able to determine  isGround without traversing the term by checking the size of the set.

Also, do you have any statistics on the effect this diff has on performance? I'm assuming you did it because it was faster but I'd be interested to see numbers before we merge this.

andreistefanescu:
The old version is about 50% slower. This does not mean much for concrete execution, but it's useful for verification.

java-backend/src/main/java/org/kframework/backend/java/kil/JavaSymbolicObject.java
```diff
@@ -113,39 +110,51 @@ public JavaSymbolicObject substitute(Variable variable, Term term, TermContext c
...
+        return variableSet;
     }
 
-     /**
+    /**
+     * Returns {@code true} if this JavaSymbolicObject does not contain any variables.
+     */
+    public boolean isGround() {
+        if (isGround == null) {
+            if (variableSet == null || !variableSet.isEmpty()) {
```
dwightguth: `if (variableSet == null) {`

java-backend/src/main/java/org/kframework/backend/java/kil/JavaSymbolicObject.java
```diff
@@ -113,39 +110,51 @@ public JavaSymbolicObject substitute(Variable variable, Term term, TermContext c
...
+                new IsGroundFieldInitializer().visitNode(this);
+            } else {
+                isGround = true;
```
dwightguth: `isGround = variableSet.isEmpty();`

andreistefanescu:
On a side note, `hashCode` is a lot faster than `isGround`, so all the boilerplate code is slowing things down. Unfortunately, it's not feasible to replace all the visitor/transformer classes with handcrafted code.

dwightguth:
still one minor fix to make, then you can merge.

dwightguth:
lgtm. merge if tests pass.

PR #1531, opened 2015-06-09
Fast unification3

andreistefanescu:
do not review


andreistefanescu:
@dwightguth please review

dwightguth:
what is the effect of this diff on execution and verification performance?

andreistefanescu:
The effect on performance is minimal. This is just refactoring the common code in pattern matching and unification. I moved from double dispatch to if-then-else to mirror Yilong's work in `NonACPatternMatcher`, but I'm not seeing any noticeable improvements in performance. Next steps are to eliminate `NonACPatternMatcher` and to add an unification based `KAbstractRewriteMachine`.

dwightguth:
looks great!

PR #1533, opened 2015-06-11
removed NonACPatternMatcher from use

andreistefanescu:
do not review


andreistefanescu:
@dwightguth please review

dwightguth:
Can you run the code before and after this diff on one of the larger examples in the tutorial and tell me the effect on the performance of the pattern matching rewrite engine?

andreistefanescu:
This slightly improves performance: collaz (n=500) went from 1.45m to 1.25m.

grosu:
Which language are you talking about?  IMP, SIMPLE, KOOL, C, Java?  Please say one of the latter :)  The performance seems low though, even for C.  How many rewrites were applied?  nailgun?  Thanks, Grigore.

andreistefanescu:
This numbers are for kool-typed-dynamic. It takes 1,692,296 steps.

grosu:
So about 20,000 steps/s, right?  How about running it with Nailgun enabled?

andreistefanescu:
About 23,000 steps/s. I did not get a chance to run it with nailgun yet, but I don't think it would improve by very much.

PR #1534, opened 2015-06-12
fast rewriting optimizations applied to unification/narrowing

andreistefanescu:
do not review yet


andreistefanescu:
@dwightguth please review

dwightguth:
something feels funny to me here... I don't see how we can unify this diff with the version of the rewriter that rewrites over KORE, and without that common element, I'm worried that we're tying ourselves into a corner with respect to the KORE transition. I'm not necessarily opposed to proceeding this way if we need to have results to show, but I'm worried we're adding technical debt here that could give us trouble later. Do you have an estimate of how much effort it would be for us to implement a tree-matching automaton that operates over not just cells but any K? There are algorithms out there like the OCAML one, so I feel like it may not actually be that much work, but I guess I don't really know what all might become involved.

I will look into the more technical parts of this diff tomorrow, but it's something we need to think about, especially when we need at some point to plan how to get multiplicity \* cells working in the rewriter for KORE. In the meantime, can you also please try to tell me what the effect this diff has on performance of verification is?

grosu:
I haven't looked over the code, but the language in @dwightguth 's comment above concerns me.  There should be no "rewriter that rewrites over KORE".  (Are you referring to @cos's tiny rewriter?)  The point is that all rewriters should implement one interface, the "K rewriter" interface, and that should only manipulate KORE terms anyway.  The debugger and any other clients of a rewrite engine should only interact with the rewrite engine by means of that interface.  Switching from one rewriter to another should be as simple as creating one instance or another of a class that implements the (fixed and well-documented) K rewriter interface.  If a particular rewriter decides to implement fancy indexing or matching-automata techniques, then that should be its own internal business in order to achieve what it wants (e.g., speed), but not something that needs to be exported above the K rewriter interface.

dwightguth:
@grosu , the issue is that this diff operates over a class CellCollection which the rewriter uses to represent the old cells that aren't of sort K. By "rewriter over KORE" I mean the version of the rewriter that accepts KORE terms as input instead of KIL terms. That code does not currently support multiplicity \* cells, and my interest is in determining how we can do so in a fashion consistent with this diff. I'm just trying to figure out how the optimizations in this diff can be unified with the differences in the structure of our term representation that is involved in the KORE transition.

andreistefanescu:
@dwightguth this diff gets collatz for 500 with `--symbolic-execution` to about 2.5m, down from more than 10m. The effect on verification performance depends on how many semantics steps vs smt calls. For `kernelc` this is maybe 20-25%. 

@dwightguth @grosu in the short term, the backend could continue to maintain a dedicated class for cells, even with the new KORE. In the long term, we should probably implement some pattern match automaton to get better performance, and that would apply to all `K` terms, not only cells.

andreistefanescu:
@dwightguth @grosu this patch makes the c verification 2-3 times faster.

grosu:
OK, great numbers and thanks for the clarifications, @andreistefanescu and @dwightguth !  I agree that we will probably want the rewriter to know which KLabels correspond to cells anyway.  In fact, from what I remember from discussions with @yilongli , the rewriter may even be faster if we do _not_ apply the configuration concretization step at kompile time, but instead modify the rewriter to work directly with K rules as they are.  The reason is that, intuitively, the indexing sort of undoes the cell completion.  The K rules already tell the rewriter, in a minimal way, what it should look for when it does the matching.

dwightguth:
lgtm. If @grosu is satisfied with the changes you made to the tutorial, you can merge this.

k-distribution/tutorial/2_languages/4_logik/basic/tests/reverse-fast-3.logik.out
```diff
@@ -2,6 +2,6 @@ Search results:
 
 Solution 1:
 Solution:Map -->
-L:Term |-> cons ( c, cons ( _0:Term, cons ( a, nil ) ) )
-T:Term |-> cons ( a, cons ( _0, cons ( c, nil ) ) )
-X:Term |-> _0
+L:Term |-> cons ( c, cons ( X:Term, cons ( a, nil ) ) )
+T:Term |-> cons ( a, cons ( X, cons ( c, nil ) ) )
+X:Term |-> X
```
grosu, line 7: Not a big deal, but I thought that we do not want to display mappings of the form `X |-> X`.

k-distribution/tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.k
```diff
@@ -771,7 +769,7 @@ the \textsf{tasks} cell has been removed. */
 
 // cleanup tasks and output a success message when done
 
-  rule (<tasks> .Bag </tasks> => .)
+  rule (<T>... <tasks> .Bag </tasks> ...</T> => .)
```
grosu, line 772: Why did you do that?  Was it cleaner to get rid of the entire T cell in your view, or you did it in order to avoid some other problems?

k-distribution/tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.k
```diff
@@ -556,15 +556,13 @@ instance of this rule). */
...
 //@ \paragraph{Check for unique class names}
 
-// The rule below is also more complicated than needed because of the
-// configuration abstraction which does not seem to work right.  We
-// should not need to mention the classes cell.
-  rule <T> (<classes>... <className> C </className>
-                         <className> C </className>
-            ...</classes> _ => .)
-           <out>... .List => ListItem("Class \"" +String Id2String(C)
+  rule (<T>...
```
grosu, line 559: Nice that this worked!

k-distribution/tutorial/1_k/5_types/lesson_6/tests/polymorphic-identity-1-fails.lambda.out
```diff
@@ -2,8 +2,12 @@ Search results:
...
 Solution 1:
 V:K -->
-int
+(int -> bool) -> bool = (_0:Type -> _0) -> _1:Type
 
 Solution 2:
 V:K -->
-int -> int = bool -> bool
+_0:Type
+
```
grosu, line 10: I don't understand Solution 2 here.  What does it mean for V to be bound to a variable?

k-distribution/tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.k
andreistefanescu, line 772: This original rule is incorrect because it deletes `tasks` despite the fact that it does not have `multiplicity=?`. The other rules I changed have the same problem because they delete `classes`. I decided to mark `T` with `multiplicity=?` and delete it uniformly in all rules.

grosu, line 772: OK, nice.  Makes sense, thanks.

k-distribution/tutorial/1_k/5_types/lesson_6/tests/polymorphic-identity-1-fails.lambda.out
andreistefanescu, line 10: `V` is bound to whatever is on the `k` cells. Here there are 3 `k` cells. My guess it that `_0` would eventually become the type of the program, if there were no typing issues.

grosu, line 10: Interesting.  It could be indeed that a pattern is matched many different ways on the same configuration.  Are we reporting these different matches as different solutions?  I'm assuming the three solutions in this case are due to three different paths, and not to the three different matches, right?  If that is the case, I would merge this code but I would submit a new issue that we should distinguish different matches of the same pattern as different solutions.

PR #1532, opened 2015-06-10
New Debugger Beginnings

msaxena2:
Logic for beginning of the new Kore based debugger. 

@dwightguth Please review. This PR also has commits from your branch (https://github.com/dwightguth/k/tree/ocaml8).

PR #1536, opened 2015-06-17
Kore debugger

msaxena2:
The code for the debugger API, and the KORE based implementation. Carries over code from #1532 .

@grosu @dwightguth This pull request is not yet ready for merging. Please feel free to review. More commits to be added.


kernel/src/main/java/org/kframework/debugger/Checkpoint.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Class representing a Checkpoint in the State.
+ *
+ * Recall that the state represents a branch in the execution tree.
+ * This class stores some intermediate K's encountered in that branch.
+ */
+public class Checkpoint {
+    private K checkpointK;
+
+    private int checkpointID;
```
dwightguth: I don't think you need this field.

kernel/src/main/java/org/kframework/krun/modes/DebugExecutionMode.java
```diff
@@ -0,0 +1,166 @@
...
+
+import static org.kframework.Collections.*;
+import static org.kframework.definition.Constructors.*;
+
+/**
+ * Created by Manasvi on 6/10/15.
+ * <p>
+ * Execution mode class for the Kore based Debugger
+ */
+public class DebugExecutionMode implements ExecutionMode<Void> {
```
msaxena2, line 33: Don't pay attention to this class. This code is going to be removed. It's there to help me develop the API for now.

kernel/src/main/java/org/kframework/debugger/KDebug.java
```diff
@@ -0,0 +1,32 @@
...
+ * Debugger Api. Any Interaction to the Debugger
+ * happens through the methods described in this
+ * interface.
+ */
+public interface KDebug {
+
+    /**
+     * Take a certain number of steps from the current state.
+     */
+    public KDebugOpResult step(int steps);
```
dwightguth: public State step(int steps, State from);

kernel/src/main/java/org/kframework/debugger/KDebug.java
```diff
@@ -0,0 +1,32 @@
...
+     * Global Setting, will persist across states.
+     */
+    public void setCheckpointInterval(int interval);
+
+    /**
+     * Go back from a given number of steps to a predefined position.
+     *
+     * You can only select a state that you've already encountered
+     */
+    public KDebugOpResult select(int stateNum);
```
dwightguth: public List<State> currentStates();

kernel/src/main/java/org/kframework/debugger/KDebugOpResult.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.NavigableMap;
+import java.util.Optional;
+
+/**
+ * Created by Manasvi on 6/16/15.
+ * <p>
+ * The result of using a debugger operation
+ */
+public class KDebugOpResult {
```
dwightguth: public class State

kernel/src/main/java/org/kframework/debugger/KDebugOpResult.java
```diff
@@ -0,0 +1,32 @@
...
+
+    private Optional<K> finalK;
```
dwightguth: why is this optional?

kernel/src/main/java/org/kframework/debugger/KDebugOpResult.java
```diff
@@ -0,0 +1,32 @@
...
+
+    private Optional<NavigableMap<Integer, Checkpoint>> checkpointMap;
```
dwightguth: why is this optional?

kernel/src/main/java/org/kframework/debugger/KDebugState.java
```diff
@@ -0,0 +1,65 @@
...
+ * The Debugger can have multiple states at the same time,
+ * but only one state is active.
+ *
+ * Every State has a Checkpoint Enabled History.
+ *
+ * A State essentially represents a specific branch in the
+ * execution tree of a program.
+ *
+ */
+public class KDebugState {
```
dwightguth: don't need this class.

kernel/src/main/java/org/kframework/debugger/KoreKDebug.java
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Kore Based Debugger Implementation.
+ */
+public class KoreKDebug implements KDebug {
+
+    private final int DEFAULT_ID = 0;
+    private final int DEFAULT_CHECKPOINT_SIZE = 50;
+
+    private Set<KDebugState> stateSet;
```
dwightguth: List probably so each element will have an index.

kernel/src/main/java/org/kframework/debugger/KoreKDebug.java
```diff
@@ -0,0 +1,89 @@
...
+    private KDebugState activeState;
```
dwightguth: don't need this field.

kernel/src/main/java/org/kframework/debugger/KDebugOpResult.java
```diff
@@ -0,0 +1,32 @@
...
+    public KDebugOpResult(K finalK, NavigableMap<Integer, Checkpoint> checkpointMap) {
+        this.finalK = Optional.ofNullable(finalK);
+        this.checkpointMap = Optional.ofNullable(checkpointMap);
+    }
+
+    public Optional<K> getFinalK() {
+        return finalK;
+    }
+
+    public Optional<NavigableMap<Integer, Checkpoint>> getCheckpointMap() {
```
dwightguth: should be package-private

kernel/src/main/java/org/kframework/debugger/KDebug.java
msaxena2: I thought there's always going to be an active State, which will be used to step from. And I don't think we should return the State itself. The object that I'm returning will in the future be used to record more info, which is irrelevant for the state. For e.g., if we want to have a log of what happened in the last operation. Plus, there can be operations that won't change the state, but will still be valid operations. This information can be easily returned using the class that I created. I think that it's better to keep it this way, and let state be something internal to the debugger. 

There can also be invalid operations that won't change the state as well. If we return the state, the state will have to be analyzed for the operation. Whereas in this case, we can have a field for depicting success/failure.

kernel/src/main/java/org/kframework/debugger/KoreKDebug.java
```diff
@@ -0,0 +1,89 @@
...
+     *
+     * @param initialK The initial Configuration.
+     * @param rewriter The Rewriter being used.
+     */
+    public KoreKDebug(K initialK, Rewriter rewriter) {
+        this.stateSet = new HashSet<>();
+        this.checkpointInterval = DEFAULT_CHECKPOINT_SIZE;
+        this.rewriter = rewriter;
+        KDebugState initialState = new KDebugState(initialK);
+        initialState.addCheckpoint(new Checkpoint(initialK, DEFAULT_ID), DEFAULT_ID);
```
dwightguth: I think instead of having a separate Checkpoint class what I would suggest is to have a State class that has a RewriterState class and a NavigableMap, and the RewriterState class would have the same fields as the Checkpoint, ie, K, and eventually Rule/substitution/etc.

kernel/src/main/java/org/kframework/debugger/KoreKDebug.java
```diff
@@ -0,0 +1,89 @@
...
+    public KDebugOpResult step(int steps) {
+        K currentK = activeState.getCurrentK();
+        int activeStateCheckpoint = activeState.getActiveStateId();
+        int lastCheckpoint = activeState.getlastMapCheckpoint();
+
+        /* Not enough steps for a new checkpoint */
+        if (activeStateCheckpoint + steps < activeState.getlastMapCheckpoint() + checkpointInterval) {
+            currentK = rewriter.execute(currentK, Optional.of(new Integer(steps)));
+            activeStateCheckpoint += steps;
+            activeState.setActiveStateId(activeStateCheckpoint);
```
dwightguth: I think it's better to keep the State class immutable for now... We can adjust that later if it becomes a serious performance issue.

kernel/src/main/java/org/kframework/krun/modes/DebugExecutionMode.java
```diff
@@ -0,0 +1,166 @@
...
+
+    }
+
+    private static String unparseTerm(K input, Module test) {
+        return KOREToTreeNodes.toString(
+                new AddBrackets(test).addBrackets((ProductionReference)
+                        KOREToTreeNodes.apply(KOREToTreeNodes.up(input), test)));
+    }
+
+    public static void prettyPrint(CompiledDefinition compiledDef, K result) {
```
dwightguth: don't duplicate this. Instead, call KRun.prettyPrint.

kernel/src/main/java/org/kframework/debugger/KoreKDebug.java
msaxena2: Here's my justification for setting it up this way. I think it's simple and works fine. 
State - The list of States the debugger is in. There are multiple States, but the debugger is in a single state at a given time, which is the active State. All operations happen using the active State. Switching a state, thus, completely switches context. 

Checkpoint - Info about a checkpoint. Has the previous rule and all that stuff. Can add more stuff into it later. 

KDebugOpResult - The result of a call to the API. It has optional fields since not all operations require all fields. See one of the comments above for justification.

kernel/src/main/java/org/kframework/debugger/KDebug.java
dwightguth: Do you have an example of something that the step command would return which would not be stored in the State field?

kernel/src/main/java/org/kframework/krun/modes/DebugExecutionMode.java
msaxena2: As I said, not too concerned about this class at the moment, so won't be changing it. This code will go soon. It's only to help me see what's going on.

kernel/src/main/java/org/kframework/debugger/KoreKDebug.java
dwightguth: all operations should not return the same type. Rather, each operation should return only the information specific to that operation. The purpose of a checkpoint is to be able to reconstruct everything needed in order to create a new State. Thus I call it RewriterResult because it represents both the result of the current operation, and the result of previous operations at checkpoints. A State would not expose its RewriterResult (RewriterResult would be a package-private class), but it would be used internally for all the fields of the State except the checkpoint map.

kernel/src/main/java/org/kframework/debugger/KDebug.java
msaxena2: If I recall correctly, a step-all command would result in multiple states. You may need to specify the list of states, and the transitions that happened, and the final active state. I think it's easier to store that info in the result class. I guess we can have it return a list of states, if you'd rather want it that way, but just feel that it would be useful in the future, for example like storing a log or something like that. In any case, it's not too important.

dwightguth: yes, so:

```
Set<State> stepAll(int steps, State from);
State {
  Rule lastAppliedRule;
  K finalK;
  Map<KVariable, K> substitution;
}
```

```
```

kernel/src/main/java/org/kframework/debugger/KoreKDebug.java
msaxena2: So, you want renaming the checkpoint class RewriterState and making it package-private?

kernel/src/main/java/org/kframework/debugger/KDebug.java
msaxena2: Yeah, I guess this could work. Will Change it.

kernel/src/main/java/org/kframework/debugger/KoreKDebug.java
dwightguth: yes plus the two fields of State should be a RewriterState and a checkpoint map. The RewriterState would not be exposed by the getters (which would only expose the fields inside the RewriterState), but there would be a package-private constructor that takes a RewriterState. And then the checkpoint map would be of type `NavigableMap<Integer, RewriterState>`

kernel/src/main/java/org/kframework/debugger/DebuggerState.java
```diff
@@ -0,0 +1,59 @@
...
+ * A State essentially represents a specific branch in the
+ * execution tree of a program.
+ */
+public class DebuggerState {
+
+    private NavigableMap<Integer, RewriterCheckpoint> checkpointMap;
+
+    private K currentK;
+
+    private int activeStateId;
```
dwightguth: since this seems to represent a step number,  can we just call it step or stepNumber?

msaxena2:
Jenkins: test this please

k-distribution/src/test/java/org/kframework/debugger/TstKDebugOnKORE_IT.java
```diff
@@ -0,0 +1,88 @@
...
+        K expectedResult = utils.stepRewrite(parsed, Optional.ofNullable(new Integer(steps)));
+        assertEquals("Normal and Debug results don't match, when both allowed to run for some random steps", expectedResult, debugResult);
+    }
+    @Test
+    public void jumpBackTest() {
+        /* Going Back on Debugger */
+        int forward = getRandomSteps(20, 150);
+        int backward = getRandomSteps(1, forward);
+        debuggerSession.step(forward);
+        K debugResult = debuggerSession.backStep(backward).getCurrentK();
```
dwightguth: I think you need to apply the KtoKORE here too otherwise you might end up with a problem if the tests apply in the other order.

kernel/src/main/java/org/kframework/krun/modes/DebugExecutionMode.java
```diff
@@ -0,0 +1,171 @@
+// Copyright (c) 2015 K Team. All Rights Reserved.
```
dwightguth, line 1: I believe we weren't going to include this file in the diff?

dwightguth:
lgtm, one minor thing to do, plus gut the internal of DebugExecutionMode so that it's not cluttered with temporary code, and then you can merge it if the tests pass.

dwightguth:
lgtm. You can merge this if you are done working on it.

PR #1537, opened 2015-06-17
Adding timing messages for verbosity level.

radumereuta:
@dwightguth please review. Here is a sample output:

```
Parse command line options                                   =    18
Importing: D:\work\test\test.k
Importing: D:\work\k\k-distribution\include\builtin\kast.k
Importing: D:\work\test\test.k
Importing: D:\work\k\k-distribution\include\builtin\domains.k
Parse definition [0/62 rules]                                =  1560
Apply compile pipeline                                       =   126
Save to disk                                                 =    64
Cleanup                                                      =     0
Total                                                        =  1768
```

### DIFF
--- a/k-distribution/src/test/java/org/kframework/Konsole.java
+++ b/k-distribution/src/test/java/org/kframework/Konsole.java
@@ -33,10 +33,9 @@ public class Konsole {
         String programModuleName = args[2];
 
         KExceptionManager kem = new KExceptionManager(new GlobalOptions());
-        Stopwatch sw = new Stopwatch(new GlobalOptions());
 
         CompiledDefinition compiledDef =
-                new Kompile(new KompileOptions(), FileUtil.testFileUtil(), kem, sw, false).run(new File(definitionFilename), mainModuleName, programModuleName, Sorts.K());
+                new Kompile(new KompileOptions(), FileUtil.testFileUtil(), kem, false).run(new File(definitionFilename), mainModuleName, programModuleName, Sorts.K());
 
         Module module = compiledDef.executionModule();
         BiFunction<String, Source, K> programParser = compiledDef.getProgramParser(kem);
--- a/k-distribution/src/test/java/org/kframework/Konsole.java
+++ b/k-distribution/src/test/java/org/kframework/Konsole.java
@@ -10,6 +10,7 @@ import org.kframework.kompile.KompileOptions;
 import org.kframework.kore.K;
 import org.kframework.main.GlobalOptions;
 import org.kframework.tiny.Rewriter;
+import org.kframework.utils.Stopwatch;
 import org.kframework.utils.errorsystem.KExceptionManager;
 import org.kframework.utils.file.FileUtil;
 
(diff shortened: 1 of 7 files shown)


dwightguth:
Instead of modifying every call site to add an object that is only used by --verbose, even though --verbose isn't being specified, can you add a constructor to Kompile that doesn't require that argument?

radumereuta:
Is this what you had in mind?

kernel/src/main/java/org/kframework/utils/errorsystem/KExceptionManager.java
```diff
@@ -24,6 +24,10 @@
 public class KExceptionManager {
     private final List<KException> exceptions = Collections.synchronizedList(new ArrayList<>());
 
+    public GlobalOptions getOptions() {
```
dwightguth: don'

dwightguth: This is a bad way of accessing GlobalOptions. Get it from KompileOptions.global

k-distribution/src/test/java/org/kframework/Konsole.java
```diff
@@ -10,6 +10,7 @@
 import org.kframework.kore.K;
 import org.kframework.main.GlobalOptions;
 import org.kframework.tiny.Rewriter;
+import org.kframework.utils.Stopwatch;
```
dwightguth: revert these three files

radumereuta:
Did I get it right this time?

radumereuta:
Jenkins retest this please

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -81,19 +82,34 @@
...
+        this.kem = kem;
+        this.kompileOptions = kompileOptions;
+        this.parser = new ParserUtils(files, kem, kompileOptions.global);
+        this.cacheParses = cacheParses;
+        this.loader = new BinaryLoader(kem);
+        this.sw = sw;
+    }
 
     public Kompile(KompileOptions kompileOptions, FileUtil files, KExceptionManager kem, boolean cacheParses) {
         this.files = files;
```
dwightguth: can you make this constructor this() the one above?

kernel/src/main/java/org/kframework/utils/errorsystem/KExceptionManager.java
```diff
@@ -101,7 +101,17 @@ public void addKException(KException kex) {
     }
 
     public void addAllKException(Collection<KException> kex) {
-        exceptions.addAll(kex);
+        if (this.options.warnings == GlobalOptions.Warnings.NONE)
```
dwightguth: take a look at how registerInternal works. You shouldn't need to have any kind of if/else clause, just do a filter on the method that that method calls.

radumereuta:
Tests passed, and I think I've addressed all the comments.

PR #1550, opened 2015-06-20
fix set matching

dwightguth:


dwightguth:
This seems to take the time for KOOL/collatz 500 from 1.6 minutes to 2 minutes. I still wonder whether only calling hashCode and equals on some classes isn't the way to go... It's less nice but optimizations are often less nice.

andreistefanescu:
@dwightguth looks good, thanks!

dwightguth:
I will merge this. Will create another diff if it doesn't completely fix the C semantics.

PR #1552, opened 2015-06-23
more improvements to ocaml backend

dwightguth:
@andreistefanescu @cos @bmmoore @radumereuta please review


kore/src/main/scala/org/kframework/kore/interface.scala
```diff
@@ -17,6 +19,8 @@ trait K extends Serializable {
   override def hashCode = cachedHashCode
 
   def computeHashCode: Int
+
+  def attEquals(other: Any, @varargs attNames: Array[String]): Boolean
```
cos: I would prefer that `attEquals` doesn't become part of the KORE classes interface. It could live very well as a static method of a class, e.g., `AttCompare.equals(a,b)`. The KORE interface is small enough and it should be stable enough that polymorphism/a visitor pattern is not needed. A simple recursive function should be enough.

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -65,8 +71,33 @@ public int run(CompiledDefinition compiledDef, KRunOptions options, Function<Mod
...
         if (result instanceof K) {
             prettyPrint(compiledDef, options.output, s -> outputFile(s, options), (K) result);
-        }
 
+            if (options.exitCodePattern != null) {
+                Rule exitCodePattern = pattern(options.exitCodePattern, options, compiledDef, Source.apply("<command line: --exit-code>"));
+                List<Map<KVariable, K>> res = rewriter.match((K) result, exitCodePattern);
+                if (res.size() != 1) {
+                    kem.registerCriticalWarning("Found " + res.size() + " solutions to exit code pattern. Returning 112.");
+                    return 112;
```
radumereuta, line 80: What are these codes? Is there a place where they are documented?

kernel/src/main/javacc/KoreParser.jj
```diff
@@ -0,0 +1,202 @@
+// Copyright (c) 2013-2015 K Team. All Rights Reserved.
```
radumereuta, line 1: Is this new? Can you describe the grammar in BNF format, in a comment at the beginning of this file?
It would make it much easier to read.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/DefinitionToOcaml.java
```diff
@@ -999,7 +1379,21 @@ public void apply(KLabel klabel) {
...
     private boolean isLookupKLabel(KApply k) {
         return k.klabel().name().equals("#match") || k.klabel().name().equals("#mapChoice") || k.klabel().name().equals("#setChoice");
     }
+
+    private boolean isList(K item, boolean klist, boolean rhs, VarInfo vars) {
+        return !klist && ((item instanceof KVariable && getSortOfVar((KVariable)item, vars).equals("K")) || item instanceof KSequence
+                || (item instanceof KApply && (functions.contains(((KApply) item).klabel()) || ((anywhereKLabels.contains(((KApply) item).klabel()) || ((KApply) item).klabel() instanceof KVariable) && rhs))))
+                && !(!rhs && item instanceof KApply && collectionFor.containsKey(((KApply) item).klabel()));
+    }
+
```
radumereuta, line 1398: I just want to reiterate that this file looks horrible. (you don't have to fix it now)

ocaml-backend/src/main/java/org/kframework/backend/ocaml/OcamlRewriter.java
```diff
@@ -52,44 +51,121 @@ public Rewriter apply(Module module) {
...
+            Thread err = new Thread(() -> {
+                int count;
+                byte[] buffer = new byte[8192];
+                try {
+                    while ((count = p2.getErrorStream().read(buffer)) >= 0) {
+                        System.err.write(buffer, 0, count);
+                    }
+                } catch (IOException e) {}
+            });
+            in.start();;
```
radumereuta: double semicolon

radumereuta:
Done reviewing for now. Mostly, I want more comments to explain some of the things you added.

andreistefanescu:
@dwightguth this is a large diff; I have scrolled through it, but if you want me to look at anything in particular, please specify.

Overall it looks ok, but I think the huge strings and maps should be moved in prelude.ml/properties files/... For now it's ok, but in the long term I want those refactored.

kernel/src/main/java/org/kframework/krun/KRun.java
dwightguth, line 80: it's basically just randomly chosen. Unfortunately we don't really have any way of using an exit code that we can guarantee doesn't overlap with the one returned by the program being executed. Exit codes don't really have any consistent across-application meaning aside from zero/nonzero anyway. We will put these exit codes in the manual when that time comes.

radumereuta, line 80: can you put a TODO marker here so we don't forget about it? And any other place where we have such things.

dwightguth, line 80: it's not going to get forgotten. Part of writing the manual will consist of documenting the behavior of all the options to krun, of which this is one. We will naturally come to this behavior when documenting how this particular feature works.

bmmoore:
It looks okay to merge to me. The changes outside the OCaml backend are all fine, and though I didn't work carefully throught the new logic in DefinitionToOcaml it looks reasonable enough as long as it still works. Do check with Remy about trying to merge in his refactoring of the same code (to prepare for generating OCaml by way of a proper AST).

PR #1555, opened 2015-07-08
Kore Based Rewriter API's Search Functionality Related Code

msaxena2:
@grosu @dwightguth .Support for search in the Kore based Rewriter API. WIP, do not review yet.


tiny/src/main/scala/org/kframework/tiny/Rewriter.scala
```diff
@@ -35,6 +36,17 @@ object SimpleIndex extends (K => Option[String]) {
 class Rewriter(module: definition.Module, index: K => Option[String] = KIndex) extends org.kframework.Rewriter {
   def this(module: definition.Module) = this(module, KIndex)
 
+
+  /**
+   * Execute a search of the Transition System.
+   * @param initialConfig The State to begin searching from
```
grosu: Try to avoid swapping `Configuration` with `State`, because that can be quite confusing to others.  
I would personally always use the former.  Also, I would go for long, non-abbreviated names for the API-related names; `initialConfiguration` is just fine (instead of `initialConfig`).

ocaml-backend/src/main/java/org/kframework/backend/ocaml/OcamlRewriter.java
```diff
@@ -51,6 +55,11 @@ public Rewriter apply(Module module) {
         }
         return new Rewriter() {
             @Override
+            public List<? extends Map<? extends KVariable, ? extends K>> search(K initialConfig, Optional<Integer> depth, Optional<Integer> bound, Rule pattern) {
+                return null;
```
dwightguth: throw an UnsupportedOperationException here

dwightguth:
I merged my pull request so if you rebase your own on top of master, I will review this now.

k-distribution/src/test/resources/convertor-tests/cd
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) 2014-2015 K Team. All Rights Reserved.
```
dwightguth: this looks like a stray file?

ocaml-backend/src/main/java/org/kframework/backend/ocaml/OcamlRewriter.java
```diff
@@ -65,6 +70,23 @@ public K execute(K k, Optional<Integer> depth) {
                 String ocaml = converter.match(k, rule, files.resolveTemp("run.out").getAbsolutePath());
                 files.saveToTemp("match.ml", ocaml);
                 String output = compileAndExecOcaml("match.ml");
+                String[] lines = output.split("\n");
```
dwightguth: this looks like a merge gone wrong?

ocaml-backend/src/main/java/org/kframework/backend/ocaml/OcamlRewriter.java
```diff
@@ -53,6 +53,11 @@ public Rewriter apply(Module module) {
         }
         return new Rewriter() {
             @Override
+            public List<? extends Map<? extends KVariable, ? extends K>> search(K initialConfig, Optional<Integer> depth, Optional<Integer> bound, Rule pattern) {
+                return null;
```
dwightguth: throw a KEMException here

dwightguth:
Couple small things but overall it seems aright for now, although I note the search command isn't working yet. But that's not that critical for me. Can you merge your latest changes into this pull request for me to review though?

msaxena2:
Jenkins:test this please

msaxena2:
Jenkins: test this please

msaxena2:
Jenkins: test this please

msaxena2:
@dwightguth Please review.

kernel/src/main/java/org/kframework/debugger/KDebug.java
```diff
@@ -19,28 +28,98 @@
...
+     * specified pattern
+     *
+     * @param startStateId  StateId to begin searching for a state from. If not specified, the default state of the
+     *                      the current Debugger instance is used.
+     * @param searchPattern Pattern the use wishes to search for.
+     * @param depth         Max depth to explore in the model. Empty parameter enables unbounded depth.
+     * @param bounds        Max number of solutions to report. Empty parameter means unbounded.
+     * @return
+     */
+    public List<? extends Map<? extends K, ? extends K>> search(Optional<Integer> startStateId, Rule searchPattern, Optional<Integer> depth, Optional<Integer> bounds);
```
dwightguth: I think for consistency either all of these methods should take an Optional<Integer> or all of them should take an int for the state id.

kernel/src/main/java/org/kframework/debugger/KoreKDebug.java
```diff
@@ -48,79 +48,141 @@ public void setCheckpointInterval(int checkpointInterval) {
...
+    }
+
+    private DebuggerState getDebuggerState(int currentStateIndex, int activeStateCheckpoint, RewriterResult result, NavigableMap<Integer, K> checkpointMap) {
+        activeStateCheckpoint += result.rewriteSteps().get();
+        DebuggerState nextState = new DebuggerState(result.k(), activeStateCheckpoint, checkpointMap);
+        stateList.add(currentStateIndex, nextState);
+        return nextState;
+    }
+
+    private boolean isLeafState(int steps, RewriterResult result) {
```
dwightguth: can we call this isFinalState? that's a bit clearer.

kernel/src/main/java/org/kframework/debugger/DebuggerState.java
```diff
@@ -19,30 +21,31 @@
  */
 public class DebuggerState {
 
-    private NavigableMap<Integer, RewriterCheckpoint> checkpointMap;
+    private NavigableMap<Integer, K> checkpointMap;
```
dwightguth: can you make these fields final please?

kernel/src/main/java/org/kframework/debugger/KoreKDebug.java
```diff
@@ -48,79 +48,141 @@ public void setCheckpointInterval(int checkpointInterval) {
...
+        stateList.add(currentState);
+        if (configurationNum >= lastKey) {
+            return step(initialStateNum, configurationNum - lastKey);
+        }
+        return backStep(initialStateNum, lastKey - configurationNum);
+    }
+
+    @Override
+    public List<? extends Map<? extends K, ? extends K>> search(Optional<Integer> startStateId, Rule searchPattern, Optional<Integer> depth, Optional<Integer> bounds) {
+        if (startStateId.isPresent()) {
```
dwightguth: oh I see. I don't think we need this feature; it can just start from the current step number.

kernel/src/main/java/org/kframework/debugger/KoreKDebug.java
```diff
@@ -48,79 +48,141 @@ public void setCheckpointInterval(int checkpointInterval) {
...
+    }
+
+    @Override
+    public List<DebuggerState> getStates() {
+        return new ArrayList<>(stateList);
+    }
+
+
+    @Override
+    public DebuggerState setState(int stateNum, Optional<Integer> configurationNum) {
```
dwightguth: I don't really follow what these two fields are but this method should only take one parameter which is an integer index in stateList.

kernel/src/main/java/org/kframework/debugger/KoreKDebug.java
```diff
@@ -48,79 +48,141 @@ public void setCheckpointInterval(int checkpointInterval) {
...
+
+    @Override
+    public DebuggerState createCopy(int stateNum) {
+        DebuggerState newState = new DebuggerState(stateList.get(stateNum));
+        stateList.add(newState);
+        return newState;
+    }
+
+    @Override
+    public DebuggerState peek(Optional<Integer> stateNum, Optional<Integer> configurationNum) {
```
dwightguth: this method should not ever have to call the rewriter because it should only ever just be showing the current state.

kernel/src/main/java/org/kframework/kil/loader/Context.java
```diff
@@ -341,7 +341,7 @@ private void computeSubsortTransitiveClosure(Poset<Sort> subsorts) {
             throw KEMException.criticalError(msg);
         }
         subsorts.transitiveClosure();
-        // detect if lists are subsorted (Vals Ids < Exps)
+        // detect if lists are subsorted (Vals ids < Exps)
```
dwightguth: this looks like a mistake?

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -77,6 +77,7 @@ public int run(CompiledDefinition compiledDef, KRunOptions options, Function<Mod
             prettyPrint(compiledDef, options.output, s -> outputFile(s, options), (K) result);
 
             if (options.exitCodePattern != null) {
+
```
dwightguth: revert this file too?

kernel/src/main/java/org/kframework/krun/KRunDebuggerOptions.java
```diff
@@ -23,6 +23,9 @@
     public final CommandSetCheckpoint setCheckpoint = new CommandSetCheckpoint();
```
dwightguth, line 23: shouldn't this class get deleted now that we aren't using JCommander?

kernel/src/main/java/org/kframework/krun/modes/DebugMode/Commands.java
```diff
@@ -0,0 +1,129 @@
...
+
+        @Override
+        public void runCommand(KDebug session, CompiledDefinition compiledDefinition) {
+            int activeStateId = session.getActiveStateId();
+            DebuggerState prevState = session.getStates().get(activeStateId);
+            DebuggerState steppedState = session.step(activeStateId, stepCount);
+            int effectiveStepCount = steppedState.getStepNum() - prevState.getStepNum();
+            if (effectiveStepCount < stepCount) {
+                System.out.println("Attempted " + stepCount + " step(s). " + "Took " + effectiveStepCount + " steps(s).");
+                System.out.println("KDebug may have encountered a final configuration.");
```
dwightguth: may seems too weak. How about "Reached a final state."

kernel/src/main/java/org/kframework/krun/modes/DebugMode/Commands.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        @Override
+        public void runCommand(KDebug session, CompiledDefinition compiledDefinition) {
+            DebuggerState nextState = session.setState(stateNum.orElse(session.getActiveStateId()), configurationNum);
+            if (nextState == null) {
+                System.out.println("Error Selecting State. State unchanged");
+                return;
+            }
+            System.out.println("Selected State " + session.getActiveStateId());
```
dwightguth: this isn't really a state id. It would be better if we changed all the terminology in the code base so that we don't call it that, but at the very least we shouldn't expose this language to users. How about "Jumped to step 523"

kernel/src/main/java/org/kframework/krun/modes/DebugMode/DebugExecutionMode.java
```diff
@@ -0,0 +1,99 @@
...
+
+        while (true) {
+            try {
+                String input = reader.readLine("KDebug> ");
+                Command command = parseCommand(input);
+                if (command instanceof Commands.QuitCommand) {
+                    break;
+                }
+                command.runCommand(debugger, compiledDefinition);
+            } catch (ParseException parseException) {
```
dwightguth, line 91: you should also here catch a KEMException and print it.

kernel/src/main/java/org/kframework/krun/KRunDebuggerOptions.java
dwightguth, line 23: oh never mind I guess it's still used by the old KIL pipeline, but you should still revert the changes to the file since KORE isn't using them.

kernel/src/main/java/org/kframework/parser/concrete2kore/ParseInModule.java
```diff
@@ -39,7 +39,7 @@
     private final Module disambModule;
     /**
      * The exact module used for parsing. This can contain productions and sorts that are not
-     * necessarily representable in KORE (sorts like Ne#Ids, to avoid name collisions).
+     * necessarily representable in KORE (sorts like Ne#ids, to avoid name collisions).
```
dwightguth: same here. Did you do some kind of search/replace? You need to go through everything to make sure that you catch all the cases where you did stray changes like this one.

kernel/src/main/javacc/DebuggerCommandParser.jj
```diff
@@ -0,0 +1,92 @@
...
+}
+
+
+TOKEN: /*RESERVED TOKENS */
+{
+       <STEP: (["s", "S"])| "step" >
+    |  <BACKSTEP:  (["b", "B"])| "back-step">
+    |  <JUMPTO: (["j", "J"])| "jump-to">
+    |  <QUIT: "quit" | "abort" | "exit">
+    |  <CHECKPOINT: "set-chkpt">
```
dwightguth: I would call this just "checkpoint" or "ch"

kernel/src/main/javacc/DebuggerCommandParser.jj
```diff
@@ -0,0 +1,92 @@
...
+    |  <RESUME: "resume">
```
dwightguth: r

kernel/src/main/java/org/kframework/krun/modes/DebugMode/DebugExecutionMode.java
```diff
@@ -0,0 +1,99 @@
...
+                System.out.println(parseException.getMessage());
+            } catch (NumberFormatException numberException) {
+                System.out.println("Error Formatting Number");
```
dwightguth: I would go with something a bit more user friendly like "Could not parse "foo" as an integer"

kernel/src/main/java/org/kframework/krun/modes/DebugMode/DebugExecutionMode.java
```diff
@@ -0,0 +1,99 @@
...
+            } catch (IOException inputException) {
+                System.out.println("Error Interacting with Console");
```
dwightguth: I would throw a KEMException here with something like "Failed to read input from console"

kernel/src/main/java/org/kframework/krun/KRunDebuggerOptions.java
msaxena2, line 23: I had made minor changes to the old debugger as well, when the current one was dependent on this file. I'd rather not spend time on this file. It'll be deleted pretty soon anyway.

kernel/src/main/java/org/kframework/krun/modes/DebugMode/Commands.java
msaxena2: So what's an alternative?  We're already using 'step' to refer to the configuration the user arrives upon after taking a certain number of steps. The rationale behind this was that the user can save his execution path at any point, copy the "state", and change the copied state, without affecting the state he copied from. I can see that the word state might be misleading, but I can't think of any better alternatives.

msaxena2: There can be some confusion I guess. I'll change it to say something like jumped to state "x", configuration "y", when the state's supplied, and just jumped to configuration "y" when the state is not supplied. I can't think of a better alternative for the word state though. I'm willing to change it if you have something in mind.

dwightguth: I don't understand why you keep trying to have this idea of state/configuration. There are exactly 2 types of integers present in debugger commands. Step numbers and state numbers. A step number is either a number of steps to take, or the number of steps the configuration has taken since its initial configuration. A state number is the integer offset of one of the current states in the list of all current states. Why do we keep coming back to this issue of language?

msaxena2:
Jenkins: test this please

kernel/src/main/java/org/kframework/debugger/DebuggerState.java
```diff
@@ -19,30 +21,31 @@
  */
 public class DebuggerState {
 
-    private NavigableMap<Integer, RewriterCheckpoint> checkpointMap;
+    private final NavigableMap<Integer, K> checkpointMap;
```
dwightguth, line 24: it's good that you are creating a copy of the map in the constructor, but can you go one step further and make sure that the copy isn't altered by making all views into the map from outside this class be unmodifiable? Basically just wrap the getter with a Maps.unmodifiableNavigableMap call.

ocaml-backend/src/main/java/org/kframework/backend/ocaml/OcamlRewriter.java
```diff
@@ -56,11 +57,11 @@ public Rewriter apply(Module module) {
         }
         return new Rewriter() {
             @Override
-            public K execute(K k, Optional<Integer> depth) {
+            public RewriterResult execute(K k, Optional<Integer> depth) {
                 String ocaml = converter.execute(k, depth.orElse(-1), files.resolveTemp("run.out").getAbsolutePath());
                 files.saveToTemp("pgm.ml", ocaml);
                 String output = compileAndExecOcaml("pgm.ml");
-                return parseOcamlOutput(output);
+                return new RewriterResult(Optional.<Integer>empty(), parseOcamlOutput(output));
```
dwightguth, line 64: I fully intend to fix this so that it returns the number of steps taken, but until then, what behavior can I expect from running the debugger on the OCAML backend? will it crash? or will it work but just not be able to distinguish between final states and nonfinal states?

dwightguth:
looks okay. Feel free to merge now or wait until you have the patterns and the watches.

msaxena2:
Thanks Dwight, I'm merging this PR now. It gives a good indication of what the new debugger is going to be like, and also gets rid of the old debugger frontend, which was unnecessarily inconvenient to use. Merging this now will also help keep the size of the next diff reasonably small.

PR #1557, opened 2015-07-08
Cons list

radumereuta:
@dwightguth please review.
Sorry it took so long, I had to refactor a few things.

Hasn't been tested too much, I expect I will have to fix a few things, and add more comments.
I left out a few details from the K3 implementation (for example looking through rewrites).
I can add those cases back when we get into problems, since I don't really understand the current code, and I don't want to just take it as is.

### DIFF
--- a/k-distribution/include/builtin/kast.k
+++ b/k-distribution/include/builtin/kast.k
@@ -74,6 +74,7 @@ endmodule
 
 module RULE-CELLS
   imports KCELLS
+  imports RULE-LISTS
   // if this module is imported, the parser automatically
   // generates, for all productions that have the attribute 'cell' or 'maincell',
   // a production like below:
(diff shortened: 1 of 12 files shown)


dwightguth:
This is not what we agreed on. The idea was to generate a grammar that parses strings correctly.

radumereuta:
I'm not sure if that would work. What happens when you want to add the list terminator when you have a rewrite as a child of a list?
What we agreed, was that I would give you the correct seed module for pretty printing, and for parsing I can do whatever works.

dwightguth:
Well yes you can generate whatever grammar you want, but it has to /be/ a
grammar. The only way the unparsing algorithm works correctly is if the
grammar contains 100% of the information needed to generate the correct AST
for a string. That can't happen when you have things like this. As for
rewrites, the parser just has to be able to correctly type terms with
rewrites in them, which is a required feature for correctness anyway.

radumereuta:
Only one test failing. I tried to see why it fails, but I'm wasting too much time. I need help from someone who understand the backend better.

dwightguth:
you are correct on the first count. However, the second is correct as is because I am comparing the outer production against a cast, not the inner one.

dwightguth:
@radumereuta the tests seem to pass and when I try this patch with the C semantics on the OCAML backend, it seems to execute correctly. Are you okay to merge this? We will solve the unparsing issue later since it is somewhat more complicated than I originally thought.

radumereuta:
lgtm, but since we both commited here, we should request someone else to review it.

dwightguth:
Closing since this has been reviewed as part of #1558 and will be merged as such

PR #1558, opened 2015-07-10
Finalize OCAML backend - All C tests pass

dwightguth:
This is not ready for review yet but I am posting it so that @taktoa and @sebmathguy can get a look at the general shape of it. I will still be posting new bug fixes to this diff as needed, but I expect to be done by Monday because I think I will have all tests passing by then. Any remaining changes will likely also be minor because the OCAML backend actually works quite well and seems to have only minor glitches remaining.

If you want to jump the gun, @radumereuta can review the parser code, @andreistefanescu can review the changes I made to the hooks, and @cos and @bmmoore can review the changes to the KORE pipeline and to OCAML. Note that I haven't completely cleaned this up yet though so there may be some funkiness.

I suggest if you do review now that you review specifically the last commit because all of the others are part of the pending pr #1557.


dwightguth:
@radumereuta @andreistefanescu @cos @bmmoore please review.

k-distribution/include/builtin/domains.k
```diff
@@ -229,33 +229,33 @@ module INT
...
                > left:
-                 Int "xorInt" Int               [function, left, latex({#1}\mathrel{\oplus_{\scriptstyle\it Int}}{#2}), hook(#INT:_xorInt_)]
+                 Int "xorInt" Int               [function, left, latex({#1}\mathrel{\oplus_{\scriptstyle\it Int}}{#2}), hook(INT.xor)]
                > left:
-                 Int "|Int" Int                 [function, left, latex({#1}\mathrel{|_{\scriptstyle\it Int}}{#2}), hook(#INT:_|Int_)]
-  syntax Int ::= "minInt" "(" Int "," Int ")"   [function, smtlib(int_min), hook(#INT:minInt)]
-               | "maxInt" "(" Int "," Int ")"   [function, smtlib(int_max), hook(#INT:maxInt)]
-               | "absInt" "(" Int ")"           [function, smtlib(int_abs), klabel(absInt), hook(#INT:absInt)]
+                 Int "|Int" Int                 [function, left, latex({#1}\mathrel{|_{\scriptstyle\it Int}}{#2}), hook(INT.or)]
+  syntax Int ::= "minInt" "(" Int "," Int ")"   [function, smtlib(int_min), hook(INT.min)]
```
grosu, line 256: You know you can use the functional notation `syntax Int ::= minInt(Int,Int)`, right?

dwightguth, line 256: Yeah, this is just left over from a long time ago. I didn't bother changing it because this diff is just focused on the hooks,

k-distribution/include/builtin/domains.k
```diff
@@ -293,61 +293,61 @@ module FLOAT
...
+                 | expFloat(Float)              [function, hook(FLOAT.exp)]
+                 | logFloat(Float)              [function, hook(FLOAT.log)]
+                 | sinFloat(Float)              [function, hook(FLOAT.sin)]
+                 | cosFloat(Float)              [function, hook(FLOAT.cos)]
+                 | tanFloat(Float)              [function, hook(FLOAT.tan)]
+                 | asinFloat(Float)             [function, hook(FLOAT.asin)]
+                 | acosFloat(Float)             [function, hook(FLOAT.acos)]
+                 | atanFloat(Float)             [function, hook(FLOAT.atan)]
+                 | atan2Float(Float, Float)     [function, hook(FLOAT.atan2)]
+                 | maxFloat(Float, Float)       [function, smtlib(max), hook(FLOAT.max)]
```
grosu, line 333: I wonder, instead of having different hook attributes for smt and for execution, can't we have just one hook, say `hook(FLOAT.max)`, and then somewhere else have tables for each backend saying how they map the abstract hooks like `FLOAT.max` to their particular functionality?

k-distribution/include/builtin/kast.k
```diff
@@ -162,6 +164,13 @@ module PROGRAM-LISTS
   //      | Es#Terminator      // if the list is *
 endmodule
 
+module RULE-LISTS
```
grosu, line 167: Not sure I understand the comment below in this module.  This module seems to be empty.  I hope we don't do things based on module names.

k-distribution/include/io/io.k
```diff
@@ -10,27 +10,27 @@ module K-IO
 
   // B, N, and P are really supposed to be Nat
   syntax KItem ::= "#open" "(" String ")" [function]
-                 | "#open" "(" String "," String ")" [function, hook(#IO:#open), impure]
-                 | "#tell" "(" Int ")" [function, hook(#IO:#tell), impure]
-                 | "#getc" "(" Int ")"             [function, hook(#IO:#getc), impure]
-                 | "#read" "(" Int "," Int ")"    [function, hook(#IO:#read), impure]
+                 | "#open" "(" String "," String ")" [function, hook(IO.open), impure]
```
grosu, line 13: I didn't know about the `impure` attribute.  There is nothing in the manual about it, yet it is critical for verification.  Who added it?  Can that person add a sentence or two about it in the manual?

k-distribution/include/builtin/kast.k
radumereuta, line 167: Actually, we are doing things based on module names.
If I want to make a parser for rules, based on a specific module, I need to tell the parser to take that specific module, analyze it and generate all sorts of other productions. Instead of adding loads of options to a single generating function, I create a new empty module which imports the original module and a few other flag modules.

These flag modules tell the parser to generate: subsorts, casts, program lists, special productions for configurations/rules. This is also easy to extend in the future, since we only add code in a single place.

But the main advantage of doing this, is that it can be used directly in K with module algebra:
`parse(TEST + LATTICE + CASTS + RULE-CELLS, InputString, StartSymbol)`
This looks ugly so we can make it look a bit nicer with a rule:
`parseRule(TEST, InputStirng, StartSymbol) => parse(TEST + LATTICE + CASTS + RULE-CELLS, InputString, StartSymbol)`
Everything about the parser that is available in Java, is now available in K as well, making K in K a step closer.

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Grammar.java
```diff
@@ -369,6 +369,9 @@ public boolean equals(Object obj) {
         }
 
         public State(String name, NonTerminal nt) {
+            synchronized (State.class) {
+                unique = counter++;
+            }
```
radumereuta, line 374: Doing this fixed the performance issues?

radumereuta:
lgtm

k-distribution/include/builtin/kast.k
grosu, line 167: @radumereuta We need to talk about this, to make sure we are on the same page.  It is OK for the frontend or even kompiler to generate new modules in the process, but it is NOT ok for users to have to remember that certain module names mean certain things, like we used to do with the `-SYNTAX` suffix.  The benefit we got for this is too minor to be worth the confusion and uncertainty it created to users.  I'll look for you some time today.

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIntOperations.java
```diff
@@ -33,6 +33,16 @@ public static IntToken div(IntToken term1, IntToken term2, TermContext context)
         }
     }
 
+    public static IntToken ediv(IntToken term1, IntToken term2, TermContext context) {
```
andreistefanescu, line 36: is there a erem?

kernel/src/main/java/org/kframework/kore/compile/ConvertDataStructureToLookup.java
```diff
@@ -280,33 +287,41 @@ private K convertList(KApply k, KLabel collectionLabel, List<K> components) {
...
-                        K element = elementsRight.get(i);
-                        state.add(KApply(KLabel("#match"), element, KApply(KLabel("List:get"), list, KToken(Integer.toString(i-elementsRight.size()), Sorts.Int()))));
+                        list = newDotVariable();
+                        // Ctx[ListItem(5) Frame ListItem(X) ListItem(foo(Y))] => Ctx [L]
+                        // requires Frame := range(L, 1, 2)
+                        // andBool 5 := L[0]
+                        // andBool X := L[-2]
+                        // andBool foo(Y) := L[-1]
+                        if (frame != null) {
+                            state.add(KApply(KLabel("#match"), frame, KApply(KLabel("List:range"), list,
```
andreistefanescu, line 307: why `List:range` and not `List.range`?

kernel/src/main/java/org/kframework/kore/compile/ConvertDataStructureToLookup.java
```diff
@@ -280,33 +287,41 @@ private K convertList(KApply k, KLabel collectionLabel, List<K> components) {
...
+                        } else {
+                            KLabel unit = KLabel(m.attributesFor().apply(collectionLabel).<String>get("unit").get());
+                            // Ctx[.List] => Ctx[L] requires L ==K range(L, 0, 0)
+                            state.add(KApply(KLabel("_==K_"), KApply(unit), KApply(KLabel("List:range"), list,
+                                    KToken(Integer.toString(elementsLeft.size()), Sorts.Int()),
+                                    KToken(Integer.toString(elementsRight.size()), Sorts.Int()))));
+                        }
+                        for (int i = 0; i < elementsLeft.size(); i++) {
+                            K element = elementsLeft.get(i);
+                            state.add(KApply(KLabel("#match"), element, KApply(KLabel("List:get"), list, KToken(Integer.toString(i), Sorts.Int()))));
```
andreistefanescu, line 319: see above

andreistefanescu:
lgtm

bmmoore:
Aside from comments above, everything in ocaml-backend, kore, and kernel:kore.compile looks good.

I've also looked less carefully over the rest of the commit, without seeing any big issues.

cos:
lgtm. The changes to kore are minimal, and I don't really know much about the ocaml part.

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Grammar.java
dwightguth, line 374: yes

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIntOperations.java
dwightguth, line 36: it's called emod but yes, it exists. It is currently implemented by a K rule in the java backend.

kernel/src/main/java/org/kframework/kore/compile/ConvertDataStructureToLookup.java
dwightguth, line 307: I didn't change any of the klabels. There is a slight chance if we do this we might break someone's definition, which is much more likely than the associated risk with changing the hook names, which should only break people if they are using their own version of the prelude. I didn't think it was worth making the change.

dwightguth:
@bmmoore please re-review. @andreistefanescu @radumereuta are you okay if I merge once Brandon is satisfied?

radumereuta:
I don't see any changes int he stuff I'm interested in, so lgtm.

dwightguth:
@radumereuta I think the changes were mainly in this commit: 5cd52dd409d72607faa5b427cce9e2ccc499478a

radumereuta:
I had another quick look, and I haven't seen anything out of place. Lgtm.

bmmoore:
I think the comments are sufficient to merge. Considedring how it's used in, I think the call to `getCellBagSort` in `SortCells.java` should be replaced with a method that just returns the bag that participates in the configuration (maybe just changing the role of getCellBagSort), but you've got enough error checking around it that it will at least fail loudly. It would be nice if you can change this along with fixing the merge conflicts, but I won't mind if you merge as-is.

dwightguth:
Jenkins: test this please

PR #1559, opened 2015-07-16
Add minimum Java release number to README

bmmoore:
Some code no longer compiles with older Java 8 JDK releases.
This commit adds a precise version number to the README.md,
specifically 8u45 which is known to work because that's
what Jenkins is currently using.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@ This is a readme file for the developers.
 
 # Prerequisites
 
-## Java Development Kit (required JDK8 or higher)
+## Java Development Kit (required JDK8 version 8u45 or higher)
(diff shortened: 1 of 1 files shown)


bmmoore:
The current master doesn't build anymore with 8u20, I had to upgrade.

cos:
It may be unreasonable to ask users to have a specific version of Java 8. Was there a bug in Java itself preventing compilation?

bmmoore:
Yes, the 8u20 compiler failed to infer types for some uses of generic functions which are handled just fine by more recent releases. Also, the readme still says "or newer", and I know at least 8u51 also works because that's what I actually upgraded my own system to.

PR #1567, opened 2015-07-28
Support IO operations for KORE pipeline

daejunpark:
do not review yet


daejunpark:
Use case:

f.k:

```
require "domains.k"

module F-SYNTAX
imports STRING
syntax Stmt ::= "run" | "read" "(" String ")" | "print" "(" K ")" | "output"
endmodule

module F

imports F-SYNTAX
imports STDIN-STREAM
imports STDOUT-STREAM
imports INT 

configuration <T> 
  <k> $PGM:K </k>
  <stdin/>
  <stdout/>
//<stderr/>
</T>

rule <k> print(S:String) => .K ...</k>
     <stdout>
       ... 
       `.List => ListItem(S) ListItem("\n")`
     </stdout>

rule <k> read(Sort:String) ... </k>
     <stdin>
       `.List => ListItem(#parseInput(Sort, " \n\t\r"))`
       ListItem(#buffer(_:String))
       ListItem(_)
       ListItem(#istream(_:Int))
     </stdin>

rule <k> read(_) => V ... </k>
     <stdin>
       ListItem(V:K) => .List
       ... 
     </stdin>
     when isStream(V) =/=K true

rule run => read("Int") ~> output ~> read("Int") ~> output ~> read("String") ~> output

rule I:Int ~> output => print(Int2String(I))
rule S:String ~> output => print(S)

endmodule
```

1.f:

```
run
```

kompile:

```
$ kompile --kore f.k
```

krun with real IO on:

```
$ krun --kore --smt none --io on 1.f <<<"1 2 abc"
1
2
abc

<T> <k> .::K </k> <stdin> ListItem ( #buffer ( "" ) ) ( ListItem ( "on" ) ListItem ( #istream ( 0 ) ) ) </stdin> <stdout> ListItem ( #ostream ( 1 ) ) ( ListItem ( "on" ) ListItem ( #buffer ( "" ) ) ) </stdout> </T>
```

krun with real IO off (e.g., `--search` or `--debugger` mode)

```
$ krun --kore --smt none --io off 1.f <<<"1 2 abc"
<T> <k> .::K </k> <stdin> ListItem ( #buffer ( "" ) ) ( ListItem ( "off" ) ListItem ( #istream ( 0 ) ) ) </stdin> <stdout> ListItem ( #ostream ( 1 ) ) ( ListItem ( "off" ) ListItem ( #buffer ( "1\n2\nabc\n\n" ) ) ) </stdout> </T>
```

Note that the output is stored in `<stdout>` cell.

daejunpark:
@dwightguth please review.

This is an initial tryout to see if I correctly understood your requirement. Please see the above use case.

Note that my mental model is to have a minimal IO support enough to pass all the tutorial tests, since I think we will have to have a better design of IO in the end.

There are several notes and issues:
- Instead of `$noIO`, I used an explicit `$IO` flag which has a value either `on` or `off`. The initial reason is that there was an issue regarding BuiltinList append operation ($noIO somehow became KItem which is not compatible with BuiltinList) which seems not trivial to fix. But I think the current form seems not that bad, so we can go with this for now. Let me know if you really want to keep the original `$noIO` style flag.
- As we discussed, for now, an user is required to write two explicit rules for `read` operation (see the above usage), which was implicitly done by the java backend in KIL pipeline. We have to discuss how and where to implement this feature.
- For now, I disabled `<stderr>`, since the current design would result in duplicating several rules for each `<stdout>` and `<stderr>`. Indeed, stderr was not supported in KIL pipeline, nor used in tutorials.
- I used `InitialConfigurationProvider.getStdinBuffer` to get stdin buffer when real IO is disabled (e.g, `--search` or `--debugger`). The problem is that the method is parameterized with `tty.stdin`, but I have no idea how to support `tty` in KORE pipeline. So, for now, I simply ignore the value. Could you please support tty in KORE pipeline so that I can use that information, or tell me how to implement it.
  - Also, I made it as a public static method, and used it KRun.java, which may be a bad design; let me know if you have a better idea. 
- There is an issue regarding injection in KORE pipeline. I suspect it is not implemented yet. Since the `#write` builtin hook function returns an injected KSequence.EMPTY (`.K`), it simply doesn't work with the current KORE pipeline. We will have to discuss how to implement it. For now, if you want to run the above use case, we will need to the following (unsafe) change:

```
diff --git a/java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIOOperations.java b/java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIOOperations.java
index 6194b12..3cf7ddd 100644
--- a/java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIOOperations.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIOOperations.java
@@ -109,7 +109,8 @@ public class BuiltinIOOperations {
     public Term write(IntToken term1, StringToken term2, TermContext termContext) {
         try {
             fs.get(term1.longValue()).write(term2.byteArrayValue());
-            return KLabelInjection.injectionOf(KSequence.EMPTY, termContext);
+            return KSequence.EMPTY;
         } catch (CharacterCodingException e) {
             throw new IllegalArgumentException(e);
         } catch (IOException e) {
```

k-distribution/include/builtin/domains.k
```diff
@@ -144,6 +144,7 @@ module LIST
   /*@ Construct an empty List: */
   syntax List ::= ".List"                 [function, hook(List:.List), smtlib(smt_seq_nil), latex(\dotCt{List})]
                 //| "."
+  syntax List ::= "(" List ")" [bracket]
```
dwightguth: I would suggest using ```` instead of littering the concrete syntax of all programming languages.

k-distribution/include/builtin/domains.k
```diff
@@ -523,21 +524,39 @@ module K-IO
                   | #ostream(Int)
   syntax Stream ::= "#noIO"
 
-  //TODO(dwightguth): add back support for stream cells
-/*
-  rule ListItem(#ostream(_))
+  configuration <stdin>  ListItem(#buffer($STDIN:String)) ListItem($IO:String) ListItem(#istream(#stdin)) </stdin>
```
dwightguth: I would prefer having separate modules for each stream so that we don't cause conflicts for applications that don't want to use the streams IO. That means the K-IO module would have no stream stuff in it, just the low-level posix system calls and maybe also the #buffer production. Then there would be a STDIN-STREAM and STDOUT-STREAM modules each with their respective component of the configuration.

k-distribution/include/builtin/kast.k
```diff
@@ -86,7 +86,10 @@ endmodule
 module CONFIG-CELLS
   imports KCELLS
   syntax #CellName ::= r"[a-zA-Z0-9\\-]+"  [token]
+
   syntax Cell ::= "<" #CellName #CellProperties ">" K "</" #CellName ">" [klabel(#configCell)]
+  syntax Cell ::= "<" #CellName "/>" [klabel(#ioCell)]
```
dwightguth: klabel(#externalCell)

kernel/src/main/java/org/kframework/kore/compile/GenerateSentencesFromConfigDecl.java
```diff
@@ -105,6 +105,19 @@
...
                 throw KEMException.compilerError("Malformed cell in configuration declaration.", term);
+            } else if (kapp.klabel().name().equals("#ioCell")) {
+                if (kapp.klist().size() == 1) {
+                    K startLabel = kapp.klist().items().get(0);
+                    if (startLabel instanceof KToken) {
+                        KToken label = (KToken) startLabel;
+                        if (label.sort().equals(Sort("#CellName"))) {
+                            String cellName = label.s();
+                            Sort sort = Sort(getSortOfCell(cellName));
+                            return Tuple3.apply(Set(), Lists.newArrayList(sort), KApply(KLabel(getInitLabel(sort)), KVariable("Init")));
```
dwightguth: this feels shaky to me. Specifically, if I externally declare a cell with no configuration variables, this block of code acts like they exist and will likely cause the configuration to get stuck. I suggest you examine the productions in the module you are currently in, looking for ones with the "initializer" tag and the return sort equal to the cell name. They will tell you whether a Map is required or not.

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -149,6 +149,15 @@ private K parseConfigVars(KRunOptions options, CompiledDefinition compiledDef) {
             K configVar = externalParse(parser, value, sort, Source.apply("<command line: -c" + name + ">"), compiledDef);
             output.put(KToken("$" + name, Sorts.KConfigVar()), configVar);
         }
+        if (options.io()) {
+            output.put(KToken("$STDIN", Sorts.KConfigVar()), KToken("\"\"", Sorts.String()));
+            output.put(KToken("$IO", Sorts.KConfigVar()), KToken("\"on\"", Sorts.String()));
+        } else {
+            // TODO(daejunpark): support tty or have generic getStdinBuffer
+            String stdin = InitialConfigurationProvider.getStdinBuffer(true);
```
dwightguth: I think you are going to have to pass this boolean value as an argument to the constructor for KRun.java honestly. Having this value be true all the time will likely break K if you run it with nailgun, which is not something that I want to have happen since it will have an immediate impact on my developer throughput.

dwightguth:
looks okay but it needs testing. you need to create a version of IMP++ for kore somewhere inside k-distribution/tests and wire it into ktest.  You probably won't be able to get all the tests passing because we don't currently support superheat/supercool in KORE, but you will have to at the very least add tests for both execution and search.

daejunpark:
@dwightguth please review again.

As you suggested, a new regression test is created, which essentially IMP (not IMP++) plus IO operations. I failed to make IMP++ work due to several orthogonal issues such as parsing problem of `List{Stmt,""}` and `<threads>` cell issue. I think that should be handled separately.

Regarding the injection issue, I will make a separate pull request.

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIOOperations.java
```diff
@@ -109,7 +113,12 @@ public Term putc(IntToken term1, IntToken term2, TermContext termContext) {
     public Term write(IntToken term1, StringToken term2, TermContext termContext) {
         try {
             fs.get(term1.longValue()).write(term2.byteArrayValue());
-            return KLabelInjection.injectionOf(KSequence.EMPTY, termContext);
+            // TODO(daejunpark): hack: should implement the injection mechanism in kore
+            if (kRunOptions.experimental.kore) {
```
dwightguth, line 117: oh, that works too.

k-distribution/include/builtin/domains.k
```diff
@@ -14,6 +14,7 @@ module MAP
   syntax Map ::= Map Map                        [left, function, hook(MAP.concat), klabel('_Map_), assoc, comm, unit(.Map), element(_|->_)]
   /*@  Construct an empty Map */
   syntax Map ::= ".Map"                         [function, hook(MAP.unit), latex(\dotCt{Map})]
+  syntax Map ::= "(" Map ")" [bracket]
```
dwightguth: I still feel ambivalent about this. You said it's because of unparsing but I don't really understand what was wrong with the way it worked before this change.

k-distribution/tests/regression/kore-io/imp.k
```diff
@@ -0,0 +1,124 @@
...
+  rule <k> read() => I ...</k>
+       <stdin> ListItem(I:Int) => .List ...</stdin>
+       when isStream(I) =/=K true
+       [read]
+
+  // TODO(daejunpark): need to be implicitly done by kompile?
+  rule <k> read() ... </k>
+       <stdin>
+         `.List => ListItem(#parseInput("Int", " \n\t\r"))`
+         ListItem(#buffer(_:String))
```
dwightguth, line 115: what happens if you remove this line of code? does it still work?

dwightguth:
looks okay just a few more minor things.

@grosu btw we get composable configurations for free from this too!

k-distribution/include/builtin/domains.k
daejunpark: Indeed, it came from Brandon's commit f13341920c7d6786fab3080ea4c009e3f1bca6f1. @bmmoore can you say what was the problem regarding this?

k-distribution/tests/regression/kore-io/imp.k
daejunpark, line 115: It doesn't work if you remove the above line `ListItem(#buffer(_:String))`, because this rule can apply over and over again without doing #parseInput. I think this is the most simplest rule for now.

java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinIOOperations.java
daejunpark, line 117: ok, then I'll leave it for now, and just open an issue regarding the injection problem.

k-distribution/tests/regression/kore-io/imp.k
dwightguth, line 115: oh I see. Yeah, that's fine then, but we should definitely create a compiler pass to generate this rule automatically because it's really not very nice.

k-distribution/include/builtin/domains.k
daejunpark: I dropped the brackets both from Map and List, since somehow it works now without that. I guess the problem was fixed in the latest consList branch.

dwightguth:
Jenkins: test this please

k-distribution/tests/regression/kore-io/imp.k
daejunpark, line 115: I opened another issue for this.

daejunpark:
You can merge if you want.

k-distribution/tests/regression/kore-io/imp.k
grosu, line 115: @daejunpark @dwightguth The rule above is not only not really nice, it is horrible.  For me this is just nightmare-ish deja vu; we had similar discussions and similar ugly rules when we started out existing design of the IO.  The history repeats and nothing is learned from it.  As I said in my message to the k-list, unless somebody proposes a better design for IO, we will stick to the one that we made several years ago.  Lots of us spent a lot of time coming up with it.  A lot more time than Daejun spent now, a lot more people.  Therefore, I'm not going to just accept to throw a nice working away for some adhoc design, justified only by (simpler) implementation.  @daejun, you said somewhere that your objective is just to make this work, because "we need a better design of this".  How do you know it so well that we need a better design?  I still haven't heart any good criticism of the existing design (I heard some criticism of its implementation, which is different thing).  And why deliberately waste weeks from your life doing something adhoc that will be thrown away?  You should never do that, never!  If you really wanted to have a quick patch in order to move on, then you should have simply translated the existing KIL-based implementation ad-literam to KORE (as I suggested at some point, but you ignored that silently), as ugly as it was.

Let me repeat the current _design_ of the IO:
1) Cells can be linked to streams, which can come from stdin, stdout, sockets, files, etc., and in general from any URIs.  This link-age of a cell to a stream is the only thing that the user has to do, and that is done when the configuration is defined.
2) Rules can match (or read) or write items in those cells, just like in any other cells.  When a rule matches (or reads), a term of a certain sort in a streamed cell, for example, the parser will be invoked to read an appropriate sequence of characters from the streamed cell.  When a rule writes a term of a certain sort to a streamed cell, then the pretty-printer will be invoked to stream than term as a sequence of characters.

No, I don't buy it that this is doomed to be slow!  It will be slow only if implemented naively.  There is no reason to not "parse" primitive values in an input stream using the fastest POSIX or whatever specialized library functions are available on the current platform, and similarly for writing things to cells.  If one wants to read an expression of sort `Exp` or a program of sort `Pgm`, well, that's a different story, and things can be slower in that case (it depends of the complexity of the grammar; see, for example, our paper on monitoring CFG specifications: http://fsl.cs.illinois.edu/index.php/Efficient_Monitoring_of_Parametric_Context-Free_Patterns).

I have added a summary of the above to the manual, too, hopefully this way will not get forgotten (again).

grosu, line 115: Just to clarify my position here.  I strongly think that in the long term we should have a nice, powerful and general design.  In fact, I believe we already have it, or at least I will continue to consider so until somebody will comes up with real criticisms and/or with a better design.  In the short term, I agree with you @daejunpark that the objective is to make the tutorial and our other existing language definitions work.  I am Ok with your `<stdin/>` and `<stdout/>` notation in the configuration for the time being, but we should avoid the user writing that ugly rule that produces items in the `<stdin/>` cell when no more and requested.  My advice, again, is: take as much as you can, all except the KIL part, from the old implementation.  For example, it took @traiansf  a few good days to make the output to `<stdout/>` work efficiently.  I remember how you could initially see each character being printed to the screen and taking 1000 factorial about 1 minute to display; then after buffering the input/output, it was printed instantaneously.  It is at best uneconomical to re-discover the same things the hard way, instead of just taking all these non-KIL related things from the old implementation.

daejunpark, line 115: @grosu 

I don't think that I came up with a something new and ignored all of the previous efforts. Indeed, as you said, I took as much as I can from the previous KIL-style implementation, and the current implementation is exactly the same with the previous except the following two things:
- Stream configuration declaration (e.g., `<stdin/>` instead of `<in stream="stdin"> .List </in>`): This is Dwight's requirement, and I assumed that you already agreed with that. If you don't like this style and want to go back to the previous style of declaration, let me know. But, I think this style of declaration seems not that bad for users, nor any performance degradation exists; rather it has a side advantage of the configuration composition, which is a different story though.
- Automatic `#parseInput` rule generation: The ugly rules for read operation are exactly the same with what the previous KIL-style implementation automatically generates for a given user's rule. The automatic generation of the ugly rules is not supported yet in this pull request, and will be implemented as a separate pull request. But I'm not sure if this is the best we can do even in the short term, and I wanted to discuss on this first, and then implement it. 

Note that the above two things have nothing to do with the performance or underlying IO operation mechanism. They are only about usability, which will be eventually matched to the previous KIL-style one.

PR #1569, opened 2015-07-29
C verification4

andreistefanescu:
@dwightguth please review

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/symbolic/ConjunctiveFormula.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/symbolic/ConjunctiveFormula.java
@@ -169,6 +169,54 @@ public class ConjunctiveFormula extends Term implements CollectionInternalRepres
     }
 
     public ConjunctiveFormula add(Equality equality) {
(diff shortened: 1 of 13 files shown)


java-backend/src/main/java/org/kframework/backend/java/symbolic/ConjunctiveFormula.java
```diff
@@ -155,11 +169,60 @@ public ConjunctiveFormula add(Substitution<Variable, Term> substitution) {
     }
 
     public ConjunctiveFormula add(Equality equality) {
+        /* simplify andBool */
+        if (equality.leftHandSide() instanceof KItem
```
dwightguth, line 173: shouldn't there be a better way to deal with this than this?

andreistefanescu, line 173: in theory we could add them as rewrite rules for formulae simplification, but in practice we don't have the infrastrucutre for that yet, and without the proper infrastrucure, it would quite slow

dwightguth:
Lgtm

PR #1572, opened 2015-07-30
fix races reported by rv-predict

dwightguth:
**do not review yet**

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/kil/Variable.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/kil/Variable.java
@@ -161,11 +161,13 @@ public class Variable extends Term implements Immutable, org.kframework.kore.KVa
     Object readResolve() {
         if (anonymous) {
             int id = Integer.parseInt(name.substring(VARIABLE_PREFIX.length()));
-            if (id < counter) {
-                return deserializationAnonymousVariableMap.get(Pair.of(id, sort), this::getFreshCopy);
-            } else {
-                counter = id + 1;
-                return this;
+            synchronized (Variable.class) {
+                if (id < counter) {
+                    return deserializationAnonymousVariableMap.get(Pair.of(id, sort), this::getFreshCopy);
+                } else {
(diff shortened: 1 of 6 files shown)


dwightguth:
Jenkins: test this please

dwightguth:
So I really have no idea what the hell is happening here. NoClassDefFoundError? @yilongli if you have any idea you are welcome to suggest otherwise I am probably inclined not to bother with the Java Backend parts of this diff because honestly it's not worth the time spent figuring what the hell happened.

yilongli:
@dwightguth Can we hold on here please? I would like to send in the patch for the data race myself once I am absolutely sure it's not a false positive. Right now, you can best help me by commenting on the issue telling me whether it looks like a false positive. Thanks!

java-backend/src/main/java/org/kframework/backend/java/kil/Variable.java
```diff
@@ -161,11 +161,13 @@ public ASTNode accept(Transformer transformer) {
     Object readResolve() {
         if (anonymous) {
             int id = Integer.parseInt(name.substring(VARIABLE_PREFIX.length()));
-            if (id < counter) {
-                return deserializationAnonymousVariableMap.get(Pair.of(id, sort), this::getFreshCopy);
-            } else {
-                counter = id + 1;
-                return this;
+            synchronized (Variable.class) {
```
yilongli, line 164: Which race are you trying to fix here?

dwightguth, line 164: There was a race reported on counter. But I am inclined to suspect an atomic integer operation is probably better here too.

yilongli, line 164: Can you file an issue with the race report attached? I would like to take a look. Besides, we want to keep track of the real races found by RV-Predict.

dwightguth:
Closing. Yilong can fix these when he has time and get my help if he is not familiar enough with the code.

Issue #1543, opened 2015-06-19
data race on field org.kframework.kompile.KompileOptions.files


yilongli:
Reported by RV-Predict:

```
Data race on field org.kframework.kompile.KompileOptions.files: {{{
    Concurrent write in thread T57 (locks held: {Monitor@6f030508})
 ---->  at org.kframework.kompile.KompileOptions.setFiles(KompileOptions.java:40)
        at org.kframework.utils.inject.DefinitionLoadingModule$$FastClassByGuice$$6c55efa6.invoke(<generated>:n/a)
        at com.google.inject.internal.ProviderMethod$FastClassProviderMethod.doProvision(ProviderMethod.java:230)
        at com.google.inject.internal.ProviderMethod.get(ProviderMethod.java:148)
        at com.google.inject.internal.InternalFactoryToInitializableAdapter.provision(InternalFactoryToInitializableAdapter.java:54)
        at com.google.inject.internal.ProviderInternalFactory.circularGet(ProviderInternalFactory.java:66)
        at com.google.inject.internal.InternalFactoryToInitializableAdapter.get(InternalFactoryToInitializableAdapter.java:46)
        at com.google.inject.internal.SingleParameterInjector.inject(SingleParameterInjector.java:38)
        at com.google.inject.internal.SingleParameterInjector.getAll(SingleParameterInjector.java:62)
        at com.google.inject.internal.ConstructorInjector.provision(ConstructorInjector.java:107)
        at com.google.inject.internal.ConstructorInjector.construct(ConstructorInjector.java:88)
        at com.google.inject.internal.ConstructorBindingImpl$Factory.get(ConstructorBindingImpl.java:269)
        at com.google.inject.internal.SingleParameterInjector.inject(SingleParameterInjector.java:38)
        at com.google.inject.internal.SingleParameterInjector.getAll(SingleParameterInjector.java:62)
        at com.google.inject.internal.ConstructorInjector.provision(ConstructorInjector.java:107)
        at com.google.inject.internal.ConstructorInjector.construct(ConstructorInjector.java:88)
        at com.google.inject.internal.ConstructorBindingImpl$Factory.get(ConstructorBindingImpl.java:269)
        at com.google.inject.internal.SingleParameterInjector.inject(SingleParameterInjector.java:38)
        at com.google.inject.internal.SingleParameterInjector.getAll(SingleParameterInjector.java:62)
        at com.google.inject.internal.ConstructorInjector.provision(ConstructorInjector.java:107)
[...]

PR #1579, opened 2015-07-31
fix race in KompileOptions.files

dwightguth:
@yilongli please review. The net effect of this is that mainDefinitionFile is always accessed in kompile and therefore we don't need the FileUtil in this class in krun.

fixes https://github.com/kframework/k/issues/1543

### DIFF
--- a/kernel/src/main/java/org/kframework/kompile/KompileOptions.java
+++ b/kernel/src/main/java/org/kframework/kompile/KompileOptions.java
@@ -26,11 +26,17 @@ public class KompileOptions implements Serializable {
     @Parameter(description="<file>")
     private List<String> parameters;
 
+    private File mainDefinitionFile;
+
     public File mainDefinitionFile() {
-        if (parameters == null || parameters.size() == 0) {
-            throw KEMException.criticalError("You have to provide exactly one main file in order to compile.");
+        File m = mainDefinitionFile;
+        if (m == null) {
+            if (parameters == null || parameters.size() == 0) {
+                throw KEMException.criticalError("You have to provide exactly one main file in order to compile.");
+            }
+            m = mainDefinitionFile = files.get().resolveWorkingDirectory(parameters.get(0));
         }
-        return files.get().resolveWorkingDirectory(parameters.get(0));
+        return m;
     }
 
     private transient Provider<FileUtil> files;
--- a/kernel/src/main/java/org/kframework/kompile/KompileOptions.java
+++ b/kernel/src/main/java/org/kframework/kompile/KompileOptions.java
@@ -28,15 +28,14 @@ public class KompileOptions implements Serializable {
 
     private File mainDefinitionFile;
 
-    public File mainDefinitionFile() {
-        File m = mainDefinitionFile;
-        if (m == null) {
+    public synchronized File mainDefinitionFile() {
+        if (mainDefinitionFile == null) {
             if (parameters == null || parameters.size() == 0) {
                 throw KEMException.criticalError("You have to provide exactly one main file in order to compile.");
             }
-            m = mainDefinitionFile = files.get().resolveWorkingDirectory(parameters.get(0));
+            mainDefinitionFile = files.get().resolveWorkingDirectory(parameters.get(0));
         }
-        return m;
+        return mainDefinitionFile;
     }
 
     private transient Provider<FileUtil> files;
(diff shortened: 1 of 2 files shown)


kernel/src/main/java/org/kframework/kompile/KompileOptions.java
```diff
@@ -26,11 +26,17 @@
     @Parameter(description="<file>")
     private List<String> parameters;
 
+    private File mainDefinitionFile;
+
     public File mainDefinitionFile() {
-        if (parameters == null || parameters.size() == 0) {
-            throw KEMException.criticalError("You have to provide exactly one main file in order to compile.");
+        File m = mainDefinitionFile;
```
yilongli: why do you need the local variable `m`?

yilongli: Is this method going to be called by multiple threads?

dwightguth: it should just be called by one thread during Kompile, but I wanted to write it in a thread safe manner just in case.

yilongli: OK, but this is not thread-safe. There is a race on `mainDefinitionFile`. And since you want to make sure `files.get().resolveWorkingDirectory(parameters.get(0))` is called only once, this is not a benign race like the one when computing `hashCode`. If you are certain that it's unlikely to be called by multiple threads, I think the best way here is to simply declare the method as `synchronized` and get rid of the local variable `m`. If you are concerned about the performance, you can also declare `mainDefinitionFile` as volatile and then use double-checked locking.

dwightguth: It's benign. I don't need the call to happen only once; I just need it to
happen at least once during kompile, which it will.

yilongli: And `files.get().resolveWorkingDirectory(parameters.get(0))` is side-effect free, right?

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
```diff
@@ -26,11 +26,17 @@
...
+        if (m == null) {
+            if (parameters == null || parameters.size() == 0) {
+                throw KEMException.criticalError("You have to provide exactly one main file in order to compile.");
+            }
+            m = mainDefinitionFile = files.get().resolveWorkingDirectory(parameters.get(0));
```
yilongli: m = mainDefinitionFile => mainDefinitionFile = m

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
dwightguth: Well it might potentially trigger Guice to create the object more than
once, but Guice should be thread-safe. The instantiation process won't have
side effects though, all it's really doing is calling a couple
constructors. In any event, the odds of this method being called by more
than one thread are low enough that I am not worried. The real bug was the
possibility that the race would lead to using the wrong working directory
to locate the file, which shouldn't happen as long as the method gets
called at least once by kompile, which it does.

yilongli: OK, in this case, my suggestion is still to just declare the method as `synchronized` and let the JVM do lock elision. It's really not worth to use (potentially) benign race here. It's hard to read and reason about. Besides, you have to configure the race detector to suppress the warning.

yilongli: The Java semantics under data race is really tricky. For example, you wrote `m = mainDefinitionFile` instead of `mainDefinitionFile = m` below. And since `mainDefinitionFile` is in race, I believe it is legal for the read in `m = mainDefinitionFile` to see null value and then you get a NPE later.

yilongli: OK, I just checked the JLS and I was wrong about the read in `m = mainDefinitionFile` can see null value. It can only happen if the last line is written as `return mainDefinitionFile`. Anyway, I don't think it's a good idea to assume Guice is absolutely thread-safe when there is a simpler and equally efficient solution available.

dwightguth:
@yilongli please re-review

yilongli:
lgtm

PR #1581, opened 2015-07-31
improvements on SMT translation

andreistefanescu:
@dwightguth please review


dwightguth:
I'm inclined to propose what Grigore had to say when I did my last OCAML pull request. Do we really need the SMT attribute anymore? Can we just use the hook name as the name of the SMT operation and then have one single prelude that defines the names of all the relevant hooks?

andreistefanescu:
We can have a properties file mapping hook names to SMT operations names, the same way we have for builtin functions.

PR #1584, opened 2015-08-03
Nice error for configuration needing Map module

bmmoore:
Replace a bare .get with a better check that a Map
module is available, when processing the configuration
in Kompile (which assumes it can use the standard Maps if desired).


bmmoore:
Jenkins: test this please

bmmoore:
@dwightguth please review.

PR #1585, opened 2015-08-03
Avoid race on parser statistics

### LINKED ISSUE #1568: [tutorial-kore branch] data race on field org.kframework.kompile.Kompile.parsedBubbles
@bmmoore I ran `kompile --kore imp.k` on your tutorial-kore branch and got the following race report:

```
Data race on field org.kframework.kompile.Kompile.parsedBubbles: {{{
    Concurrent write in thread T14 (locks held: {})
 ---->  at org.kframework.kompile.Kompile.performParse(Kompile.java:407)
        at org.kframework.kompile.Kompile.lambda$resolveBubbles$101(Kompile.java:315)
        at org.kframework.kompile.Kompile$$Lambda$161.apply(Unknown:n/a)
    T14 is created by n/a

    Concurrent read in thread T13 (locks held: {})
 ---->  at org.kframework.kompile.Kompile.performParse(Kompile.java:407)
        at org.kframework.kompile.Kompile.lambda$resolveBubbles$101(Kompile.java:315)
        at org.kframework.kompile.Kompile$$Lambda$161.apply(Unknown:n/a)
    T13 is created by n/a
}}}
```

I took a quick look at the source code: https://github.com/kframework/k/blob/e2d6df353dfa237443139b1f95d8d912305817d9/kernel/src/main/java/org/kframework/kompile/Kompile.java
It looks like a real race to me.

bmmoore:
Now uses AtomicInteger for counters, rather than
just incrementing an ordinary field.
Fixes #1568

### DIFF
--- a/kernel/src/main/java/org/kframework/kompile/Kompile.java
+++ b/kernel/src/main/java/org/kframework/kompile/Kompile.java
@@ -58,6 +58,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -86,8 +87,8 @@ public class Kompile {
     private final KompileOptions kompileOptions;
     private final Stopwatch sw;
 
-    private int parsedBubbles = 0;
-    private int cachedBubbles = 0;
+    private final AtomicInteger parsedBubbles = new AtomicInteger(0);
+    private final AtomicInteger cachedBubbles = new AtomicInteger(0);
 
     public Kompile(KompileOptions kompileOptions, FileUtil files, KExceptionManager kem, Stopwatch sw, boolean cacheParses) {
         this.files = files;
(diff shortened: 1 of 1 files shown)


bmmoore:
jenkins: retest this

bmmoore:
jenkins: test this please

bmmoore:
Jenkins: test this please

bmmoore:
@yilongli please review. This is very small.

yilongli:
This is not the best fix. The real problem here is the use of stateful, interfering function in the [`Stream.flatMap`](https://github.com/kframework/k/pull/1585/files#diff-f2f0da219ae6d8b30590d7ad56d6b7a0R316) method, which is a clear violation of its API specification. I think the right way to fix this is to refactor `performParse` to not depend on the cache and then reorganize the [stream pipeline](https://github.com/kframework/k/pull/1585/files#diff-f2f0da219ae6d8b30590d7ad56d6b7a0R316) to filter out cached bubbles before applying `performParse`. As a side benefit, the cache map doesn't have to be a synchronized map any more.

bmmoore:
Using a properly synchronized cache doesn't violate the definition of stateful at http://docs.oracle.com/javase/8/docs/api/java/util/stream/package-summary.html#Statelessness - the caching doesn't change the result the function returns, and the counters are just statistics that are not really part of the result (any more than timing the operation might also change run to run). It also talks about properly synchronizing access mutable state instead of completely forbidding it.

"interference" doesn't apply here, the only collection being modified is the cache, which isn't the source of the stream.
http://docs.oracle.com/javase/8/docs/api/java/util/stream/package-summary.html#NonInterference

yilongli:
You are right about "stateless" and "interference". Still, the better way is to restructure the stream pipeline to avoid stateful behavioral parameters as suggested at the end of http://docs.oracle.com/javase/8/docs/api/java/util/stream/package-summary.html#Statelessness. I wouldn't insist you make the change in this pull request though.

bmmoore:
Well, it's not a "stateful behavioral parameter" when the cache is used to get exactly the same result faster, but I just don't see a good way to restructure the pipeline to avoid access to the cache.
Splitting it out as you suggest sounds like it would turn one pass into at least three, to filter cache entries, to parse them, and then to update the cache again - with the last step either being sequential, or again being a parallel operation that updates the cache, plus the intermediate results of the second stage would need to be a bit larger to keep the results paired with the input, plus it might lose some benefits from caching during different uncached bubbles in a single operation (it's unlikely for rule bodies to be repeated, but some simple requires clauses might repeat).

If you have a better suggestion or an argument that splitting it up like that isn't as bad as it sounds to me I'll consider a bigger cleanup, otherwise is this ok to merge?

yilongli:
OK, I think you are right. I thought there was an easy way to remove bubbles that have identical content first so that in the last step you don't need a cache. I found some alternatives [here](http://stackoverflow.com/questions/27870136/java-lambda-stream-distinct-on-arbitrary-key) but they are probably not much better than the current one.

Issue #1575, opened 2015-07-30
[kore] Parsing failed for List with empty separator


daejunpark:
_[Already reported to @radumereuta but just for tracking purpose.]_

For the following definition minimized from IMP++:

```
  1 require "domains.k"                                                                                                               
  2                                                                                                                                   
  3 module T-SYNTAX                                                                                                                   
  4   syntax BExp                                                                                                                     
  5   syntax Block ::= "{" Stmts "}"                                                                                                  
  6   syntax Stmt  ::= Block                                                                                                          
  7                  | "if" "(" BExp ")" Block "else" Block                                                                           
  8                  | "while" "(" BExp ")" Block                                                                                     
  9   syntax Stmts ::= List{Stmt,""}
 10 endmodule                                                                                                                         
 11                                                                                                                                   
 12 module T                                                                                                                          
 13   imports T-SYNTAX
 14   configuration <T> <k> $PGM:K </k> </T>
 15   rule while (B) S => if (B) {S while (B) S} else {.Stmts}  [structural]
 16 endmodule
```

Parsing got stuck at line 15 since it failed to recognize `S` and `while (B) S` are separated by `""`.

```
org.kframework.utils.errorsystem.KEMException: [Error] Compiler: Had 1 parsing errors.
    at org.kframework.utils.errorsystem.KEMException.create(KEMException.java:97)
    at org.kframework.utils.errorsystem.KEMException.compilerError(KEMException.java:63)
    at org.kframework.kompile.Kompile.parseDefinition(Kompile.java:254)
    at org.kframework.kompile.Kompile.run(Kompile.java:125)
    at org.kframework.kompile.KompileFrontEnd.run(KompileFrontEnd.java:89)
    at org.kframework.main.FrontEnd.main(FrontEnd.java:52)
    at org.kframework.main.Main.runApplication(Main.java:109)
    at org.kframework.main.Main.runApplication(Main.java:99)
    at org.kframework.main.Main.main(Main.java:51)
[Error] Compiler: Had 1 parsing errors.
[Error] Inner Parser: Parse error: unexpected character '}'.
    Source(/Users/daejunpark/kore/tmp/././t.k)
    Location(15,44,15,45)
```

Notes:
- If we have `{S S}` instead of `{S while (B) S}`, it works.
- If we have `List{Stmt, ","}` instead of `""` as the list separator, it works.

bmmoore:
[...]

PR #1588, opened 2015-08-05
Kore cell fragments

bmmoore:
This patch hopefully completes the implementation of cell fragments, following the design at https://github.com/kframework/k/wiki/KAST-and-KORE

It extends the translation from the configuration declaration into sentances to also generate the necessary sorts and productions for representing fragments, and the kore SortCells pass to make use of these fragments.
More work was necessary to support fragments in definitions, because of an old hack that let the user write sorts like ThreadCellFragment and siletly replaced them with Bag.
There's a bit of a cyclic dependency in parsing configuration declarations, worked around with a bit of a hack involving the new Module.usedCellSorts field.

The new features are covered by unit tests, and an example in regression/kore-cell-fragments/imp_fragments added to ktest suite.

### DIFF
--- a/kernel/src/main/java/org/kframework/kore/compile/GenerateSentencesFromConfigDecl.java
+++ b/kernel/src/main/java/org/kframework/kore/compile/GenerateSentencesFromConfigDecl.java
@@ -28,8 +28,11 @@ import org.kframework.utils.errorsystem.KEMException;
 import scala.Option;
 import scala.Tuple2;
 import scala.Tuple3;
+import scala.collection.Seq;
 import scala.collection.immutable.Set;
 
+import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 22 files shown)


grosu:
Thanks, Brandon, I really appreciate it!  Since it is still fresh in your mind, would you please revise the wiki page https://github.com/kframework/k/wiki/KAST-and-KORE to be consistent with the implementation, if needed?

bmmoore:
It should be entirely up to date now, except we don't have the Cell" suffix on all the examples of generated productions,

bmmoore:
By the way, the included imp++ compiles, but issue #1575 prevents testing any example programs.

dwightguth:
looks okay. The thing with the sort check being disabled on *Cell and *CellFragment sorts is kind of less than ideal, but I guess it can do for now. We need to add an entire compiler pass after parsing that does nothing but checks anyway, we can probably handle it then.

@cos do you have any issues?

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -184,7 +186,7 @@ case class Module(name: String, imports: Set[Module], localSentences: Set[Senten
   // check that non-terminals have a defined sort
   private val nonTerminalsWithUndefinedSort = sentences flatMap {
     case p@Production(_, items, _) =>
-      val res = items collect { case nt: NonTerminal if !definedSorts.contains(nt.sort) => nt }
+      val res = items collect { case nt: NonTerminal if !definedSorts.contains(nt.sort) && !usedCellSorts.contains(nt.sort) => nt }
```
cos, line 189: We probably want to eventually have a separate kore definition for the "flexible" syntax so we don't have to do this sort of checks inside kore classes. But that's not short term. The PL looks good to me as is.

kernel/src/main/java/org/kframework/kore/compile/GenerateSentencesFromConfigDecl.java
```diff
@@ -246,10 +258,16 @@ public K apply(KToken k) {
                 .map(Constructors::NonTerminal)), Stream.of(Terminal("</" + cellName + ">")))
                 .collect(Collectors.toList());
 
+        java.util.Set<Sentence> sentences = new HashSet<Sentence>();
+
+        // syntax Cell ::= "<cell>" Children... "</cell>" [cell, cellProperties, configDeclAttributes]
+        Production cellProduction = Production("<" + cellName + ">", sort, items.stream().collect(Collections.toList()),
+                cellProperties.addAll(configAtt));
+        sentences.add(cellProduction);
```
bmmoore, line 266: This caused #1612 - this doesn't remove or replace the ealier version of cellProduction added without the attribute.

PR #1593, opened 2015-08-06
Documentation fix for krun -p option

bmmoore:
The short form is just `-p`, not `-cp` like in the example in the `--help` text.


bmmoore:
@dwightguth You can merge this, I'm only waiting because I'm not sure why it wasn't working for me or if more changes are needed to the description.

PR #1596, opened 2015-08-07
fix several simple data races

yilongli:
@dwightguth please review


yilongli:
@dwightguth please review

PR #1600, opened 2015-08-07
minor optimization of computing Term.hashCode

yilongli:
@dwightguth please review


dwightguth:
lgtm. Maybe we should replace all the other caches with this paradigm? I think we are currently using a class MapCache that I wrote which may be less efficient...

yilongli:
Yes, I think we should replace them with ConcurrentHashMap whenever possible. I fixed two more non-thread-safe cache maps detected by RV-Predict.

Issue #1603, opened 2015-08-07
Type inference ignoring ambiguity


bmmoore:
In this example the last rule ends up with an inferred type of `X` for `X`, despite `A` and `B` also being legal sorts and none greater than the others. There are no warnings besides the informative messages that some variables are assigned inferred types.

```
require "domains.k"
module TEST-SYNTAX
syntax T ::= "wrap" M [klabel(wrapM)] | "wrap" X [klabel(wrapX)]
syntax A
syntax B
syntax M ::= A | B
syntax N ::= A | B | X // make A and B both maximal in the intersection of N and M
syntax X
syntax KItem ::= label(N,T)
endmodule
module TEST
imports MAP
imports TEST-SYNTAX
configuration <k>$PGM:K</k>
rule X => label(X,wrap X:A)
rule X => label(X,wrap X:B)
rule X => label(X,wrap X:X)
rule X => label(X,wrap X)
endmodule
```

This happens because VariableTypeInference filter handles parsing ambiguities (between the two "wrap" productions in this case) and type checking together by producing multiple sets of type constraints corresponding to combinations of ambiguous parses, throws out some sets of constraints, and then succeeds if a single uniquely-solvable set of constraints remains.
However, instead of just throwing out inconsistent sets of constraints it also throws out constraints which have multiple solutions but no greatest solution (e.g, X can be either A or B in the constraints coming from the parse of "wrap" as wrapM), before seeing if a single solution remains.

grosu:
This is a very interesting and well thought of example, @bmmoore! As a user, I'd like to see an error/warning message here telling me that variable X can have any of the sorts A, B, and X.

bmmoore:
That's exactly what you will get with changes in #1604.

grosu:
@bmmoore I think this example is nice to include in the manual.  Can you please add it there, with some text extracted from our recent discussions?  Now it is the right moment to do it, because it is still fresh in our mind.  We'll have to soon start writing the manual, we have too much pressure on our shoulders from users.

bmmoore:
Ok, it's added next to the other text on inference.

bmmoore:
(I thought I put the magic text in the #1604 message to automatically close this issue when it merged, looks like it didn't read past the "and")

PR #1604, opened 2015-08-08
Type inference fix

### LINKED ISSUE #1586: Type inference fails with overloading, for variables and wildcards
I thought type inference was supposed to give variables the most general types that let a rule parse, and then assign other labels and terms the least sorts after that.
This doesn't seem to always work, and also I get different errors about variables and wildcards. Here's a sample definition, to be saved as `test.k`

```
require "domains.k"
module TEST-SYNTAX
syntax A ::= B | "a"
syntax B ::= "b"
syntax TA ::= TB | t(A) [klabel(t)]
syntax TB ::= t(B) [klabel(t)]
endmodule
module TEST
imports MAP
imports TEST-SYNTAX
configuration <k>$PGM:K</k>
rule t(X)::TA => .K
rule t(Y) => .K
rule t(_) => .K
endmodule
```

Compiling it produces as `kompile --kore test.k` produces this output:

```
[Error] Compiler: Had 1 parsing errors.
[Error] Critical: Could not infer a unique sort for variable 'Y'. Possible
sorts: A, B
    Source(/home/brandon/test/k/wildcard/././test.k)
    Location(13,6,13,16)
1: syntax TB ::= "t" "(" B ")"
[Location(#token("6",Int),#token("15",Int),#token("6",Int),#token("30",Int))
Source(#token("/home/brandon/test/k/wildcard/././test.k",KString))
klabel(#token("t",AttributeValue)) klabel(#token("t",KString))]
    t(#SemanticCastToB(#token(#KVariable,"_")))
2: syntax TA ::= "t" "(" A ")"
[Location(#token("5",Int),#token("20",Int),#token("5",Int),#token("35",Int))
Source(#token("/home/brandon/test/k/wildcard/././test.k",KString))
klabel(#token("t",AttributeValue)) klabel(#token("t",KString))]
    t(#SemanticCastToA(#token(#KVariable,"_")))
    Source(/home/brandon/test/k/wildcard/././test.k)
    Location(12,6,12,10)
```

Note that there's an error about `Y` but not `X` even though `Y` could be typed just like `X`, and that the wildcard gets only a warning even though the ambiguity is equivalent to the ambiguity about `Y` (and it just happens to default to the less general type).

I expected all of X, Y, and _ to end up having inferred type A.

bmmoore:
This updates the type inference code to calculate sorts also for _ variables, and to more carefully combine sort constraints from multiple possible parses.

Now a variable is assigned an inferred sort only if that sort is a (nonstrict) supersort of the sort the variable has in every possible way of assigning legal variable sorts in any of the possible ambiguous parses. An error will be reported if any variable doesn't have a unique maximal sort, or if it's not possible to simultaneously assign all those sorts. Also, an ambiguity error will be reported a bit after this code if assigning these inferred types isn't enough to reduce the possible parses down to a single KORE term (debugging VariableTypeInferenceFilter may show some ambiguity nodes, because this code is still working in terms of productions where e.g. overloaded labels may lead to mutliple syntax trees that still eventually flatten to the same AST).

@radumereuta please review.

Fixes #1586 and #1603.

### DIFF
--- a/kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/VariableTypeInferenceFilter.java
+++ b/kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/VariableTypeInferenceFilter.java
@@ -30,9 +30,12 @@ import scala.util.Either;
 import scala.util.Left;
 import scala.util.Right;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
+import java.util.ListIterator;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
@@ -56,6 +59,69 @@ public class VariableTypeInferenceFilter extends SetsGeneralTransformer<ParseFai
         this.sortSet = sortSet;
         this.productions = productions;
         this.inferSortChecks = inferSortChecks;
+    }
+
+    /** return the subset of maximal elements of {@code sorts} */
+    private Set<Sort> maximal(Collection<Sort> sorts) {
+        Set<Sort> maxSorts = new HashSet<>();
+
+        nextsort:
+        for (Sort vv1 : sorts) {
+            for (Sort vv2 : sorts)
+                if (subsorts.lessThan(vv1, vv2))
+                    continue nextsort;
+            maxSorts.add(vv1);
+        }
+        return maxSorts;
+    }
+
+    /** Return the set of all known sorts which are a lower bound on
+     * all sorts in {@code bounds}, leaving out internal sorts below "KBott".
+     */
+    private Set<Sort> lowerBounds(Collection<Sort> bounds) {
+        Set<Sort> mins = new HashSet<>();
+        nextsort:
+        for (Sort sort : iterable(sortSet)) { // for every declared sort
+            if (subsorts.lessThanEq(sort, Sort("KBott")))
+                continue;
+            for (Sort bound : bounds)
+                if (!subsorts.greaterThanEq(bound, sort))
+                    continue nextsort;
+            mins.add(sort);
+        }
+        return mins;
+    }
+
+    // When passed a mutable List {@code sets} of nonempty subsets of {@code universe},
+    // returns a set containing at least one item in common with each of the sets.
+    // Empties {@code sets}.
+    private <T> Set<T> hittingSet(Set<T> universe, List<Set<T>> sets) {
+        assert sets.stream().allMatch(s -> !s.isEmpty());
+        Set<T> hittingSet = new HashSet<>();
+        while (!sets.isEmpty()) {
+            T maxItem = null;
+            int maxCount = 0;
+            for (T item : universe) {
+                int count = 0;
+                for (Set<T> s : sets) {
+                    if (s.contains(item)) {
+                        ++count;
+                    }
+                }
+                if (count > maxCount) {
+                    maxItem = item;
+                    maxCount = count;
+                }
+            }
+            hittingSet.add(maxItem);
+            ListIterator<Set<T>> li = sets.listIterator();
+            while (li.hasNext()) {
+                if (li.next().contains(maxItem)) {
+                    li.remove();
+                }
+            }
+        }
+        return hittingSet;
     }
 
     static final class VarKey {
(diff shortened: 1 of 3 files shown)


bmmoore:
Passes when I run that test by hand, or run ktest locally.

bmmoore:
Jenkins: retest this please

dwightguth:
lgtm. When you parse the C semantics with this are you able to successfully run tests in the OCAML backend?

bmmoore:
C is actually broken in master, from a bug filed as #1612. I'm looking into it, I suspect it was a missed problem in the cell fragments change.

bmmoore:
That's been fixed now. Just to match what I'm testing C against on my local machine, I'm retesting this branch.

bmmoore:
Jenkins: retest this please.

bmmoore:
@dwightguth The `test-build` target passed. (I had to fix the C makefile to not lose K's increased stack size when setting K_OPTS to avoid a stack overflow).

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/VariableTypeInferenceFilter.java
```diff
@@ -58,6 +61,111 @@ public VariableTypeInferenceFilter(POSet<Sort> subsorts, scala.collection.immuta
...
+
+        nextsort:
+        for (Sort vv1 : sorts) {
+            for (Sort vv2 : sorts)
+                if (subsorts.lessThan(vv1, vv2))
+                    continue nextsort;
+            maxSorts.add(vv1);
+        }
+        return maxSorts;
+    }
```
radumereuta: Wait, how does this work? Continue goes back to the beginning... What is the exit condition?
Also, are jumps allowed?
I usually implemented maximal like this:

```
for (Sort vv1 : sorts) {
    boolean maximal = true;
    for (Sort vv2 : sorts) {
        if (subsorts.lessThan(vv1, vv2)) {
            maximal = false;
            break;
        }
        if (maximal)
            maxSorts.add(vv1);
    }
}
```

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/VariableTypeInferenceFilter.java
```diff
@@ -58,6 +61,111 @@ public VariableTypeInferenceFilter(POSet<Sort> subsorts, scala.collection.immuta
...
+        for (Sort sort : iterable(sortSet)) { // for every declared sort
+            if (subsorts.lessThanEq(sort, Sort("KBott")))
+                continue;
+            for (Sort bound : bounds)
+                if (!subsorts.greaterThanEq(bound, sort))
+                    continue nextsort;
+            mins.add(sort);
+        }
+        return mins;
+    }
```
radumereuta, line 79: Shouldn't these be part of `POSet`?
They feel like general operations over a poset. I think in KIL we had them implemented like that, but in KORE we haven't ported all the operations. Also, the ones we have right now are implemented really inefficiently.
And again, I don't understand how this one finishes if you implement it like this with jumps.

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/VariableTypeInferenceFilter.java
```diff
@@ -95,55 +203,20 @@ public VariableTypeInferenceFilter(POSet<Sort> subsorts, scala.collection.immuta
...
-                            solution.putAll(key, maxSorts);
-                        else {
-                            failsAmb = maxSorts;
-                            failsAmbName = key;
-                            solution.clear();
-                            break;
-                        }
                     } else {
-                        solution.putAll(key, mins);
+                        solution.putAll(key, maximal(mins));
```
radumereuta: Ah, ok. I get why you put the code in this file.

kernel/src/test/java/org/kframework/parser/concrete2kore/RuleGrammarTest.java
```diff
@@ -347,4 +347,36 @@ public void test20() {
...
+                "module TEST\n" +
+                "syntax T ::= \"wrap\" M [klabel(wrapM)] | \"wrap\" X [klabel(wrapX)]\n" +
+                "syntax A\n" +
+                "syntax B\n" +
+                "syntax M ::= A | B\n" +
+                "syntax N ::= A | B | X // make A and B both maximal in the intersection of N and M\n" +
+                "syntax X\n" +
+                "syntax KItem ::= label(N,T)\n" +
+                "endmodule";
+        parseRule("X => label(X,wrap X)", def, 0, true);
```
radumereuta, line 380: This should have a warning. X is not declared so it should contain a hidden warning.

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/VariableTypeInferenceFilter.java
bmmoore: It's continuing the outer loop. It's a named loop, not a jump label: https://docs.oracle.com/javase/tutorial/java/nutsandbolts/branch.html

kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/VariableTypeInferenceFilter.java
bmmoore, line 79: Maybe it should. This code used to be in the body of the apply method, and I only took it as far as extracting it into a method - also the new POSet is written in scala so I couldn't just refactor the method to there.

kernel/src/test/java/org/kframework/parser/concrete2kore/RuleGrammarTest.java
bmmoore, line 380: I'm expecting an error (with `true`) rather than assign X a sort, as discussed at #1603. Making type inference behave that way is one of the points of making this pull request.

bmmoore:
FYI, @cos maybe you want to look at the POSet method?

radumereuta:
I just tested the issue myself. If this is the wanted behavior, then go ahead with merging.
lgtm

PR #1571, opened 2015-07-30
Substantial optimizations to OCAML backend

dwightguth:
@bmmoore @radumereuta @cos please review

This is probably my last commit to the OCAML backend for now. Everything seems to work well and it performs at roughly 250,000x C, which is roughly 90-100x the Java backend. Substantial further performance results will have to come from something like partial compilation to decrease the number of rules tried and applied.

PR #1605, opened 2015-08-08
Refactor parseString to return K.

bmmoore:
@dwightguth please review. This is just a small refactoring cleaning up something from #1571.

All callers immediately applied TreeNodesToKORE.apply,
so it's convenient to add it into parseString.

Also, a recent change to AmbFilter makes the result somewhat
unpredictable as a Term by picking an arbitrary resolution of
any ambiguity where the unequal Terms become equivalent
after applying TreeNodesToKORE.apply, so it's safer not
to expose the Term-returning version.


bmmoore:
Jenkins: retest this please.

PR #1620, opened 2015-08-13
several optimizations to startup time for OCAML backend

dwightguth:
@bmmoore @andreistefanescu please review.

Basically, I fixed the inefficiency in AbstractUnifier.flushTaskBuffer, added a definition-scoped cache for parses of patterns, and wrote code to parse K terms inside OCAML in order to decrease time spent compiling programs.

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/symbolic/AbstractUnifier.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/symbolic/AbstractUnifier.java
@@ -1,36 +1,15 @@
 // Copyright (c) 2013-2015 K Team. All Rights Reserved.
 package org.kframework.backend.java.symbolic;
 
-import org.kframework.backend.java.kil.Bottom;
-import org.kframework.backend.java.kil.BuiltinList;
-import org.kframework.backend.java.kil.BuiltinMap;
-import org.kframework.backend.java.kil.BuiltinSet;
-import org.kframework.backend.java.kil.CellCollection;
-import org.kframework.backend.java.kil.Hole;
-import org.kframework.backend.java.kil.InjectedKLabel;
-import org.kframework.backend.java.kil.KCollection;
-import org.kframework.backend.java.kil.KItem;
-import org.kframework.backend.java.kil.KLabelConstant;
-import org.kframework.backend.java.kil.KLabelInjection;
-import org.kframework.backend.java.kil.KList;
-import org.kframework.backend.java.kil.KSequence;
-import org.kframework.backend.java.kil.Kind;
-import org.kframework.backend.java.kil.Term;
-import org.kframework.backend.java.kil.TermContext;
-import org.kframework.backend.java.kil.Token;
-import org.kframework.backend.java.kil.Variable;
-import org.kframework.backend.java.util.Profiler;
-import org.kframework.utils.errorsystem.KEMException;
+import com.google.common.collect.Multimap;
+import org.apache.commons.lang3.tuple.Pair;
+import org.kframework.backend.java.kil.*;
+import org.kframework.backend.java.util.DoubleLinkedList;
 
-import java.util.ArrayDeque;
 import java.util.ArrayList;
-import java.util.Deque;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-
-import org.apache.commons.lang3.tuple.Pair;
-import com.google.common.collect.Multimap;
 
 
 /**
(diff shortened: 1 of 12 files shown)


java-backend/src/main/java/org/kframework/backend/java/symbolic/AbstractUnifier.java
```diff
@@ -68,8 +47,167 @@ protected void fail(Term term, Term otherTerm) {
         }
     }
 
-    private final Deque<Pair<Term, Term>> tasks = new ArrayDeque<>();
-    private final Deque<Pair<Term, Term>> taskBuffer = new ArrayDeque<>();
+    private final DoubleLinkedList<Pair<Term, Term>> tasks = new DoubleLinkedList<>();
+    private final DoubleLinkedList<Pair<Term, Term>> taskBuffer = new DoubleLinkedList<>();
+
+    private static class DoubleLinkedList<E> {
```
andreistefanescu: move this class in its own file

andreistefanescu:
lgtm

java-backend/src/main/java/org/kframework/backend/java/symbolic/AbstractUnifier.java
```diff
@@ -213,9 +192,7 @@ protected boolean unify() {
     }
 
     private void flushTaskBuffer() {
-        while (!taskBuffer.isEmpty()) {
```
bmmoore, line 216: Is this change necessary? Each item moves only once from taskBuffer to tasks, and ArrayDeque promises push and pop run in amortized constant time. It also provides an addAll method. Let's not copy and paste a whole collection class unless it's really a huge performance improvement.

bmmoore:
I'd like some measurement justifying the queue change, the rest looks good.

java-backend/src/main/java/org/kframework/backend/java/util/DoubleLinkedList.java
```diff
@@ -0,0 +1,173 @@
+// Copyright (c) 2015 K Team. All Rights Reserved.
```
bmmoore, line 1: You can't copy from the Java library and claim it's ours.

bmmoore:
This little test consistently shows your DoubleLinkedList running slower than the existing ArrayDeque code.
https://gist.github.com/bmmoore/c693e532aa934d4eb7d0

java-backend/src/main/java/org/kframework/backend/java/symbolic/AbstractUnifier.java
dwightguth, line 216: On the example I tested it reduced the time spent processing macros from 1600 ms to 1300 ms. It's not a dramatic change but still significant especially because this runs on every single krun invocation. I wasn't able to reproduce this effect on programs in the java backend itself, but it also wasn't noticably slower, either.

bmmoore:
That's probably okay, but I'd prefer we avoid defining a whole collection, and I'm still not convinced it's a performance improvement in general, as ArrayDeque docs claims significantly better constant factors than linked lists.

Instead, how about bootstraping by managing the task buffers with a queue of queues. I updated the gist linked above with an example of that.

bmmoore:
Also, can you find parameters that make the microbenchmark show a linked-list solution actually being faster than ArrayDeque?

bmmoore:
I updated the gist again with your latest class. Now I see results like this:

DoublyLinkedList 196665429 <- your linked list
ArrayDeque       124859368 <- the original code
ArrayDequeBoot   98433690 <- queue of queues, done directly in the loop
ArrayDequeBoot2  192451247 <- queue of queues, abstracted into a class
ArrayDequeBoot3  177424203 <- your latest code

So, obviously this benchmark isn't reproducing the distribution where you saw a speedup, but it does seem to show the array-based implementations being faster, and a pretty hefty abstraction penalty.

bmmoore:
This one finally consistently beats the old ArrayDeque code on my benchmark.

bmmoore:
Now I hear it doesn't work so well on real-world examples. Well, I'm fine with adding code like this nested queue stuff or with just reverting to the old array deque. Can you provide a reproducible benchmark showing advantage for whichever way you want to go?

Also, let's just rebase out the version with JDK-based code.

dwightguth:
the linkedlist example performs better than the list of queues on kool-collatz, and on par with the previous implementation. It also shows consistent gains of 300-400 ms on kcc tests/unit-pass/20041011-1.c in time spent in macro expansion. I am going to go with this implementation.

bmmoore:
I'd be a lot more comfortable if we had a microbenchmark where the new implementation isn't slower than the old one, but I guess it's okay to merge. As a final sanity check, let's make sure the ktest run is as fast as usual.

dwightguth:
Jenkins: test this please

PR #1623, opened 2015-08-14
rename the subsort instead of the supersort

andreistefanescu:
@daejunpark this fixes a bug in the previous commit; please review


daejunpark:
lgtm

PR #1625, opened 2015-08-14
add MINT in domains.k

daejunpark:
@dwightguth @andreistefanescu  please review.

FYI, I added `[bracket]` in INT module, suggested by Grigore in https://github.com/kframework/wasm-semantics/pull/2

> It should be OK to use parentheses here, because they are a bracket for Int, aren't they? If they are not, then we should add them. We want to use the backticks for grouping of K code to help the parser, and not as brackets in the object languages.


daejunpark:
Jenkins: test this please

dwightguth:
Lgtm

daejunpark:
Jenkins: test this please.

daejunpark:
Hmm, it doesn't crash with segfault any longer once I dropped the bracket.

PR #1630, opened 2015-08-18
Add regression test for wasm semantics

daejunpark:


dwightguth:
I feel kind ambivalent about having this code in two places. If you want regression testing, can't you just configure it to kick off a build after every build of k-framework-master, like we do for the other semantics?

daejunpark:
Close this via https://github.com/kframework/wasm-semantics/pull/12

PR #1631, opened 2015-08-18
Make K_OPTS extend defaults

bmmoore:
Changes the master k script so setting the K_OPT environment
variable does not completely remove K's default JVM options,
but instead lets settings in K_OPT override individual defaults.

This was noticed while working on the C semantics, where the makefile setting K_OPTS to increase heap size lost the default 32m stack size, leading to stack overflow errors.

@dwightguth please review.


dwightguth:
merge if tests pass

PR #1632, opened 2015-08-18
Simplify follow regex construction

bmmoore:
This changes Terminal to take the follow pattern as a list of strings
rather than a single regular expression, and to use the specialized
method `BasicAutomaton.makeStringUnion` to build the matcher,
rather than escaping the patterns to make regexps which match
only that string, appending them together in quadratic time, and
passing it through the RegExp parser which needs linear space
(which was leading to stack overflows trying to build the C semantics
under default JVM options).

@radumereuta, @cos please review (@cos especially for how Down/Up handles that Seq - it seems to work but I don't know if the Product case handles it well).

### DIFF
--- a/kernel/src/main/java/org/kframework/parser/concrete2kore/generator/RuleGrammarGenerator.java
+++ b/kernel/src/main/java/org/kframework/parser/concrete2kore/generator/RuleGrammarGenerator.java
@@ -214,8 +214,7 @@ public class RuleGrammarGenerator {
                                     });
                             // add follow restrictions for the characters that might produce ambiguities
                             if (!follow.isEmpty()) {
-                                String restriction = follow.stream().map(StringUtil::escapeAutomatonRegex).reduce((s1, s2) -> "(" + s1 + ")|(" + s2 + ")").get();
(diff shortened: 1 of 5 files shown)


dwightguth:
What impact if any does this have on performance?

bmmoore:
I haven't noticed any performance impact. Just C compiling consistently at default stack size or even `-Xss512k`.

dwightguth:
Okay, cool. looks good to me but feel free to get the other reviews you were looking for.

bmmoore:
@radumereuta It might be possible to share a single set of states for all auto follow restrictions and just set different start states for different tokens, but that seems to go beyond what the public interface of the dk.brics.automaton package allows.

radumereuta:
talked with @bmmoore and lgtm.

cos:
Up/Down should handle the Seq just fine. Not sure if there's a test for it, but if it'll fail some test in the feature, I'll fix Up/Down as necessary. The PL lgtm.

bmmoore:
Up/Down does work, I'm just not sure treating a Seq as a Product makes a nice flat term, or maybe adds unnecessary nested layers.

Issue #1634, opened 2015-08-18
JVM Segfault in libz3


bmmoore:
JVM crashes like this seem to be getting more common. It happened three times in a row trying to build a recent pull request

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00002b4cd3730106, pid=810, tid=47609081427712
#
# JRE version: Java(TM) SE Runtime Environment (8.0_45-b14) (build 1.8.0_45-b14)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.45-b02 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# C  [libz3.so+0xf09106]  scoped_timer::imp::sig_handler(sigval)+0x6
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
#
```

Full logs at

http://office.runtimeverification.com:8080/rv-jenkins/job/k-framework-pull-requests/1727/artifact/k-distribution/hs_err_pid810.log

http://office.runtimeverification.com:8080/rv-jenkins/job/k-framework-pull-requests/1726/artifact/k-distribution/hs_err_pid20930.log

http://office.runtimeverification.com:8080/rv-jenkins/job/k-framework-pull-requests/1726/artifact/k-distribution/hs_err_pid20930.log
(this one doesn't recognize the stack frame, but register values look like we're in Z3)

@andreistefanescu, IIRC you built our libz3. Can you figure out what's causing this?

andreistefanescu:
This is a known bug in libz3, caused by a race between the timeout handle and the some cleanup code. You can use `--z3-executable` which creates a new process instead of the jni.

bmmoore:
I can, but that doesn't help in Jenkins. Looks like #1633 has failed on this four times in a row now.

bmmoore:
Please link to the upstream issue.

bmmoore:
@dwightguth is there any way to work around this in Jenkins?

dwightguth:
@andreistefanescu this is causing too much disruption, it's time to fix the underlying cause. We need to either get the z3 guys to fix the segfault, or else remove all calls to the library from ktest and make the executable the default.

andreistefanescu:
[...]

PR #1635, opened 2015-08-18
Revert "Kore Based Rewriter API's Search Functionality Related Code"

msaxena2:
Reverts kframework/k#1555.


bmmoore:
Why was this reverted?

msaxena2:
Looks like the PR may have been the reason behind builds issues on the Jenkins server. I don't understand why it passed in the first place though. In any case, I'll make a new PR, and track down the issue, and then re-merge.

bmmoore:
I think the biggest issue is #1634, which doesn't seem to have improved since you reverted (and don't think this code could possbily have triggered that).

PR #1636, opened 2015-08-18
Sort cells rewrites

bmmoore:
The SortCells pass did not recurse to find cell rewrites hidden under
labels such as "#cells", and was thus incorrectly computing the sets of
neighboring cells of some cell fragment variables.
This fixes that. It also simplifies the logic by adding
a small pass to lift any rewrite inside a non-leaf cell to occur
as an immediate child of the parent cell.
It also slightly cleans how configuration information is passed
to the sub-passes inside the ConcretizeCells constructor.

### DIFF
--- a/kernel/src/main/java/org/kframework/kore/compile/SortCells.java
+++ b/kernel/src/main/java/org/kframework/kore/compile/SortCells.java
@@ -31,6 +31,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static org.kframework.Collections.*;
 import static org.kframework.definition.Constructors.*;
@@ -55,7 +56,6 @@ import static org.kframework.kore.KORE.*;
  * expression involving the split variables, using the cell fragment productions
  * introduced along with the cell labels.
  */
-// TODO handle cell rewrites
 public class SortCells {
     private final ConcretizationInfo cfg;
     private final KExceptionManager kem;
@@ -116,7 +116,7 @@ public class SortCells {
             if (parentCell == null) {
                 parentCell = cell;
             } else if (!parentCell.equals(cell)) {
-                throw KEMException.criticalError("Cell variable used under two cells, "
+                throw KEMException.criticalError("Cell variable "+var+" used under two cells, "
                         + parentCell + " and " + cell);
             }
             if (remainingCells == null) {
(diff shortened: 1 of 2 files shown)


bmmoore:
@cos, @dwightguth, please review.

kernel/src/main/java/org/kframework/kore/compile/LiftCellRewrites.java
```diff
@@ -0,0 +1,87 @@
...
+ * of that cell's klabel.
+ *
+ * This is done to simplify SortCells.
+ *
+ * This must be called after CloseCells and ResolveAnonymousVariables,
+ * otherwise a single "_" or "..." occuring outside any rewrite might
+ * be split into two independent variables instead of representing a
+ * single term which is not modified by a rule.
+ */
+public class LiftCellRewrites extends TransformKORE {
```
dwightguth: does this pass have any effect on where the rewrites will be in the code which results from the SortCells pass? or will the rewrites still be in the same place they were before?

bmmoore:
This could affect rewrites in cases like

```
<parent>... ``<cell>Stuff</cell> `<thread>A</thread> => <thread>B</thread><thread>C</thread>` ``
```

I suppose it would be better not to unnecessarily raise rewrites. (though I think most examples would have failed on an NPE before this patch anyway).

dwightguth:
I would prefer we not unnecessarily raise rewrites, yes. The design of the new compiler is to make the translation passes be as modular and independent from one another as possible; you are introducing a dependency here.

How much work would it be to keep the rewrites the way they are, but still fix the errors that you encountered?

bmmoore:
I replaced the branch. Turns out there was no real cost -  now SortCells always traverses parent cell contents twice, for left and right of any potential rewrites, but it avoids the earlier pass. Do note we already have internal dependencies among the stages of ConcretizeCells, like whether cells are represented with `#dots()`, or are expected to be sorted.

bmmoore:
Jenkins: test this please.

bmmoore:
Incidentally, 86996a8040afd97b646b852a6fbac4aaabca46e4 has a small diff from 7ef302bb62e2d0980dce8fb9662f25ec4df71226, I rebased mostly to avoid ever introducing an extra file or touching ConcretizeCells.

bmmoore:
Jenkins: retest this please.

cos:
lgtm.

Issue #1574, opened 2015-07-30
[kore] KLabelInjection not supported


daejunpark:
`org.kframework.backend.java.kil.KLabelInjection` is not supported in KORE pipeline.
See #1567 and f4cf76717429878438694d52ae92db46765ebea4

PR #1639, opened 2015-08-19
added support for cells with multiplicity=* in the kore to java-backe…

andreistefanescu:
…nd translation

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/compile/KOREtoBackendKIL.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/compile/KOREtoBackendKIL.java
@@ -126,7 +126,7 @@ public class KOREtoBackendKIL extends org.kframework.kore.AbstractConstructors<o
         Assoc.flatten(klabel, klist.items(), module).stream().forEach(k -> {
             if (k instanceof KApply) {
                 builder.put(
-                        kLabel2cellLabel(((KApply) k).klabel().name()),
+                        CellLabel.of(((KApply) k).klabel().name()),
                         KList(((KApply) k).klist().items()));
             } else if (k instanceof KVariable) {
                 // TODO(AndreiS): ensure the ... variables do not have sort K
@@ -150,9 +150,9 @@ public class KOREtoBackendKIL extends org.kframework.kore.AbstractConstructors<o
             Option<Att> attOption = module.attributesFor().get(klabel);
             if (attOption.isDefined() && attOption.get().contains(Attribute.CELL_BAG))
                 return KLabelInjection.injectionOf(CellCollection(klabel, klist), context);
-            else if (definition.cellMultiplicity(kLabel2cellLabel(klabel.name())) == ConfigurationInfo.Multiplicity.STAR)
+            else if (definition.cellMultiplicity(CellLabel.of(klabel.name())) == ConfigurationInfo.Multiplicity.STAR)
                 return KLabelInjection.injectionOf(
-                        CellCollection.builder(definition).put(kLabel2cellLabel(klabel.name()), KList(klist.items())).build(),
+                        CellCollection.builder(definition).put(CellLabel.of(klabel.name()), KList(klist.items())).build(),
                         context);
             else
                 return KApply1(klabel, klist, k.att());
@@ -201,10 +201,6 @@ public class KOREtoBackendKIL extends org.kframework.kore.AbstractConstructors<o
 
     }
 
-
-    public static CellLabel kLabel2cellLabel(String kLabel) {
-        return CellLabel.of(kLabel.substring(1, kLabel.length() - 1));
-    }
 
     public static ConfigurationInfo.Multiplicity kil2koreMultiplicity(Cell.Multiplicity multiplicity) {
         switch (multiplicity) {
--- a/java-backend/src/main/java/org/kframework/backend/java/compile/KOREtoBackendKIL.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/compile/KOREtoBackendKIL.java
@@ -5,10 +5,14 @@ package org.kframework.backend.java.compile;
 import org.kframework.attributes.Att;
 import org.kframework.attributes.Location;
 import org.kframework.attributes.Source;
+import org.kframework.backend.java.kil.CellCollection;
+import org.kframework.backend.java.kil.CellLabel;
+import org.kframework.backend.java.kil.Definition;
 import org.kframework.backend.java.kil.InjectedKLabel;
 import org.kframework.backend.java.kil.KCollection;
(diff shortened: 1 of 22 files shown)


andreistefanescu:
@bmmoore please review

andreistefanescu:
@daejunpark @dwightguth please review

java-backend/src/main/java/org/kframework/backend/java/compile/KOREtoBackendKIL.java
```diff
@@ -160,4 +200,24 @@ public Rule convert(Optional<Module> module, org.kframework.definition.Rule rule
                 context);
 
     }
+
+
+    public static CellLabel kLabel2cellLabel(String kLabel) {
+        return CellLabel.of(kLabel.substring(1, kLabel.length() - 1));
```
dwightguth: do you know for sure that the code will break if you don't have this? Obviously it looks a little funnier, but not all cells are technically guaranteed to have cell labels beginnning with `<` and ending with `>`.

kernel/src/main/java/org/kframework/kore/compile/GenerateSentencesFromConfigDecl.java
```diff
@@ -348,6 +348,7 @@ public K apply(KToken k) {
                 bagAtt = bagAtt.add(Attribute.IDEMPOTENT_KEY, "");
             case "Bag":
                 bagAtt = bagAtt.add(Attribute.COMMUTATIVE_KEY, "");
+                bagAtt = bagAtt.add(Attribute.CELL_BAG, "");
```
dwightguth, line 351: I personally don't have any objection to this attribute but I believe @cos didn't like it when I added it originally which is why I went around it the way that I did. Discussion?

dwightguth:
looks okay but I haven't dug too deep into it. I assume you've tried some KORE definition with multiplicity \* on it?

kernel/src/main/java/org/kframework/kore/compile/GenerateSentencesFromConfigDecl.java
bmmoore, line 351: This pull request works by turning cell collections into the existing `CellCollection` class of the backend, and needed some way to recognize which sorts are actually the bags used in the configuration. I suppose we could compute that somehow, but with an actual use case adding an attribute seems reasonable.

java-backend/src/main/java/org/kframework/backend/java/compile/KOREtoBackendKIL.java
andreistefanescu: @dwightguth it seems you are right, thanks for the suggestion!

andreistefanescu:
I tried it on the kore version of imp++

bmmoore:
Execution may be improved, but something is broken trying to print terms now:

```
$ krun --kore --debugger --smt none --parser "kast --kore -s Stmts" ../../../1_k/2_imp/lesson_1/tests/sum.imp
KDebug> p
initTCell ( .Map $STDIN |-> "
" $IO |-> "off" $PGM |-> ` int n , ` sum , .Ids ` ;  ` ( n = 100 ) ;  ` ( sum = 0 ) ;  ` ` while ( ! ` n <= 0 ` ) { ( sum = sum + n ) ;  ` ( n = n + -1 ) ;  .Stmts ` } `  .Stmts ` ` ` ` )

KDebug> s
1 Step(s) Taken.
KDebug> p
[Error] Internal: Uncaught exception thrown of type ClassCastException.
Please rerun your program with the --debug flag to generate a stack trace, and
file a bug report at https://github.com/kframework/k/issues
```

daejunpark:
@andreistefanescu Returning `KLabelInjection` might be problematic, until #1574 is fixed, right? I suspect @bmmoore's error is due to that.

kernel/src/main/java/org/kframework/kore/compile/GenerateSentencesFromConfigDecl.java
cos, line 351: If the backend would support general bags (not just for cells), would this still be needed?

dwightguth, line 351: Likely not but there are matching patterns supported on cells right now that are not supported for e.g. maps and sets.

andreistefanescu, line 351: The backend represents bags of cells as multimaps from cell names to cell contents. We could have an `key` attribute for indexing in addition to `assoc, comm, id`, and then the solution would be general.

grosu, line 351: But why do we even have "bags of cells" at all?  According to https://github.com/kframework/k/wiki/KAST-and-KORE#cells-configurations-and-configuration-abstraction, once we get into KORE, all we need is an attribute `cell` to the productions defining cells.  Then based on that, the kompiler should do its configuration abstraction, which includes adding the missing cells, sorting them, etc.  "Bags of cells" are monsters of the past.  I'd like to remind you that we should keep the implementation and wiki in synch, otherwise we end up misscommunicating and having a hard time putting together the manual.  If for some reason you decide to depart in the implementation from the design described in the wiki, then please bring the issue up for discussion, and then we update the wiki, if needed.

andreistefanescu:
@bmmoore I fixed the issue you reported.

kernel/src/main/java/org/kframework/kore/compile/GenerateSentencesFromConfigDecl.java
andreistefanescu, line 351: This pull request addresses the java backend internal representation of cells with `multiplicity=*`. The extra attribute helps identify such cells. There are of course other ways of doing that, and I chose the simplest one.

andreistefanescu:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

andreistefanescu:
@bmmoore @daejunpark I added the conversion from `CellCollection` to kore and fixed the race in issue #1658.

bmmoore:
I think we should merge this. I've tested that it didn't break the kore version of IMP, and it clearly improves the situation on IMP++.

I can still raise some exceptions, most worryingly when spawn.imp tries to actually make a thread, but I had to fix some problems my kore IMP++ first, and it's clearly doing a lot better than it did before.
Let's get this into master so it can benefit from other things like the parsing fixes, and I'll file separate issues.

bmmoore:
Commenting in case you didn't get a notification that this was reverted.

PR #1641, opened 2015-08-19
move native libraries code to kernel + add functionality for it to oc…

dwightguth:
…aml backend

@bmmoore please review

Basically if you pass --native-libraries "pthread" to krun then the ocaml backend will use -lpthreads during the linker.


bmmoore:
It looks reasonable, but when is this needed? Doesn't ocamlfind also provide linker options for libraries with a native component?

dwightguth:
This  is used indirectly by the proprietary native linker code for the c semantics, which needs to make sure that the binary is linked against any native libraries the user specifies with -l to kcc.

bmmoore:
Ok, that's good. This shouldn't ever be necessary for O'Caml libraries, at worst we may need to pass `-dllpath-all` to ocamlfind.

PR #1642, opened 2015-08-20
Debugger Watch and Match Capabilities

msaxena2:
Not ready for review yet. 

Gives the Debugger capability to add watches.


java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
```diff
@@ -116,9 +116,10 @@ public RewriterResult execute(K k, Optional<Integer> depth) {
 
         @Override
         public List<Map<KVariable, K>> match(K k, org.kframework.definition.Rule rule) {
-            throw new UnsupportedOperationException();
+            return (List<Map<KVariable, K>>) search(k, Optional.of(0), Optional.empty(), rule);
```
dwightguth: I suggest instead to make it be a List<? extends Map<? extends KVariabale, ? extends K>> because technically this is not a safe cast.

kernel/src/main/java/org/kframework/debugger/DebuggerState.java
```diff
@@ -23,18 +25,22 @@
 
     private final NavigableMap<Integer, K> checkpointMap;
 
+    private final List<DebuggerMatchResult> watchList;
```
dwightguth, line 30: I still want these two fields unmodifiable from the outside; we skipped over that part last time so can you get to it now?

dwightguth:
I have not gone into it in depth but the design looks okay to me. I will provide more in depth feedback tomorrow.

kernel/src/main/java/org/kframework/kompile/CompiledDefinition.java
```diff
@@ -97,12 +100,18 @@ public Module getExtensionModule(Module module) {
     }
 
     public Rule compilePatternIfAbsent(FileUtil files, KExceptionManager kem, String pattern, Source source) {
```
cos, line 102: If I understand this correctly, the public `compilePatternIfAbsent` and `parsePatternIfAbsent` essentially allow a compiled definition to have rules which have distinct `FileUtil`s and `KExceptionManager`s. Furthermore, as the patterns are only cached by the pattern `String`, when asking for a pattern with a specific `FileUtil`, `KExceptionManager`, and `Source`, you may get back a cached rule that has different ones.

dwightguth, line 102: So, FileUtil and KExceptionManager are used to parse the rule but not
stored as part of the result, nor will the same string parse differently
depending on their values. Thus the cache is safe with respect to those
parameters. If you are running a server and you parse something that has
warnings, you will only get the warnings once, but that's not a big deal
and is an acceptable price to pay for the performance gains this provides.
You are correct that the Source should be part of the cache key though; if
you make an issue and assign it to me, I will get to this after this pull
request is merged.

cos, line 102: added #1648

kernel/src/main/java/org/kframework/debugger/DebuggerState.java
msaxena2, line 30: done

msaxena2:
@dwightguth Please review.

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -400,7 +400,7 @@ private boolean addSearchResult(
         if (depth == 0) {
             addSearchResult(searchResults, initCnstrTerm, pattern, bound);
             stopwatch.stop();
-            System.err.println("[" + visited.size() + "states, " + step + "steps, " + stopwatch + "]");
```
dwightguth, line 403: actually can you make this run if the --statistics flag is set?

kernel/src/main/java/org/kframework/debugger/KoreKDebug.java
```diff
@@ -173,4 +238,48 @@ public DebuggerState createCopy(int stateNum) {
     public DebuggerState getActiveState() {
         return stateList.get(activeStateIndex);
     }
+
+    @Override
+    public DebuggerMatchResult match(String pattern) {
+        String DebuggerSource = "Debugger TUI";
```
dwightguth: `<command line>`

kernel/src/main/java/org/kframework/krun/modes/DebugMode/DebugExecutionMode.java
```diff
@@ -61,41 +65,81 @@ private ConsoleReader getConsoleReader() {
...
     public Void execute(K k, Rewriter rewriter, CompiledDefinition compiledDefinition) {
-        KDebug debugger = new KoreKDebug(k, rewriter, checkpointInterval);
-        ConsoleReader reader = getConsoleReader();
+        KDebug debugger = new KoreKDebug(k, rewriter, checkpointInterval, files, kem, kRunOptions, compiledDefinition);
 
+        ConsoleReader reader = getConsoleReader();
         while (true) {
             try {
                 String input = reader.readLine("KDebug> ");
+                if (input.isEmpty()) {
```
dwightguth, line 97: is this handling the ^D case? ^D should exit the debugger, whereas ^C should eventually terminate the current command and print a new prompt.

dwightguth:
I don't see anything obviously wrong but I also haven't dug super into this. Can you provide me with a transcript of a debugger session that tests all the new features you added?

msaxena2:
Jenkins: test this please

dwightguth:
lgtm

PR #1643, opened 2015-08-20
update documentation in preparation for K 3.6 release

dwightguth:
@grosu please review, let me know if you think of anything else that should be in the changelog for K 3.6.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,11 +11,42 @@
   Java backend is deemed a suitable complete replacement for maude.
 - Removed the `K_NAILGUN` environment variable: K now uses nailgun whenever
   it is able to connect to a K server.
+- Added an option `-w2e` to convert all warnings that are not suppressed by the current warning level into errors.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -11,11 +11,42 @@
...
+
+- Users may now use both "requires" and "require", and both "imports" and "import".
+- Non-anonymous variables may now contain underscore characters.
+- Added an "initial" cell attribute whose effect is to ensure that a multiplicity ? or *
+  cell is part of the initial configuration even if it contains no configuration variables.
+
+## Kompile ##
+
+- The `--directory` option now defaults to a value equal to the directory containing
+  the main definition file, rather than the current working directory.
```
grosu, line 29: I don't remember why we had it the other way before, but I do remember that there was a discussion about it and then a decision has been made.  When we have flexibility, we typically follow the Java approach.

grosu:
LGTM, but I'm assuming that you have added all the major changes.  It may not be a bad idea to ask the k-list developers if they want to add anything they've done to the changelog.

PR #1644, opened 2015-08-20
fix random number generation in ocaml backend

dwightguth:
@bmmoore please review

Note that we probably eventually want the random function to take a minimum and maximum value as well, but this is basically as good as it worked in the past.

### DIFF
--- a/k-distribution/include/ocaml/prelude.ml
+++ b/k-distribution/include/ocaml/prelude.ml
@@ -414,6 +414,8 @@ struct
   let hook_rfindChar c lbl sort config ff = raise Not_implemented
 end
 
+open Gmp.RNG
+
 module INT =
 struct
   let hook_tmod c lbl sort config ff = match c with
@@ -478,6 +480,11 @@ struct
     | _ -> raise Not_implemented
   let hook_min c lbl sort config ff = match c with
       [Int a], [Int b] -> [Int (Z.min a b)]
+    | _ -> raise Not_implemented
+  let hook_random c lbl sort config ff = match c with
+      [Int seed] -> Random.init (Z.to_int seed);
+          let i64 = Random.int64 (Int64.max_int) in
+          [Int (Z.of_int64 i64)]
     | _ -> raise Not_implemented
   let hook_ediv c lbl sort config ff = raise Not_implemented
   let hook_emod c lbl sort config ff = raise Not_implemented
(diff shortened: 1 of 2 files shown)


dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

bmmoore:
It looks very strange to be reinitializing the random number generator at every call.

bmmoore:
As it is, this hook is a deterministic function of the Int argument. I don't think a fixed somewhat arbitrary `Int -> Int` function is a good thing to name random.

Maybe some of the functions here would be useful?
http://caml.inria.fr/pub/docs/manual-ocaml/libref/Random.State.html

dwightguth:
Okay. Since this is going to be part of the 4.0 release anyway, I will fix this so that it has a much better interface.

dwightguth:
@bmmoore please re-review

Note that this diff will require an update to the MLGMP module to be installed because randseed wasn't part of the module before.

k-distribution/include/ocaml/prelude.ml
```diff
@@ -479,6 +481,12 @@ struct
   let hook_min c lbl sort config ff = match c with
       [Int a], [Int b] -> [Int (Z.min a b)]
     | _ -> raise Not_implemented
+  let hook_rand c lbl sort config ff = match c with
+      [Int max] -> let mpz = Gmp.Z.urandomm Gmp.RNG.default (from_zarith max) in
+          [Int (to_zarith mpz)]
+    | _ -> raise Not_implemented
+  let hook_srand c lbl sort config ff = match c with
+      [Int seed] -> let () = Gmp.Z.randseed Gmp.RNG.default (from_zarith seed) in []
```
daejunpark, line 489: @dwightguth Why didn't you provide the any-value-matching case for `hook_srand`?

dwightguth, line 489: oh, mistake. Thanks, I will fix that.

dwightguth:
@bmmoore please re-review

bmmoore:
I'd like to discuss the design more, but this code looks good for what it is and you can merge, especially if you are using it.

In particular, using global state like this obviously makes debugging and search and such more difficult, but then again I'm not sure anything other than debugging and simple execution make sense on a definition that calls a random number generator (or at least a definition that's happy to get some arbitrary RNG from our library rather than explicitly defining a particular generator as part of the system). Also, by switching to GMP and leaving the default algorithm of Mersenne Twister you've made the state quite large and expensive to copy. When I suggested Random.State I definitely thinking of also changing the signature of the hooks to explicitly pass around RNG state values, but I'm not sure if that's the best design.

PR #1645, opened 2015-08-21
remove test that is broken on windows

dwightguth:
@cos please review

I would prefer to fix the test, but unfortunately it's basically a broken test design and I have no idea how to do that. If you have suggestions, let me know, otherwise, I would like to point out that this test is calling an entirely deprecated parsing stack.


cos:
If I remember correctly, it had a path problem. I agree we can comment it for now. Could you add an issue for me, also pasting the exception?

dwightguth:
See #1649

PR #1650, opened 2015-08-21
try to fix frequent jvm segfaults

dwightguth:
**do not review yet**


dwightguth:
Jenkins: test this please

dwightguth:
I am merging this since it seems to fix the frequent test failures in master that we have had.

@andreistefanescu if you want to add back this test you will have to solve the segfault issue. I tried just changing it to --z3-executable but that apparently causes the test to fail. I think you should seriously consider whether you want to keep spending effort writing code that only works for a mode of verification that can cause an extremely simple test like this to fail over 50% of the time.

PR #1651, opened 2015-08-21
propagate counter between preprocessing and verification

andreistefanescu:
@daejunpark please review


daejunpark:
lgtm once you add 1 to the last counter value.

PR #1652, opened 2015-08-22
Kil kill

dwightguth:
@andreistefanescu @cos @bmmoore @radumereuta @daejunpark @grosu please review
### things I deleted
- sdf parser
- old krun, kompile, kast frontends
- old compiler
- old debugger, executor, model checker, prover frontends
- old unparser
- unparser and unflatten backend
- `--kore` flag (now the default)
- dead code arising from deletion of the above
### things I did not delete
- html, latex, pdf, doc, coq backends
- checks that need to be rewritten for kore
- kil hierarchy
### things I moved
- various static things that were part of a class I deleted but used by the new pipeline
- various packages with kore in their name to a better package name
### things I commented out
- most of ktest. unfortunately we're still not in a good shape to pass any of the tests in ktest, even imp, mostly because the unparser isn't done yet. Once we implement a collection normalization filter and the format attribute it should pass though.

The diff looks pretty intimidating but it's actually much less bad than it seems. Most of it is just deleted files and small renames to packages. The remainder is all grouped in a single commit that you should be able to easily review separately, and is also mostly deletions, but at the less-than-file level. That said, you may wish to review the first commit on the command line just because github truncated the diff and it's actually moderately distracting to see the full body of every file that is deleted in the diff.

If we need to bring anything back, we can always do that; I tested the code which remains against IMP, the KORE tests in ktest, and the C semantics and they all passed; I don't think anything got deleted that was part of the KORE pipeline.


grosu:
Wow, this is big!  Many will be affected, but it had to happen, otherwise we would continue to drag our feet.  This way we know that no time is being wasted on KIL, and developers who want functionality not present in KORE have motivation to make that functionality work.  I rely on @andreistefanescu , @cos, @bmmoore , and @daejunpark for checking the technical details of the pull request.

@bmmoore , I see that the tutorial tests have been deactivated.  I understand why, but you had one or two tests that worked in your branch.  Can you please add those back?

@dwightguth , was it necessary to remove most of ktest?  I think its functionality should be orthogonal to KIL vs. KORE.  Yes, we do not have a pretty-printer yet, I know, but we could use the labeled form until we get one.  Actually, we should get a pretty-printer asap.

Otherwise it all LGTM.

dwightguth:
I have no particular objection to us modifying temporarily the outputs of krun for the tutorial to give us better test coverage. Given that the test coverage is severely limited by this diff, I think that's probably a good idea. But I think that that is better done by Brandon in a followup pull request than as part of this giant monster.

cos:
lgtm. I see you also renamed several packages. I'm ok with the changes.

bmmoore:
With #1575 still open, removing SDF parsing may be premature. My only other concern is being able to run old definitions, which is probably handled well enough by keeping independent K installations.

dwightguth:
Closing since we agreed to do this later.

PR #1654, opened 2015-08-24
fix bug in OCAML imp + make ocaml install much easier

dwightguth:
@andreistefanescu @bmmoore please review


andreistefanescu:
lgtm

daejunpark:
@dwightguth Just for a minor note. Don't we need to support Windows? OPAM isn't easy to be installed in Windows yet.

grosu:
That is not minor, many people only use Windows these days, including myself.  Also, our Japanese collaborators at MSYS seem to prefer Windows.  So we need to make it easy to work with Windows, too.

dwightguth:
@grosu I will look into the install process for OCAML on Windows but I suspect we are going to have to tell people to install Cygwin or MinGW. In any event, the deadline the MSYS guys gave us for figuring this out is the 10th so I will have something installable on Windows by then.

PR #1657, opened 2015-08-26
added support for non-deterministic lookups

andreistefanescu:
@daejunpark please review


daejunpark:
for a record, by this change, this term `(foo(0) |-> 0, foo(1) |-> 1)[bar(X:Int)]` will be split in two terms: `0 if foo(0) = bar(X)` and `1 if foo(1) = bar(X)`.

daejunpark:
lgtm

PR #1663, opened 2015-08-28
Automatic follow restriction should not be calculated for empty strings.

radumereuta:
Fixing #1575. 
@andreistefanescu or @daejunpark please review.
@bmmoore was right. It was related to automatic follow restriction.
What delayed me from seeing this, is that there is no representation in KORE for certain parts of the grammar. I think we got carried away again with some over optimization.


andreistefanescu:
lgtm

radumereuta:
@andreistefanescu please review the unit test.

PR #1664, opened 2015-08-28
fix comparison between multiple patterns issue

daejunpark:
when multiple patterns exist, comparison should be made with considering their klabels as well as their inputs.


andreistefanescu:
ups... I forgot about that

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicUnifier.java
```diff
@@ -388,7 +388,8 @@ private void unifyMap(BuiltinMap map, BuiltinMap otherMap) {
         List<KItem> otherRemainingPatterns = new ArrayList<>();
         for (KItem pattern : patterns) {
             for (KItem otherPattern : otherPatterns) {
-                if (pattern.getPatternInput().equals(otherPattern.getPatternInput())) {
+                if (pattern.getPatternInput().equals(otherPattern.getPatternInput())
```
andreistefanescu: I would do the `KLabel` check first

daejunpark:
@andreistefanescu please review again

andreistefanescu:
lgtm

PR #1677, opened 2015-08-31
Revert "added support for cells with multiplicity=* in the kore to ja…

dwightguth:
Reverts kframework/k#1639

This seems to have broken ExpandMacros for the OCAML backend.

### DIFF
--- a/java-backend/pom.xml
+++ b/java-backend/pom.xml
@@ -36,7 +36,7 @@
     <dependency>
       <groupId>uk.com.robust-it</groupId>
       <artifactId>cloning</artifactId>
-      <version>1.9.2</version>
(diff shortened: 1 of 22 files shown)


bmmoore:
Reverting that definitely breaks Kore execution. Did you update any bugs to attempt to notify @andreistefanescu  or I that you were doing this? The not of this revert on the pull request doesn't seem to have been enough to generate a notification.

dwightguth:
Yeah, I informed Andrei that I was doing this.

bmmoore:
Ok, I suppose that's not unreasonable. This does interfere with my attempts to work on Kore as well.

PR #1680, opened 2015-08-31
Tutorial kore

bmmoore:
Let's merge this now that we've made a 3.6 release.

This adds a tutorial-kore directory, with an updated vesion of 2_imp/lesson_5 that seems to entirely work, and a version of 4_imp++/lesson_8 which can at least run the simple sum.imp under the --smt none option.


dwightguth:
I do not at all like the idea of maintaining separate source trees. That seems like a great way for things to become out of sync to me. What we should do is maintain a single source tree and iteratively work on examples until they pass the ktest tests.

bmmoore:
How do you suggest maintaining the ktest configurations?

dwightguth:
If our primary goal is to convert the tutorial, then just keep the existing ktest configurations, comment them out, and gradually reintroduce them as they start passing.

Issue #1668, opened 2015-08-31
Finalize streams I/O


dwightguth:
There is still some issue of the compiler pass for streams IO which adds #parseInput to the stdin cell.

bmmoore:
We are also planning to change the syntax back to use an attribute, rather than a reserved cell name.

daejunpark:
@dwightguth Do you agree with this change of syntax back to the original style? I just wanted to be confirmed before implementing it.

bmmoore:
Specifically, we discussed this in the K meeting on Monday, and couldn't come up with any compelling arguments for breaking backwards compatibility.

dwightguth:
I mean, I did it the way that I chose to in order to make it simpler to implement because it means that you don't have to keep track of which cells have streams in them and then run a compiler pass to generate the stream rules from the IO module. I don't particularly have any objection to reimplementing all that junk because it doesn't really affect me one way or the other, but this solution is a lot cleaner and requires much more minimal intervention from the compiler.

grosu:
Why is the current solution "a lot cleaner"?  We want to be able to stream data to cells in K in the long term.  Your solution, @dwightguth, does not get us any closer to that.  Unless you or anybody else comes up with a better design, we stick to what we have now.

As a general rule of thumb, from here on I want to be part of the loop when changes are being made to the design of K that require changing the tutorial and the movies.

grosu:
@daejunpark , can you please give us an ETA on this?  We do not want to modify the definitions to work with `<stdin/>`, etc., only to have to change them back in a few hours or days.  If the problem is difficult, please let us know.  But the point is that, for now, you should not do anything conceptually new or different from what we had.

daejunpark:
I'm hoping to finish today or early next week. I'll get the temporary modification back to the original once I fix this.

grosu:
@daejunpark so it looks like we are done with this, right?  Is there a way to mark this task as done in the milestone, without removing it?

daejunpark:
Yes, we are done with this. I have no idea how to mark it in such a way.

bmmoore:
If we just close this issue, it will show up under the "Closed" tab on the milestone's page.

daejunpark:
Closed via #1716

PR #1684, opened 2015-09-01
Tutorial kore

bmmoore:
This startus updating the tutorial to use kore. IMP works, up to formatting changes in expected output.

There are various execution problems with IMP++ and lambda++.

### DIFF
--- a/k-distribution/include/builtin/domains.k
+++ b/k-distribution/include/builtin/domains.k
@@ -1,4 +1,27 @@
 // Copyright (c) 2015 K Team. All Rights Reserved.
+module DOMAINS-COMMON
+  imports EMPTY-ID
+  imports INT-SYNTAX
+  imports BOOL-SYNTAX
+  imports STRING-SYNTAX
+endmodule
+
+module DOMAINS-SYNTAX
+  imports DOMAINS-COMMON
+  imports ID
+endmodule
+
+module DOMAINS
+  imports INT
+  imports BOOL
+  imports STRING
+  imports K
+  imports K-IO
+  imports LIST
+  imports MAP
+  imports SET
+endmodule
+
 module MAP
   imports LIST
   imports SET
--- a/k-distribution/include/builtin/domains.k
+++ b/k-distribution/include/builtin/domains.k
@@ -39,6 +39,7 @@ module MAP
   syntax Map ::= ".Map"                         [function, hook(MAP.unit), latex(\dotCt{Map})]
                // breaks klabel uniqueness
                //| "."                            [function, hook(MAP.unit)]
+  syntax Map ::= "(" Map ")" [bracket]
   /*@ Construct a singleton Map (a Map with only one key/value pair). The key
    is on the left and the value is on the right */
   syntax Map ::= K "|->" K                      [function, hook(MAP.element), latex({#1}\mapsto{#2})]
--- a/k-distribution/tutorial/1_k/2_imp/lesson_5/imp.k
+++ b/k-distribution/tutorial/1_k/2_imp/lesson_5/imp.k
@@ -1,4 +1,5 @@
 // Copyright (c) 2014-2015 K Team. All Rights Reserved.
+require "domains.k"
 
 /*!
 \setlength{\parindent}{1em}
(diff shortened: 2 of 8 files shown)


bmmoore:
@andreistefanescu @daejunpark please review

k-distribution/include/builtin/domains.k
```diff
@@ -16,6 +39,7 @@ module MAP
   syntax Map ::= ".Map"                         [function, hook(MAP.unit), latex(\dotCt{Map})]
                // breaks klabel uniqueness
                //| "."                            [function, hook(MAP.unit)]
+  syntax Map ::= "(" Map ")" [bracket]
```
andreistefanescu, line 42: If we are going to go this way, we could add `()` to lists and sets as well.

k-distribution/tutorial/1_k/2_imp/lesson_5/imp.k
```diff
@@ -193,7 +201,7 @@ make the second rule structural, thinking of dissolving the residual
 empty \texttt{int;} declaration as a structural cleanup rather than as
 a computational step. */
 
-  rule <k> int (X,Xs => Xs);_ </k> <state> Rho:Map (.Map => X|->0) </state>
-    when notBool (X in keys(Rho))
+  rule <k> int `X,Xs => Xs`;_ </k> <state> Rho:Map (.Map => X|->0) </state>
+      when notBool `X in keys(Rho)`
```
andreistefanescu, line 205: I think @grosu  would prefer `requires`.

andreistefanescu:
lgtm

k-distribution/include/builtin/domains.k
bmmoore, line 42: Probably, but the definitions so far haven't used `()` on those sorts.

k-distribution/tutorial/1_k/2_imp/lesson_5/imp.k
bmmoore, line 205: I'm just making minimal changes to get things working, @grosu is planning to more thoroughly revise the tutorials after that.

daejunpark:
@bmmoore Your IMP++ assumed #1668 fix regarding IO, right? Otherwise, lgtm.

k-distribution/include/builtin/domains.k
daejunpark, line 42: @bmmoore @dwightguth would not like this. I remember that at some point it worked without this bracket annotation after some change into master. I'm not sure if it still works in the latest master, though.

bmmoore, line 42: What worked? If we have an explicit `Map ::= "(" Map ")"` production it should have a bracket annotation, and in the kore pipeline only the ``` notation is built into K syntax, no parentheses.

bmmoore:
I haven't actually gotten to the point of testing or trying to fix IO in IMP++.

k-distribution/include/builtin/domains.k
bmmoore, line 42: @dwightguth says he doesn't mind as long as it doesn't break the C semantics with new ambiguities.

grosu, line 42: I think it is OK to add user-space brackets wherever we can in the pre-defined data-types.  Of course, we should do it in a reasonable way, to not add ambiguities.

k-distribution/tutorial/1_k/2_imp/lesson_5/imp.k
grosu, line 205: Yes, `requires` please.

Not sure whether I should get into this parsing issue here, but I just realized that if add parentheses as brackets automatically to lists, then we may not need to replace them with backticks here.  I have mixed feelings about this, though, because in that case the parser would have to stop when it cannot parse anymore to the left or to the right, and thus end up with undesired rewrites ... @radumereuta ?

grosu, line 205: But what @bmmoore said is OK, for now just make minimal changes and let me go over the entire thing at the end and make all the other stylistic changes.

PR #1688, opened 2015-09-01
Fix inRHS

bmmoore:
Repairing pull request #1636. This fixes an error in tracking whether we are under a rewrite. @dwightguth @cos, please review.


bmmoore:
@dwightguth , this also passes `test-build` on the C semantics at the commit which merged kframework/c-semantics#118.

dwightguth:
lgtm

PR #1695, opened 2015-09-01
Ignore sort predicates on macros in OCaml's ExpandMacros

bmmoore:
This makes the OCaml backend work on its regression test without `--non-strict`, by just blindly ignoring sort predicates on side conditions of macro rules.


bmmoore:
@dwightguth please review

PR #1697, opened 2015-09-02
pre-summer changes

cos:
changes from before this summer -- most of them only affect tiny
as @dwightguth is busy, @bmmoore or @andreistefanescu please have a look


cos:
now passing

bmmoore:
That first commit looks like a mix of several changes. Maybe you could split it more clearly into changes to the tiny rewriter, and cleanups elsewhere.

cos:
the changes elsewhere are enabling the changes to the tiny rewriter, and they're very limited.

bmmoore:
You should also remove the imports you added for the unsupported method.

bmmoore:
Looks good otherwise

k-distribution/src/test/java/org/kframework/kore/convertors/TstTinyOnKORE_IT.java
```diff
@@ -32,33 +33,67 @@
 
 
     protected File testResource(String baseName) throws URISyntaxException {
+        System.out.println(baseName);
         return new File(TstTinyOnKORE_IT.class.getResource(baseName).toURI());
     }
 
     @Test @Ignore
```
dwightguth, line 40: if you fixed it, should you unignore this?

cos, line 40: it was working before summer but something changed in the meantime (or the way I merged) broke it. It will take time to figure out but I don't want to postpone merging this -- I already spent too much time merging and I don't want to have the master move too much without me.

cos:
Jenkins, test this please

PR #1698, opened 2015-09-02
Kore3

andreistefanescu:


andreistefanescu:
@bmmoore please review

bmmoore:
Looks good except for the comments above

bmmoore:
Ok, if you don't intent to restore the CELL_BAG attribute, then it's not easy to do better than the existing code. PR now looks good.

However, do note that cellBagSorts seems to be computed as any sort with an associative operator that subsorts a cell sort, instead of just contiaining sorts that are actually used as part of the configuration. Like, if a user definition defined their own bag of thread cells as part of explicitly implementing a scheduler or something, I think you'd end up with two sorts in the set of cell bag sorts for thread, and confuse these `size == 1` checks.

Issue #1691, opened 2015-09-01
Cell sorts not used in sort inference?


bmmoore:
In this definition, the variable `I` is given sort `K`, instead of inferring sort `Int` from the sort of the `<count>` cell.

```
require "domains.k"

module TEST-SYNTAX
 imports TEST-COMMON
 imports ID
endmodule

module TEST-COMMON
 imports DOMAINS-COMMON

 syntax Pgm ::= "start"
endmodule

module TEST
 imports TEST-COMMON
 imports DOMAINS

 configuration <k>$PGM:K</k><count>0::Int</count>

 rule start => 1
 rule <k>`I => .K` ...</k><count>_ => I</count>
endmodule
```

@radumereuta, I'll look into this, but if it's an issue with the loose cell grammar you are generating for parsing you should take it over.

bmmoore:
I see the RULE_CELLS case in RuleGrammarGenerator indeed ignores the sort of leaf cells, but I think I can fix it.

bmmoore:
Fixed in #1703

PR #1703, opened 2015-09-03
Use leaf cell sorts in inference, parsing

bmmoore:
@radumereuta, @andreistefanescu  please review.

The productions generated for parsing cells in rule bodies
always parsed the contents as K, even for leaf cells which
have a specific sort from the configuration.
This commit generates more specific productions for leaf cells,
using the actual expected sort of the contents.

This allows removing a type annotation that was added to IMP++
only to work around this issue.

The test in RuleGrammarTest needed an extra attribute in
handwritten cell productions to avoid an assertion error,
because ConfigurationInfoFromModule is pretty fragile.

In passing, I also changed the unit test to complain first about
an unexpected (lack of) parse error before checking the
precise warning count.

### DIFF
--- a/kernel/src/main/java/org/kframework/parser/concrete2kore/generator/RuleGrammarGenerator.java
+++ b/kernel/src/main/java/org/kframework/parser/concrete2kore/generator/RuleGrammarGenerator.java
@@ -170,15 +170,15 @@ public class RuleGrammarGenerator {
         }
         extensionProds.addAll(prods);
 
-        boolean add_rule_cells;
+        boolean addRuleCells;
         if (baseK.getModule(RULE_CELLS).isDefined() && mod.importedModules().contains(baseK.getModule(RULE_CELLS).get())) { // prepare cell productions for rule parsing
             // make sure a configuration actually exists, otherwise ConfigurationInfoFromModule explodes.
-            add_rule_cells = mod.sentences().exists(func(p -> p instanceof Production && ((Production) p).att().contains("cell")));
+            addRuleCells = mod.sentences().exists(func(p -> p instanceof Production && ((Production) p).att().contains("cell")));
         } else {
-            add_rule_cells = false;
+            addRuleCells = false;
         }
         Set<Sentence> parseProds;
-        if (add_rule_cells) {
+        if (addRuleCells) {
             ConfigurationInfo cfgInfo = new ConfigurationInfoFromModule(mod);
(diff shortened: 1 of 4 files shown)


andreistefanescu:
lgtm

bmmoore:
This fixes #1691

kernel/src/main/java/org/kframework/parser/concrete2kore/generator/RuleGrammarGenerator.java
```diff
@@ -166,23 +169,38 @@ public ParseInModule getCombinedGrammar(Module mod) {
             }
         }
         extensionProds.addAll(prods);
-        Set<Sentence> parseProds;
+
+        boolean add_rule_cells;
```
radumereuta: The java standard uses camecase names for variables. This is C. Please fix.

kernel/src/main/java/org/kframework/parser/concrete2kore/generator/RuleGrammarGenerator.java
```diff
@@ -166,23 +169,38 @@ public ParseInModule getCombinedGrammar(Module mod) {
...
                     assert p.items().head() instanceof Terminal || p.items().head() instanceof RegexTerminal;
                     assert p.items().last() instanceof Terminal || p.items().last() instanceof RegexTerminal;
-                    Seq<ProductionItem> pi = Seq(p.items().head(), NonTerminal(Sort("#OptionalDots")), NonTerminal(Sort("K")), NonTerminal(Sort("#OptionalDots")), p.items().last());
+                    Seq<ProductionItem> pi;
+                    if (cfgInfo.isLeafCell(p.sort())) {
+                        ProductionItem body = p.items().tail().head();
+                        pi = Seq(p.items().head(), NonTerminal(Sort("#OptionalDots")), body, NonTerminal(Sort("#OptionalDots")), p.items().last());
+                    } else {
+                        pi = Seq(p.items().head(), NonTerminal(Sort("#OptionalDots")), NonTerminal(Sort("K")), NonTerminal(Sort("#OptionalDots")), p.items().last());
+                    }
```
radumereuta, line 194: These are terribly long lines, you could initiate `body = NonTerminal(Sort("K"))`, have the if, and then have that huge line only once. Could make it easier to maintain in the future.

radumereuta:
only a few minor code style, other than that, lgtm

PR #1704, opened 2015-09-03
Add IO to IMP++

bmmoore:
IO for IMP++, using the temporary `<stdin/>` and `<stdout>` interface. Also works around the list subsorting issues. Examples such as `lesson_1/tests/io.imp` work. @andreistefanescu, @daejunpark please review.


daejunpark:
lgtm. I'll change it back to the original style of IO, once I fix #1668

bmmoore:
The list stuff is only there to work around #1705.

andreistefanescu:
lgtm

PR #1709, opened 2015-09-03
Extend rather than override K_OPTS also on Windows.

bmmoore:
Missed the windows script  in #1631. Now I just need to make sure this gets tested on windows.


bmmoore:
Verified using `-XX:+PrintFlagsFinal` in `K_OPTS` to check that our default flags are now still applied of `K_OPTS` is set, and that putting options in `K_OPTS` overrides the `K` defaults.

bmmoore:
@dwightguth please review, I see you wrote the batch script.

PR #1710, opened 2015-09-03
Make maincell (usually <k>) always have contents K

### LINKED ISSUE #1708: Sort of <k> cell should be special-cased to K
To allow us to set the start symbol by writing like `<k>$PGM:Pgm</k>` without breaking parsing, we used to have a special case that the contents of the `<k>` cell were always expected to be just `K`, instead of taking the sort from the provided initializer. We should restore this (unless perhaps we completely change how we handle `$PGM`).

bmmoore:
@radumereuta Please review.

Fixes  #1708


dwightguth:
lgtm

radumereuta:
lgtm

PR #1712, opened 2015-09-04
kernelc execution examples work

andreistefanescu:


andreistefanescu:
@bmmoore please review

java-backend/src/main/java/org/kframework/backend/java/compile/KOREtoBackendKIL.java
```diff
@@ -204,6 +247,16 @@ public Rule convert(Optional<Module> module, org.kframework.definition.Rule rule
 
     }
 
+    public static void flatten(Term term, KLabelConstant label, List<Term> list) {
```
bmmoore: Looks pretty similar to org.kframework.kore.Assoc.flatten (one overload relies on attributes, but another explicitly takes the binary and unit KLabels).

java-backend/src/main/java/org/kframework/backend/java/symbolic/ConjunctiveFormula.java
```diff
@@ -172,14 +172,14 @@ public ConjunctiveFormula add(Substitution<Variable, Term> substitution) {
     public ConjunctiveFormula add(Equality equality) {
         /* simplify andBool */
         if (equality.leftHandSide() instanceof KItem
-                && ((KItem) equality.leftHandSide()).kLabel().toString().equals("'_andBool_")
+                && ((KItem) equality.leftHandSide()).kLabel().toString().endsWith("_andBool_")
```
bmmoore, line 175: This is pretty ugly. Is there any way you could be more precise about what labels you want?

kernel/src/main/java/org/kframework/kore/compile/ConvertDataStructureToLookup.java
```diff
@@ -159,13 +159,13 @@ K addSideCondition(K requires) {
         List<Tuple2<KApply, KApply>> edges = new ArrayList<>();
         for (KApply k1 : state) {
             Multiset<KVariable> rhsVars = HashMultiset.create();
-            if (k1.klabel().name().equals("_in_")) {
+            if (k1.klabel().name().equals("Set:in")) {
```
bmmoore, line 162: This should work based on the hooks instead of label names too.

java-backend/src/main/java/org/kframework/backend/java/symbolic/ConjunctiveFormula.java
andreistefanescu, line 175: I want the labels with and without `'`. It's a big hack.

bmmoore:
I don't know how the rewriter internals are supposed to work, but the changes outside of that look pretty reasonable, except for relying on KLabel names instead of hook attributes.

kernel/src/main/java/org/kframework/kore/compile/ConvertDataStructureToLookup.java
andreistefanescu, line 162: This pass seems to be work in progress. I think it's best if @dwightguth cleans it up when he has the time.

java-backend/src/main/java/org/kframework/backend/java/symbolic/ConjunctiveFormula.java
bmmoore, line 175: It would be a bit less of a hack if you made fields for the expected labels, or something like that.

kernel/src/main/java/org/kframework/kore/compile/ConvertDataStructureToLookup.java
bmmoore, line 162: Ok then, open an issue to track the cleanup and I won't complain any more about it here. I see we already had hooks on most of those productions in `domains.k` but for some reason this was hardcoding the default Klabels.

andreistefanescu, line 162: #1715 tracks this issue

bmmoore:
Looks good.

Issue #1705, opened 2015-09-03
Need subsorting between related lists.


bmmoore:
The kore kompile and/or krun don't automatically recognize lists of related elements as related sorts. This can be worked around with explicit subsort productions. @andreistefanescu , @radumereuta , where was this handled before?

radumereuta:
The code that handles this in K3 is found in
 `org.kframework.compile.transformers.CompleteSortLatice:122`.
For the definition only, I'm looking for lists where the separator is the same, and if the elements are subsorted, then I automatically add a subsort to the module.

I can have a look at it tomorrow.

grosu:
So, is this issue closed now?

cos:
done

PR #1713, opened 2015-09-04
FIxing issue #1705. 

radumereuta:
Calculating automatic subsorts for overloaded lists.
This works only at parsing, disambiguation and type inference level.

@bmmoore, @andreistefanescu  please review, but do not merge yet.
It works, but I'm not sure why. I used to have a place where I would take into consideration overloaded operators, but it works without that step, so I think we need more tests for this.

Please test it in the meantime..

### DIFF
--- a/k-distribution/include/builtin/kast.k
+++ b/k-distribution/include/builtin/kast.k
@@ -172,6 +172,13 @@ module RULE-LISTS
   // adds the subsort production to the parsing module only:
   // Es ::= E        [userList("*")]
 
+  // automatically add subsorts between related lists
+  // Exp ::= Int
+  // Exps ::= List{Exp,","}
+  // Ints ::= List{Int,","}
+  // because the elements of the two lists are subsorted,
+  // the lists will automatically be subsorted also by adding the production:
+  // Exps ::= Ints
 endmodule
 
 module DEFAULT-CONFIGURATION
--- a/kernel/src/main/java/org/kframework/parser/concrete2kore/generator/RuleGrammarGenerator.java
+++ b/kernel/src/main/java/org/kframework/parser/concrete2kore/generator/RuleGrammarGenerator.java
@@ -312,7 +312,7 @@ public class RuleGrammarGenerator {
              */
             for (UserList ul1 : userLists) {
                 for (UserList ul2 : userLists) {
-                    if (ul1 != ul2 && ul1.separator.equals(ul2.separator)
+                    if (ul1 != ul2 && ul1.klabel.equals(ul2.klabel)
                             && mod.subsorts().$greater(Sort(ul1.childSort), Sort(ul2.childSort))) {
                         res.add(Production(Sort(ul1.sort), Seq(NonTerminal(Sort(ul2.sort)))));
                     }
--- a/kernel/src/main/java/org/kframework/parser/concrete2kore/generator/RuleGrammarGenerator.java
+++ b/kernel/src/main/java/org/kframework/parser/concrete2kore/generator/RuleGrammarGenerator.java
@@ -297,11 +297,26 @@ public class RuleGrammarGenerator {
 
         if (baseK.getModule(RULE_LISTS).isDefined() && mod.importedModules().contains(baseK.getModule(RULE_LISTS).get())) {
             java.util.Set<Sentence> res = new HashSet<>();
-            for (UserList ul : UserList.getLists(parseProds)) {
+            java.util.List<UserList> userLists = UserList.getLists(parseProds);
+            for (UserList ul : userLists) {
                 org.kframework.definition.Production prod1;
                 // Es ::= E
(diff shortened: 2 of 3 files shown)


bmmoore:
@radumereuta  says the place in question is in `TypeInferenceSupremumFilter`, at `boolean isSubsorted(Production big, Production small)`

k-distribution/include/builtin/kast.k
```diff
@@ -172,6 +172,13 @@ module RULE-LISTS
   // adds the subsort production to the parsing module only:
   // Es ::= E        [userList("*")]
 
+  // automatically add subsorts between related lists
```
andreistefanescu, line 175: why is this commented out code here?

bmmoore, line 175: It's only there for documentation - importing `RULE-LISTS` is a marker that tells the parser to generate the right productions for parsing user lists.

andreistefanescu:
lgtm

bmmoore:
These rules look good, but I think we should just generate them into the module so every pass can use them.

radumereuta:
Well, I'm not sure exactly where to generate them.
Most likely I don't want those productions to be present when I'm generating the parser for programs.

bmmoore:
Good catch on the klabel thing. Also, does not adding the rules to the actual modules mean we can end up with incorrect isSort predicates? I wouldn't mind just adding them for parsing, if I was sure the subsort relationship in the rest of parsing came out right.

There are some funny possibilities about where to generate rule. What if we have a definiton like this:

```
module A
imports ...
syntax Xs ::= List{X,","}
syntax Ys ::= List{Y,","}
...rules here...
endmodule

module B
imports A
syntax X ::= Y
...more rules...
endmodule
```

I guess then we'd want to put `Xy ::= Ys` in module B? Should the compiler process rules from A and B under a different subsort relation or what? (we could even have one module make a declaration `Xs ::= Middle` and another with `Middle ::= Ys`, and then have a frontier of multiple independent modules that import both of those and learn that `Xs ::= Ys`).

bmmoore:
Actually, all these issues about which modules first add different subsort relationships arise even without considering lists, so hopefully the compiler passes don't depend too much on the subsort relationship.

radumereuta:
I think they will at one point. I just didn't add enough tests.
So, other opinions... where would it be the best place to add these subsorts to the definition?

bmmoore:
Well, we also wouldn't want to put them somewhere that they could mess up program parsing. If we have two lists with the same klabel but a different separator, we don't want to start letting you switch separators in the middle of a list in the concrete syntax.

bmmoore:
So, "just generate them into the module" isn't nearly as simple as I hoped, and I think we should postpone that. My only other concern is whether letting type inference use relationships that don't actually exist according to the definition can lead to tricky misparses. Remember those bugs I turned up a while back about confusing the "syntactic subsort" relationship with the semantic subsort relation?

bmmoore:
We actually need parsing and disambiguation to have different productions. I'm not sure parsing actually needs any of these subsorts, but if it does they would have to relate lists with the same separator. On the other hand, sort inference and other processing like that needs to subsort between lists with the same klabel.

radumereuta:
It's easier to add the subsort in the parser also, because it will then give you all the type restrictions for variables. You don't have to look in the context, it's enough to just traverse the parse forest, and collect each instance of a variable with the appropriate list wrapper and context. The invalid ones (depending on the rest of the context will discarded).

grosu:
@radumereuta please take a look at #1733 as well and let us know if that clarifies things or not.

bmmoore:
From that discussion, I think what we should do here is not worry about explicitly representing these subsort relationships by adding sentences into modules, but instead make the code which calculates the subsort lattice add them into the computed relation (as it used to, now justified as a very limited case of proof). Then the parser elaboration passes can add whatever rules they each need to get correct parses.

radumereuta:
@bmmoore that's pretty much what I've implemented in this pull request. I'm considering the automatic subsort relation only where I need it to parse.

@grosu I'm not sure if I understand what you are trying to say in #1733 very well. You say semantics module, and syntax module. Right now, every module can have syntax and semantics. I can call `#parse` on any module, and get a valid parser, so I feel a bit reluctant in changing the module, since the user might want to access it at one point and expecting the original content. I'm not sure how to loo, at this problem, from a modularity point of view, compiler, parser... It seems to involve all of them.

grosu:
All we need for now is to make sure what we had before works.  That is, we do not want users to have to add redundant productions that we did not have before, such as `Exps ::= Ints`.  So please make sure that is indeed the case.

@radumereuta: if you do not understand something specifically, then go ahead and comment or ask questions there, making them as specific as possible so I can address them.  The way you put the question here I cannot figure it out which of my comments you are referring to.  And even if I do, I do not want to answer a question from an issue into another issue.

radumereuta:
Better fix in #1891.

PR #1716, opened 2015-09-06
KORE IO

daejunpark:
wip, do not review

### DIFF
--- a/k-distribution/include/builtin/domains.k
+++ b/k-distribution/include/builtin/domains.k
@@ -564,6 +564,12 @@ module K-IO
 
 endmodule
 
+// NOTE: DO NOT DIRECTLY IMPORT *-STREAM MODULES
+// These stream modules will be automatically instantiated and implicitly imported
+// into the main module when `stream` attributes appear in configuration cells.
+// Only `Stream` productions and `[stream]` rules will be imported.
+// The cell name will be replaced with the one of the main configuration.
+
 module STDIN-STREAM
   imports K-IO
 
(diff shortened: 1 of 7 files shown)


daejunpark:
@dwightguth please review.
@bmmoore I've changed the imp++ back to original for IO. Could you check if it works?

daejunpark:
this will fix #1668 and #1704.

daejunpark:
@dwightguth For a quick explanation of this design, it essentially instantiates the stream modules into the main module when stream-attributed cells are found. Also it generates the stdin stream unblocking rules, while it supports only a specific pattern that is supported by the previous implementation in KIL pipeline. Overall, this pull request provides the same interface/behavior with the previous one. 

Instantiating template modules, however, are quite ad-hoc for now. We will need to have a notation for fine-controling the instantiation later on. The unblocking rule generation is also fragile as the previous one was.

daejunpark:
I've made it more robust for both importing (and instantiating) stream modules and generating unblocking rules by having additional annotations, `[stream]` and `[unblock]` respectively.

~~I'll implement one more improvement, which is to add `initializer` attributes to config cell initialization rules. It will make it much robust to find such a initializer rule. Let me know if it would be problematic.~~

bmmoore:
IMP++ IO seems to be working fine now.

dwightguth:
looks fine to me although I haven't looked too closely. As long as it successfully handles the IO cases, I am fine with it being merged.

PR #1721, opened 2015-09-09
Various fixes for K on windows

dwightguth:
@bmmoore please review


bmmoore:
I haven't tried installing OCaml on windows to test this changes, but reading throught the changes it looks good aside from the comments above.

bmmoore:
Ok, looks good. The conflict is trivial, just keep both functions and merge if it passes.

PR #1722, opened 2015-09-09
Serialize2

dwightguth:
@andreistefanescu @bmmoore please review


bmmoore:
You should document the public methods, especially those whose interface or behavior is affected by this change. Otherwise it looks good.

andreistefanescu:
This is a hack to bypass the bug in the handling of sets of cells. Please fix the root cause of the problem.

dwightguth:
It's not something specific to sets; I was able to reproduce the exact same bug without sets using only the Java Backend. I don't really want to go digging deep into the rewriter code to fix something that has nothing to do with OCAML. I can add a TODO to remove the hack if and when you fix the underlying bug if you'd like, though.

andreistefanescu:
If that is the case, please make an issue with a minimal K definition
exhibiting the bug and I will look into it. However, I will not merge this
hack.

Andrei

dwightguth:
Jenkins: test this please

bmmoore:
That interface and comments looks a lot better. I just wonder where you are getting these mode strings from.

PR #1726, opened 2015-09-10
make KApply be a KCollection

cos:
small but important change

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/kil/KItemRepresentation.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/kil/KItemRepresentation.java
@@ -6,6 +6,9 @@ import org.kframework.kore.*;
 import org.kframework.kore.KLabel;
 import org.kframework.kore.KList;
 
+import java.util.List;
+import java.util.stream.Stream;
+
 public interface KItemRepresentation extends KoreRepresentation, KApply {
     default Term kLabel() {
         return ((KItem) toKore()).kLabel();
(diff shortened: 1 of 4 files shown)


cos:
@dwightguth, @bmmoore small one, please have a look.

dwightguth:
Is this really needed? You can always just access the KList directly... If you really think this is important though, I don't necessarily have a complaint, but you need to make the methods that you added to KItemRepresentation work correctly. It shouldn't be hard.

cos:
> Is this really needed?

Unfortunately `KSeq` and `KRewrite` are not `KApply`s -- I would preferred them to be but it was beyond me. I want to be able to not care about what I'm traversing and `KCollection` gives me that.

> If you really think this is important though, I don't necessarily have a complaint, but you need to make the methods that you added to KItemRepresentation work correctly. It shouldn't be hard.

The backend is smaller than before but still has quite a few classes, so it's not trivial. I prefer to wait for Andrei to do it. I'll add an issue for this and bug him.

bmmoore:
I don't really see the point of this, but at the very least you need to document the KCollection class if you do.

cos:
Jenkings, test this please

cos:
@andreistefanescu has already picked up the backend implementation of the missing methods. If nobody complains, I would merge this now so I don't get stuck with a subsequent pull request.

cos:
Jenkins, test this please

dwightguth:
lgtm

PR #1727, opened 2015-09-11
Feature/split k cell from top cell transform

cos:


cos:
@bmmoore, please review when it passes the tests

bmmoore:
The label constants should follow the style we've used everywhere else, of exposing a static final string field named in uppercase. Looks okay otherwise. (In particular, calling it as an accessor I'm not quite sure if those values are sometimes meant to vary).

cos:
Jenkins, test this please.

bmmoore:
Looks good.

PR #1737, opened 2015-09-18
fix issue 1724

andreistefanescu:
@dwightguth please review


andreistefanescu:
@dwightguth where is the place for regression tests in kore?

dwightguth:
I don't believe we have such a place right now, but what I suggest you do is to create a file regression-kore.k in tests/regression, and add individual tests to this definition.

PR #1738, opened 2015-09-18
Issue #1682 fix. 

### LINKED ISSUE #1682: List Parsing adding terminator badly
While trying to fix IMP++ with kore, writing the while rule as

```
rule while (B) S => if (B) {`__`(S,`__`(while (B) S,.Stmts))} else {.Stmts}  [structural]
```

parsed the inside of the first block as

```
`__`(`__`(S,`__`(while (B) S,.Stmts)),.Stmts)
```

with an extra terminator.

msaxena2:
Fixes #1682 . Initial Commit. Not Yet Ready to Review.

### DIFF
--- a/kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/AddEmptyLists.java
+++ b/kernel/src/main/java/org/kframework/parser/concrete2kore/disambiguation/AddEmptyLists.java
@@ -74,7 +74,7 @@ public class AddEmptyLists extends SetsGeneralTransformer<ParseFailedException, 
                         || tc.production().klabel().get().name().startsWith("#SemanticCastTo")
                         || tc.production().klabel().get().name().equals("#InnerCast")))) {
                     newItems.add(child);
-                } else if (childSort.equals(Sorts.K()) || !subsorts.lessThan(childSort, expectedSort)) {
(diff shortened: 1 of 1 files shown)


msaxena2:
@dwightguth @bmmoore @radumereuta Please Review.

dwightguth:
looks okay but I'd prefer to see this change as part of a fix that can be tested by testing SIMPLE. Without that I don't really know for sure that this isn't going to break stuff.

bmmoore:
Looks like an improvement, but why did you change `lessThan` to `lessThanEq`?

bmmoore:
If you don't have a good reason, I'd rather not change the function.

For Dwight's request, rather than waiting any longer, how about testing that this doesn't break the other definitions which have been successfully translated to kore, currently IMP, IMP++, and lambda++.

daejunpark:
@bmmoore I guess that IMP++ will fail even on the current master due to the sort lattice issue #1749. The failure has not been seen, because ktest is disabled with those translated languages. Let me make a pull request to make ktest work with the new kore unparser output, so that we can see master doesn't break the translated languages as well.

bmmoore:
You should add some regression tests with examples from the list parsing bug. Other than that, you can get enough general Kore coverage by merging from master to get @daejunpark's pull request.

bmmoore:
This breaks a few test cases, and isn't the right place to fix list terminators. I'm making a new pull request for this.

PR #1740, opened 2015-09-20
hotfix for intermittent bug

cos:


cos:
@andreistefanescu have a look

andreistefanescu:
lgrm

daejunpark:
@cos could you please explain which bug is fixed, and why?

cos:
There's a bug in the IntelliJ Scala compiler which only pops up in certain interleavings of the Java-Scala mixed compilation. When it pops up, certain invocations of KList cannot find the method anymore. This simplification should fix it. If you've never encountered the bug, don't worry. If it ever pops up again, let me know.

PR #1741, opened 2015-09-22
Tutorial lambda++

daejunpark:
wip

todo: npe, macro


daejunpark:
@bmmoore please review.

Comments and questions:
- I've modified all the lessons except lesson_1 which uses the substitution.
- I've written a separate config-kore.xml file to test them, which can be referred to fix the original config.xml in the end.
- The lessons from 2 to 5 do not have a separate syntax module and have the `Id` parsing issue, so I made a syntax module for each of them by rearranging syntax declarations. Do you have a better idea for that?

bmmoore:
I doubt we can avoid splitting the syntax into multiple modules. Maybe the rules could be written in the common module? I'm not sure that's any cleaner.

bmmoore:
Just a warning, the `--parser "kast --kore --sort ..."` option doesn't work properly on windows. For some reason you need to explicity write `kast.bat` even though `kast` alone works from the command line. I don't know if ktest can somehow pass different options on different platforms.

bmmoore:
Looks good, except for the number in that comment in config.xml. Merge if tests pass after adjusting that.

daejunpark:
@bmmoore I fixed the comment in the main config.xml. For the config-kore.xml, you're right, but it would be ok, since those config files are temporary, and the problematic option will be dropped once the starting symbol issue is fixed.

daejunpark:
@bmmoore your quick idea doesn't work. 

I added the main module back to the syntax module as below, but those syntaxes defined in `DOMAINS` make an ambiguity when parsing programs. For example, the following program has `a` as an `Id` but it is ambiguous between `Id` and `KLabel`:

```
$ cat ../../1_lambda/lesson_1/tests/free-variable-capture.lambda 
a (((lambda x.lambda y.x) y) z)
$ kast --kore --sort Exp ../../1_lambda/lesson_1/tests/free-variable-capture.lambda 
1: syntax KBott ::= KLabel "(" KList ")"
...
2: syntax Exp ::= Exp Exp
...
```

Definition without separating program syntaxes:

```
module LAMBDA-SYNTAX                                                                           
  imports LAMBDA                                                                               
  imports ID                                                                                   
endmodule                                                                                      

module LAMBDA                                                                                  
  imports DOMAINS                                                                              
  imports EMPTY-ID                                                                             

  syntax Exp ::= Id                                                                            
               | "lambda" Id "." Exp  [latex(\lambda{#1}.{#2})]                                
               | Exp Exp              [strict, left]                                           
               | "(" Exp ")"          [bracket]                                                

  configuration <T>                                                                            
                  <k> $PGM:Exp </k>                                                            
                  <env> .Map </env>                                                            
                  <store> .Map </store>                                                        
                </T>                                                                           

  syntax Val ::= closure(Map,Id,Exp)                                                           
  syntax Exp ::= Val                                                                           
  syntax KResult ::= Val                                                                       

  rule <k> lambda X:Id . E => closure(Rho,X,E) ...</k>                                         
       <env> Rho </env>                                                                        
    [structural]                                                                               
  rule <k> closure(Rho,X,E) V:Val => E ~> Rho' ...</k>                                         
       <env> Rho' => Rho[X <- !N] </env>                                                       
       <store>... .Map => (!N:Int |-> V) ...</store>                                           
  rule <k> X => V ...</k>                                                                      
       <env>... X |-> N ...</env>                                                              
       <store>... N |-> V ...</store>                                                          

// Auxiliary task                                                                              
  rule <k> _:Val ~> `Rho => .K` ...</k> <env> _ => Rho </env>                                  
    [structural]                                                                               
endmodule
```

Any more idea?

bmmoore:
No more ideas, sorry. I see I was forgetting about the functions on those library sorts, which are absolutely necessary to write the rules we want, but pollute the grammar.

Issue #1692, opened 2015-09-01
NPE in KILtoSMTLib.renameSort without --smt none


bmmoore:
Trying to run `sum.imp` under the kore IMP++ from `lesson_8` on the tutorial-kore branch (at  4b1f3935c84db5941451fa443f86c9a83183254c) as
`krun --debug --kore --parser "kast --kore --sort Stmts" ../../2_imp/lesson_1/tests/sum.imp` throws

```
java.lang.NullPointerException
        at org.kframework.backend.java.symbolic.KILtoSMTLib.renameSort(KILtoSMTLib.java:368)
        at org.kframework.backend.java.symbolic.KILtoSMTLib.getSortAndFunctionDeclarations(KILtoSMTLib.java:246)
        at org.kframework.backend.java.symbolic.KILtoSMTLib.translateConstraint(KILtoSMTLib.java:176)
        at org.kframework.backend.java.symbolic.SMTOperations.checkUnsat(SMTOperations.java:40)
        at org.kframework.backend.java.symbolic.ConjunctiveFormula.checkUnsat(ConjunctiveFormula.java:580)
        at org.kframework.backend.java.kil.ConstrainedTerm.unify(ConstrainedTerm.java:224)
        at org.kframework.backend.java.symbolic.SymbolicRewriter.computeRewriteStepByRule(SymbolicRewriter.java:206)
        at org.kframework.backend.java.symbolic.SymbolicRewriter.lambda$computeRewriteStep$86(SymbolicRewriter.java:154)
        at org.kframework.backend.java.symbolic.SymbolicRewriter$$Lambda$81/976725249.apply(Unknown Source)
        at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
        at java.util.Spliterators$IteratorSpliterator.tryAdvance(Spliterators.java:1812)
        at java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:126)
        at java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:529)
        at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:516)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:502)
        at java.util.stream.FindOps$FindOp.evaluateSequential(FindOps.java:152)
        at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
[...]

PR #1743, opened 2015-09-23
Fix strict hole sort

daejunpark:


daejunpark:
todo: fix sort lattice issue

daejunpark:
this will fix the recurring issue #1692

daejunpark:

dwightguth:
Hold on a sec. This is a change being made to the core compilation pipeline in order to solve an issue in the Java backend. That seems strange to me, and in particular I don't know how this change will affect the C semantics. Can you justify the change to the core of the K language by explaining how the passes worked before and after, and why you think they behaved incorrectly before?

andreistefanescu:
This change preserves the sort information for heating/cooling rules, thus making the sort information consistent across all rules. You are free to discard it in the ocaml backend.

dwightguth:
Okay, that sounds like a good change, but I was looking for more
information. More concretely, what is the change made to the semantics of
strict and contexts? Like, what rules were generated before and after the
change? How was consistency violated in the old semantics?

andreistefanescu:
Let's say you would manually write a heating rule:
`HOLE + K => HOLE ~> freezerPlus(K)    requires notBool isKResult(HOLE)    [heat]`
Then the sort inferencer would infer the sorts `Exp` for `K` and `HOLE`, which the java backend would use, and the ocaml backend would discard. Before the change, the heating rules were generated without this inferred sort information. This leads to a bunch of small issues, that both me and @daejunpark hit. This change should add the inferred sort information to the rule.

dwightguth:
Oh, okay, I see. In that case, can you please disable this change if the
user passes the --non-strict flag?

daejunpark:
The sort preservation is disabled, thus this change has no effect, when `--non-strict` option is given.

dwightguth:
lgtm

Issue #1744, opened 2015-09-24
Cell Fragments Related Issue


msaxena2:
I ran into the following problem while trying to work with Simple. Here are the relevant details:
Consider the following configuration:

```
<T>
    <a> <b> .List </b> <c> .List </c> </a>
</T>
```

A pattern of the form `<a> <b> X </b> Y:ACellFragment </a>` compiles to: ``<T>`(`<a>`(`<b>`(X), Y)) requires __andBool__(isList(X), isACellFragment(Y))`. 

A match on any configuration using this pattern fails as the inferred sort of Y during the match is CCell, and isACellFragment(Y) fails as CCell isn't a cell subsort of ACellFragment.

cos:
@msaxena2, is this still failing?

PR #1745, opened 2015-09-24
fix to the cell fragments issue

msaxena2:
Deals with #1744 . @bmmoore please review.


dwightguth:
This looks like it could potentially have impact beyond the issue of cell fragments. Can we please hold off on merging this until we've had a chance to evaluate its impact on C?

dwightguth:
Oh wait, never mind. I am looking at the issue that you reported and this is the wrong fix. The correct issue is that if a variable matches only one cell in the parent cell, then it should be treated as a variable for that Cell rather than a cell fragment. IE, 

A pattern of the form `<a> <b> X </b> Y:ACellFragment </a>` compiles to: ``<T>`(`<a>`(`<b>`(X), Y)) requires __andBool__(isList(X), isCCell(Y))`.

bmmoore:
That's not a clear description of a correct fix. If there's exactly one cell captured by a fragment variable we can keep the name and just change the sort, but we also have to handle other occurances, for example if the fragment is saved on a stack. When this is the rule

```
syntax Stack ::= "emtpy" | frame(ACellFragment,Stack)
<a> <b> Stk => frame(Y,Stk) </b> Y:ACellFragment </a>
```

it could compile to

```
`<a>`(`<b>`(Stk => frame(`<a>-fragment`(`noBCell`(),Y),Stk)),Y) requires isCCell(Y)
```

bmmoore:
Your code change seems good so far, but incomplete. Look at the code starting `if (k.klabel().name().equals("isBag")) {` in processVars. - if @dwightguth wants you to reduce sort-predicate applications, then that's the place to do it.

msaxena2:
@bmmoore I merged the [PR](https://github.com/msaxena2/k/pull/1) you'd made into my fork, and your commits now will appear as a part of this PR.

bmmoore:
Jenkins: test this please

bmmoore:
Well, I wrote these latest patches - @msaxena2, @dwightguth can you look over them?
@dwightguth, our fslwork/fslexp are currently acting up. Can you please try this against whatever version of C you want checked?

dwightguth:
If you can't test it beforehand I'm inclined to just let the PR go through and revert it if it breaks something in the master build of the C semantics. Are you okay with that?

bmmoore:
There's something about an `isBag` predicate. @dwightguth, where does that come from?

```
Apply compile pipeline                                       =  6065
java.util.NoSuchElementException: key not found: isBag
    at scala.collection.MapLike$class.default(MapLike.scala:228)
    at scala.collection.AbstractMap.default(Map.scala:59)
    at scala.collection.MapLike$class.apply(MapLike.scala:141)
    at scala.collection.AbstractMap.apply(Map.scala:59)
    at org.kframework.backend.ocaml.DefinitionToOcaml.forEachKLabel(DefinitionToOcaml.java:701)
    at org.kframework.backend.ocaml.DefinitionToOcaml.printKType(DefinitionToOcaml.java:686)
```

bmmoore:
Perhaps this will work with C after #1754?

dwightguth:
This looks like you have somehow failed to transform one or more isBag
predicates into their actual form where they check whether specific cells
are correctly formed. This is causing it to fail because it does not
recognize this klabel as one of the klabels in the definition. I believe
you need to fix this, although I don't know for sure whether it will be
fixed by #1754

bmmoore:
Why should `isBag` be satisfied by a cell fragment?

bmmoore:
With a patch to accept `isBag` as a cell fragment predicate, it's running into an error because SortCells is incorrectly promoting a variable explicitly typed as `L:LocalCell` into a cell fragment.

bmmoore:
@msaxena2 I have another pull request against your branch that fix the remaining problems here.

dwightguth:
@bmmoore says C works so lgtm

PR #1746, opened 2015-09-25
revert #1741, changes to lambda++

daejunpark:
Revert #1741, the ugly changes to lessons 2 to 5 of lambda++. Now only the last lesson of lambda++ works in KORE, as the other language imp and imp++ do.

@grosu @bmmoore please review.


bmmoore:
I think it's important that we keep one definition that works in kore, and reminds us how we had to disorganize it to get it working, but I don't mind reverting the other lessons. Looks good.

daejunpark:
I close this. Instead I'll write a comment for the existing changes.

PR #1750, opened 2015-09-28
fix context hole bug

daejunpark:
now the error occurs only when a context has multiple holes with different sort casting except KItem.

this will fix the issue on translating SIMPLE and KOOL.

@msaxena2 please try if it fix the issue.

### DIFF
--- a/kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
+++ b/kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
@@ -114,10 +114,13 @@ public class ResolveContexts {
                 if (k.klabel().name().startsWith("#SemanticCastTo")) {
                     K child = k.klist().items().get(0);
                     if (child instanceof KVariable && ((KVariable) child).name().equals("HOLE")) {
-                        if (holeWithSemanticCast != null) {
-                            throw KEMException.compilerError("Cannot compile a context with multiple holes.", context);
+                        if (!ResolveSemanticCasts.getSortNameOfCast(k).equals("KItem")) {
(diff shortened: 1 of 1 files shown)


kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
```diff
@@ -114,10 +114,13 @@ public Void apply(KApply k) {
...
                     K child = k.klist().items().get(0);
                     if (child instanceof KVariable && ((KVariable) child).name().equals("HOLE")) {
-                        if (holeWithSemanticCast != null) {
-                            throw KEMException.compilerError("Cannot compile a context with multiple holes.", context);
+                        if (!ResolveSemanticCasts.getSortNameOfCast(k).equals("KItem")) {
+                            if (holeWithSemanticCast != null && !holeWithSemanticCast.equals(k)) {
+                                throw KEMException.compilerError("Cannot compile a context with multiple holes" +
+                                        " having different sort casting.", context);
+                            }
+                            holeWithSemanticCast = k;
```
andreistefanescu, line 122: is it ok not to set `holeWithSemanticCast` on the else branch?

andreistefanescu:
lgtm

kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
daejunpark, line 122: yes, it is set only when HOLE with non-KItem sort casting. in the below, line 131(old) or 134(new), the default HOLE with KItem sort casting will be used.

Issue #1673, opened 2015-08-31
Nondeterminism in heating/cooling


dwightguth:
Need nondeterministic heating and cooling to work correctly (either superheat/supercool or something involving strategies)

grosu:
@cos, if I remember correctly, you said this will be solved with strategies, right?  But if that is your plan, then this will probably take quite a bit or time to be done, won't it?  Is there something quick that we can do now to transit with all the tests passing, and do the strategies afterwards?

cos:
The hack-in extreme is to special-case for superheating/cooling like before. The clean extreme is to wait for a clean implementation of strategies (which depends on several other things). I'm working on a middle ground: have the superheating/cooling executed by strategies, but not expose the mechanism to the users just yet -- we'll have to figure out modularization and configuration merging before that. This approach will be ready in a few days for the Java backend.

grosu:
Alright, this sounds like a good plan.  This will also allow you to better understand what you need from the final strategy language that you will eventually come up with, so it fulfills multiple roles.  Thanks.

grosu:
@cos , and ETA on this?

cos:
If everything goes well, I'll make a pull request today. I might need something from Dwight and then I'll make it tomorrow.

cos:
as discussed, we're moving this to 4.1

Issue #1749, opened 2015-09-26
sort lattice java backend


daejunpark:
java backend requires sort lattice is complete, i.e., glb exists for any set of sorts.

for IMP++, no GLB exists for AExp and KResult, since LB of AExp and KResult is {Int, String}.
https://github.com/kframework/k/blob/3b490a898d05e7fd9828bc79e1681b9fbb557cfa/k-distribution/tests/regression/kore-io/imp.k

this makes `isAExp(T) andBool isKResult(T)` be `false` for any `T`, which is a condition of a cooling rule.

bmmoore:
This definition used to be working when run by hand.
@andreistefanescu , do you have any idea why this might have changed?

dwightguth:
Was that before #1743 was merged? The condition mentioned in the issue would have been added only by that pull request.

daejunpark:
@bmmoore @dwightguth yes, it happens after #1743 merged.

bmmoore:
Ok, that explains everything.

PR #1751, opened 2015-09-29
enable ktest imp, imp++, lambda++ for kore translation

daejunpark:
since kore translation takes a while, we need to enable ktest again for kore to make sure that master doesn't break the existing kore translated languages. currently, ktest are enabled for imp, imp++, and lambda++ that have been translated so far.

to enable ktest without any potential side effects to other languages that are not yet translated, a separate directory `tests-kore` is generated for each `tests` in which an adjusted `config.xml` file is created, new `*.out` files are carefully generated if they are different with the original only due to the new unparser.

however, some test programs are excluded until #1673 is fixed. see TODOs in config.xml files.

@bmmoore please review.


daejunpark:
see also https://github.com/kframework/k/pull/1738#issuecomment-143905151

daejunpark:
also note that imp++ is updated due to #1749:
https://github.com/kframework/k/pull/1751/files#diff-50d3be08d61055952993aa6dded0b5df

bmmoore:
Looks good. Having the separate config files for now is probably the best we can do.
I don't know why IMP++ wasn't working now, when we were able to run programs such as spawn by hand before, but that's another problem.

Issue #1747, opened 2015-09-26
cell fragment parsing issue


daejunpark:
For a rule that matches a cell fragment:

```
configuration <T> 
  <k> $PGM:K </k>
  <x> 
    <a> .K </a>
    <b> .K </b>
    <c> .K </c>
  </x>
</T>

syntax KItem ::= frag(XCellFragment)

rule frag(<x>-fragment `<a> A </a>` noBCell noCCell </x>-fragment) => .K
```

It generates a parsing error:

```
$ kompile --kore
[Error] Compiler: Had 1 parsing errors.
[Error] Critical: Unexpected sort Cell for term
<a>(#noDots(),#token(#KVariable,"A"),#noDots()). Expected ACellOpt.
```

@bmmoore Do you have any idea?

daejunpark:
The problem is that `<a> A </a>` should be of `ACell`, but is inferred as just `Cell`.

Sort casting is an workaround:

```
`<a> A </a>`:>ACell
```

Is there any better way?

daejunpark:
see also #1744

daejunpark:
see also #1748

bmmoore:
I think this is just a little problem with how the concrete syntax for cells in rules is generated.
The actual labels of the cells are defined by productions as described on the KAST and KORE wiki page, but to allow configuration abstraction in the full K syntax we generate a looser grammar using a generic `Cell` sort. I think the problem is just that we are declaring the syntax of specific cells with one production like `syntax Cell ::= "<a>" Cells "</a>"` rather than making two production like `syntax ACell ::= "<a>" Cells "</a>"` and `syntax Cell ::= ACell`.

I've worked on that code before, if I'm right about the cause this should be a quick fix.

bmmoore:
@daejunpark Has this been fixed?

daejunpark:
yes.

PR #1754, opened 2015-09-30
Parse pre-completion cells in proper sorts.

### LINKED ISSUE #1748: cell fragment as an argument
Consider the following example that is simplified from KOOL.

```
configuration <T> 
  <k> $PGM:K </k>
  <x> 
    <a> 1 </a>
    <b> 2 </b>
    <c> 3 </c>
  </x>
</T>

syntax KItem ::= frag(XCellFragment)

rule frag(<x>-fragment `<a> A </a>`:>ACell `<b> B </b>`:>BCell C </x>-fragment) => A +Int B

rule <k> run1 => frag(X) ... </k>
     <x> 
       X:XCellFragment
       <c> _:Int </c>
     </x>

rule <k> run2 => frag(A B) ... </k>
     <x> 
       A:ACell
       B:BCell
       <c> _:Int </c>
     </x>
```

Rule `run1` works, but `run2` does not. 

@bmmoore  Any idea?

bmmoore:
When generating the modified productions that allow incomplete cells
with features such as dots and omitting parent cells,
the modified productions for indvidual cells were places into the
generic Cell sort (used for flexibility such as allowing parent cells),
this instead keeps the original more specific FooCell sorts,
for which subsorts into Cell were already generated.

@daejunpark  this should fix ~~#1748~~ #1747

I need to add tests, but you can check if this commit helps.

### DIFF
--- a/k-distribution/tests/regression/config.xml
+++ b/k-distribution/tests/regression/config.xml
@@ -73,8 +73,8 @@
            programs="ocamlbackend/tests"
            results="ocamlbackend/tests" />
 
-  <include file="kore-cell-fragments/imp_fragments/config.xml"
-           directory="kore-cell-fragments/imp_fragments/" />
+  <include file="kore-cell-fragments/config.xml"
+           directory="kore-cell-fragments/" />
 
   <include file="issue1724/config.xml"
            directory="issue1724"
--- a/k-distribution/tests/regression/kore-cell-fragments/cell_arguments/config.xml
+++ b/k-distribution/tests/regression/kore-cell-fragments/cell_arguments/config.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright (c) 2015 K Team. All Rights Reserved. -->
+  <tests>
+    <test definition="test.k">
+      <kompile-option name="--kore"/>
+    </test>
+  </tests>
--- a/k-distribution/tests/regression/kore-cell-fragments/cell_arguments/test.k
+++ b/k-distribution/tests/regression/kore-cell-fragments/cell_arguments/test.k
@@ -1,0 +1,20 @@
+// Copyright (c) 2015 K Team. All Rights Reserved.
+require "domains.k"
+module TEST-SYNTAX
+imports TEST
+endmodule
+module TEST
+imports MAP
+configuration <T>
+  <k> $PGM:K </k>
+  <x>
+    <a> .K </a>
+    <b> .K </b>
+    <c> .K </c>
+  </x>
+</T>
+
+syntax KItem ::= frag(XCellFragment)
+
+rule frag(<x>-fragment <a> A </a> noBCell noCCell </x>-fragment) => .K
+endmodule
--- a/k-distribution/tests/regression/kore-cell-fragments/config.xml
+++ b/k-distribution/tests/regression/kore-cell-fragments/config.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright (c) 2015 K Team. All Rights Reserved. -->
+
+<tests>
+  <include file="imp_fragments/config.xml"
+           directory="imp_fragments/" />
+
+  <include file="cell_arguments/config.xml"
+           directory="cell_arguments/" />
+</tests>
+
--- a/kernel/src/main/java/org/kframework/parser/concrete2kore/generator/RuleGrammarGenerator.java
+++ b/kernel/src/main/java/org/kframework/parser/concrete2kore/generator/RuleGrammarGenerator.java
@@ -190,7 +190,7 @@ public class RuleGrammarGenerator {
                     if (cfgInfo.isLeafCell(p.sort())) {
                         body = p.items().tail().head();
                     } else {
-                        body = NonTerminal(Sort("K"));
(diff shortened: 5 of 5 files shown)


bmmoore:
With this code there are no parse error on your test module, with necessary imports added as below:

```
require "domains.k"
module TEST
imports MAP
imports INT

configuration <T>
  <k> $PGM:K </k>
  <x>
    <a> 1 </a>
    <b> 2 </b>
    <c> 3 </c>
  </x>
</T>

syntax KItem ::= frag(XCellFragment) | "run1" | "run2"

rule frag(<x>-fragment <a> A </a> <b> B </b> C </x>-fragment) => A +Int B

rule <k> run1 => frag(X) ... </k>
     <x>
       X:XCellFragment
       <c> _:Int </c>
     </x>

rule <k> run2 => frag(<x>-fragment A B noCCell </x>-fragment) ... </k>
     <x>
       A:ACell
       B:BCell
       <c> _:Int </c>
     </x>

endmodule
```

dwightguth:
I can vaguely recall trying this in the past and running into some kind of problem on the C semantics. Can you test to make sure that make test-build still passes?

bmmoore:
I'm testing it now. I wouldn't be too surprised if the C semantics needs some work to move away from the old assumption that cell fragments are Bags.

bmmoore:
@daejunpark please review.

It did turn out to be a one-line change in the code.

daejunpark:
lgtm

bmmoore:
@dwightguth A bunch of errors seem to all involve the `<curr-scope>` cell, starting with these two

```
[Error] Critical: Unexpected sort Scope for term
#SemanticCastToScope(#token(#KVariable,"OldScope")). Expected FileScope.
    Source(/home/brandon/code/k/c-semantics/semantics/././language/common/dynamic.k)
    Location(208,24,208,38)
[Error] Critical: Unexpected sort Scope for term Scope. Expected FileScope.
    Source(/home/brandon/code/k/c-semantics/semantics/././language/common/dynamic.k)
    Location(212,29,212,34)
```

and then many instances like these of parse errors on uses of that cell:

```
[Error] Inner Parser: Parse error: unexpected character '<'.
    Source(/home/brandon/code/k/c-semantics/semantics/././language/translation/decl/global.k)
    Location(111,32,111,33)
[Error] Inner Parser: Parse error: unexpected character '<'.
    Source(/home/brandon/code/k/c-semantics/semantics/././language/translation/decl/global.k)
    Location(123,32,123,33)
[Error] Inner Parser: Parse error: unexpected character '<'.
    Source(/home/brandon/code/k/c-semantics/semantics/././language/translation/decl/global.k)
    Location(53,32,53,33)
[Error] Inner Parser: Parse error: unexpected character '<'.
Source(/home/brandon/code/k/c-semantics/semantics/././language/translation/decl/global.k)
Location(61,32,61,33)
```

There seem to be a bunch of unrelated errors from not allowing ",," anymore.

bmmoore:
@dwightguth Considering I don't change at all the items expected inside the generated production, are you sure `test-build` works correctly on the current master of K?

dwightguth:
Oh, I guess maybe it doesn't. Try running it from the following pull
request: https://github.com/kframework/c-semantics/pull/120

bmmoore:
Jenkins: test this please.

bmmoore:
The errors seem to come from C still using `.Bag` and `:Bag` for the absence of cells and to annotate cell bags or fragments, which is an error in the C semantics that will have to be fixed.

bmmoore:
This commit leaves one remaining parsing ambiguity in the `thrd-join` rule in the threads library of C, but it happens to be resovled in favor of `#cells`, so test-build passes even without #1791 after adding this commit.

dwightguth:
What's the ambiguity? And will #1791 fix it?

bmmoore:
The ambiguity is between parsing a pair of identical cells, like the `<thread>...</thread><thread>...</thread>` using either production

```
Bag ::= Bag Bag [klabel("#cells")]
```

from the loose front-end syntax, or

```
ThreadCellBag ::= ThreadCellBag ThreadCellBag [klabel("_ThreadCellBag_")]
```

from the more strict generated syntax.

The latest commit vastly reduces the scope of the ambiguity by making the generated front end syntax of parent cells parse the bodies as Bag rather than K, so the second production can't apply, but an ambiguity is still possible if this situation appears at the very top of a rule, which is parsed as K. This happens in the rule I mentioned above - and that's the only case in the C semantics.

The problem #1791 fixes is that SortCells would fail with an exception on the second parse. It doesn't touch parsing, but it does fix SortCells to handle either parse equivalently. `test-build` only passes without #1791 because the ambiguity is coincidentally resolved in favor of the parse as `#cells`.

bmmoore:
@daejunpark Do you have any comments on the commits since your approval?

bmmoore:
Jenkins: test this please.

daejunpark:
lgtm.

dwightguth:
looks fine to merge to me.

daejunpark:
@bmmoore I cannot find this merge in the latest master. For example, there is no cell_arguments directory in the regression test:
https://github.com/kframework/k/tree/master/k-distribution/tests/regression/kore-cell-fragments

Do you know what happened?

bmmoore:
The lost merge has been sorted out.

PR #1755, opened 2015-10-01
a patch to evade unnecessary AssertionError on KList when kompile

laurayuwen:


laurayuwen:

dwightguth:
is this actually correct, though? fail() isn't really an error; it just means that the two terms don't match. an assertion error means that calling the function in this particular way is not valid. where is the context where the error happened that you are trying to fix, and which interpretation is correct in that case?

laurayuwen:
The assertion error made the following fail to kompile. It used to kompile before. 

syntax KListWrap ::= "[" KList "]"
syntax KItem ::= getLastKListElement( KListWrap )                       [function]
rule getLastKListElement([Ho:K,, Hd:K,, Tl:KList]) => getLastKListElement([Hd,,Tl])
rule getLastKListElement([Hd:K,, .KList]) => Hd
rule getLastKListElement([.KList]) => .K

dwightguth:
This is not allowed in KORE anyway. KLists are not first order terms anymore so you can't have a KList variable. If you replace this code with something legal that uses a List type, do you still get the error?

andreistefanescu:
@dwightguth the assertion violation occurs at `add` call sites from the code inherited by `PatternMatcher` from `AbstractUnifier`. Changing how `add` works is simpler then duplicating the code from `AbstractUnifier` and adapting it to fail instead of calling `add`. 
@laurayuwen looks good

PR #1756, opened 2015-10-01
several fixes to ocaml backend to make it work better with kdebug

dwightguth:
@bmmoore @msaxena2 please review

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
@@ -145,9 +145,9 @@ public class InitializeRewriter implements Function<Module, Rewriter> {
         }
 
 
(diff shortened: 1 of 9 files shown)


dwightguth:
Note that there is still a bug dealing with set and map choices that has an effect on KDebug, but I don't have time to fix it right now; I'll get to it after the evaluation is ready.

msaxena2:
lgtm.

bmmoore:
Looks good.

I found the logic of the new Transformer class in DeconstructIntegerAndFloatLiterals a bit subtle - passinmg the constructor argument striaght through as the lhs variable works only given the invariant that side conditions do not contain rewrites. You might consider adding a comment somewhere around Transformer or convert/convertLookups in DeconstructIntegerAndFloatLiterals.java, but as a private class used only in private methods it's sufficiently encapsulated that I won't insist.

PR #1762, opened 2015-10-05
fix IntelliJ warnings

yilongli:
do not review yet


yilongli:
Nothing major. I already restrain myself to only fixing the most obvious warnings.

@daejunpark should I include you in pull requests regarding the java-backend? I am not entirely clear about who is in charge of what at this moment :)

andreistefanescu:
lgtm

daejunpark:
@yilongli You don't need to include me. :) I'm just auditing changes to keep update.

yilongli:
@dwightguth ?

java-backend/src/main/java/org/kframework/backend/java/symbolic/KILtoSMTLib.java
```diff
@@ -503,7 +502,7 @@ public ASTNode transform(KItem kItem) {
             // smtlib expression instead of operator
             String expression = label;
             for (int i = 0; i < kList.getContents().size(); i++) {
-                expression = expression.replaceAll("\\#" + (i + 1) + "(?![0-9])", ((SMTLibTerm) kList.get(i).accept(this)).expression());
+                expression = expression.replaceAll("#" + (i + 1) + "(?![0-9])", ((SMTLibTerm) kList.get(i).accept(this)).expression());
```
dwightguth, line 505: you're sure # is not a special character in Java regexes?

dwightguth:
Does this remove 100% of the unused imports in the codebase? If so, we should probably add a checkstyle for that...

java-backend/src/main/java/org/kframework/backend/java/symbolic/KILtoSMTLib.java
bmmoore, line 505: Only under special flag settings:

http://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#COMMENTS

dwightguth:
okay lgtm but you should probably also look into the checkstyle thing so that we don't have regressions of unused imports.

yilongli:
@dwightguth Nope, this pull request doesn't address all the unused imports. I only run the scan over the java-backend module.

dwightguth:
oh okay. lgtm then.

PR #1767, opened 2015-10-06
fix ocaml backend on windows

dwightguth:
@bmmoore please review


bmmoore:
Looks fine. Was the temp directory always '.'?

dwightguth:
well it would have worked on linux before because we are setting the working directory on the same line. but it broke on windows because of the forward slash

PR #1774, opened 2015-10-14
Fix inconsistent TermContext counters

yilongli:
do not review yet


yilongli:

dwightguth:
okay it looks like this change has undone the fact that we no longer had redundant counter values in spawn.imp. Why is that?

Also,  if you are going to go with a long rather than a BigInteger then you need to make the rewriter throw an error if the counter overflows, otherwise you may end up with configurations reusing the value of a fresh counter.

yilongli:
@dwightguth what is the redundant counter value you are talking about?

As for the counter, I really don't think our rewriter will ever get to point it can overflow.

yilongli:
Closing it because I don't think it's a good idea to share the same counter between instances of TermContext anymore. It just introduces more non-determinism when performing search, which makes CFG generation unnecessarily harder.

PR #1777, opened 2015-10-15
Fix TermContext.counter

yilongli:
do not review yet


yilongli:

This pull request only cleans up the code related to `TermContext.counter`. I am leaving the fix for clashing counter values during CFG generation to the next pull request. On a high level, the changes here try to reduce the number of `TermContext` instances created. Besides, I avoid creating `TermContext.counter` as much as possible in order to reduce the risk of clashing counter values.

java-backend/src/main/java/org/kframework/backend/java/kil/TermContext.java
```diff
@@ -1,52 +1,65 @@
...
     private final GlobalContext global;
 
-    private Term topTerm;
-    private ConjunctiveFormula topConstraint;
+    /**
+     * This class shall not be accessed by more than one thread,
+     * so it is made non-thread-safe intentionally.
+     */
+    private static class FreshCounter implements Serializable {
+        private long value;
```
dwightguth: I disagree with your assessment that overflowing a long can never happen. It would require a very long running process, but imagine if we use RV-Match to run some server that does not terminate but simply accepts new requests? If the program runs long enough, eventually a counter may overflow because the fresh counter will be incremented every single time a variable is allocated on the stack, which could be many, many times per request.

dwightguth:
The changes you are making to conflate GlobalContext and TermContext do not seem right to me. A single GlobalContext is supposed to remain unchanged across multiple calls to the Rewriter interface. InitializeRewriter relies on this in order for it to work correctly. By adding the counter to that class you are making a change that I can't predict the effects of and which may introduce very subtle bugs. Can you explain why you thought it was necessary to do this?

java-backend/src/main/java/org/kframework/backend/java/kil/TermContext.java
yilongli: OK, I can change it back to `BigInteger`.

yilongli:
@dwightguth please re-review

java-backend/src/main/java/org/kframework/backend/java/kore/compile/ExpandMacros.java
```diff
@@ -123,7 +123,8 @@ private Context expand(Context context) {
     }
 
     public K expand(K term) {
-        TermContext tc = TermContext.of(rewriter.rewritingContext);
+        // YilongL: is it correct to create a fresh TermContext every time?
```
dwightguth: yes that is actually fine because macro expansion never invokes any functions and therefore cannot be impacted in any way by the value of the counter.

dwightguth:
looks okay but you should get Andrei to review also.

java-backend/src/main/java/org/kframework/backend/java/kore/compile/ExpandMacros.java
yilongli: This is what I thought before so I initially did not associate a counter with those TermContext's that are used for macro expansion. However, several tests failed because of that: http://office.runtimeverification.com:8080/rv-jenkins/job/k-framework-pull-requests/org.kframework.k$k-distribution/2092/testReport/junit/(root)/_var_lib_jenkins_jobs_k-framework-pull-requests_workspace_k-distribution_tutorial_1_k_5_types_lesson_7/tricky_2/. Looks like macro expansion can involve generating fresh constants.

dwightguth: I believe this is a bug. function evaluation is supposed to occur only
after macro expansion has finished, and macro expansion is not supposed to
have any notion of functions or side conditions. This was something decided
relatively recently though so the Java backend was not made in alignment
with this decision. ExpandMacros.java is currently only used by the OCAML
backend.

yilongli: @andreistefanescu can you take a look at the failing test? It seems to be related to the your binder substitution.

PR #1780, opened 2015-10-16
Separate termcontext whenever possible

yilongli:
do not review yet


yilongli:

There are two major changes in this pull request:
- replace `TermContext` with `GlobalContext` as much as possible; the key insight here is that `KItem` doesn't need the fresh counter inside `TermContext` to compute its sort so that a `TermContext` could be recreated from the `GlobalContext` to feed to `MetaK.matchable`
- modify `CopyOnWriteTransformer` to use the `GlobalContext` carried by the parameter of the `transform` method by default so that most substitution related operations do not have to take a `TermContext` or `GlobalContext` argument

dwightguth:
Are you sure this is actually safe to do? GlobalContext is actually not safe to serialize in a lot of cases. Originally it was not intended to be serialized at all but Andrei persuaded me to let it be serialized in a really hackish way that means that you can get NullPointerExceptions really easily. I think it works right now because we are not actually currently serializing any java kil classes, but that is likely to change for performance reasons eventually. I would feel much better if you either did not include Global Context as part of terms, or found a way to recover the correct GlobalContext for a term following serialization.

yilongli:
@dwightguth I am fully aware of the serialization problem of `GlobalContext`. But the previous version that carries a `TermContext` is only worse. I will keep looking for a way to eventually eliminate `GlobalContext` from terms but, for now, people should just properly set the `GlobalContext` (e.g., using a `CopyOnWriteTransformer`) inside the initial term at the beginning of krun whenever necessary.

yilongli:
@dwightguth The fundamental problem of `TermContext` and/or `GlobalContext` is the way we invoke builtin functions: it always take a `TermContext` as an argument even if the function is pure and the argument is not used. In `KItem`, the `GlobalContext` is used to compute its sort by `MetaK.matchable/unifiable`. I cannot imagine that such operation should involve evaluating any impure function (probably not even pure function?). Therefore, I think `GlobalContext` can be removed from `KItem` and thus `BuiltinMap/Set/List` but I just don't have more time to work on it now.

dwightguth:
Technically that's not true though. I can write a rule like:

```
rule isFoo(foo(I:Int)) => someImpureFunction(I)
```

yilongli:
I think it's safe to require sort predicate rules to be side-effect free.

yilongli:
@andreistefanescu could you take a look at this pull request too?

java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
```diff
@@ -74,13 +74,13 @@ public boolean equals(Object obj) {
 
     private final TermContext context;
 
-    public ConstrainedTerm(Term term, ConjunctiveFormula constraint) {
+    public ConstrainedTerm(Term term, ConjunctiveFormula constraint, TermContext context) {
         this.data = new Data(term, constraint);
-        this.context = constraint.termContext().fork();
+        this.context = context.fork();
```
andreistefanescu: are you still creating new contexts here?

java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
```diff
@@ -59,32 +59,32 @@
     private Sort sort;
     private Set<Sort> possibleSorts;
     private transient boolean enableCache; // for lazy computation
-    private transient TermContext termContext; // for lazy computation
+    private final GlobalContext global; // for lazy computation
```
andreistefanescu, line 62: could we have `Definition` instead of `GlobalContext` here?

java-backend/src/main/java/org/kframework/backend/java/symbolic/KILtoSMTLib.java
```diff
@@ -213,6 +214,9 @@ public static String translateImplication(
         return sb.toString();
     }
 
+    private final Definition definition;
+    private final GlobalContext global;
```
andreistefanescu: since `definition` is its own field, create a field for the options and delete `global`

andreistefanescu:
lgtm

java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
yilongli: Yeah, I forgot that this change was made in a later commit. But I have now included it in this PR.

java-backend/src/main/java/org/kframework/backend/java/kil/KItem.java
yilongli, line 62: Eventually yes, I think. But now it is a bit complicated. It's explained in my previous reply to Dwight.

java-backend/src/main/java/org/kframework/backend/java/symbolic/KILtoSMTLib.java
yilongli: done

PR #1791, opened 2015-10-21
Allow cell bag union labels in SortCells

bmmoore:
SortCells would previously fail some inputs using the
labels from a cell bag sort. This commit changes the pass to
recognize those labels as acceptable items under a parent cell,
and to handle them correctly when re-assembling the sorted contents.

@dwightguth please review.
This also allows the C semantics to build under #1754 -
the sort changes there added ambigous parses using the cell bag labels,
and when one of those parses was chosen the build would fail because
of this problem.

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/compile/KOREtoBackendKIL.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/compile/KOREtoBackendKIL.java
@@ -161,9 +161,9 @@ public class KOREtoBackendKIL extends org.kframework.kore.AbstractConstructors<o
             org.kframework.kore.KList klist = ((KApply) k).klist();
             if (useCellCollections && definition.configurationInfo().getCellForConcat(klabel).isDefined())
                 return KLabelInjection.injectionOf(CellCollection(klabel, klist), context);
-            if (useCellCollections && definition.configurationInfo().getCellForUnit((KApply) k).isDefined())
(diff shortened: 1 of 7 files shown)


dwightguth:
I don't see anything wrong but I also don't completely follow what this change is doing. Can you post an example of the rule whose behavior has been changed and explain why it didn't work before and what you did differently?

bmmoore:
The problem this addresses is rules failing to compile, not behavior changes. If input containing two adjacent copies of a multiplicity \* cell is parsed as a term using the cell collection's concatenation operator instead of the generic `#cells` term, the SortCells pass would previously crash rather than recognizing that as content that belongs in the appropriate slot of the completed term.

I ran into those cases from work on #1754 affecting parsing, but I think we need this code anyways in case the user explicitly writes a term in labeled form. I've since extended #1754 to eliminate most of the potential parses that triggered this bug, but one that remains is at https://github.com/kframework/c-semantics/blob/master/semantics/library/threads.k#L74 . The term like this

```
<thread>... A ...</thread> <thread>... B ...</thread>
```

can parse using either the production

```
Bag ::= Bag Bag [klabel("#cells")]
```

from the loose front-end syntax, or

```
ThreadCellBag ::= ThreadCellBag ThreadCellBag [klabel("ThreadCellBag")]
```

from the grammar generated from the configuration.

bmmoore:
Now that Jenkins finished, do you have any objections?

PR #1794, opened 2015-10-21
several minor fixes to make ocaml backend somewhat faster

dwightguth:
@bmmoore please review


bmmoore:
Looks good, besides not updating that comment.

PR #1796, opened 2015-10-22
Tidy Transformer stuff

bmmoore:
Directly call Sets.union instead of going through the silly little
wrappers used as internals in the Transformers system.

Extract helper methods for returning a warning-less result or error.

@cos, @dwightguth please review.

@cos, I changed `ChildrenMapping` from a trait to an abstract class only because if I didn't then the `simpleResult` and `simpleError` methods would end up as unimplemented abstract methods in the abstract classes that extended the trait, instead of picking up the trait's implementation.

### DIFF
--- a/kernel/src/main/java/org/kframework/parser/concrete2kore/ParseInModule.java
+++ b/kernel/src/main/java/org/kframework/parser/concrete2kore/ParseInModule.java
@@ -127,7 +127,7 @@ public class ParseInModule implements Serializable {
         getGrammar();
 
         Grammar.NonTerminal startSymbolNT = grammar.get(startSymbol.name());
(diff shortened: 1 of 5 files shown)


dwightguth:
Lgtm

radumereuta:
lgtm

cos:
the silly little wrappers were actually abstraction allowing the `Transformer` to be used for cases where the merge is not set union. In any case, we'll eventually transition these steps to kore or even rewriting, so there's no point in fussing about it now. lgtm.

bmmoore:
Yeah, those methods definitely had a use as internals in the Transformer implementation, it's just silly to call them from outside when it's just unioning sets.

PR #1797, opened 2015-10-22
Fix nulls in Modules

bmmoore:
Stop passing null to Module constructor, add assert, declare @Nonnull

These three calls are the reason toString often fails on modules while debugging.

@dwightguth, @radumereuta please review.


dwightguth:
Lgtm

radumereuta:
lgtm

PR #1798, opened 2015-10-22
Fix List terminators

### LINKED ISSUE #1682: List Parsing adding terminator badly

bmmoore:
Unit tests and serveral fixes for adding list terminators. Should resolve #1682.

@radumereuta, @dwightguth, @msaxena2 please review.


dwightguth:
looks okay but I haven't really examined it too closely. Have you tested this on SIMPLE?

radumereuta:
I don't get it, this was supposed to be simple.
Just find places in concrete syntax where you were supposed to have list terminators, and add a list terminator in that place. You didn't have to worry for example for cast term, yet I see some checks that handle such cases.

Could you write a few lines and try to explain what you are trying to do there?

bmmoore:
I'm trying to actually put the list terminators everywhere they are needed, and not put them anywhere they are not needed, both of which the previous code failed at.

About explicitly considering casts, an annotated variable often parses as an application of a production with label `#SyntacticCast` or `#SemanticCastToSort`, applied to a token of Sort `KVariable`. Adding a terminator between the label and variable would be a bad idea, which is why that check is needed and why it was there all along.

The other big issue is that anything written in labeled form is represented at this point in the parser as an application of a production with label `#KApply`, which doesn't carry nearly enough sort information to decide whether a terminator is needed around the result. Actually, I think I'm still not correctly adding terminators around the arguments of a label application.

radumereuta:
Why aren't you doing this at the TermCons level?
And secondly, this looks very ugly. Lots of steps that I had implemented in K3 looked a lot more simple written with KIL. With KORE looks very verbose and hard to work with. I have a strong feeling we are doing something wrong.

bmmoore:
I am doing this at the TermCons level, read more carefully.

One obstacle is that the pass lowering labels like `#KApply` to actual terms is only written for Kore. It also does not calculate a specific sort or production for overloaded labels, like a list cons.

bmmoore:
Jenkins: test this please.

msaxena2:
As far as the problem with SIMPLE (mentioned in #1682 ) is concerned, this solves the issue. LGTM.

bmmoore:
Jenkins: test this please.

dwightguth:
lgtm

bmmoore:
While I do agree the it's messier than I would like, I think this is the best we can do without a bigger redesign of parse tree handling, and we should get a fix for this issue into master. I've opened up #1828 to track the cleanup.

PR #1805, opened 2015-10-22
fix kore translation bug

daejunpark:


daejunpark:
@bmmoore please review.

bmmoore:
Looks good. I checked you do get inferred sort `Bool` on that rule.

PR #1807, opened 2015-10-26
add brandon's compiler patch

dwightguth:
@bmmoore please review

Let me know if you have a better suggestion for where to store the patch file.

@pdaian FYI. If we go this route and put the patch file here it's going to be very important that it doesn't get deleted or things will break.

### DIFF
--- a/k-distribution/src/main/scripts/bin/k-configure-opam
+++ b/k-distribution/src/main/scripts/bin/k-configure-opam
@@ -1,0 +1,6 @@
+#!/bin/sh
+opam init
+opam repository add k "$(dirname "$0")/../lib/opam"
+opam update
+opam switch 4.03.0+trunk
+opam install mlgmp zarith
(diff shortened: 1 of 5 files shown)


bmmoore:
Try using opam variables - the interpolation syntax like `%{lib}` should also work in the patch name. There's nothing quite as nice as a variable specifically for the current repo, but we could probably get by hardcoding the "k" name into something like `%{root}/repo/k/compilers/4.03.0/4.03.0+k/files/compiler-patch.patch`, if you add the patch at `files/compiler-patch.patch` under the switch definition in the k repo.

https://opam.ocaml.org/doc/Manual.html#Variables

Please investigate this first, before we host the patch ourselves. Hosting the patch is fine if there are no other options, but making sure it stays up will be annoying.

bmmoore:
k-configure-opam-dev is not marked executable

dwightguth:
It actually doesn't have to be because the mode is set by the assembly plugin and the script isn't intended to be run out of the src/main/scripts dir.

bmmoore:
Ok, but it's still a bit odd that the scripts differ on that.
I do see a this message in the middle of running the scripts:

```
[ERROR] k is already a remote repository
```

but it still updated the files and built the compiler ok.
(I didn't see an obvious option to suppress that message)
Looks good enough.

PR #1808, opened 2015-10-26
Binder, lambda

laurayuwen:
Dwight made some fix


dwightguth:
lgtm but @bmmoore and @grosu should maybe review the changes to lambda++.

bmmoore:
The definition is fine (I'm assuming it passes the updated tests). It is `1_lambda`, not `3_lambda++`.

The only issue is one we've seen in other definitions - using modules to control the visibility of Id concrete syntax means we can't write the rules between syntax definitions anymore, but we don't have any idea how to do better yet.

PR #1810, opened 2015-10-27
stop variable type inference from inferring variables as KList

dwightguth:
@bmmoore please review


radumereuta:
lgtm, but what is the reason for this? I don't see any reference to an issue or a discussion.

dwightguth:
It's a bug that arises when you write the following rule in the java backend:

```
syntax Int ::= foo(Int) [function]

rule `foo`(_) => 0 [owise]
```

bmmoore:
Code is fine, but please add a comment explaining the signficance of the tests - that sorts above K or at and below KBott are assumed to only come from the K-in-K grammar from kast.k.

radumereuta:
Also, are there any unit tests for this?

dwightguth:
Jenkins: test this please

bmmoore:
lgtm

PR #1811, opened 2015-10-28
Support -L in kcc

dwightguth:
@bmmoore please review

### DIFF
--- a/kernel/src/main/java/org/kframework/krun/KRunOptions.java
+++ b/kernel/src/main/java/org/kframework/krun/KRunOptions.java
@@ -312,5 +312,9 @@ public final class KRunOptions {
         @Parameter(names="--native-libraries", description="Native libraries to link the rewrite engine against. Useful in defining rewriter plugins.",
                 listConverter=StringListConverter.class)
         public List<String> nativeLibraries = Collections.emptyList();
+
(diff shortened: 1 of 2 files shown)


bmmoore:
Looks good. Can you add a comment about the options. I assume you were trying to add `--whole-archive` to be sure functions are available for dynamic use, but I think that's a pretty obscure ld flag (is there any such thing as a famous ld flag?), and the necessary multiple layers of argument escaping with `-cclib -Wl,` certainly don't help readability.

dwightguth:
I mean it works fine with GNU ld the issue is just portability. I reverted that portion because I didn't want to have to deal with the fact that that code hadn't yet been made to work on all platforms. If you look at the diff as a whole, that stuff is no longer part of what is actually getting merged.

Do you still want some kind of comment? I mean, the code is pretty straightforwardly equal to the description of the options on the command line.

bmmoore:
If you are not using that option anymore, then I'd just like a little comment mentionin that escaping with both `-cclib` and `-Wl,` is necessary to pass linker options through the ocaml compiler and then through the C compiler and on to the linker (or however exactly it works). I definitely didn't recognize `-cclib`.

dwightguth:
Okay, will do.

dwightguth:
@bmmoore please re-review

bmmoore:
It's very clear now.

PR #1814, opened 2015-10-30
Modification to Pom and Scripts for compatibility with OS X El Capitan

### LINKED ISSUE #1813: Building K fails on OS X El Capitan
calling mvn package on OS X El Capitan fails with the following error message:

```
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, <<< FAILURE! - in org.kframework.backend.java.symbolic.UseSMTTest
testGetModel(org.kframework.backend.java.symbolic.UseSMTTest)    <<< ERROR!
java.lang.UnsatisfiedLinkError: no z3java in java.library.path
    at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1865)
    at java.lang.Runtime.loadLibrary0(Runtime.java:870)
    at java.lang.System.loadLibrary(System.java:1122)
    at com.microsoft.z3.Native.<clinit>(Native.java:10)
    at com.microsoft.z3.Context.<init>(Context.java:24)
    at org.kframework.backend.java.symbolic.UseSMT.checkSat(UseSMT.java:44)
    at org.kframework.backend.java.symbolic.UseSMTTest.testGetModel(UseSMTTest.java:52)
```

msaxena2:
Fixes #1813 . WIP, Do not Review yet.


msaxena2:
Jenkins: test this please.

msaxena2:
Jenkins: test this please

k-distribution/src/main/scripts/lib/setenv
```diff
@@ -8,7 +8,7 @@ fi
 K_LIB_DIR="$(cd $(dirname "$0")/../lib; pwd)"
 if [ `uname` = 'Darwin' ]; then
   export PATH="$K_LIB_DIR/native/osx":$PATH
-  export DYLD_LIBRARY_PATH=$DYLD_LIBRARY_PATH:"$K_LIB_DIR/native/osx"
+  export LIBRARY_PATH="$K_LIB_DIR/native/osx"
```
dwightguth, line 11: this needs to include the existing DYLD_LIBRARY_PATH in order to maintain consistency with the old functionality.

k-distribution/src/main/scripts/lib/setenv
```diff
@@ -8,7 +8,7 @@ fi
...
 elif [[ `uname` == *MINGW* || `uname` == *MSYS* || `uname` == *CYGWIN* ]]; then
   export PATH="$K_LIB_DIR/native/windows":"$K_LIB_DIR/native/windows$ARCH":"$K_LIB_DIR/native/$ARCH":$PATH
```
dwightguth, line 13: you should export LIBRARY_PATH=$PATH here.

k-distribution/src/main/scripts/lib/setenv
```diff
@@ -16,7 +16,7 @@ else
     echo "Warning: unsupported OS detected. Defaulting to Linux mode, but errors may occur." 1>&2
   fi
   export PATH="$K_LIB_DIR/native/linux":"$K_LIB_DIR/native/linux$ARCH":$PATH
-  export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:"$K_LIB_DIR/native/linux$ARCH"
+  export LIBRARY_PATH="$K_LIB_DIR/native/linux$ARCH"
```
dwightguth, line 19: same here, LIBRARY_PATH needs to include the existing LD_LIBRARY_PATH.

msaxena2, line 19: @dwightguth I agree with that this would allow us to maintain consistency with the old system, but the point of this PR was to abandon modifying/using the LD_LIBRARY_PATH and DYLD_LIBRARY_PATH in favor of using the -Djava.library.path argument to the jvm itself. It seems like the solution didn't work as some of the tests Jenkins ran failed due to lack of availability of required shared libraries. At this point, I think we may have to look for another solution altogether; one that doesn't require the script to deal with el Capitan separately from other versions of OS X.

cos:
I didn't have any trouble on El Capitan. Does this break compatibility with pre-El Capitan by any chance?

dwightguth:
It shouldn't. For info on the issue see here: https://forums.developer.apple.com/message/31148

If you had System Integrity Protection disabled you might not have seen this issue.

Obviously this fix seems to be broken though so we will have to figure out how to fix this at some point.

PR #1815, opened 2015-11-03
fix java-semantics performance problem

yilongli:
@dwightguth @laurayuwen please review

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
@@ -37,7 +37,7 @@ public class SymbolicRewriter {
     private boolean transition;
     private final RuleIndex ruleIndex;
     private final KRunState.Counter counter;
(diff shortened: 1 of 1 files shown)


dwightguth:
Looks like there is another issue with the ordering of search solutions here. I think we should probably try to lexicographically sort the results in order to avoid this issue. But feel free to defer that to a later pull request if you want.

dwightguth:
lgtm otherwise

laurayuwen:
lgtm. The speed is normal now.

dwightguth:
@yilongli feel free to un-revert the change involving parallel streams also if you want.

yilongli:
@dwightguth I don't understand why the ordering of the search solutions is not sorted. In [PrintSearchResult](https://github.com/kframework/k/blob/master/kernel/src/main/java/org/kframework/backend/unparser/PrintSearchResult.java), the solutions are put into a `TreeSet`. Any idea?

dwightguth:
I believe that that test is running on the KORE pipeline where the same thing has not been done yet. Take a look at KRun.printSearchResult.

yilongli:
OK, it's a bit difficult to sort the search results in the KORE pipeline so I just revise the code to be more deterministic.

PR #1819, opened 2015-11-03
Kil Kill

dwightguth:
**please review**


dwightguth:
@andreistefanescu @cos @bmmoore @radumereuta @daejunpark @msaxena2 @laurayuwen  @grosu please review

### things I deleted
- sdf parser
- old krun, kompile, kast frontends
- old compiler
- old debugger, executor, model checker, prover frontends
- old unparser
- unparser and unflatten backend
- `--kore` flag (now the default)
- dead code arising from deletion of the above

### things I did not delete
- html, latex, pdf, doc, coq backends
- checks that need to be rewritten for kore
- kil hierarchy

### things I moved
- various static things that were part of a class I deleted but used by the new pipeline
- various classes that were part of a package that I deleted but were used by the new pipeline

### things I commented out
- all the kil tests in ktest. None of them pass still because their definitions have not yet been updated. I think we should make a major push following this merge to add back as many definitions as possible to improve test coverage..

I suggest you review the diff one commit at a time. I separated out the deletions from the changes to Java code, and from the modified ktest xml files. The deletions commit contains only deletions so you should be able to review it easily on the command line if it gets cut off. Just do something like "git show 612db45 --name-only" and it will spit out a complete list of deleted files.

If we need to bring anything back, we can always do that. I didn't comment out any of the KORE tests and as you can see they all still pass, so I don't think anything got deleted that was part of the KORE pipeline. But there are certainly things that got deleted that will need to be converted into the new pipeline, like collection normalization and lexicographic sorting of search results.

grosu:
I have checked the three commits except the deletions.  I hope we will add the examples back asap after we merge; I'll help with that, too.  Otherwise, LGTM.

@andreistefanescu @cos @bmmoore @radumereuta @daejunpark @msaxena2 @laurayuwen please review ASAP, so we can merge.

radumereuta:
I think this is a very good first cleanup.
There are still a few other things that we have to remove from the old implementation, i.e. the KIL dependency between Outer.jj and KORE, but that should be done separately, and after this.

lgtm apart from a few comments.

cos:
lgtm. it also merged almost cleanly into the pattern-match-automaton branch.

dwightguth:
Yes, I still want to do some package renames but those merge a lot less well so I left them out when I remade the diff. We can do that when there aren't so many big changes outstanding.

cos:
Noticed it doesn't contain the previous package renames. Thanks for that!

msaxena2:
lgtm.

andreistefanescu:
lgtm

daejunpark:
lgtm

cos:
@bmmoore, @laurayuwen, are you ok with merging?

bmmoore:
Looks good to merge.

From deleted attributes I'm not sure "variable" and "nelist" won't eventually be useful, but they are trivial to restore.

PR #1823, opened 2015-11-08
Cell fragment

daejunpark:
do not review.


daejunpark:
@bmmoore please review. Note that I implemented this by using your existing implementation as a black box, since as Grigore said, cell-fragment will be eventually re-implemented. Regarding cell parsing, I reported an issue #1842, which will be handled at a separate PR. Until then, I used a simple workaround of declaring additional production.

cos:
@bmmoore, as you ok with the change? Is it ok for @daejunpark to merge?

bmmoore:
I really dislike how this works by adding and removing dummy cells (and treating k specially in the process). I'm trying to work out if this pass can be done more cleanly.

cos:
@bmmoore, have you figured out a cleaner solution in the meantime by any chance? Otherwise, can we merge this for now and add an issue to track that this needs improvement?

bmmoore:
Please see #1890 for my revisions.

daejunpark:
@bmmoore @cos please review. We already reached the agreement to merge it for now, so please merge this long-standing (for a month!) PR asap. @cos, I had merge conflicts due to your transformer update. Please take a look if I made a right change: fca4f30

cos:
fca4f30 looks good to me. I agree we should merge.

bmmoore:
Merging this solution as agreed. Hopefully this can be simplified as in #1890, once I get that working.

PR #1830, opened 2015-11-09
fix hidden warning

dwightguth:
@bmmoore please review


bmmoore:
lgtm. I've noticed including `String2Id` in `EMPTY-ID` pollutes the concrete syntax of languages, but we can fix that in another request.

dwightguth:
Yes that's a good point. Probably we want ID, EMPTY-ID-SYNTAX, and ID-SYNTAX

dwightguth:
Jenkins: test this please

PR #1844, opened 2015-11-13
Simplify step-counting loop in SymbolicRewriter

bmmoore:
I thought that loop could be a bit clearer.

@dwightguth Please review.


dwightguth:
looks okay but I want @yilongli to also take a look.

yilongli:
lgtm. Could you check if `PatternMatchRewriter` can also benefit from such refactoring?

bmmoore:
No, that's already a nice for loop. It might disagree with this code on the convention for step number passed to the "compute step" function, but that's not the issue here.

PR #1845, opened 2015-11-13
add -I flag to add directory to search path

dwightguth:
@bmmoore @cos please review

### DIFF
--- a/k-distribution/include/builtin/prelude.k
+++ b/k-distribution/include/builtin/prelude.k
@@ -1,0 +1,2 @@
+require "kast.k"
+require "domains.k"
--- a/k-distribution/include/builtin/prelude.k
+++ b/k-distribution/include/builtin/prelude.k
@@ -1,2 +1,3 @@
+// Copyright (c) 2015 K Team. All Rights Reserved.
 require "kast.k"
 require "domains.k"
--- a/k-distribution/src/test/java/org/kframework/kore/convertors/TstTinyOnKORE_IT.java
+++ b/k-distribution/src/test/java/org/kframework/kore/convertors/TstTinyOnKORE_IT.java
@@ -58,8 +58,7 @@ public class TstTinyOnKORE_IT {
                 "<T>(<k>(while(_)_(_<=_(0,n:Id),{_}(__(_=_;(s:Id,_+_(s:Id,n:Id)),_=_;(n:Id,_+_(n:Id,-1)))))),<state>(_Map_()))");
     }
 
-    // tiny backend does not work with domains.k
-    @Test @Ignore
+    @Test @Ignore("tiny backend does not work with domains.k")
     public void simpleNestedFunctions() throws IOException, URISyntaxException {
         executeTest("FUNC", "FUNC",
                 "`foo`(bar)",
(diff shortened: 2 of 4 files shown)


bmmoore:
looks good, If @cos is okay with the remaining test coverage on the tiny backend.

cos:
I'm ok with removing that test of tiny. 

Still, I don't like the implicit `requires domains.k`. Adding one line to your definition is not such a big deal. I prefer less magic here. Worst case, add another option for "implicit requires" and default it to `domains.k`. This will provide a way to not use the implicit `require`.

dwightguth:
K definitions literally do not work without domains.k because they rely on having a configuration which relies on having a Map. I would prefer to not have one more way that a novice developer can write code that doesn't compile for no particular gain. If you don't want to use the implicit kast.k and domains.k, you can always provide your own in the following manner:

```
mkdir my_builtins
vim test.k
vim my_builtins/domains.k
vim my_builtins/kast.k
kompile test.k -I my_builtins
```

cos:
`domains.k` is not part of the K language. Let's try to avoid adding something to the language which can be solved by either writing one more line in a definition, or having an extra command line parameter.

dwightguth:
Again though, I say, the K framework does not work without the Map module to implement the functionality  required by configuration variables. I see no reason why we should require a specific line of code to be present in every single K definition when we can make it implicit instead. That sounds like "part of the language" to me.

Also I should note there is actually precedent for even more automatically included things than this. Java automatically imports the java.lang package, OCAML automatically imports the Pervasives module, gcc automatically imports /usr/include/stdc-predef.h. Maude automatically requires files containing prelude modules as well, but does not import them, perhaps being the most analogous case here.

cos:
> Again though, I say, the K framework does not work without the Map module to implement the functionality required by configuration variables. I see no reason why we should require a specific line of code to be present in every single K definition when we can make it implicit instead. That sounds like "part of the language" to me.

Not all languages require `Map`. Furthermore, if you really want to get rid of that line, I have you a specific simple solution that does not involve hardwiring this in the compilation pipeline.

> Also I should note there is actually precedent for even more automatically included things than this. Java automatically imports the java.lang package, OCAML automatically imports the Pervasives module, gcc automatically imports /usr/include/stdc-predef.h. Maude automatically requires files containing prelude modules as well, but does not import them, perhaps being the most analogous case here.

Yes, many languages do it. Some of them have been bitten by it and are trimming it down, e.g., Scala. In our case, we have simple solution not to do it. Why insist on it then?

P.S. I also realize the futility of arguing over this detail. Unfortunately, we seem to have gotten used to this exaggerated attention to detail, sometimes in the detriment of higher-level goals. I hope we'll move away from it soon.

dwightguth:
I mean,

```
$ cat test.k
module TEST
  configuration <k> $PGM:K </k>

endmodule
$ kompile test.k
[Error] Compiler: Module Map must be visible at the configuration declaration,
in module TEST
```

But whatever, not worth arguing about. I have added back the no-prelude flag even though it's basically useless since we are only requiring files that contain modules that the tool crashes without. Can I merge this now?

cos:
The dependence on Map is an artifact of the current implementation.

Yes, I'm ok with merging.

dwightguth:
For what it's worth, I have no objection to removing domains.k from prelude.k if it stops being required in order for semantics to work.

bmmoore:
I think reasonable default includes and a way to suppress them is the best compromise for most languages.

PR #1846, opened 2015-11-13
move ocaml backend to proprietary repo

dwightguth:
@pdaian please review


dwightguth:
@pdaian looks like it successfully picks up the artifact for the OCAML backend from the maven repo where it was built by the rv-match repo. Please review.

PR #1848, opened 2015-11-15
test branch

cos:


cos:
Jenkins: test C semantics

cos:
@dwightguth, I know it is for a different reason, but the c semantics seems to be failing on master also.

PR #1853, opened 2015-11-16
auto format everything

cos:


cos:
Jenkins, test this please

java-backend/src/main/java/org/kframework/backend/java/kil/Sort.java
```diff
@@ -69,8 +68,7 @@
      * Gets the corresponding {@code Sort} from its {@code String}
      * representation.
      *
-     * @param name
-     *            the name of the sort
+     * @param name the name of the sort
      * @return the sort
      */
     public static Sort of(String name) {
```
radumereuta, line 74: The older version of this file, looked a lot better before.

java-backend/src/main/java/org/kframework/backend/java/symbolic/PatternMatcher.java
```diff
@@ -414,8 +409,8 @@ public void unify(BuiltinSet builtinSet, BuiltinSet patternSet) {
     @Override
     public void unify(CellCollection cellCollection, CellCollection otherCellCollection) {
         if (cellCollection.hasFrame()) {
-        // TODO(dwightguth): put this assertion back in once this class is constructed by
-        // the injector
+            // TODO(dwightguth): put this assertion back in once this class is constructed by
+            // the injector
 //            assert !termContext.definition().context().javaExecutionOption/*s.concreteExecution() :
 //                "the subject term should be ground in concrete execution";*/
```
radumereuta, line 415: These two lines aren't auto formatted. Why?

kernel/src/main/java/com/davekoelle/AlphanumComparator.java
```diff
@@ -102,25 +94,20 @@ public int compare(String o1, String o2)
...
+                        if (result != 0) {
                             return result;
                         }
                     }
                 }
-            } else
-            {
+            } else {
                 result = thisChunk.compareTo(thatChunk);
             }
```
radumereuta, line 113: This is external code, should we touch it? What does the License say?

kernel/src/main/java/org/kframework/compile/checks/CheckSyntaxDecl.java
```diff
@@ -69,7 +69,7 @@ else if (!(pi instanceof Terminal)) {
             }
         } else if (!node.containsAttribute(Constants.FUNCTION)
                 && (node.getSort().equals(Sort.K) ||
-                    node.getSort().equals(Sort.KLIST))) {
+                node.getSort().equals(Sort.KLIST))) {
```
radumereuta, line 72: I liked my alignment better... (see final comment).

kernel/src/main/java/org/kframework/kil/visitors/Visitor.java
```diff
@@ -5,106 +5,181 @@
...
+
     public R visit(Attribute node, P p) throws E;
+
     public R visit(KInjectedLabel kInjectedLabel, P p) throws E;
+
     public R visit(FreezerLabel freezerLabel, P p) throws E;
+
     public R visit(Freezer f, P p) throws E;
+
     public R visit(BackendTerm term, P p) throws E;
```
radumereuta, line 176: This would have looked better if it were more compact like before.

cos:
test (don't mind me)

radumereuta:
In `Grammar.java` the javadoc is broken because you didn't add the `<pre>` tag like @bmmoore suggested. (I don't know how to add comments there because the PR is so big)

Also, I'm still not convinced by this idea. Yes, there are lots of places where formatting can be improved, but an equal amount, where automatic formatting is worse than before.

Thoughts?

java-backend/src/main/java/org/kframework/backend/java/symbolic/PatternMatcher.java
cos, line 415: it probably considered that non-aligned comments are meant to start at the beginning of the line

java-backend/src/main/java/org/kframework/backend/java/kil/Sort.java
cos, line 74: while it could be argued as such for some cases like this one, they are few and far between

kernel/src/main/java/org/kframework/compile/checks/CheckSyntaxDecl.java
cos, line 72: the problems are the inconsistency of `&&` and `||`. They should be either at the beginning or the end of the line, not alternate.

cos:
Forgot to include the `<pre>` in this latest diff. It works as @bmmoore said. 

The question is of degrees. Are there any places where automatic formatting makes the code unreadable? Are there any places where manual formatting would make the code _much_ more readable? If the answer is no to both questions, then I would say that the gain from auto-formatting (i.e., the "auto") thumps any small inconveniences.

radumereuta:
Problems in files which don't show on the diff:  
**_Grammar.java**_:
38: use `<pre>` to keep the alignment in the comments. The ASCII schematics is important for understanding the code. **_This is really important to fix**_  
338: is it possible to have an option in the auto-formatter to keep single line comments when possible? That looks kind of excessive. Java is already verbose, making it occupy even more space doesn't help.  
**_Parser.java**_
338: Again, lots of comments that could be single line.

cos:
> is it possible to have an option in the auto-formatter to keep single line comments when possible? 

Yes, I'll activate it.

cos:
@radumereuta, this PR has been open for review for 11 days already. Do you consider it reviewed?

bmmoore:
As long as this is still open I suggest adding the "Field annotations" option "Do not wrap after single annotation", which would let `@Nonnull` be used without excesive wrapping.

radumereuta:
Well, I commented on lots of stuff, I don't see any commit that fixes the issues.

radumereuta:
There, I fixed the tags, but you shouldn't merge like this. Also you need to merge master into this branch.

cos:
@radumereuta, "fixing" the issues is not a simple commit but a force push that would likely "unhook" some of your comments. That's why I didn't fix anything. And yes, tagging with "reviewed" lets me know I can do this.

I'm actually leaving the actual merge for the end of the year so that we have all the kore stuff merged -- fewer chances for strange conflicts.

cos:
will do this when activity is very low at the end of the year.

PR #1858, opened 2015-11-17
inline definition

### LINKED ISSUE #1857: Lookup directories
I have the following folder structure:

```
project
 \_ src
 |    \_ common
 |    \_ rvm
 |    |   \_ rvm-syntax.k
 \_ test
```

In rvm-syntax I have something like

```
require "domains.k"
require "../common/common-syntax.k"
...
```

If I go to src/common/rvm folder and try to kompile rvm-syntax it works fine, however If I go to test folder and try to kompile ../src/rvm/rvm-syntax.k I get:

```
Lookup directories:[./../src/rvm,
/home/ali/FSL/kbuild3/lib/java/../../include/builtin]
    Source(./../src/rvm/rvm-syntax.k)
    Location(2,1,2,35)

```

I'm not sure it is a problem or a feature!

cos:
should fix #1857

### DIFF
--- a/kernel/src/main/java/org/kframework/kompile/Kompile.java
+++ b/kernel/src/main/java/org/kframework/kompile/Kompile.java
@@ -283,7 +283,7 @@ public class Kompile {
         }
         Definition definition = parser.loadDefinition(
                 mainModuleName,
-                mainProgramsModule, prelude + "require " + StringUtil.enquoteCString(definitionFile.getPath()),
+                mainProgramsModule, prelude + FileUtil.load(definitionFile),
                 Source.apply(definitionFile.getPath()),
                 definitionFile.getParentFile(),
                 ListUtils.union(kompileOptions.includes.stream()
--- a/kernel/src/main/java/org/kframework/kompile/Kompile.java
+++ b/kernel/src/main/java/org/kframework/kompile/Kompile.java
@@ -5,6 +5,7 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import com.google.inject.Inject;
 import org.apache.commons.collections15.ListUtils;
+import org.apache.commons.io.FileUtils;
 import org.kframework.Collections;
 import org.kframework.attributes.Source;
 import org.kframework.builtin.BooleanUtils;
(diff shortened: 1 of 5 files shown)


kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -276,14 +277,14 @@ public Module parseModule(CompiledDefinition definition, File definitionFile, bo
...
-    public Definition parseDefinition(File definitionFile, String mainModuleName, String mainProgramsModule, boolean dropQuote) {
+    public Definition parseDefinition(File definitionFile, String mainModuleName, String mainProgramsModule, boolean dropQuote) throws IOException {
         String prelude = REQUIRE_PRELUDE_K;
         if (kompileOptions.noPrelude) {
             prelude = "";
         }
         Definition definition = parser.loadDefinition(
                 mainModuleName,
-                mainProgramsModule, prelude + "require " + StringUtil.enquoteCString(definitionFile.getPath()),
+                mainProgramsModule, prelude + FileUtils.readFileToString(definitionFile),
```
dwightguth: don't bubble up exceptions like this, it's bad design. Use FileUtil.load instead.

cos: It's actually bad design not to... the point of the exception mechanism is to bubble them up to the level you want to address them. The fact we need our own `FileUtil` in the first place is a sign of bad design. But anyhow, it's already there, so I adjusted it for the sake of consistency and moving forward.

dwightguth: Yes I agree however there are serious problems associated with checked exceptions that make it quite a mess from a code maintenance perspective to have to pass checked exceptions through many levels of the call stack. If this becomes an issue in the future we can create some kind of UncheckedIOException class or something, catch the checked exception in the FileUtil class and rethrow it as an unchecked exception, then catch it appropriately higher up. But this is good enough for now.

dwightguth:
Can you add a test for this case?

cos:
Where should the test live? Are there any other tests for krun behavior?

dwightguth:
probably in k-distribution/tests/regression. That's where most of the one-off tests are. You can make it an integration test.

cos:
most of those are run through ktest which takes care of paths. Are there any other naked krun tests I can use as a model?

dwightguth:
Hmm. No I guess we don't really have anything like that. But it shouldn't be too hard to mock something up that calls Kompile and KRun. The tricky part will be creating the directory structure since you don't control JUnit's working directory.

cos:
I would prefer to not break new ground here now -- not enough time for it. Are you ok with merging without the test?

PR #1859, opened 2015-11-18
Clean up symbolic rewriter

yilongli:
do not review yet

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/kil/Definition.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/kil/Definition.java
@@ -51,7 +51,6 @@ import java.util.stream.Collectors;
 
 import static org.kframework.kore.KORE.Sort;
 import static org.kframework.Collections.*;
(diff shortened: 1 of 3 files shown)


yilongli:
I mistakenly thought that the `FastRuleMatcher` already worked for unification so I was trying to connect it to my CFG generation. Anyway, I separated the potentially useful change in this pull request. @andreistefanescu please review

andreistefanescu:
Why is side condition evaluation moved in the matcher?

yilongli:
Because I think the automata should eventually take into account side conditions of the rules. For example, the same side condition should be evaluated only once for different rules. For now, it only simplifies the code on the SymbolicRewriter side a little bit.

andreistefanescu:
You can merge it, but this part of the code is work in progress, so it is very likely to go through several changes in the future.

dwightguth:
@yilongli just a note for the future I agree we should eventually be able to use an automata to evaluate side conditions, but it will have to happen only for functions that are not marked impure and do not recusively call functions marked impure.

PR #1865, opened 2015-11-20
Feature/refactoring of scala use

cos:


cos:
@dwightguth, please have a look.

cos:
review after #1871 is merged

cos:
I adde a `*new` migration method and another one crashed. This is not a viable path.

dwightguth:
After talking with @grosu we have concluded that the best path here is to continue with the migration pattern that we have talked about but to only make changes to the K API (ie the semantics and signature of public methods that when changed cause downstream projects to break) if a clear case can be made that some kind of change is needed and cannot be achieved any other way. Given that I think long term, if we're going to be substantially changing the signatures of these methods, we should be using classes that are not part of the Scala standard library, I would be interested in hearing why it is necessary to change these signatures from one Scala class to another.

dwightguth:
Jenkins: test this please

dwightguth:
So to summarize offline conversations, we've decided to make the OCAML backend depend on a stable beta release of K so that K can evolve without worrying about these types of constraints. I will be responsible for periodically releasing betas in order to keep the OCAML backend up to date with relevant changes to K.

cos:
Jenkins, test this please.

PR #1871, opened 2015-11-27
compile refactoring

cos:
**please review**

PR #1872, opened 2015-11-27
basic strategies

cos:


cos:
builds upon #1871

cos:
@msaxena2 , I ignored several of your debugger tests as they are flaky, i.e., they give different results on different runs. The extra non-determinism is exposed (it was there before, just not visible) by executing kore_imp using heating/cooling strategies.

To fix this, either make the tests robust using the same definition or make a copy of `kore_imp.k` which does not activate strategies -- i.e. use the version just before this PR.

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
```diff
@@ -101,12 +101,14 @@ public synchronized Rewriter apply(Module module) {
 
     public static class SymbolicRewriterGlue implements Rewriter {
 
-        private final SymbolicRewriter rewriter;
+        private SymbolicRewriter rewriter;
```
dwightguth, line 104: can you explain why this was done?

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -158,6 +165,55 @@ public KRunState rewrite(ConstrainedTerm constrainedTerm, int bound) {
...
     }
 
+    /**
+     * This method adds a #STUCK item on the top of the strategy cell of the stuck configuration and returns
+     * the resulting configuration. If the configuration already had the #STUCK flag, it returns Optional.empty()
+     * to end the rewriting.
+     */
+    private Optional<ConstrainedTerm> addStuckFlagIfNotThere(ConstrainedTerm subject) {
+        Optional<K> theStrategy = ((KApply) subject.term()).klist().stream()
+                .filter(t -> t instanceof KApply && ((KApply) t).klabel().name().contains("<s>"))
```
dwightguth: I think if we are going to privilege the `<s>` cell we should not do it by name but by an attribute like "strategycell" on the production similar to the "maincell" attribute.

k-distribution/tutorial/1_k/4_imp++/lesson_8/tests-kore/config.xml
```diff
@@ -4,7 +4,7 @@
 
     <include file="../../lesson_7/tests/config.xml"
              skip="pdf"
-             exclude="spawn"
+             exclude="spawn div"
```
dwightguth, line 7: is there a reason this was done instead of just fixing the test?

kernel/src/main/java/org/kframework/kore/compile/GenerateSentencesFromConfigDecl.java
```diff
@@ -202,6 +202,12 @@ public Void apply(KToken k) {
             }
             return null;
         }
+
+        @Override
+        public Void apply(KVariable k) {
+            hasConfigVar = true;
```
dwightguth, line 199: why was this done?

kore/src/main/scala/org/kframework/Strategy.scala
```diff
@@ -0,0 +1,51 @@
...
+  val addStrategyCellToRulesTransformer =
+    DefinitionTransformer(
+      ModuleTransformer.fromSentenceTransformer({
+        (module, r) =>
+          val rich = kore.Rich(module)
+
+          import rich._
+
+          // hackish test for whether we're using strategies or not
+          // TODO: replace with something more principled
```
dwightguth: I'd prefer this TODO get resolved...

cos:
@dwightguth, this is work in progress. It would be great if you could have a look at #1865 first.

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
cos, line 104: I needed the KOREtoBackendKIL convertor in SymbolicRewriter (because that's the primary constructor of Java backend KORE objects), but it takes as an argument the `useCellCollections` flag which is different between execution and search/proving.

k-distribution/tutorial/1_k/4_imp++/lesson_8/tests-kore/config.xml
cos, line 7: I think this test, like a few others, fails non-deterministically.

kore/src/main/scala/org/kframework/Strategy.scala
cos: Resolved. After thinking about it, I figured this is actually a very simple and clean way of activating strategies.

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
cos: I cannot really do that considering our current configuration infrastructure. I define the strategy language separately with referring to the `<s>` cell. It would be hard to rename it afterwards. And I don't really want to add lots of code just for this -- it would be ugly.

Also, strategies will be very used, so the `<s>` cell may get a privileged status like the `<k>` cell does.

In any case, for now the strategy implementation is experimental. We can get back to it once we have better configuration inheritance support.

kernel/src/main/java/org/kframework/kore/compile/GenerateSentencesFromConfigDecl.java
cos, line 199: Because the Init should be passed downwards when encountering any kind of variable, not only config ones. It allows `<k> ... </k> InitSCell(Init)`.

cos:
@dwightguth, have a look to make sure it doesn't affect OCaml or the C semantics.

java-backend/src/main/java/org/kframework/AddConfiguraitonRecoveryFlags.java
```diff
@@ -0,0 +1,45 @@
...
+import org.kframework.kore.KORE;
+import org.kframework.kore.KVariable;
+import scala.Function1;
+
+import java.util.function.UnaryOperator;
+
+/**
+ * Puts a {@link Att#refers_THIS_CONFIGURATION()} and {@link Att#refers_RESTORE_CONFIGURATION()} marker on rules that do.
+ */
+public class AddConfiguraitonRecoveryFlags implements UnaryOperator<Sentence> {
```
dwightguth: looks like a typo?

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -160,6 +162,7 @@ private void checkDefinition(Definition parsedDef) {
 
         if (!errors.isEmpty()) {
             kem.addAllKException(errors.stream().map(e -> e.exception).collect(Collectors.toList()));
+            kem.print();
```
dwightguth: this doesn't look right. The only call to kem.print() should be in FrontEnd.java when execution ends. Why was this done?

cos:
The PR adds an experimental implementation for strategies. For now, the strategies can only determine when rules are apply, not where in the term. They are activated by importing (transitively) the `STRATEGY` module. `STRATEGY` has the primitives, while `BASIC-STRATEGY` is the implementation of a simple strategy language. There are two tests, one very small checking the functionality, and another controlling heating/cooling for `IMP` via strategies. The experimental `--heat-cool-by-strategies` flag determines to activate strategy-driven heating/cooling.

kernel/src/main/java/org/kframework/krun/KRunOptions.java
```diff
@@ -192,7 +192,7 @@ public OutputModeConverter(String optionName) {
     }
 
     @Parameter(names="--search", description="In conjunction with it you can specify 3 options that are optional: pattern (the pattern used for search), bound (the number of desired solutions) and depth (the maximum depth of the search).")
-    private boolean search = false;
+    public boolean search = false;
```
dwightguth, line 195: why was this made public?

dwightguth:
I don't obviously see anything here that should break the OCAML backend or the C semantics, but it's harrd to be 100% sure. @msaxena2 can you test this diff against the C semantics on the Java backend and make sure that it doesn't seem to introduce any regressions in the point up to where you are currently stuck?

kernel/src/main/java/org/kframework/kompile/Kompile.java
cos: For debugging. I will remove it. Generally, we should really fix the implementation to show the actual errors by default, not useless "I found 4 errors but I won't show them to you.".

kernel/src/main/java/org/kframework/krun/KRunOptions.java
cos, line 195: To set it without injection. As far as I understand DI, all classes should be fully usable without it. Having a private field with no setter or constructor parameter, thus only settable via injection, violates this.

kernel/src/main/java/org/kframework/kompile/Kompile.java
dwightguth: You should get both. If you aren't seeing the 4 errors it says it found, something really strange is going on. Do you have something I can use to reproduce that behavior?

kernel/src/main/java/org/kframework/krun/KRunOptions.java
dwightguth, line 195: okay, that's fine.

kernel/src/main/java/org/kframework/kompile/Kompile.java
cos: I simply run integration tests as junit tests, e.g., `TstBackendOnKORE_IT` or `StrategiesTest`. It may very well be that I'm not creating the injection KEM object correctly somehow.

dwightguth: oh, I see. Yeah the issue here then is that you are receiving the exception that halts execution but aren't actually printing out any other errors and warnings the KEM already contains. We should probably consider making a JUnit Rule that dumps the errors somewhere useful if the test fails.

cos: ok, I see. I'll add before after to make a fresh KEM and dump it afterwards. I created an issue for this: #1899.

I also remember Guice has some special functionality for unit tests. Feel free to pick it up if you already know how to use it.

kernel/src/main/java/org/kframework/krun/KRunOptions.java
grosu, line 195: But you should still put a one-line comment there, so others, or even you, will not wonder the same thing later on.

kernel/src/main/java/org/kframework/kompile/Kompile.java
grosu: Same here, in case you decide to keep it.  It takes only 5 seconds to add a comment "//for debugging", and that saves others and you time later.

kernel/src/main/java/org/kframework/krun/KRunOptions.java
cos, line 195: The idea here is that all functionality should be available without injection. So having the method public (or otherwise available) is the default case. If we do decide to explicitly make something available only via injection (i.e., private here) for whatever reason, we could annotate that as a special case.

kernel/src/main/java/org/kframework/kompile/Kompile.java
cos: I won't keep it. Still, I should have added the comment in the first place to mark it is temporary.

andreistefanescu:
Looks good.

PR #1873, opened 2015-11-28
configuration inheritance

cos:


dwightguth:
I don't have anything particularly against this approach, but it's worth noting that we already had code to do something similar involving something along the lines of:

```
configuration <k> $PGM:K </k>
configuration <T> <k/> </T>
```

I like your approach better because it gives more flexibility, but you should go through the code and strip out the old stuff if we're going to do this.

grosu:
Guys, this is an important design aspect of K.  I'd like to know what this is about.  We cannot just implement things without designing them first.  We need to have a longer term plan described somewhere (the manual?) and then implement that.  This is not the sort of thing that you design "as you go".

cos:
I need some way to specify partial configurations in imported modules to make strategies work without labeled syntax. I do not care that much about syntax (we can decide that afterwards) but we will need this feature one way or another. My solution here is actually very simple and doesn't introduce any special constructs (not even `<k/>` -- though I would have used it if I would have known about it) as it makes use of our decided-upon de-sugaring of configuration declarations. Have a look at [this unit test](https://github.com/kframework/k/pull/1873/files#diff-6b4f96434e198514e2b5bccd9f571b11).

Also, this is not ready for review.

k-distribution/src/test/resources/compiler-tests/configuration-inheritance.k
```diff
@@ -0,0 +1,25 @@
...
+    configuration <b> foo </b>
+
+    rule <b> foo => bar </b>
+endmodule
+
+module C
+    imports A
+    imports B
+
+    configuration <t> initKCell(Init) initBCell </t>
```
grosu, line 22: @cos, regardless of how much you needed this, you cannot just change the design of K without a discussion and without documenting it.  It is not crystal clear what is here anyway.  Some of the decisions seem adhoc, without a principle underneath.  Let's discuss this in the K meeting today.

cos, line 22: They just seem ad-hoc. Also, again, this is not ready for review.

daejunpark:
Dwight, could you tell me in what aspect Cosmin's approach gives more
flexibility than your approach? I didn't see any difference in terms of the
expressiveness power.

Daejun

dwightguth:
Theoretically Cosmin's approach gives control over the map passed to an
initializer. You could use "Init" to refer to the variables, or you could
construct any other map you choose. I don't believe this is possible with
the other approach.

cos:
waiting for #1886

cos:
the main changes:
1. the generated productions pass in Init if the cell has any (not only configuration) variables inside
2. the parsing is layered such that the configuration productions from imported modules are used in parsing the current configuration

dwightguth:
As I said, if you're going to go with this approach you are going to need to strip out the code from the old approach. I believe you will also need to persuade @grosu that we are fine going with this design. Other than that this seems okay, though.

cos:
@dwightguth, I already confirmed with @grosu that he's ok with allowing the generated productions to interleave with configuration syntax, which is essentially what this PR adds. It essentially doesn't add anything, it just fixes the fact that this wasn't allowed before.

Regarding the `<s/>` style syntax, I actually kind of like it. It looks more user-friendly than `initSCell(Init)`. I wouldn't strip it out before we decide on a final user-friendly version (which may even be `<s/>`). I'll add an issue to track this.

grosu:
Yes, I confirm that @cos and I discussed this and I agreed with it.  I think users should have the full capability of KORE in K, and that in particular means to use the configuration initialization macros manually.  So I agree with @cos that it should have worked from the beginning, so this is only fixing it.  Regarding `<s/>`, I also like it, but we need to explain its semantics on the wiki.  How do you desugar it in the LHS of a rule?  In the RHS?
(@cos: edited to fix formatting)

cos:
@grosu, it is desugared to either `initSCell` or `InitSCell(Init)` for the purpose of generating the configuration productions for the current module. These productions, along with the imported ones, are used by the configuration concretization steps down the pipeline.

PR #1875, opened 2015-11-29
rename rule variables in the new term

laurayuwen:
@andreistefanescu I feel this two line is ok, we don't really need to extract a method.


andreistefanescu:
lgtm

cos:
@laurayuwen, could you merge master into your branch so we can test and merge this back?

laurayuwen:
@cos is this ok?

PR #1878, opened 2015-11-30
add kdep tool to assist with makefiles

dwightguth:
@bmmoore @cos please review

### DIFF
--- a/kernel/src/main/java/org/kframework/parser/concrete2kore/ParserUtils.java
+++ b/kernel/src/main/java/org/kframework/parser/concrete2kore/ParserUtils.java
@@ -1,7 +1,6 @@
 // Copyright (c) 2015 K Team. All Rights Reserved.
 package org.kframework.parser.concrete2kore;
 
-import com.google.common.collect.HashMultimap;
 import com.google.common.collect.SetMultimap;
 import org.apache.commons.io.FileUtils;
 import org.kframework.attributes.Source;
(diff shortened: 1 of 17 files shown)


cos:
@dwightguth, could you remove the "Created by dwightguth" comments and add documentation for the new classes. I will re-review once there is some description of what they do.

dwightguth:
@cos please re-review

kernel/src/main/java/org/kframework/parser/concrete2kore/ParserUtils.java
```diff
@@ -109,17 +111,37 @@ public static Module parseMainModuleOuterSyntax(String definitionText, Source so
         return kilToKore.apply(def).getModule(mainModule).get();
     }
 
-    private List<org.kframework.kil.Module> slurp(
+    public static class SlurpedDefinition {
```
cos: what is the purpose of this class? It is not used anywhere and it seems odd that it mixes file system dependencies with resolved module.

dwightguth: It's used by KDepFrontEnd to compute the list of files in the definition.

cos: ok, I see now, the .modules part is used by the compile pipeline while the .dependencies by the kdep frontend. It seems a bit ugly, especially as it makes the rather involved `ParserUtil` even more complex. Can't the kdep frontend simply traverse the modules to get their Sources? They are generated anyhow.

dwightguth: It will be faster to use the information that gets computed as the files are traversed than to try to traverse everything a second time. Speed is pretty important for this tool because it runs on pretty much every invocation of make.

cos: The cost of one traversal of the module structure is completely negligible compared to the cost of parsing the modules. I strongly prefer to not let even more unnecessary complexity sneak into the K interfaces.

dwightguth:
Alright. Since this happens only at kompile-time and the difference only comes out to about 360 milliseconds per build of the C semantics, I will accept the simpler solution that I didn't think of. I was thinking that I would have to iterate through all the sentences which would take a lot longer, but I guess I don't.

PR #1882, opened 2015-12-03
check variables on rhs are bound on left

dwightguth:
@cos please review

### DIFF
--- a/kernel/src/main/java/org/kframework/kompile/Kompile.java
+++ b/kernel/src/main/java/org/kframework/kompile/Kompile.java
@@ -23,7 +23,19 @@ import org.kframework.definition.Sentence;
 import org.kframework.kore.K;
 import org.kframework.kore.KApply;
 import org.kframework.kore.Sort;
-import org.kframework.kore.compile.*;
+import org.kframework.kore.compile.AddImplicitComputationCell;
+import org.kframework.kore.compile.ConcretizeCells;
+import org.kframework.kore.compile.GenerateSentencesFromConfigDecl;
+import org.kframework.kore.compile.GenerateSortPredicateSyntax;
+import org.kframework.kore.compile.ResolveAnonVar;
+import org.kframework.kore.compile.ResolveContexts;
+import org.kframework.kore.compile.ResolveFreshConstants;
+import org.kframework.kore.compile.ResolveHeatCoolAttribute;
+import org.kframework.kore.compile.ResolveIOStreams;
+import org.kframework.kore.compile.ResolveSemanticCasts;
+import org.kframework.kore.compile.ResolveStrict;
+import org.kframework.kore.compile.SortInfo;
+import org.kframework.kore.compile.checks.CheckRHSVariables;
 import org.kframework.main.GlobalOptions;
 import org.kframework.parser.TreeNodesToKORE;
 import org.kframework.parser.concrete2kore.ParseCache;
@@ -130,12 +142,25 @@ public class Kompile {
         Definition parsedDef = parseDefinition(definitionFile, mainModuleName, mainProgramsModuleName, true);
         sw.printIntermediate("Parse definition [" + parsedBubbles.get() + "/" + (parsedBubbles.get() + cachedBubbles.get()) + " rules]");
 
+        checkDefinition(parsedDef);
+
         Definition kompiledDefinition = pipeline.apply(parsedDef);
         sw.printIntermediate("Apply compile pipeline");
 
         ConfigurationInfoFromModule configInfo = new ConfigurationInfoFromModule(kompiledDefinition.mainModule());
 
         return new CompiledDefinition(kompileOptions, parsedDef, kompiledDefinition, programStartSymbol, configInfo.getDefaultCell(configInfo.topCell()).klabel());
+    }
+
+    private void checkDefinition(Definition parsedDef) {
+        CheckRHSVariables checkRHSVariables = new CheckRHSVariables(errors);
+        stream(parsedDef.modules()).forEach(m -> stream(m.localSentences()).forEach(checkRHSVariables::check));
+
+        if (!errors.isEmpty()) {
+            kem.addAllKException(errors.stream().map(e -> e.exception).collect(Collectors.toList()));
+            throw KEMException.compilerError("Had " + errors.size() + " structural errors.");
+        }
+
     }
 
     public Function<Definition, Definition> defaultSteps() {
@@ -242,7 +267,7 @@ public class Kompile {
             loader.saveOrDie(files.resolveKompiled("cache.bin"), caches);
         }
         if (!errors.isEmpty()) {
-            kem.addAllKException(errors.stream().map(e -> e.getKException()).collect(Collectors.toList()));
+            kem.addAllKException(errors.stream().map(e -> e.exception).collect(Collectors.toList()));
             throw KEMException.compilerError("Had " + errors.size() + " parsing errors.");
         }
         return parsedMod;
@@ -305,14 +330,14 @@ public class Kompile {
             loader.saveOrDie(files.resolveKompiled("cache.bin"), caches);
         }
         if (!errors.isEmpty()) {
-            kem.addAllKException(errors.stream().map(e -> e.getKException()).collect(Collectors.toList()));
+            kem.addAllKException(errors.stream().map(e -> e.exception).collect(Collectors.toList()));
             throw KEMException.compilerError("Had " + errors.size() + " parsing errors.");
         }
         return parsedDef;
     }
 
     Map<String, ParseCache> caches;
-    java.util.Set<ParseFailedException> errors;
+    java.util.Set<KEMException> errors;
     RuleGrammarGenerator gen;
 
     private Module resolveConfig(Module module, Definition def) {
--- a/kernel/src/main/java/org/kframework/kore/compile/RewriteAwareVisitor.java
+++ b/kernel/src/main/java/org/kframework/kore/compile/RewriteAwareVisitor.java
@@ -1,0 +1,35 @@
+package org.kframework.kore.compile;
+
+import org.kframework.kore.KRewrite;
+
+/**
+ * A visitor designed to track whether we are currently in the left hand side or right hand side of a term.
+ *
+ * This visitor provides two boolean methods, isRHS() and isLHS(). Outside of a rewrite, both are true, signifying
+ * that the term being visited is part of both the LHS and the RHS. Inside a rewrite, only one is true. It is an error
+ * for both to be false.
+ */
+public class RewriteAwareVisitor extends VisitKORE {
+
+    private boolean isRHS = true;
+    private boolean isLHS = true;
+
+    public boolean isLHS() {
+        return isLHS;
+    }
+
+    public boolean isRHS() {
+        return isRHS;
+    }
+
+    @Override
+    public Void apply(KRewrite k) {
+        isRHS = false;
+        apply(k.left());
+        isRHS = true;
+        isLHS = false;
+        apply(k.right());
+        isLHS = true;
+        return null;
+    }
+}
--- a/kernel/src/main/java/org/kframework/kore/compile/RewriteAwareVisitor.java
+++ b/kernel/src/main/java/org/kframework/kore/compile/RewriteAwareVisitor.java
@@ -12,8 +12,19 @@ import org.kframework.kore.KRewrite;
  */
 public class RewriteAwareVisitor extends VisitKORE {
 
-    private boolean isRHS = true;
-    private boolean isLHS = true;
+    public RewriteAwareVisitor(boolean isBody) {
+        if (isBody) {
+            isRHS = true;
+            isLHS = true;
+        } else {
+            isRHS = true;
+            isLHS = false;
+        }
+    }
+
+
+    private boolean isRHS;
+    private boolean isLHS;
 
     public boolean isLHS() {
         return isLHS;
--- a/kernel/src/main/java/org/kframework/kore/compile/RewriteAwareVisitor.java
+++ b/kernel/src/main/java/org/kframework/kore/compile/RewriteAwareVisitor.java
@@ -1,3 +1,4 @@
+// Copyright (c) 2015 K Team. All Rights Reserved.
 package org.kframework.kore.compile;
 
 import org.kframework.kore.KRewrite;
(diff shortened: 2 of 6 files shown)


grosu:
Of course, we want this to be a warning, because several definitions do add unbound variables in their RHS.  For example, see the Damas-Milner type system.  Those variables are then unified during execution, and thus we obtain narrowing, etc.

dwightguth:
@grosu in those cases we expect the user to use ? in the name of their variable. This allows the full generality while at the same time continuing to fail the build when the user has simply made a typo. This is the behavior the tool has had in the past and it is the same behavior the tool has in this pull request.

grosu:
Yes, true, they should use ?X for the RHS variable that is unbound.  We should tell them in the error message (sorry, I didn't check the code).  Mathematically, ?X is just like any undound variable in the RHS, but for practical reasons we require the ?.  Thanks for the clarification.

dwightguth:
The error message you would get if X is unbound on the rhs would be:

```
Found variable X on right hand side of rule, not bound on left hand side. Did you mean "?X"?
```

grosu:
perfect!

cos:
@dwightguth, considering the priority you give to improving your own code after merge, I am going to make sure the code laves little to be improved.

cos:
At a high level, my main complaint is that you add 190 lines in order to add make one very simple check.

kernel/src/main/java/org/kframework/kore/compile/RewriteAwareVisitor.java
```diff
@@ -0,0 +1,36 @@
...
+import org.kframework.kore.KRewrite;
+
+/**
+ * A visitor designed to track whether we are currently in the left hand side or right hand side of a term.
+ *
+ * This visitor provides two boolean methods, isRHS() and isLHS(). Outside of a rewrite, both are true, signifying
+ * that the term being visited is part of both the LHS and the RHS. Inside a rewrite, only one is true. It is an error
+ * for both to be false.
+ */
+public class RewriteAwareVisitor extends VisitKORE {
```
cos, line 13: While the class is nicely documented, is it really needed? Also, the name is slightly misleading. All K visitors are "aware" of rewrites. The better name for it would be "VisitorWithKRewriteLhsAndRhsMarkers". Still, the name being this verbose suggests that adding such a visitor for use by others (it is public) is making our code base and API larger with little benefit.

kernel/src/main/java/org/kframework/kore/compile/checks/CheckRHSVariables.java
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Checks a sentence to determine if it declares any variables in the RHS that are not bound in the LHS.
+ *
+ * More specifically, it performs the following checks: for each anonymous variable, if it is on the right hand side
+ * of a rewrite operator, and does not signify a fresh variable or constant, it is an error. For each non-anonymous
+ * variable, if it is on the right hand side of a rewrite operator, it is an error if it does not appear anywhere on the
+ * left hand side of the rule, and does not signify a fresh variable or constant.
+ */
+public class CheckRHSVariables {
```
cos, line 26: Even with the extra visitor, this class is way to verbose for its purpose. An 125 lines class for checking a simple property. I sketched a simpler solution here: https://github.com/kframework/k/blob/dwight/rhs/kernel/src/main/java/org/kframework/kore/compile/checks/CheckRHSVariables.java#L58-L80

The solution uses a new `FoldKIntoSet` class (added by #1886) which will be useful for many other passes, especially when we translate Radu's frontend.

kernel/src/main/java/org/kframework/kore/compile/checks/CheckRHSVariables.java
```diff
@@ -0,0 +1,125 @@
...
+    private final Set<KEMException> errors;
+
+    public CheckRHSVariables(Set<KEMException> errors) {
```
cos, line 29: Passing this parameter in is very C-style. Simply create a new `Set` inside and have a method for accessing it.

kernel/src/main/java/org/kframework/kore/compile/checks/CheckRHSVariables.java
```diff
@@ -0,0 +1,125 @@
...
+        this.errors = errors;
+    }
+    private void check(Rule rule) {
```
cos, line 32: Instead of `check(Rule...` and `check(Context...` (which may be enriched with others in the future, simply have a `DefinitionTransformer` pass that does nothing but this check, or use the `AbstractReflectiveVisitor`.

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -138,6 +152,17 @@ public CompiledDefinition run(File definitionFile, String mainModuleName, String
...
+        CheckRHSVariables checkRHSVariables = new CheckRHSVariables(errors);
+        stream(parsedDef.modules()).forEach(m -> stream(m.localSentences()).forEach(checkRHSVariables::check));
+
+        if (!errors.isEmpty()) {
+            kem.addAllKException(errors.stream().map(e -> e.exception).collect(Collectors.toList()));
+            throw KEMException.compilerError("Had " + errors.size() + " structural errors.");
+        }
+
+    }
+
```
cos, line 165: Let's not add any more low-level code to `Kompile` directly. Could this stay as a static method in `CheckRHSVariables`?

kernel/src/main/java/org/kframework/kore/compile/RewriteAwareVisitor.java
dwightguth, line 13: We are doing this in a bunch of places throughout the codebase. Having a central mechanism for all those places is always going to be better than duplicating code in each location. We can quibble about the name if you want, but we already have an enormous amount of copy-pasted code in the code base because we've been too afraid to centralize common mechanisms associated with writing compiler passes. I'd prefer to do something about that rather than continuing to let the problem get worse.

If you like, though, I can go through the codebase and try to find as many places as possible where this pattern is used and replace the existing code with this mechanism. I didn't in the interests of time constraints, but I can do it if it will set your mind at ease.

kernel/src/main/java/org/kframework/kore/compile/checks/CheckRHSVariables.java
dwightguth, line 26: It's not at all clear to me that your version even works correctly in all the cases mine has been written to handle. Most of the line length of what I wrote consists of things to deal with the lack of a useful visiting infrastructure for sentences or klabels, and while I would welcome solutions to these problems, so far as I can determine your code does not actually provide them. If you strip out all of this you end up with about 23 lines of quite simple code. That is actually less code than what you provide because your code is also 23 lines but does less actual percentage of the total work than mine. Your code is also considerably more complex and less straightforward to read. I am not particularly opposed to having a better infrastructure for visiting klabels and sentences in order to make this class less verbose, but your solution so far as I can tell primarily serves to make the implementation more complex for a gain that is not apparent to me, given that it does not appear to address the real source of complexity in the original file at all.

kernel/src/main/java/org/kframework/kore/compile/checks/CheckRHSVariables.java
dwightguth, line 32: if you can help me understand how to write a method in this fashion that does what I need, I would be willing to do it. But the property in question isn't a local property to K terms, it's a local property to sentences. So I don't see how DefinitionTransformer currently has the infrastructure to support this. And I don't really understand how AbstractReflectiveTransformer works, so if you want me to use it to do this you're going to have to explain how I can do so.

kernel/src/main/java/org/kframework/kore/compile/checks/CheckRHSVariables.java
dwightguth, line 29: having a single error set across multiple checks vastly reduces the need to tediously iterate through all the checks and retrieve their errors individually. If you can suggest to me a way that I can write this functionality in no more than 3 lines per check, while still maintaining the stateless property you are hoping for, I am open to suggestions. But this seems like the simplest solution to me, and it's not like it makes this class any harder to construct from outside of the pipeline. Hell we can have two constructors if you really want.

kernel/src/main/java/org/kframework/kompile/Kompile.java
dwightguth, line 165: what part of this exactly do you want to put in CheckRHSVariables? As far as I can tell if any single part of this function is moved there we will actually have to duplicate even more code as new checks are added.

kernel/src/main/java/org/kframework/kore/compile/checks/CheckRHSVariables.java
cos, line 26: > It's not at all clear to me that your version even works correctly in all the cases mine has been written to handle.

I haven't checked it but feel free to identify those cases. 

> Most of the line length of what I wrote consists of things to deal with the lack of a useful visiting infrastructure for sentences or klabels, and while I would welcome solutions to these problems, so far as I can determine your code does not actually provide them.

There is sufficient infrastructure for writing this in an elegant, short fashion.  `DefinitionTransformer.fromSentenceTransformer` can be used for visiting sentences -- you already did in many other cases. I'm not sure what you mean by the klabels part.

> If you strip out all of this you end up with about 23 lines of quite simple code. 

You already created an extra class to support your class. What more is needed to make your code 23 lines? I want to see it.

>  That is actually less code than what you provide because your code is also 23 lines but does less actual percentage of the total work than mine. 

What doesn't it cover. Please be specific. If it does miss something, I would be really surprised if anything it omits cannot be done in under 5 lines.

> Your code is also considerably more complex and less straightforward to read. 

How is this complex?
1. Gather all rewrites (5 lines)
2. Define a KVariable gatherer (5 lines)
3. Gather all KVariables on the lhs (1 line)
4. Gather all Variables on the rhs (1 line)
5. Check variables on the rhs against those on the lhs (10 lines essentially copy-pasted from your code).

Can you make this any more simple?

Yes, I also "cheated" here and used a new `FoldToKSet`. Without it I would have needed to use `VisitKORE` and mutability, which would have taken 2 more lines, one for each of the first two steps. 

So no, there is nothing fundamental stopping one from writing small, clear code with what is already available. The primary source of complexity in the original file is the complex approach to solving a very simple problem.

kernel/src/main/java/org/kframework/kore/compile/checks/CheckRHSVariables.java
cos, line 29: I don't know what you're talking about. Simply create a new `Set` inside `CheckRHSVariables` and make it a return of the `check` method. On the `Kernel` side, do set union.

kernel/src/main/java/org/kframework/kore/compile/checks/CheckRHSVariables.java
dwightguth, line 26: Your code does not correctly handle side conditions or klabel variables. I can use DefinitionTransformer.fromSentenceTransformer if you really want me to, but it does not make this class any smaller. This class would still need to specifically enumerate over sentence types and deal with the details of the different parts of a sentence, which your code does not currently do.

I added a class, yes, but it's a class dealing with functionality used in a number of different passes that need to be aware of whether they are on the rhs or lhs when visiting. To make the code roughly 23 lines, you would extend the infrastructure to deal with klabels correctly (which is needed regardless for correctness), and you would extend the infrastructure again to provide a common mechanism for deciding how to visit parts of a sentence.

kernel/src/main/java/org/kframework/kore/compile/checks/CheckRHSVariables.java
dwightguth, line 29: set union of what? we would still have to manually union each individual set together one at a time because there would be no collection to do any kind of flatMap over.

kernel/src/main/java/org/kframework/kore/compile/checks/CheckRHSVariables.java
cos, line 32: You can use `DefinitionTransformer.fromSentanceTransformer` for the traversal. Yes, `DefinitionTransformer.fromRuleBodyTransformer` doesn't include side conditions, but this is a good opportunity to add one transformer helper that does, say `DefinitionTransformer.fromRuleTransformer`, that would take `(K, K, K) -> (K, K, K)` (body, requires, ensures).

kernel/src/main/java/org/kframework/kore/compile/checks/CheckRHSVariables.java
dwightguth, line 29: Oh wait, never mind, I think I see what you're saying. Alright, but that would not work with multiple checks. You would have to update the visiting infrastructure as well to fold together the results of multiple checks. It wouldn't be as simple as an andThen anymore.

kernel/src/main/java/org/kframework/kore/compile/checks/CheckRHSVariables.java
dwightguth, line 32: Hmm. That could work I guess. But how would you propose to deal with the unblock restriction or the gatherVars call?

kernel/src/main/java/org/kframework/kompile/Kompile.java
cos, line 165: I don't think it is a good idea to add a method to `Kompile` for each new thing we do. As for exceptions, we should have a function somewhere else that adds all exceptions -- I've seen this pattern many time throughout the code base.

dwightguth, line 165: it's not "for each new thing" it's for a critical piece of the compilation process (sanity checks) that was entirely missing before. It seems to me more like arguing in favor of parseDefinition not being in this class either because it's "a new thing".

kernel/src/main/java/org/kframework/kore/compile/checks/CheckRHSVariables.java
cos, line 26: > Your code does not correctly handle side conditions 

Correct. It does now. Check it out.

> or klabel variables.

They are already visited by the fold.

> This class would still need to specifically enumerate over sentence types and deal with the details of the different parts of a sentence, which your code does not currently do.

We can definitely add this. See below.

cos:
Jenkins, test this please.

cos:
Jenkins, test this please

Issue #1866, opened 2015-11-20
Add support for Variable KLabels to the java backend.


msaxena2:
The java backend doesn't support KLabels that are also KVariables.

cos:
@msaxena2, did you get this to work?

msaxena2:
Not yet, but still working on it.

PR #1885, opened 2015-12-05
Support for matching patterns that have variable KLabels.

msaxena2:
Intended solution for #1866 . Not yet ready for review. @cos, feel free to have a look, but bear in mind this is still a work in progress.


kore/src/main/scala/org/kframework/kore/compile/MergeRules.scala
```diff
@@ -77,7 +77,7 @@ class MergeRules(c: Constructors[K]) extends (Module => Module) {
       .map { p => (normalizeKSeq(p._1), p._2) }
 
     val disjunctionOfKApplies: Iterable[(K, K)] = normalizedTerms
-      .collect({ case (x: KApply, ruleP) => (x, ruleP) })
+      .collect({ case (x: KApply, ruleP) if ! x.klabel.isInstanceOf[KVariable] => (x, ruleP) })
```
cos, line 80: it seems you could do away with all the changes in this file

cos, line 80: it seems to me that the new code achieves exactly the same thing as the old one

java-backend/src/main/java/org/kframework/backend/java/kil/RuleAutomatonDisjunction.java
```diff
@@ -52,13 +59,23 @@ public RuleAutomatonDisjunction(List<Pair<Term, BitSet>> children, GlobalContext
...
+                    KItem item = ((KItem) p.getLeft());
+                    if (item.kLabel() instanceof Variable) {
+                        Integer arity = new Integer(item.klist().size());
+                        if (!varKItemDisjunctionsMap.containsKey(arity)) {
+                            varKItemDisjunctionsMap.put(arity, new ArrayList<>());
+                        }
+                        varKItemDisjunctionsMap.get(arity).add(new ImmutablePair<>((KItem) p.getLeft(), p.getRight()));
+                    } else {
+                        this.kItemDisjunctionsArray[((KLabelConstant) ((KItem) p.getLeft()).kLabel()).ordinal()] = (Pair<KItem, BitSet>) (Object) p;
+                    }
```
cos, line 76: this looks good

java-backend/src/main/java/org/kframework/backend/java/symbolic/FastRuleMatcher.java
```diff
@@ -184,10 +190,9 @@ private BitSet match(Term subject, Term pattern, BitSet ruleMask, scala.collecti
 
         if (subject instanceof KItem && pattern instanceof KItem) {
             KItem kitemPattern = (KItem) pattern;
-
             KLabelConstant subjectKLabel = (KLabelConstant) ((KItem) subject).kLabel();
-            KLabelConstant patternKLabel = (KLabelConstant) kitemPattern.kLabel();
-            if (subjectKLabel != patternKLabel) {
+            KLabelConstant patternKLabel = kitemPattern.kLabel() instanceof Variable ? null : (KLabelConstant) kitemPattern.kLabel();
+            if (patternKLabel != null && subjectKLabel != patternKLabel) {
```
cos: how should this work? why do you want the `null` patternKLabel to go through? Where is the klabel variables added to the substitution?

cos:
Looks good so far, but it's not there yet. You still have to actually add the klabel variable to the substitution.

kore/src/main/scala/org/kframework/kore/compile/MergeRules.scala
msaxena2, line 80: @cos The line ensures that the disjunction is pushed down only when the KApply's klabel is not also a variable. If this check didn't exist, there's going to be an issues since we assume that all KApplies with the same KLabels must have the same number of children, which doesn't apply in the case of KApplies with variable KLabels.

cos, line 80: ok, correct

msaxena2:
@cos I've made the changes requested + added a test. Please re-review.

java-backend/src/main/java/org/kframework/backend/java/symbolic/FastRuleMatcher.java
```diff
@@ -136,8 +137,14 @@ private BitSet match(Term subject, Term pattern, BitSet ruleMask, scala.collecti
...
             if (subject instanceof KItem) {
-                // main match of KItems
+                // main match of KItem
                 matchInside(subject, ruleMask, path, returnSet, automatonDisjunction.getKItemPatternForKLabel((KLabelConstant) ((KItem) subject).kLabel()));
+                List<Pair<KItem, BitSet>> varLabelPatterns = automatonDisjunction.getKItemPatternByArity(((KItem) subject).klist().size());
+                if (!(varLabelPatterns == null)) {
+                    for (Pair<KItem, BitSet> p : varLabelPatterns) {
+                        matchInside(subject, ruleMask, path, returnSet, p);
+                    }
+                }
```
cos: could this be moved to a method? (starting with line 142)

k-distribution/src/test/resources/convertor-tests/kore_varlabel.k
```diff
@@ -0,0 +1,34 @@
...
+  imports VARLABEL-COMMON
+
+
+  configuration <T color="yellow">
+                  <k color="green"> $PGM:Holder </k>
+                </T>
+
+  rule Y:KLabel(X + 1) => Y(X + 2)
+
+  rule Z:KLabel(X + 2) => finish
```
cos: it shouldn't require `:KLabel`

also, it would be good to see if the injection mechanism works. something like: `rule Z(...) =>  foo(Z)`. I don't remember whether it should be `foo(Z)` or `foo(#klabel(Z))`. Search though Grigore's note in the wiki pages for the answer -- it may require some digging.

cos:
It would be good to also test the injection mechanism. Still, overall, nice clean work! :-)

msaxena2:
@cos How does the ready-for-review and reviewed tag system work for people without any access to the repository? As far as I know, they'd not have the ability to assign labels to their pull request. And if a reviewer puts the reviewed tag, they may not even be able to change it back to ready-for-review, even if they've made the changes requested.

cos:
KLabels are injected when they appear in a position other than the "head" of a KApply. You can have a look into it if you like, otherwise, I am also ok with merging as it is.

cos:
Just let me know.

cos:
Regarding your review-system question, we can put the labels ourselves when dealing with outside contributors. Also, generally, feel free to change labels and assignments to make sure we follow the workflow. E.g., if I forgot to flip the ready-for-review flag but assigned you as the reviewer, feel free to flip it yourself. Or to assign yourself if I only flipped the flag without changing assignments.

msaxena2:
@cos It'd be great if you could merge this for now. I can return to the issue of changing the tests to also include injected klabel mechanism as soon as I have time to look into it. Thanks.

msaxena2:
Also, regarding the review/ready-for-review tags: I like the fact that we can see all the pull requests marked by meaningful color codes. I was just wondering if it would be possible to allow owners of pull request who don't have any privileges to just have privileges for two actions : to assign the ready-for-review tag, and then change the reviewed tag back to ready-for-review. I'm not even sure github allows such finely grained control.

cos:
Jenkins, test this please

cos:
Afaik, github doesn't allow this. Also, it might not be worth the effort to teach our internal processes to infrequent contributors. We can do it for them.

cos:
Seems like Jenkins doesn't want to wake up without a commit. Could you merge with the latest master to make sure everything still works?

dwightguth:
You typed it wrong

dwightguth:
Jenkins: test this please

cos:
Thanks!

PR #1886, opened 2015-12-05
better KORE transformers

cos:


cos:
@bmmoore, I moved and refactored a bit your `VisitK`and `TransformK` classes -- just to make things uniform. Also added a `FoldK` class.

bmmoore:
Looks good after the commits I added. Did you move the `KRewrite` interface intentionally?

It's a bit unfortunate that the monoid for fold has to be declared just by implementing methods rather than something more reusable, but I guess that's the best we can do.

cos:
no, moving `KRewrite` was not intentional. Thanks for removing the namespace references!

cos:
shall we merge?

PR #1888, opened 2015-12-06
refactor lhs-rhs variable check a bit

cos:


kore/src/main/scala/org/kframework/kore/transformers.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  def apply(k: KRewrite): T = merge(apply(k.left), apply(k.right))
+
+  def apply(k: KToken): T = unit
+
+  def apply(k: KVariable): T = unit
+
+  def apply(k: KSequence): T = k.items.asScala.map(apply).fold(unit)(merge)
+
+  def apply(k: InjectedKLabel): T = unit
```
dwightguth: I do not see how this visits a klabel variable

kore/src/main/scala/org/kframework/kore/transformers.scala
```diff
@@ -0,0 +1,108 @@
...
+  def apply(k: InjectedKLabel): T
+}
+
+/**
+  * Folds a K term into a T. T must be a monoid with the identity defined by unit and the operation by merge.
+  */
+trait FoldK[T] extends KTransformer[T] {
+
+  def apply(k: KApply): T = merge(
+    k.klabel match { case v: InjectedKLabel => apply(v); case _ => unit },
```
dwightguth: or this

kernel/src/main/java/org/kframework/kore/compile/checks/CheckRHSVariables.java
```diff
@@ -0,0 +1,163 @@
...
+
+        Set<KVariable> rhsVars = new FoldKIntoSet<KVariable>() {
+            public Set<KVariable> apply(KRewrite rw) {
+                return apply(rw.right());
+            }
+
+            public Set<KVariable> apply(KVariable v) {
+                return Set(v);
+            }
+        }.apply(KORE.KApply(KORE.KLabel(KLabels.ML_AND), KORE.KApply(KORE.KLabel(KLabels.ML_AND), k, requires), ensures));
```
dwightguth, line 79: I also do not see how this correctly treats variables in the requires clause as not part of the LHS.

kernel/src/main/java/org/kframework/kore/compile/checks/CheckRHSVariables.java
```diff
@@ -0,0 +1,163 @@
...
+
+        stream(rhsVars).forEach(v -> {
+            if (v.equals(ResolveAnonVar.ANON_VAR)
```
dwightguth, line 82: this also does not seem to correctly handle the case where an anonymous variable exists in part of a term that is both LHS and RHS.

cos:
Actually, reverted those commits. I don't think `InjectedKlabel` should ever have a variable inside.

kernel/src/main/java/org/kframework/kore/compile/checks/CheckRHSVariables.java
cos, line 79: this is the visitor for the RHS. The better name for it might the `otherVariables`

dwightguth, line 79: oh, I see

kernel/src/main/java/org/kframework/kore/compile/checks/CheckRHSVariables.java
cos, line 82: correct. simply add `&& !lhsVars.contains(v)`

dwightguth:
Not true. In fact, we already have definitions where this will occur. Consider:

```
rule Lbl:KLabel(.KList) => .K
       requires #klabel(Lbl) in setOfKLabels
```

not being able to do this severely restricts the ability to do anything useful with klabel variables.

cos:
We went back and forth on this last year and I remember the conclusion was that they will get auto-injected. But I might be wrong. In any case, it's very easy to make the visitors check for it.

dwightguth:
Can I close this? You are welcome to refactor the check however you like as long as it  doesn't break the build for the C semantics.

cos:
Thanks! I'll integrate same of it in the next few days. Waiting for #1886. Leave it open so I don't forget.

cos:
might get back to it at some point

PR #1890, opened 2015-12-08
Handle cell fragment labels in SortCells

bmmoore:
Besides the changes in SortCells, this also needs to
add some information fo ConfigurationInfo.
This also reverts earlier changes to again pass a KExceptionManager
throught to SortCells, and use it to report some errors.

The kool-untyped example is updated to again use cell fragments,
and a regression test is added.
(these parts were done by Daejun)


bmmoore:
@daejunpark Can you look at the logic in SortCells. I think the test failures are from some mistake I added in ConfigurationInfoFromModule

daejunpark:
@bmmoore What do you mean by 'looking at'? Do you want to me fix your pull request to pass the test?

daejunpark:
@bmmoore could you write comments for the unit tests? I'm trying to examine if the unit tests are sufficient, but it is not straightforward for me to parse the unit test. It would be useful if we have a comment in the concrete cell syntax for each unit tests.

bmmoore:
Do you want a description of the purpose of each test, or just a comnent giving the corresponding concrete syntax? It would be nice to at least use the KAST parser for making the terms for these tests.

daejunpark:
For me for now, the corresponding concrete syntax will be sufficient. But I like your idea of employing KAST parser for these kinds of unit tests.

bmmoore:
Ok, those comments will be easy to add. Using KAST here is a little more complicated because I also need the sort attributes on variables for these tests, which we don't have any concrete syntax for.

cos:
@bmmoore, is there any chance to still merge this?

daejunpark:
Close deferred PR. Labeled with `deferred` for later search purpose.

PR #1827, opened 2015-11-09
Simple 

msaxena2:
WIP. Doesn't use Cell Fragments yet. Some issues with I/O need to be fixed.

Issue #1733, opened 2015-09-15
List subsorting and modular parsing


grosu:
@bmmoore sent the following message to k-list (guys, let's have these discussions on github, so we can keep track of them and close as we clarify).

---

We've gotten a bit stuck trying to figure out how to combine automatic list subsorting with the new modular parsing.

Following established behavior, and also how `isSort` predicates would execute, we want

```
syntax Exp ::= Int
syntax Ints ::= List{Int,","}
syntax Exps ::= List{Exp,","}
```

to also make `Ints` a subsort of `Exps`.

The problem comes when we consider explicitly applied klabels. From the point of view of ASTs and semantics, we have exactly the same relationship in

```
syntax Exp ::= Int
syntax Ints ::= List{Int,","} [klabel(cons)]
syntax Exps ::= List{Exp,";"} [klabel(cons)]
```

but if we try to reflect the relationship by automatically generating the production

```
syntax Exps ::= Ints
```

we have to keep the parser from using it, or the concrete parser will expand to
allow input that mixes the different terminators, which ought to be illegal:

```
1+2;3+4;5,6,7
```

We want to be able to write any definition in kore, so it's not an answer just to say
that we should generate the subsort production at a later stage in the pipeline than
the parser draws from.

There's also no need to generate a clause of the `isExps` function which directly calls
`isInts` - walking down the spine of the list and calling `isExp` on the items already would
return `true` on a list of `Ints` without it.

I think a solution respecting all the design principles we have been aiming for may
finally require using an attribute to say a production isn't supposed to be used
for parsing at all. `onlyLabel` is a bit funny of a funny name on a subsort production
that doesn't actually have a label. Note that this is much better behaved than attributes
like `notInRules` - those lead to multiple different grammars from a module, while what
we want is to not add a production to the single grammar of the module (i.e, to write
a "syntax" production entirely for semantic effects like introducing a klabel).

A smaller issue is where to put a subsort declaration - now that we take into account modules, we can have situations like the lists being declared in unrelated modules, or even adding the subsort between the items later on. I think this demonstrates all the worst cases:

```
module ELTS
syntax Int ::= ...
syntax Exp ::= ...
syntax Foo
endmodule

module INTS
imports ELTS
syntax Ints ::= List{Int,","}
-- can't see Exps here
endmodule

module EXPS
imports ELTS
syntax Exps ::= List{Exp,","}
-- can't see Ints here
endmodule

module SUB1
imports INTS
imports EXPS
[...]

PR #1891, opened 2015-12-08
fix for #1705

cos:


radumereuta:
@bmmoore please review.
I've been working with @cos to fix issue #1705.
This replaces an older try of mine which wasn't really clean.
Basically we've extended the implementation of Module to automatically add the subsort production between two lists which have related elements. This is done automatically, and is consistent with what @grosu was saying in #1733.

The only problem with doing this automatically is that we don't want that production to be included in the parser for programs, and I solve this by a simple check when I'm generating the grammar for programs.

This should also solve some recent issues with Simple.

bmmoore:
The line I commented on looks like a mistake, please explain it. I didn't see a compensating change elsewhere, so I don't know how both versions can even pass tests, unless that attribute is unused or we have insufficient tests.

Have you updated any definitions to depend on the extra subsorting? It seems like this shouldn't affect parsing at all, is it useful for anything yet?

My commit should be a simple merge to catch up with changes in master, but please check over it. There was a merge conflict in `outer.scala`.

cos:
@bmmoore, yes, that bug slipped in, thanks!

@radumereuta, if you know where such sub-sorting appears in the tutorial, could you have a go at it? Otherwise, I can make the changes.

radumereuta:
The test that I added doesn't work without the inferred subsort.
In the tutorial, this is needed at least for Simple. (see #1827 @daejunpark comment from Nov 10).

bmmoore:
The fix looks fine.

cos:
correct

radumereuta:
So, ok to be merged now?

bmmoore:
Add a test that would have caught that bug. I'm still surprised `mvn verify` passed, I thought it would parse all the nils as the wrong label!

radumereuta:
Which bug? The subsort to KResult? That should have been taken care by the `kool` programs.
But what baffles me right now, is that the test failed, then I merged master and they passed, although they shouldn't, then I've re-added the subsort, and passed again.

bmmoore:
I mean the one fixed in 8a85bec, where you had been giving the nil production the same klabel as the cons productions.

radumereuta:
Does this test look good?

bmmoore:
That looks good, as long as actually it fails if you (temporarily) undo 8a85bec. I'll try running it.

bmmoore:
That test doesn't work, it's the `Es#Terminator ::= "" [klabel('.Es)]` production which was affected.

radumereuta:
Yes, you are right, I forgot to test for programs also.

bmmoore:
That test also doesn't find any problems. You should be able to run a command like this and get the last message:

```
$ mvn package && git revert --no-edit 8a85bec && ! mvn package && echo "Test is strong enough"
```

Actually, the apply function in `definition.Production` adds an additional klabel attribute based on the first argument, so some code might work after reverting 8a85bec depending on which klabel attribute the set returns. But, your test still "passes" even if after reverting that commit I also edit RuleGrammarGenerator.java#274 to construct the production as

```
prod1 = Production(ul.klabel, Sort(ul.sort + "#Terminator"), Seq(Terminal("")),
        newAtts.add(Constants.ORIGINAL_PRD, ul.pTerminator))
```

We need a regression test that would actually test that these productions are actually used correctly in the parser, or perhaps to explain why the klabel assigned there is never actually used.

radumereuta:
Right again.
I made sure this time that all the productions generated for program lists are verified in the test, for both `*` and `+`.

cos:
@radumereuta, this is more your PR than mine. 

If you consider the issue fixed, please mark it as `ready-for-review` so @bmmoore can merge it.

bmmoore:
Okay, these tests still don't actually catch the bug fixed in 8a85bec, but everything else seems to work (and we've fixed that bug), so I'm going to merge this and open a new pull request for sorting out the regression tests.

bmmoore:
I made pull request #1917, which starts by reverting 8a85bec, to continue figuring out why those tests don't work.

PR #1895, opened 2015-12-09
fix module import in DOMAINS module

dwightguth:
We probably don't want to import the entire syntax of K into the definition since the up'd syntax doesn't really exist at runtime.


dwightguth:
@cos @bmmoore please review

cos:
considering the size of the PR, I assumed @cos OR @bmmoore :-)

PR #1896, opened 2015-12-10
auto-format Module

cos:
to make it easier to review #1891


cos:
again, just auto-format, nothing interesting

PR #1900, opened 2015-12-10
module-qualified sorts

cos:
I don't expect this to work in the near future... just tracking progress.


cos:
@radumereuta, this is a step forward. 

Still, your commits have regressed the branch in two ways:
1. The transformers are again in one huge method sharing information. Please break it apart again so that any information sharing is explicit.
2. Many tests that were passing before are now failing: see `org.kframework.parser.concrete2kore.RuleGrammarTest` for example.

cos:
Again, it is nice that the passes use transformers now but, from the tests, it is not clear that it actually works.

cos:
@radumereuta, l-am adus la un numar mic de erori, vezi 84d4ff4b5d00465cf0c57cf47d526d9e52ed9d54, dupa care am facut merge cu master-ul.

Acum sunt ceva mai multe erori dar cred ca ne apropiem. Mare parte din teste au trecut end-to-end.

Daca nu-ti place sa lucrezi pe varianta merged, fa un branch din 84d4ff4b5d00465cf0c57cf47d526d9e52ed9d54 pe care-l vom remergui aici. Totusi, cred ca putem tinti si direct sa mearga pe master. Suntem aproape.

cos:
(apologies to non-Romanian speakers for the above message -- no point in translating it as it's outdated already)

@radumereuta, the merged with master version already has only a few errors. We are very close to making it work. Have a look at the parsing errors and I'll deal with the rest tomorrow.

cos:
also, ktest is twice as slow as before, but I think we should deal with the functional bugs first.

andreistefanescu:
revert this hack if no longer necessary: https://github.com/kframework/k/blob/kore-verification/kernel/src/main/java/org/kframework/parser/concrete2kore/ParserUtils.java#L188

k-distribution/tutorial/2_languages/2_kool/2_typed/1_dynamic/kool-typed-dynamic.k
```diff
@@ -442,19 +442,18 @@ current status of the \textsf{return} cell and then pop it when the
 method returns.  We have to do the same with the current object
 environment, so we group them together in the stack frame.  */
 
-  syntax KItem ::= (Map, K, Bag) // TODO: change Bag into ControlCellFragment once issue #1842 is fixed
+  syntax KItem ::= fstackFrame(Map, K, List, Type, K)
```
radumereuta, line 445: Does the owner of the language agree with this change?

grosu: I don't like it, but it is OK for now, we need to move on.  This needs to be done properly, with a list cell, when the time comes.

cos:
@radumereuta, the merge conflict is in the area you fixed the cell parsing issue in master. You know best what's in there. Could you do the merge?

PR #1887, opened 2015-12-06
unification with kore

andreistefanescu:

PR #1906, opened 2015-12-13
unification with kore

cos:
this is a clone of @andreistefanescu's code in #1887 -- moved here so @laurayuwen and I can commit as well


cos:
@dwightguth, could you please help with this:
As I understand from @andreistefanescu the main issue here is that the backend uses the wrong set of builtin functions at runtime -- it uses the set meant for kompile. I'm trying to fix it but, as the `GlobalContext` is injected, I'm mostly going on guesses here. I tracked it to the `getBuiltinTable` method in `JavaSymbolicCommonModule` module but kind of got stuck there. When navigating injections I'm relying on the injected class but (let me know if there is a better way) but, as here I've reached the rather common `Map<String, String>`, I'm mostly stuck.

dwightguth:
So, that's what the annotations are for. You can search for all the places in the code where `@Builtins` is used and look for a place where it's annotating a provider for that type. In this case you will find that it's populated by the MapBinder in the configure method of JavaSymbolicCommonModule.

cos:
Ok. That method loads the method names only from `/org/kframework/backend/java/symbolic/hooks.properties`, so it doesn't look like it's dependent on `kompile` vs `krun`. Any ideas?

dwightguth:
Take a look at BuiltinFunction line 73. This is where the
ImpureFunctionException gets inserted. But it occurs only for
Stage.INITIALIZING. When we actually rewrite, the GlobalContext that gets
created gets created with Stage.REWRITING. What you need to figure out is
why the old hooks are lingering around. My guess is it is because of a
TermContext that got stored in a term during the initializing stage.

java-backend/src/main/java/org/kframework/backend/java/symbolic/FastRuleMatcher.java
```diff
@@ -72,38 +76,58 @@ public FastRuleMatcher(GlobalContext global, int ruleCount) {
...
                 }
             }
+//            assert constraints[i].isSubstitution();
+//            Substitution<Variable, Term> subst = RewriteEngineUtils.evaluateConditions(rule, constraints[i].substitution(), context);
+//            if (subst != null) {
+//                theResult.add(Pair.of(subst, i));
+//                if (computeOne) {
+//                    break;
+//                }
+//            }
```
laurayuwen: Could you explain why is substitution replaced with constraints? and are the commented code here needed?

cos: Unification can have (possibly partial) results which cannot be represented by a substitution, e.g., `X + Z = Y * Z`. The change shouldn't have any significant impact when doing only matching. 

I removed the commented out code -- thanks for catching that!

cos:
@dwightguth, please have a look to make sure everything will be ok with the C semantics.

cos:
Because we likely have non-determinism somewhere. And very likely is not "new" non-determinism.

cos:
Jenkins: test this please

laurayuwen:
lgtm. Waiting for @dwightguth

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
```diff
@@ -89,7 +89,7 @@ public InitializeRewriter(
 
     @Override
     public synchronized Rewriter apply(Module module) {
-        TermContext initializingContext = TermContext.builder(new GlobalContext(fs, javaOptions, globalOptions, krunOptions, kem, smtOptions, hookProvider, files, Stage.INITIALIZING))
```
dwightguth, line 92: This doesn't look at all right. I haven't reviewed the rest of this code and won't until I get into work in the morning, but this will very likely break the way impure functions are handled, unless you have completely rewritten this code somehow.

cos, line 92: Yes. This a design bug to begin with. The terms themselves have a link to the kompilation/running stage they are in. So the terms generated by this pass are not behaving correctly at `KRun`. The problem likely doesn't break anything because there is no "active" impure function in the generated terms, which is a good thing to begin with.

It might be good to keep things this way and not differentiate between "initialization" and "rewriting". Still, I would argue for leaving this decision for another day.

Bottom line, if you agree that we shouldn't make that distinction, this is just leaving in some harmless redundant code. If you consider that we should keep the distinction, this replaces a serious bug that prevented functions from being executed with a harmless one that would only pop up in very few situations.

dwightguth, line 92: This is actually not safe though. We do partial evaluation during the initialization pass which is why the handling for impure functions was done in the first place. There was a bug wherein impure functions with ground arguments applied statically instead of at every rule application. Unless you can persuade me that this case is handled somehow, this is not harmless at all and will in fact break the C semantics.

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -223,23 +228,24 @@ private boolean strategyIsSave(ConstrainedTerm subject) {
...
                 definition.automaton.leftHandSide(),
                 allRuleBits,
                 computeOne,
                 subject.termContext());
-        for (Pair<Substitution<Variable, Term>, Integer> pair : matches) {
-            // start the optimized substitution
-            Rule rule = definition.ruleTable.get(pair.getRight());
-
+        for (Triple<ConjunctiveFormula, Boolean, Integer> triple : matches) {
+            assert triple.getLeft().isSubstitution();
```
laurayuwen, line 238: This assertions is giving me errors when I try to run logik test

cos, line 238: Can't see the code now, but try to remove it -- might be leftover from my merge.

(sent from mobile phone)

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
dwightguth, line 92: as I expected, this breaks a number of other obscure corner cases that rely on detecting the current stage. You need to find a different solution for this problem, preferably one that solves the underlying issue of impure functions in an actually correct fashion.

cos, line 92: hmm... I sent the message below yesterday by replying to the github email notification but somehow it didn't get posted. Anyhow, it is still relevant:

I understand all that. But, to be clear, this is not a choice between working and not working for the C semantics. It seems like a choice between no impure functions executing and some impure functions executing early. So between not working at all and somewhat working.

The issue is deeper -- essentially this shouldn't have been controlled by a flag on the terms in the first place, but by one in the executor or, better yet, executing statically with a definition that doesn't have the impure rules at all. The latter solution would allow the backend to be oblivious to this problem.

In any case, several people are stuck waiting for this PR. Could we merge for now? I'll fix the early evaluation issue separately by the end of the week.

dwightguth, line 92: No, we can't. If you want a rapid fix, feel free to modify BuiltinFunction.java to change its behavior to do something similar to this simple. But this is actually going to introduce new bugs for functionality that is currently working perfectly. You need to revert this change.

cos, line 92: @dwightguth, please be explicit, which functionality is working perfectly and will be affected by this?

dwightguth, line 92: do a search for all the places in the code where the Java backend decides something based on a Stage and take your pick. I count at least 3.

cos, line 92: Actually, when I look at those places, most of them seem like hacks to avoid fixing some other issue. Considering that all tests (that I can see) pass, I see no reason to hold back from merging.

dwightguth, line 92: Cosmin, you're not the one who's going to have to fix this when it inevitably breaks something. It should take you five minutes to modify this slightly to be a one line change in BuiltinFunction.java instead of this class. Do **NOT** merge this version.

cos, line 92: @dwightguth, sure, let's go with that. Would would be that one-line change in `BuiltinFunction`?

dwightguth, line 92: so, BuiltinFunction checks the stage to decide whether to throw an ImpureFunctionException or to use the real hook. You would just remove that check and always use the real hook. That's essentially what inlining REWRITING here does except doing it here has a bunch of other side effects also.

grosu, line 92: Guys, let me see if I understand the big picture of this.

1) To make the tutorial work, we needed unification.  That is crystal clear.

2) Andrei implemented support for unification and made a PR, say U.

3) As part of that, for some reason, something else (dealing with impure functions) and unrelated to unification which was there all the time was uncovered, say F, which would normally need to be changed anyway in the long term.  But until now it created no problems, because it was used in a restricted way.

4) Now Dwight refuses to approve U because of an imperfect solution for F, which is now used in a more general setting that it was originally intended, so its more general use is overall incorrect.

5) Cosmin offered to fix F, but he wants to do it later, because merging U now would unblock other people, for example those translating the tutorial, so that would speed things up.

So here the conflict between 4 and 5.  Is that relatively accurate?  If yes, then both arguments are valid.  The cheapest solution would be to not open the F problem now, and thus separate concerns.  We first merge U without opening F even if that means that U would be somewhat restricted in its full generality, and then we take F separately as a task in itself and merge a complete solution for it afterwards.  If we can achieve this with a line of code, that would be great.

cos, line 92: What @dwightguth is proposing is still not solving the the static evaluation of impure function bug, but it does reduce the chance for other unforeseen side effects. I'll push the change now.

grosu, line 92: Please put a comment saying so in the code, too, so we know that we have to get back to it, eventually.

While under normal circumstances we'd like perfect solutions to everything, I think that is acceptable in our case here.  It gets us closer to our goal, by unblocking other people, and separating concerns (unification from impure functions).  So I'm OK with it if Dwight is.

cos, line 92: Yes, I did.

laurayuwen:
I think Cosmin's last commit 3055027 addressed the change in BuiltinFunction.java proposed by Dwight and the comment request from Grigore, although don't hear further from them, I merged this PR.

dwightguth:
Yes, I'm satisfied.

PR #1908, opened 2015-12-13
add badges

cos:


cos:
Added a Jenkins badge as well -- thanks @dwightguth for #1909.

PR #1915, opened 2015-12-15
add todos for kore hack

cos:


daejunpark:
@cos I made more changes to be consistent with other todo notes.

Can I merge this, even if I made such a small fix?

PR #1917, opened 2015-12-15
Excercise regression test by reintroducing bug

bmmoore:
This intentionally reintroduces the bug which was fixed in
8a85beccaae21702cec078db5da40bb94e713c10,
so we can develop a unit test that will actually catch it.

@radumereuta, please continue with this branch. Why don't all those tests you added actually catch this bug? Maybe the klabel attributes of these productions are never actually used?


cos:
@radumereuta, is there anything to be done about this?

daejunpark:
Close deferred PR. Labeled with `deferred` for later search purpose.

PR #1918, opened 2015-12-15
Inheritance breaks fragments

bmmoore:
These first two commits demonstrate that a parent cell whose children all have inherited definitions does not get a cell fragment, and that then adding another explicit child is enough to cause the expected fragment productions to be generated and the rules to pass.

@cos, @dwightguth this was introduced in #1873.


bmmoore:
This branch and pull request was made just to host these commits. You're welcome to use it to develop a fix, or just to close it.

cos:
Thanks @bmmoore, I will try to look into it soon. Does it break anything for now?

cos:
closing the pull request but will leave the branch for fixing later

PR #1920, opened 2015-12-16
Tutorial simple

daejunpark:


daejunpark:
@bmmoore please review.

This change is a subset of the one made for KOOL.

Currently, it passes all tests except `threads_{07,10,11}`, which will be handled in a separate PR.

bmmoore:
Looks good. We might want to think about adding parentheses back to certain sorts in `domains.k`.

Issue #1842, opened 2015-11-12
parsing cells in cell fragment terms


daejunpark:
For example, suppose `<x>` cell consists of `<a>`, `<b>`, and `<c>` cells.

```
syntax KItem ::= foo(XCellFragment)
rule foo(A:ACell B:BCell) => .
```

The above rule fails to be parsed, because Cell is not sub-sort of XCellFragment. But if the sub-sort relationship is added, parsing takes unreasonably long.

dwightguth:
Doesn't cell fragments have its own concrete syntax that is different from this?

bmmoore:
cell fragments in K3 just had the syntax of bags, I think this is the concrete syntax we want to give them.

What subsort relationship are you adding?

daejunpark:
```
syntax XCellFragment ::= Cell  [avoid]
```

grosu:
What is `Cell`?  It does not appear in your original grammar, where you only had `ACell` and `BCell`.  I still think that we are making things more complicated than they need to be by keeping track of sorts here at all.  I would just assume that all cell fragment labels can take any arguments of sort K and produce something of sort K, and then resolve errors due to the too permissive syntax afterwards.  In same sense the same like in K3, but with K instead of Bag.

daejunpark:
@bmmoore, I assigned it to @radumereuta, assuming that you are not working on this.

bmmoore:
`Cell` is a sort from `kast.k` which is used for defining the nice front-end syntax of uncompleted cells.
For more than one cell you actually want the name `Bag` (which is poorly named, I just opened #1942 to rename it to `Cells`). We should never refer to that in ordinary rules generated by orginary compiler passes, but that seems like the right way to go inside of `RuleGrammarGenerator` where the complete sugared front-end syntax is being calculated from the input Module.

grosu:
I agree with Cells over Bag, but I am not sure that we even need to complicate ourselves with Cells at all.  We want to allow any operations to be structural, not only cells.  And once they are structural, they can take any arbitrary number of arguments and of any sorts that can ever appear underneath them.  That is just yelling for one generic sort for their arguments and results, which can be used anywhere, which is what K is for.  However, if we are very close to something that works, I would not want to make the transition to 4.0 depend on making the above crystal clear in the implementation.  But I'd like to avoid definitions that are very hard to explain to students, especially by other instructors than me (keep it in mind that Elsa will teach CS422 next semester).

grosu:
[...]

Issue #1801, opened 2015-10-22
partial functions in symbolic rewriting


daejunpark:
In java-backend's symbolic rewriting mode, if a condition of a rule contains a partial function which failed to evaluate, the rule is simply applied with a constraint having the unresolved function, which has little sense. We have two different situations in this regard:
- We may want to drop the current execution path entirely, especially for searching or proving. e.g., divide-by-zero
- We may want to consider those unresolved functions as bottom and the whole condition as false, especially for normal concrete execution. e.g., getKLabel applied to a token.

One possible solution would be to ask user to make every partial function as total, e.g., using `[owise]` rule. This will work for the second situation above, but may not for the first one. It would be better to discuss a general solution.

bmmoore:
I agree paths should be dropped if the function is definitely stuck on (all ground instances of) the given arguments, but I think it makes plenty of sense to have unevaluated functions in side conditions in more general cases, if you consider such a constraint as also asserting that the function is actually defined on those arguments, and _some_ way of checking whether a function is defined seems to be necessary for handling function applications with symbolic arguments (I suppose splitting states based on different clauses of a function definition might sometimes work, but it's not safe for recursive functions).

At the other side of tracking conditions in execution, we need to make sure that functions failing to evaluate is also considered wherever we need to calculate what conditions make a term stuck.

daejunpark:
https://github.com/kframework/k/pull/1925/files#diff-577ed67708e4facf3a50380408fa9c30R949

daejunpark:
@cos could you take care of this? I think you can easily fix it at least for the case of KOOL:
https://github.com/kframework/k/pull/1925/files#diff-577ed67708e4facf3a50380408fa9c30R949

cos:
@daejunpark, `[owise]` can be easily handled by strategies. They are not active by default now, and it wouldn't be wise to activate them for the tutorial just now. Fixing this by other means would mean adding another special case in the backend, and I would rather avoid it. I'll mark this to be fixed later.

PR #1925, opened 2015-12-17
tutorial-kool

daejunpark:
@grosu Please review the final changes made into KOOL.

NOTE: This is an artificial pull request for the review purpose only. You can leave comments in the same way we did for a normal pull request, but should NOT merge it into master.


k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -171,8 +178,13 @@ for the KOOL object-oriented constructs.
                | "(" Id ")" Exp          [strict(2)]    // KOOL  cast
                | "new" Id "(" Exps ")"   [strict(2)]    // KOOL
                | Exp "." Id                             // KOOL
+               // TODO(KORE): drop klabel once parsing issues #1682 are fixed
+               > Exp "[" Exps "]"        [strict,klabel("farray")]
```
grosu: yes, the klabels here are ugly.  issue #1682 seems to be closed, though.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -197,7 +209,11 @@ for the KOOL object-oriented constructs.
                > "spawn" Block
                > Exp "=" Exp             [strict(2), right]
 
+  // TODO(KORE): drop klabel once issues #1913 are fixed
+  syntax Ids  ::= List{Id,","}           [klabel("'_,_::Ids")]
```
grosu: So why do we still have quoted KLabels?  @radumereuta , are you doing _anything_ special for quotes in label names?  If not, then we should drop the quotes.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -282,38 +299,42 @@ member declarations (\textsf{declarations}).
...
                     <thread multiplicity="*" color="yellow">
-                      <k color="green"> ($PGM:Stmts ~> execute) </k>
-                      <br/>
+                      <k color="green"> `$PGM:Stmts ~> execute` </k>
+                    //<br/> // TODO(KORE): support latex annotations #1799
                       <control color="cyan">
                         <fstack color="blue"> .List </fstack>
                         <xstack color="purple"> .List </xstack>
-                        <br/>
+                      //<br/> // TODO(KORE): support latex annotations #1799
```
grosu, line 293: We need to have somebody work on kdoc ASAP.  @edgar-pek is now working for RV, and RV has no interest whatsoever in a Latex/PDF viewer of K definitions.  So this needs to be done at UIUC.  This looks like a perfect task for a more junior member of the K team.  @laurayuwen , does this look like something you'd like to work on before the next semester starts?  We definitely have no time for the Java semantics, so you could work on kdoc in parallel with learning about test-case generation.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -349,13 +370,18 @@ object and method closures. */
   syntax Vals ::= List{Val,","}
   syntax Exp ::= Val
   syntax KResult ::= Val
```
grosu, line 358: Do we need to subsort Val to KResult once we subsorted Vals to KResult?

daejunpark, line 358: I think @radumereuta is best to answer.

FYI, it was a part of PR #1891. @cos attempted to drop this subsort in 15c36dad90dc838ffe763406d16df64e2b6ddadf but @radumereuta put it back in c07db8ffea8fb9a8eab76f0a0e6bd3c1d7e4a97b.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
daejunpark: This is related to lines 963-965 which don't work. Initially, I thought that it was due to the issue #1682, but it still doesn't work even if the issue has been fixed. It might be due to that the issue is not completely fixed, or due to another issue. Indeed, I didn't look into this that much, since I thought that we spent too much time on kool-untyped, and might be better to move on to another language, hoping that the hidden issue will be fixed in the meantime.

@grosu Do you think that I should figure out this hidden issue first before working on kool-typed-dynamic?

grosu: yes, @daejunpark , let us first be happy with kool-untyped and then move to other languages.  If others know what's going on can quickly help here, that'd be great.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -349,13 +370,18 @@ object and method closures. */
...
+  syntax KResult ::= Vals
 
 /*@ The semantics below are taken verbatim from the untyped SIMPLE
 definition. */
 
   syntax KItem ::= "undefined"  [latex(\bot)]
 
+  // TODO(KORE): Change .K back to . throughout this definition #1772
+  rule <k> var X:Id; => .K ...</k>
```
grosu: @radumereuta can we now have polymorphic dots?  Note that people can define their own languages with polymorphic constants, so appending sorts to dots does not solve the inherent problem, it only patches it temporarily.

In the meanwhile, to not have to modify the existing definitions too much, I propose to also add `.` as a constant to the sort `K`.  So we will write `.Map`, etc., for all the other empty collections, but we take the freedom to write `.` instead of `.K` if we want to.  I think this will allow us to still have definitions which look nice and at the same time to transit to 4.0 without a perfect solution to polymorphic constants.  What do you think?

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -363,10 +389,10 @@ definition. */
 
   context var _:Id[HOLE];
 
-  rule <k> var X:Id[N:Int]; => . ...</k>
+  rule <k> var X:Id[N:Int]; => .K ...</k>
        <env> Env => Env[X <- L] </env>
        <store>... .Map => L |-> array(L +Int 1, N)
-                          (L +Int 1) ... (L +Int N) |-> undefined ...</store>
+                          `L +Int 1` ... `L +Int N` |-> undefined ...</store>
```
grosu: Let's try to avoid using backticks, because we decided to replace them back with parentheses in the long term.  However, let's not do it now, to avoid unnecessary ambiguities.  Let us instead add parentheses as brackets to all builtin domains where it makes sense, and thus we don't have to use backticks except for truly exceptional situations.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -432,7 +458,7 @@ interestingly, the semantics of return stays unchanged. */
   rule <k> return(V:Val); ~> _ => V ~> K </k>
        <control>
          <fstack> ListItem((Env,K,C)) => .List ...</fstack>
-         (_ => C)
+         `_ => C`
```
grosu: Same here, can we add parenthese as brackets for cell fragment sorts, and thus avoid the backticks?

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -445,17 +471,17 @@ interestingly, the semantics of return stays unchanged. */
 
   context (HOLE => lvalue(HOLE)) = _
 
-  rule <k> loc(L) = V:Val => V ...</k> <store>... L |-> (_ => V) ...</store>
+  rule <k> loc(L) = V:Val => V ...</k> <store>... L |-> `_ => V` ...</store>
```
grosu: Do we really need the backticks here?

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -464,27 +490,29 @@ interestingly, the semantics of return stays unchanged. */
 
   rule <k> print(V:Val, Es => Es); ...</k> <out>... .List => ListItem(V) </out>
     [print]
-  rule print(.Vals); => .  [structural]
+  rule print(.Vals); => .K  [structural]
 
 
   syntax KItem ::= (Id,Stmt,K,Map,ControlCellFragment)
+  // TODO(KORE): drop the additional production once parsing issue #1842 is fixed
+                 | (Id,Stmt,K,Map,K)
```
grosu, line 480: The above is ugly and I have a strong feeling that we are overcomplicating our existence with the cell fragments by not making them K to K operations.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -464,27 +490,29 @@ interestingly, the semantics of return stays unchanged. */
...
 
   syntax KItem ::= "popx"
 
-  rule <k> (try S1 catch(X) {S2} => S1 ~> popx) ~> K </k>
+  rule <k> `try S1 catch(X) {S2} => S1 ~> popx` ~> K </k>
```
grosu: Don't we have a bracket on sort K, so we don't need the backtick?

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -494,43 +522,51 @@ semantics of thread spawning will be defined shortly.  However,
 interestingly, the other concurrency constructs keep their semantics
 from SIMPLE unchanged. */
 
-  rule (<thread>... <k>.</k> <holds>H</holds> <id>T</id> ...</thread> => .)
+  // TODO(KORE): ..Bag should be . throughout this definition #1772
+  rule `<thread>... <k>.K</k> <holds>H</holds> <id>T</id> ...</thread> => .Bag`
```
grosu: Can we have a bracket on cells as well?  Or maybe the bracket on cell fragments is good enough?  We wouldn't need such a bracket if we made the cell fragment labels take K to K.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -494,43 +522,51 @@ semantics of thread spawning will be defined shortly.  However,
...
 
-  rule <k> join T:Int; => . ...</k>
+  rule <k> join T:Int; => .K ...</k>
        <terminated>... SetItem(T) ...</terminated>
 
-  rule <k> acquire V:Val; => . ...</k>
+  rule <k> acquire V:Val; => .K ...</k>
        <holds>... .Map => V |-> 0 ...</holds>
-       <busy> Busy (.Set => SetItem(V)) </busy>
+       <busy> Busy `.Set => SetItem(V)` </busy>
```
grosu: Same here, a bracket on sets should be needed and then we need no backticks.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -494,43 +522,51 @@ semantics of thread spawning will be defined shortly.  However,
...
-  rule <k> rendezvous V:Val; => . ...</k>
-       <k> rendezvous V; => . ...</k>  [rendezvous]
+  rule <k> rendezvous V:Val; => .K ...</k>
+       <k> rendezvous V; => .K ...</k>  [rendezvous]
 
 //@ \subsection{Unchanged auxiliary operations from untyped SIMPLE}
 
   syntax Stmts ::= mkDecls(Ids,Vals)  [function]
-  rule mkDecls((X:Id, Xs:Ids), (V:Val, Vs:Vals)) => var X=V; mkDecls(Xs,Vs)
+  rule mkDecls(`X:Id, Xs:Ids`, `V:Val, Vs:Vals`) => var X=V; mkDecls(Xs,Vs)
```
grosu: Looks like it would be nice to also have brackets on Lists of each sort.  However, for this to work smoothly we'd need to give priority to the K parentheses, because in the case above, we want to make sure that the list associativity does not start regrouping the parentheses.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -547,7 +583,7 @@ from SIMPLE unchanged. */
   syntax Map ::= Int "..." Int "|->" K
     [function, latex({#1}\ldots{#2}\mapsto{#3})]
   rule N...M |-> _ => .Map  when N >Int M
-  rule N...M |-> K => N |-> K (N +Int 1)...M |-> K  when N <=Int M
+  rule N...M |-> K => N |-> K `N +Int 1`...M |-> K  when N <=Int M
```
grosu: parentheses should work above.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -494,43 +522,51 @@ semantics of thread spawning will be defined shortly.  However,
...
+  /*
   syntax KItem ::= lookup(Int)
+  */
   rule <k> lookup(L) => V ...</k> <store>... L |-> V:Val ...</store>  [lookup]
 
   syntax KItem ::= env(Map)
-  rule <k> env(Env) => . ...</k>  <env> _ => Env </env>  [structural]
-  rule (env(_) => .) ~> env(_)  [structural]
+  rule <k> env(Env) => .K ...</k>  <env> _ => Env </env>  [structural]
+  rule `env(_) => .K` ~> env(_)  [structural]
```
grosu: Hm, how can we make sure that the second rule above applies before the first one?
Probably we'll deal with this using strategies, eventually.  For now, let's not worry about it. because we had that problem there before, too.  Can you @daejunpark just add a TODO comment?

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -591,6 +627,8 @@ to the language, namely object and method closures: */
 
   syntax Val ::= objectClosure(CrntObjCellFragment)
                | methodClosure(Id,Int,Ids,Stmt)
+  // TODO(KORE): drop the additional production once parsing issue #1842 is fixed
+               | objectClosure(K)
```
grosu, line 615: @radumereuta , lots of ugliness in this semantics is due to issue #1842.  Are you taking care of it, or you are expecting somebody else to?  I just want to make 100% sure that it is being taken care of.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -663,12 +703,17 @@ environment. */
 Initially, the classes forming the program are moved into their
 corresponding cells: */
 
-  rule <k> class Class1 extends Class2 { S } => . ...</k>
-       <classes>... (. => <class>
+  syntax KItem ::= ClassBody(K, K)
```
grosu, line 690: What is this monstrosity?  :)  Let's please revisit this as soon as we have AC.  Look at how beautiful it was before.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -743,11 +788,16 @@ members (both fields and methods) of that class. */
...
+       <classes>...`Class |-> ClassBody(Class1:Id, S)`...</classes>
+       [structural]
+       /*
        <className> Class </className>
        <extends> Class1:Id </extends>
        <declarations> S </declarations>  [structural]
+       */
 
-  rule <k> create(Object) => . ...</k>  [structural]
+  rule <k> create(Object) => .K ...</k>  [structural]
```
grosu: Yes, really, please have the default `.` be `.K`, as I mentioned before.  This way we'll not have to change lots of rules, and things will work smoothly once we add polymorphism.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -904,8 +954,21 @@ method call or the array access.  */
        <envStack> ListItem((Class':Id,_)) => .List ...</envStack>
     when Class =/=K Class'  [structural]
 
+  // TODO(KORE): drop klabel once parsing issues #1682 are fixed
```
grosu: What is all this stuff and why do we need it, when we didn't need it before?

grosu:
I added all my comments to kool-untyped.  Let's please try to address them asap, so we can move on to translating the rest of the tutorial.

cos:
@grosu, please flip the label to `reviewed` and assign the issue to @daejunpark -- both options are on the rhs column.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
cos: tracked by #1927

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
cos, line 358: tracked by #1928

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
radumereuta: If you can make a small test for it, I can have a look in the next few days.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
radumereuta: Double quotes are not needed.
The general rule is, that you can put anything you want inside an attribute value, as long as you have balanced parenthesis and no double quotes. If you want to do that anyway, you must put it inside a string. The current implementation works pretty well, since we have lots of latex attributes.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
radumereuta, line 358: Well... I don't know. Do we want to set this rule: if the elements of a list are KResult, then the entire list is a KResult?
It's doable, it's just something extra that needs to be done and documented.

grosu, line 358: No, we do not want to set that.  We used to have the "hybrid" attribute before, but we decided to drop it for many reasons.  What I was saying was the opposite: if Vals is a KResult, then why have to say it that Val is also a KResult?

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
grosu: That's good to know, but not what I meant.  I was concerned about the quote in `'_,_`, which is an artifact of the old Maude backend.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
radumereuta: This is an interesting discussion that I also had with @dlucanu.
If we decide on a solution for this, I can have the implementation working in a day or two.
I'm going to describe this briefly here and make a wiki/issue page with the entire description of the problem early next week.

Currently we have a mechanism for overloaded operators, which we use for lists.
This filter is applied on an ambiguity, and I'm looking to eliminate productions which may be overloaded, and leave only the top most one.
I'm defining a partial order over productions in the following manner:

```
P1: S1 ::= NT1 T1 [klabel(kl1)]
P2: S2 ::= NT2 T2 [klabel(kl2)]
P1 < P2 iff S1 <= S2 /\ NT1 <= NT2 /\ T1 == T2 /\ kl1 == kl2
```

The problem with dots, is that they all have different klabels. We decided to check for klabels when I'm doing this disambiguation, because if the users wants to specify that such two lists are not related, then I should warn them about something being wrong there (by reporting the ambiguity, and not just choose something). Generic evaluation rules (like we have now for lists) would not work in such cases.

All this argument is based on the fact that overloaded dots is the same type of ambiguity like overloaded lists.

There is another idea though. I can try and do something more special for the list terminators. I could look for the context in which it appears, and see if it is a list, and choose the exact sort for the list terminator of that list. It might be possible to also include the user lists like `.Ids`, `.Exps`.
This might be expanded even for the `...` (rest of the list), but I haven't thought this through yet.

As I said, I will follow up with more details and use cases from the tutorial, and we can have a more in depth discussion then, because the problem is not that simple.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
radumereuta: In `kast.k` line 51, there is this production:

```
syntax KBott     ::= "`" K "`"    [bracket]
```

To add parenthesis as brackets, just add (I would suggest modify and change all the current examples):

```
syntax KBott     ::= "(" K ")"    [bracket]
```

I'm not sure if it is needed for all sorts, since this accepts `K` and returns `KBott`. I remember having a post parsing filter which ensures the types are correct over the brackets (it's too late for me to check it now, if needed, I can do it next week).

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
radumereuta: Backtick right now are used as brackets.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
radumereuta: The brackets over K should work here too.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
radumereuta: Brackets over K should work here too.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
radumereuta: Again, we might get away with just one set of parenthesis over K.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
radumereuta, line 615: The issue is not assigned to me, so I wasn't looking into this problem.
@bmmoore did some work on these at one point and I thought he's taking care of it.
If you want me, I can have a look, but I will have to go back and see what was done in this regard.
Please assign the issue to me, if you want me to look into it.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
cos: tracked by #1929

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
radumereuta: You can put there almost anything you want.
I think right now there is a step which assigns a default klabel for every productions and uses the old way.
I will leave it to you to decide how to generate them.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
cos: @radumereuta, @grosu, considering how much time discussions about dots have taken so far, it may be wise to postpone them until we have K 4.0

In the interim, I've added "." as a variant of ".K" here #1926 and it seems to work. Just need to adjust the tests.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
laurayuwen, line 293: @grosu Does it have to be before next semester starts? I really want to spend more time on preparing for qual.

grosu, line 293: Yes, kdoc is an inherent part of K.  We'll find somebody else to work on this.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
daejunpark: @cos I've tested your #1926 for kool-untyped and it worked. I can change `.K` back to `.` throughout the definition, once #1926 is completed.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
daejunpark: fixed in fdc04d5976c7e7857cb40aab0f0c1c65d723eb7b

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
daejunpark: @grosu Yes, because `=>` is greedy. Without the backtick, it is considered as:

```
<store>... `L |-> _` => V ...</store>
```

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
daejunpark: @radumereuta I cannot use K bracket until #1929 is fixed. In the meantime, I attempted to fix this by adding a bracket for Set, but parsing failed again: 

```
$ kompile kool-untyped.k
[Error] Compiler: Had 1 parsing errors.
[Error] Critical: Unexpected sort Set for term Busy. Expected KLabel.
    Source(./kool-untyped.k)
    Location(538,15,538,19)
```

It seems that when we have a bracket here, parser tried to interpret the whole content `Busy (.Set => SetItem(V))` as a KApply term.

daejunpark: tracked by #1931

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
daejunpark: Will be done once #1926 is completed.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
daejunpark, line 690: @cos will keep track of it. It is a part of 2230bb3947c34b34b09e44eaa2ea2dfbd46b31cf.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
daejunpark: fixed in f3ba3bd94563d49112a0d4d9ef11b8ab5f26560e

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
daejunpark: fixed in 34c91d37ee74b06925f99e1a989472ed3ad4d2e9

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
daejunpark: fixed in f3ba3bd94563d49112a0d4d9ef11b8ab5f26560e in a way specific to Int sort.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
daejunpark: fixed in 9aa57a0f185ad4fcc00e1e11446ae176d90f61e3

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
daejunpark: done in b57519f09bf344c5169a1e408a428d8c2c7d9a57

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
cos, line 690: we'll get rid of it once we have AC

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
daejunpark: fixed in 196c6855acac3042e3b1441783fc6af366d53edf

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
daejunpark: fixed in 196c6855acac3042e3b1441783fc6af366d53edf

daejunpark:
@grosu I applied the changes made in PR #1930 into here to hide discussions that are resolved. Never mind Jenkins failure, since this is a PR for review purpose only and will not be merged.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
daejunpark: fixed in 93d369351f054cf79b0e3a0a32f7a0a3acf4163e. somehow it works now, while it didn't before.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
daejunpark: fixed in 93d369351f054cf79b0e3a0a32f7a0a3acf4163e. But note that lines 960-963 will still be there until issue #1801 is fixed.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
daejunpark, line 615: @radumereuta I assigned it to you.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
bmmoore, line 358: That is not implicit because `syntax Vals ::= List{Val,","}` makes a cons list. A singleton Vals looks like `_,_(1,`.List{"_,_"}`())`, while a single Val can just be the Int alone.

grosu, line 358: Of course, I expected that to be the reason for the bug, but note that we did not have this problem in 3.6.  Also, I believe that it should work once we have Assoc rewriting.  There should be no cons list in the semantics anymore when people use List{Sort,","}, as discussed.  The List{Sort,","} generates some productions for parsing programs (which are cons lists), others for parsing the semantics (assoc lists), and translations rules that take the former to the latter.

cos:
@daejunpark, I can close this, right?

daejunpark:
@cos No, this is still useful until we complete KOOL.

Issue #1933, opened 2015-12-19
naming scheme for modules


cos:
I think there is some confusion regarding the naming scheme for modules. I will take `IMP` as an example.

In K < 4.0, the syntax was in `IMP-SYNTAX` and the semantics in `IMP`. Thus, `IMP` imported `IMP-SYNTAX`. `IMP-SYNTAX` was used directly to generate program parsers via some Java hacks which made sure that this is possible, e.g., restricting `Id` etc.

In K >= 4.0, we have more fine-grained modules so various parsers are generated from a specialized modules. So now we have the original syntax of the language, a module for parsing programs, a module for parsing rules, etc. The problem is that it seems to me we have messed up the naming of these modules a bit. We now have:

```
module IMP-COMMON
  contains the syntax of the language
endmodule

module IMP-SYNTAX 
  // despite its name, it is actually the module used for generating the program parser and contains
  // no syntactic productions, but only two imports like the ones below
  // this module could easily be auto-generated for most cases, unless the user
  // wants to do anything very funky for their program parser
  import IMP-COMMON
  imports DOMAINS-SYNTAX
endmodule

module IMP
  import IMP-COMMON
  contains the semantics of the language
endmodule
```

From what I see, the above does not follow any of the naming schemes discussed so far (e.g., the one here: https://github.com/kframework/k/wiki/!!!-Module-System). We ended with this naming scheme by equating correct "the `-SYNTAX` module **is used for generating** the program parser" with the incorrect "the `-SYNTAX` module **is** the program parser".

We might play with naming in the long run but, for now, could we make a simple change: let's name the module containing the entire syntax of the language `-SYNTAX` and the module for the program parser `-PROGRAMS`? We'll also need to propagate the naming change to `domain.k` which also has the same issue.
So:

```
 X-COMMON => X-SYNTAX
 X-SYNTAX => X-PROGRAMS
```

I know we are on a deadline and there are bigger issues than naming, but we're propagating this throughout the tutorial and we'll confuse students next semester. E.g., @omarz1, who is much more familiar than them with everything, posted this on gitter when looking at `lambda.k`:

> I looked at the lambda.k definition. I don't understand the way you're doing things now. You used to have a main module and a syntax module. Now you have the syntax in a commons module, while the syntax module just imports basic modules or something.
> no. syntax module imports common module
> but why did you choose to make it like that?

Again, it's not a big issue, so unless we have a quick consensus, I would drop this -- we don't really have time for long discussion.

[...]

PR #1934, opened 2015-12-19
change lambda lesson_8

laurayuwen:
This is a PR for @grosu to review kore transition changes made in lambda.k 
It is easier and more efficient for me to make a PR on lesson 8 as lesson 9 is already merged.
I checked that the difference is lesson 9 added comments and div by zero. 
Please do NOT merge this.

### DIFF
--- a/k-distribution/tutorial/1_k/1_lambda/lesson_8/lambda.k
+++ b/k-distribution/tutorial/1_k/1_lambda/lesson_8/lambda.k
@@ -23,6 +23,7 @@ module LAMBDA-COMMON
                | "let" Id "=" Exp "in" Exp
                | "letrec" Id Id "=" Exp "in" Exp
                | "mu" Id "." Exp      [binder, latex(\mu{#1}.{#2})]
+  syntax KVariable ::= Id
 endmodule
 
 module LAMBDA
@@ -31,7 +32,6 @@ module LAMBDA
   imports SUBSTITUTION
 
   syntax KResult ::= Val
-  syntax Variable ::= Id
 
   rule (lambda X:Id . E:Exp) V:Val => E[V / X]
 
--- a/k-distribution/tutorial/1_k/1_lambda/lesson_8/lambda.k
+++ b/k-distribution/tutorial/1_k/1_lambda/lesson_8/lambda.k
@@ -1,38 +1,49 @@
 // Copyright (c) 2012-2015 K Team. All Rights Reserved.
-require "modules/substitution.k"
+require "domains.k"
+require "substitution.k"
 
-module LAMBDA
-  imports SUBSTITUTION
+module LAMBDA-SYNTAX
+  imports LAMBDA-COMMON
+  imports ID
+endmodule
 
+module LAMBDA-COMMON
+  imports DOMAINS-COMMON
   syntax Val ::= Id
                | "lambda" Id "." Exp  [binder, latex(\lambda{#1}.{#2})]
   syntax Exp ::= Val
                | Exp Exp              [strict, left]
                | "(" Exp ")"          [bracket]
-  syntax Variable ::= Id
-  syntax KResult ::= Val
-
-  rule (lambda X:Id . E:Exp) V:Val => E[V / X]
-
   syntax Val ::= Int | Bool
   syntax Exp ::= Exp "*" Exp          [strict, left]
                | Exp "/" Exp          [strict]
                > Exp "+" Exp          [strict, left]
                > Exp "<=" Exp         [strict]
+  syntax Exp ::= "if" Exp "then" Exp "else" Exp  [strict(1)]
+               | "let" Id "=" Exp "in" Exp
+               | "letrec" Id Id "=" Exp "in" Exp
+               | "mu" Id "." Exp      [binder, latex(\mu{#1}.{#2})]
+endmodule
+
+module LAMBDA
+  imports LAMBDA-COMMON
+  imports DOMAINS
+  imports SUBSTITUTION
+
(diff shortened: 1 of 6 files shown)


k-distribution/tutorial/1_k/1_lambda/lesson_8/lambda.k
```diff
@@ -1,38 +1,49 @@
...
+               | "mu" Id "." Exp      [binder, latex(\mu{#1}.{#2})]
+endmodule
+
+module LAMBDA
+  imports LAMBDA-COMMON
+  imports DOMAINS
+  imports SUBSTITUTION
+
+  syntax KResult ::= Val
+  syntax Variable ::= Id
```
grosu: I'm pretty sure that we agreed, long ago, to change the sort Variable into KVariable, for consistency with other Kthings.

grosu: Also, it feel like this production should go into the same place with the [binder] tag, because they really work together, so it does not make a lot of sense to separate them.  Also, this production does not seem to create any possible ambiguities or parsing complexity.  @radumereuta , @andreistefanescu , @cos, @bmmoore what do you guys think?

grosu:
Looks good, thanks @laurayuwen.  One thing that I'm not sure about, in all our definitions, is the terminology -COMMON and -SYNTAX.  I'd like to think a bit more about it, but in the meanwhile, if you guys have any suggestions I'd be glad to hear them.

cos:
@grosu, regarding module naming, please have a look at this issue: #1933

k-distribution/tutorial/1_k/1_lambda/lesson_8/lambda.k
laurayuwen: I put  `syntax Variable ::= Id` in semantics module because I feel it is there for the use of substitution and not lambda syntax. It can be moved into the same place with the [binder] tag though technically. Testing don't show ambiguities for this particular case.

I cannot use KVariable here right now because in the backend, it still checks if it is Variable.

```
  public static Term userSubstitution(Map<Term, Term> substitution, Term term, TermContext context) {
        if (substitution.isEmpty()) return term;
        for (Term var : substitution.keySet()) {
            if (!context.definition().subsorts().isSubsortedEq(Sort.VARIABLE, var.sort())) {
                throw KEMException.criticalError("All keys in substitution should be variables, found " + var + " of sort " + var.sort());
            }
        }
        UserSubstitutionTransformer transformer = new UserSubstitutionTransformer(substitution, context);
        return (Term) term.accept(transformer);
    }
```

laurayuwen: anybody have more comments on KVARIABLE/VARIABLE? @andreistefanescu @cos

grosu: > I put syntax Variable ::= Id in semantics module because I feel it is there for
> the use of substitution and not lambda syntax.

So is the binder attribute.  The binder attribute makes no sense without stating what your substitution variables are, and similarly, the variables make no sense without binders.  So they belong together.

@laurayuwen, you are part of the whole K team, so you can go ahead and make the KVariable w=thing work, can't you?  It is nice and friendly to contact the authors of that code to get involved, but if they do not answer in time, then you can go ahead and make the change yourself.

@radumereuta , do you remember why Variable was not changed to KVariable long ago when I first reported it.  @andreistefanescu ?  I remember explicitly discussing this with you, and I do remember a pathetic excuse for delaying it a bit, namely that KVariable was used "for something else" in the implementation.  `Variable` is really a bad name, for the same reason `Result` would be: there is a good chance it interferes with user-defined syntax.

But guys, especially @cos, why does the Java code has anything to do with the particular name we pick in the user-visible syntax?  Shouldn't the sort KVariable (or the current Variable) simply be hooked to a particular class name, so that we can change its user-visible name however we wish?

cos: > But guys, especially @cos, why does the Java code has anything to do with the particular name we pick in the user-visible syntax? Shouldn't the sort KVariable (or the current Variable) simply be hooked to a particular class name, so that we can change its user-visible name however we wish?

For KVariable, yes, this is probably the case -- such a change should have no or minimal impact afaik.

More generally, in theory, yes, it would be possible to never rely on any sort names in any definitions, not even for K stuff, but I'm not sure it's worth it -- we're replacing a convention on the sort or label name with a convention on the hook name. Essentially, it may be a case of over-abstraction or over-indirection. In practice, this is not done mainly because of a dependence issue: the definitions for the the builtins (both K stuff,`kast.k`, and regular, `domains.k`) are in `k-distribution` so they are not available in the `kernel`, where the rest of the code is. We might want to bring those definitions to `kernel.k` eventually but I refrained from opening a serious discussion so far.

Regarding the `K*` names in general, having the `K` prefix may actually become redundant with modularization of sort names. `Variable` could be `Variable@K`. Even nicer, forcing the parsed sort of an `X` variable would be `X@K`. Just putting this out there. In practice, as all of us are already very used to saying "K" something, changing this is likely not a good idea.

laurayuwen:
@grosu In order to pass Jenkins test, I also updated lesson_9, you can just review lesson_8
@cos Could you also have a look?

grosu:
lgtm.  I guess we can merge, but please let's not do other languages until we get rid of this --COMMON thing.

laurayuwen:
got it. And the reviewer does the merge.

Issue #1935, opened 2015-12-20
unexpected parsing ambiguities


daejunpark:
For this rule from kool-untyped:

```
  rule <k> new Class:Id(Vs:Vals) ~> K
           => create(Class) ~> storeObj ~> Class(Vs); return this; </k>
```

Ambiguities: (the second is what expected.)

```
new ( Class:Id(Vs:Vals) ~> K  => create(Class) ~> storeObj ~> Class(Vs) );
return this; 
```

vs

```
( new Class:Id(Vs:Vals) ~> K )
=>
( create(Class) ~> storeObj ~> Class(Vs); return this; )
```

How the first could be possible? 

Actual warning msg:

```
[Warning] Inner Parser: Parsing ambiguity. Arbitrarily choosing the first.
1: syntax Stmts ::= Stmts Stmts
[Location(#token("236",Int),#token("20",Int),#token("236",Int),#token("63",Int))
Source(#token("./kool-untyped.k",KString)) klabel(#token("__",KString))
right(#token("",AttributeValue))]

__(_;(new_(_)(#KRewrite(#KSequence(_(_)(#SemanticCastToId(#token(#KVariable,"Class")),#SemanticCastToVals(#token(#KVariable,"Vs"))),#token(#KVariable,"K")),#KSequence(#KSequence(create(#SemanticCastToId(#token(#KVariable,"Class"))),storeObj()),#SemanticCastToId(#token(#KVariable,"Class")))),#SemanticCastToVals(#token(#KVariable,"Vs")))),return_;(this()))
2: syntax KBott ::= K "=>" K
[Location(#token("193",Int),#token("24",Int),#token("193",Int),#token("102",Int))
Source(#token("/Users/daejunpark/kore/k/k-distribution/target/release/k/lib/java/../../include/builtin/kast.k",KString))
hook(#token("org.kframework.kore.KRewrite",AttributeValue))
klabel(#token("#KRewrite",AttributeValue)) klabel(#token("#KRewrite",KString))
non-assoc(#token("",AttributeValue))]

[...]

PR #1936, opened 2015-12-20
kool-untyped 3rd

daejunpark:


daejunpark:
@grosu please review.

daejunpark:
@grosu please review this.

daejunpark:
@grosu I think that all of your comments are either fixed or tracked by separate issues that will take more time to fix. I want you to confirm it, so that I can move on to another language.

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
```diff
@@ -581,7 +596,14 @@ from SIMPLE unchanged. */
...
   rule N...M |-> _ => .Map  when N >Int M
-  rule N...M |-> K => N |-> K (N +Int 1)...M |-> K  when N <=Int M
+  rule N...M |-> K => N |-> K `(N +Int 1)...M |-> K`  when N <=Int M
+  // TODO(KORE):
+  // The additional bracket is required for disambiguity.
+  // Otherwise, this rule is possible to be parsed as:
+  //   rule N...M |-> K => N |-> ( (K(N +Int 1))...M |-> K )  when N <=Int M
+  // since K has no sort casting, thus can be considered as a klabel
+  // Note that backtick is better than parenthesis here,
+  // because parenthesis doesn't eliminate possiblility of K being of a klabel.
```
grosu, line 606: Not sure about that comment.  It should be possible to infer that K cannot be a KLabel there.
Besides, I think we should separate map binding using a comma, but that's something we can do after Jan 18.

grosu:
As I told @laurayuwen, we should stop doing more definitions until we have what @cos and I agreed yesterday in place and working, basically the trick that allows us to not add the -COMMON modules and basically to not have to change the module structure we had before at all.

Once that happens, we will have to apply minimal changes to the tutorial languages.

Also, in terms of what else to work on, @daejunpark , what is really really needed is to make sure that, once we make the transition, we will not need to look back at 3.6.  Otherwise, by definitions, we are _not_ ready to transit.  In addition to making the tutorial examples work, which I am 100% sure that we will before the deadline, we also need to make sure that we do not lose any of the nice error messages that we had in 3.6, which we've spent many tens of hours to design.  Some of them are unnecessary now, because we do stronger consistency checks on K modules, but many of them are probably still necessary.  Without them, users/students will stumble into errors that they have no idea what they mean (e.g., uncaught exceptions).  I would say that that is probably the best thing you can really help us with, especially while you travel in Korea.  How does this sound?  Thanks!

k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
daejunpark, line 606: Isn't KLabel a subsort of K? Why do you think the variable K cannot be of a KLabel?

BTW, using commas as map separators will fix this problem.

grosu, line 606: Of course KLabel is not a subsort of K.  Check kore.k.

daejunpark, line 606: Oh, you're right. I fixed the todo comment in dc269d68c75f015373c06124c000d87d23479564, and reported this in issue #1935 as well.

PR #1944, opened 2015-12-21
change for review

laurayuwen:
This is an artificial pull request for @grosu to review changes made on last lesson of tutorial 5_types.
Do NOT merge.


k-distribution/tutorial/1_k/5_types/lesson_9/lambda.k
```diff
@@ -1,9 +1,14 @@
...
+require "unification.k"
 
-module LAMBDA
-  imports UNIFICATION
+module LAMBDA-SYNTAX
+  imports LAMBDA-COMMON
+  imports ID
+endmodule
 
+module LAMBDA-COMMON
```
grosu, line 10: But we should not have any -COMMON anymore, as discussed with @cos.  So please, @laurayuwen, stop changing definitions before we switch to the new style, which actually should be quite close 3.6.  It should be identical, in fact, in terms of module structure.

k-distribution/tutorial/1_k/5_types/lesson_9/lambda.k
```diff
@@ -36,9 +48,9 @@ module LAMBDA
   rule T1:Type  + T2:Type => T1 = int ~> T2 = int ~> int
   rule T1:Type <= T2:Type => T1 = int ~> T2 = int ~> bool
 
-  rule <k> lambda X:Id . E:Exp => E ~> ?T:Type -> HOLE ~> tenv(TEnv) ...</k>
+  rule <k> lambda X:Id . E:Exp => E ~> ?T:Type ~> tenv(TEnv) ...</k>
        <tenv> TEnv => TEnv[X <- ?T] </tenv>
-  rule T2:Type ~> T1:Type -> HOLE => T1 -> T2
+  rule T2:Type ~> T1:Type => T1 -> T2
```
grosu, line 53: This rule is wrong.  Please also read the README if it is not clear what the meaning is.

k-distribution/tutorial/1_k/5_types/lesson_9/lambda.k
```diff
@@ -36,9 +48,9 @@ module LAMBDA
   rule T1:Type  + T2:Type => T1 = int ~> T2 = int ~> int
   rule T1:Type <= T2:Type => T1 = int ~> T2 = int ~> bool
 
-  rule <k> lambda X:Id . E:Exp => E ~> ?T:Type -> HOLE ~> tenv(TEnv) ...</k>
+  rule <k> lambda X:Id . E:Exp => E ~> ?T:Type ~> tenv(TEnv) ...</k>
```
grosu, line 51: Why are you doing this?  The rule was that way before because of a bug.  That rule should not be needed at all, in fact.

k-distribution/tutorial/1_k/5_types/lesson_9/lambda.k
```diff
@@ -59,5 +71,5 @@ module LAMBDA
   rule T = T => .
 
   syntax KItem ::= tenv(Map)
-  rule <k> T:Type ~> (tenv(TEnv) => .) ...</k>  <tenv> _ => TEnv </tenv>
+  rule <k> T:Type ~> tenv(TEnv) => T ...</k> <tenv> _ => TEnv </tenv>
```
grosu, line 74: Why are you changing this rule?  It was beautiful before, because it maximally localized rewriting.

k-distribution/tutorial/1_k/5_types/lesson_9/lambda.k
grosu, line 53: In fact, this rule should not be there at all.

k-distribution/tutorial/1_k/5_types/lesson_9/lambda.k
laurayuwen, line 10: Ok, I was just showing what changes were merged.

grosu, line 10: Thanks for that, it is very useful, because it helps me look at the main thing quickly.

k-distribution/tutorial/1_k/5_types/lesson_9/lambda.k
laurayuwen, line 51: The only thing I did (actually not me, Andrei changed this when he tried if unification works) to the two rules was to remove HOLE because it is not recognized any more. And it works without HOLE as an indicator. 
Could you be more specific about the bug? I searched the whole 5_types and only find a bug description in lesson_3.

```
  syntax Exp ::= Exp "->" Exp                  [strict]
  rule <k> lambda X : T . E:Exp => (T -> E) ~> Rho ...</k>
       <tenv> Rho:Map => Rho[X <- T] </tenv>
  // Remove the sorting of E above when the bug is fixed.  Currently, the
  // inferred sort is Type, due to a bug.
```

Lesson_3 does use one less rule to infer the type of lambda expression, but it added another syntax that is not there in Lesson_9. Why did you say both rules are not needed?

andreistefanescu, line 51: This rule (and the one below) represent the heating and cooling for `->`. Alternatively, we could have `syntax Exp ::= Exp "->" Exp                  [strict]`. Either should work. I simply updated lesson 9 as it was, and removed the `HOLE` since it's not supported in the latest K. As they are now, the rules are correct, although a little bit cryptic.

grosu, line 51: Guys, please read also the READMEs and think of teaching.  The rules as they are now are not only cryptic, but they are also error prone.  If you ever happen to end up with one type after the other in the K cell, T1 ~> T2, then it will automatically assume a function type.  We should do the alternative thing, with the type arrow strict.  Then you don't need the heating and the cooling rules.  That's what I meant at the beginning.

andreistefanescu, line 51: Sure, we can do `[strict]`.

PR #1949, opened 2015-12-23
rename -COMMON modules + "main syntax module" refactoring

cos:
**the PR will break (well, fix) APIs**
- renames `*-COMMON` modules to `*-SYNTAX`
- the former `-SYNTAX` modules are now removed 
- @radumereuta, the ID importation is now done in the former `*-PROGRAM-LISTS` module. It is now called `*-PROGRAM`. Such a module is generated for **each** regular module in the definition -- this is inefficient and will hopefully become unnecessary after your work on cleaning up the frontend for sort and label modularization is finished.
- cleaned up a bit the way the "main syntax" module is identified; basically, there is no `mainSyntaxModule` in `Definition` anymore. The main syntax module is tracked by an attribute on `Definition`
- the last two changes above also solve a problem with the kast frontend using incomplete modules when parsing with a non-"main syntax module" module.

### DIFF
--- a/k-distribution/include/builtin/domains.k
+++ b/k-distribution/include/builtin/domains.k
@@ -1,6 +1,6 @@
 // Copyright (c) 2015 K Team. All Rights Reserved.
 module DOMAINS-SYNTAX
-  imports EMPTY-ID
+  imports ID
   imports INT-SYNTAX
   imports BOOL-SYNTAX
   imports STRING-SYNTAX
@@ -478,7 +478,7 @@ module ID
   syntax Id ::= r"(?<![A-Za-z0-9\\_])[A-Za-z\\_][A-Za-z0-9\\_]*"     [notInRules, token, autoReject]
 endmodule
 
-module EMPTY-ID
+module ID-PROGRAM-PARSING
   imports STRING
 
   syntax Id
--- a/k-distribution/include/builtin/domains.k
+++ b/k-distribution/include/builtin/domains.k
@@ -1,14 +1,9 @@
 // Copyright (c) 2015 K Team. All Rights Reserved.
-module DOMAINS-COMMON
+module DOMAINS-SYNTAX
   imports EMPTY-ID
   imports INT-SYNTAX
   imports BOOL-SYNTAX
   imports STRING-SYNTAX
-endmodule
-
-module DOMAINS-SYNTAX
-  imports DOMAINS-COMMON
-  imports ID
 endmodule
 
 module DOMAINS
--- a/k-distribution/include/builtin/domains.k
+++ b/k-distribution/include/builtin/domains.k
@@ -474,11 +474,11 @@ module STRING
 
 endmodule
 
-module ID
+module ID-PROGRAM-PARSING
   syntax Id ::= r"(?<![A-Za-z0-9\\_])[A-Za-z\\_][A-Za-z0-9\\_]*"     [notInRules, token, autoReject]
 endmodule
 
-module ID-PROGRAM-PARSING
+module ID
   imports STRING
 
   syntax Id
--- a/kernel/src/main/java/org/kframework/kompile/CompiledDefinition.java
+++ b/kernel/src/main/java/org/kframework/kompile/CompiledDefinition.java
@@ -54,7 +54,7 @@ public class CompiledDefinition implements Serializable {
      * A function that takes a string and the source of that string and parses it as a program into KAST.
      */
     public BiFunction<String, Source, K> getProgramParser(KExceptionManager kem) {
-        return getParser(kompiledDefinition.mainSyntaxModule(), programStartSymbol, kem);
+        return getParser(kompiledDefinition.programParsingModule(), programStartSymbol, kem);
     }
 
     /**
--- a/kernel/src/main/java/org/kframework/kompile/CompiledDefinition.java
+++ b/kernel/src/main/java/org/kframework/kompile/CompiledDefinition.java
@@ -38,17 +38,19 @@ public class CompiledDefinition implements Serializable {
     public final Definition kompiledDefinition;
     public final Sort programStartSymbol;
     public final KLabel topCellInitializer;
+    private KExceptionManager kem;
     private final Module languageParsingModule;
     private transient Map<String, Rule> cachedcompiledPatterns;
     private transient Map<String, Rule> cachedParsedPatterns;
 
 
-    public CompiledDefinition(KompileOptions kompileOptions, Definition parsedDefinition, Definition kompiledDefinition, Sort programStartSymbol, KLabel topCellInitializer) {
+    public CompiledDefinition(KompileOptions kompileOptions, Definition parsedDefinition, Definition kompiledDefinition, Sort programStartSymbol, KLabel topCellInitializer, KExceptionManager kem) {
         this.kompileOptions = kompileOptions;
         this.parsedDefinition = parsedDefinition;
         this.kompiledDefinition = kompiledDefinition;
         this.programStartSymbol = programStartSymbol;
         this.topCellInitializer = topCellInitializer;
+        this.kem = kem;
         this.languageParsingModule = kompiledDefinition.getModule("LANGUAGE-PARSING").get();
     }
 
@@ -84,11 +86,18 @@ public class CompiledDefinition implements Serializable {
         RuleGrammarGenerator gen = new RuleGrammarGenerator(parsedDefinition, kompileOptions.strict());
 
         Option<Module> userProgramParsingModule = parsedDefinition.getModule(moduleName + RuleGrammarGenerator.POSTFIX);
-        if(userProgramParsingModule.isDefined()) {
+        if (userProgramParsingModule.isDefined()) {
+            kem.registerCompilerHiddenWarning("Module " + userProgramParsingModule.get().name() + " is user-defined.");
             return userProgramParsingModule;
         } else {
             Option<Module> moduleOption = parsedDefinition.getModule(moduleName);
-            return moduleOption.isDefined() ? Option.apply(gen.getProgramsGrammar(moduleOption.get())) : Option.empty();
+            Option<Module> programParsingModuleOption = moduleOption.isDefined() ?
+                    Option.apply(gen.getProgramsGrammar(moduleOption.get())) :
+                    Option.empty();
+            if (programParsingModuleOption.isDefined()) {
+                kem.registerCompilerHiddenWarning("Module " + programParsingModuleOption.get().name() + " has been automatically generated.");
+            }
+            return programParsingModuleOption;
         }
     }
 
--- a/kernel/src/main/java/org/kframework/kompile/CompiledDefinition.java
+++ b/kernel/src/main/java/org/kframework/kompile/CompiledDefinition.java
@@ -56,7 +56,7 @@ public class CompiledDefinition implements Serializable {
      * A function that takes a string and the source of that string and parses it as a program into KAST.
      */
     public BiFunction<String, Source, K> getProgramParser(KExceptionManager kem) {
-        return getParser(programParsingModuleForModule(mainSyntaxModuleName()).get(), programStartSymbol, kem);
+        return getParser(generatedProgramParsingModuleFor(mainSyntaxModuleName()).get(), programStartSymbol, kem);
     }
 
     /**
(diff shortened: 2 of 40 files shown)


cos:
@radumereuta, made changes in your code so please have a look
@dwightguth, this will definitely break all non-adapted backends and definitions
@grosu, let me know if I made any mistakes when changing definitions

k-distribution/include/builtin/domains.k
```diff
@@ -1,16 +1,11 @@
 // Copyright (c) 2015 K Team. All Rights Reserved.
-module DOMAINS-COMMON
+module DOMAINS-SYNTAX
   imports EMPTY-ID
```
grosu: Don't we want to preserve the same conventions for IDs, that is, to rename EMPTY-ID to ID and current ID to ID-PARSING?  I thought the convention was that we will have a module M-PARSING for any module M.

grosu:
Except for the EMPTY-ID thing, looks great to me.  Happy to see it was so easy to achieve this.

I added a section in the Manual, too, currently mostly referring to #1933, called Module Naming Conventions:

https://github.com/kframework/k/wiki/Manual-(to-be-processed)

@dwightguth , hopefully this does not break the OCAML backend.  This change is the simplest way to achieve the generality that we need to pass the tutorial without dirty hacks. As soon as I get the OK from you and @radumereuta , I'll merge it.  Thanks, @cos.

kernel/src/main/java/org/kframework/kompile/CompiledDefinition.java
```diff
@@ -71,7 +73,11 @@ public Module executionModule() {
         return kompiledDefinition.mainModule();
     }
 
-    public Module syntaxModule() { return kompiledDefinition.mainSyntaxModule(); }
+    public String mainSyntaxModuleName() { return parsedDefinition.att().<String>getOptional(Att.syntaxModule()).get(); }
+
+    public Option<Module> programParsingModuleForModule(String syntaxModuleName) {
```
radumereuta: Javadoc!

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -210,10 +210,10 @@ public Definition addSemanticsModule(Definition d) {
 
         Module languageParsingModule = Module("LANGUAGE-PARSING",
```
radumereuta, line 209: @cos tells me that this is something that @dwightguth wrote. So, @dwightguth please add some Javadoc!

kernel/src/main/java/org/kframework/parser/concrete2kore/generator/RuleGrammarGenerator.java
```diff
@@ -119,8 +124,13 @@ public Module getConfigGrammar(Module mod) {
      */
     public Module getProgramsGrammar(Module mod) {
         // import PROGRAM-LISTS so user lists are modified to parse programs
-        Module newM = new Module(mod.name() + "-PROGRAM-LISTS", Set(mod, baseK.getModule(PROGRAM_LISTS).get()), Set(), Att());
-        return newM;
+        scala.collection.Set<Module> modules = org.kframework.Collections.Set(mod, baseK.getModule(PROGRAM_LISTS).get());
+
+        Option<Module> idOpt = baseK.getModule(ID);
+        if(idOpt.isDefined() && !mod.equals(idOpt.get()))
+            modules = org.kframework.Collections.add(idOpt.get(), modules);
```
radumereuta: You are not allowed to include `ID` by default in every single program parser.
There are languages, where our definition of Id differs. There are even languages which don't even have Ids (see BF in our repository).

I personally struggle to understand how you can solve this problem with only 2 modules. Until this gets clarified, this PR should not be merged.

dwightguth:
@cos do you expect this change to have any impact on definitions that do not use the ID module or kast?

cos:
@dwightguth, maybe. It changes the way the program parsing module is identified. Is it now `-PROGRAM-PARSING` instead of `-SYNTAX`.

kernel/src/main/java/org/kframework/kompile/Kompile.java
cos, line 209: We'll need to move this out of the compilation pipeline eventually, but it's not code introduced in this PR so let's leave it for now.

dwightguth:
Okay, but if we don't use the program parser at all that shouldn't really matter right?

cos:
Yes, it shouldn't matter. But I would still try out the C semantics to make sure. And the OCAML backend will definitely not compile anymore -- but the changes should be minimal.

cos:
@dwightguth, scratch the "definitely" -- it you didn't write direct `Definition` transformers and used the gives `*Transformer`s, it may still work.

kernel/src/main/java/org/kframework/kompile/CompiledDefinition.java
```diff
@@ -71,7 +73,24 @@ public Module executionModule() {
...
+     */
+    public Option<Module> programParsingModuleFor(String moduleName) {
+        RuleGrammarGenerator gen = new RuleGrammarGenerator(parsedDefinition, kompileOptions.strict());
+
+        Option<Module> userProgramParsingModule = parsedDefinition.getModule(moduleName + RuleGrammarGenerator.POSTFIX);
+        if(userProgramParsingModule.isDefined()) {
+            return userProgramParsingModule;
+        } else {
+            Option<Module> moduleOption = parsedDefinition.getModule(moduleName);
+            return moduleOption.isDefined() ? Option.apply(gen.getProgramsGrammar(moduleOption.get())) : Option.empty();
```
radumereuta: This should report a hidden warning, something like:
`[Warning] Didn't find TEST-PROGRAM-PARSING, generating one, and using TEST-SYNTAX as a source.`
In fact we do this right now in the front-end when we decide which is the default module for parsing programs.
The more I think about it, it makes more sense to move this code in the front-end, and revert the changes to this file.
This way we can have a default semantics module, and a default programs module, which is decided by `kompile`. We don't mix parsing stuff with syntactic sugar.

grosu: @radumereuta , I don't think that we should move this to the front-end.  The role of the front-end is to take the original .k definition and only parse it, nothing else.  Not even eliminate brackets.  The output of the front end should be an (EKORE) AST, which is what gets passed to the kompile pipeline.  Let's call it Level-0.  Then each kompilation stage takes a Level_n AST and generates a Level_n+1 AST.  Different tools extract the definition at different levels.  For example, the kdoc tool will extract it at Level_0, because we do not want to add anything to the user's definition when generating the poster.  The OCAML backend will probably need everything to be completed, so it will likely take the last level.  But the java backend that will do rewriting-modulo-structure will not need the configuration concretization, for example.

This is what we want to achieve eventually, and the more we delay getting there the more we suffer doing things over and over again, instead of just doing them once and for all.

Getting back to our concrete situation here, I regard the addition of these MODULE-PROGRAM-PARSING modules as a kompilation stage that creates a parallel module structure that resembles the original one, where additional stuff is added in a principled manner to the user-defined modules for additional reasons.  We will likely need to generate several similar module structures, probably one for each kind of parser that we need with krun.

radumereuta: I can go along with that.
How about my proposal for a hidden warning?
For programs, the users may want to completely control what goes into the parser. If we're doing something hidden, I think we should at least warn them. What would be the mechanism to completely control what goes into the program parser?

grosu: I agree that we should give users a way to see what's in these automatically generated modules!  There could be more than just tokens.  For example, if a user defines `Exps ::= List{Exp,","}` in a module M, then the module M-PROGRAM-PARSING will contain productions of the form:

```
module M-PROGRAM-PARSING
...
  syntax Exps ::= "" [klabel(empty)] | NeExps
  syntax NeExps ::= Exp | Exp "," NeExps
...
endmodule
```

while the module M that gets imported in the semantics M-SEMANTICS will be

```
module M
...
  syntax Exps ::= "." | Exp | Exps "," Exps  [assoc unit(.)]
...
endmodule
```

Additionally, we'll have to also generate a module

```
module M-TRANSLATE
  import M-PROGRAM-PARSING
  import M
...
  rule empty => .::Exps  [macro]
  rule (E,Es)@M-PROGRAM-PARSING => (E,Es)@M  [macro]
...
endmodule
```

and then to make sure that in the M-SEMANTICS' configuration we generate

```
... <k> #rewrite(M-TRANSLATE, #parse(M-PROGRAM-PARSING, $Pgm)) </k> ... 
```

This will give us a very powerful and flexible mechanism to deal with all sorts of syntactic sugars, at the same time making the frontend easier and more contained.  For example we can then similarly add Map{SourceSort, TargetSort, "|->"}, etc.  Ah, man, I cannot wait to have K defined in K.  Then we can do all these without even touching the frontend or the parser.

But yes, you are right @radumereuta , with all this stuff being generated automatically, we will definitely need to allow users to see what's in these automatically generated modules.  Not sure if we want to add a warning by default, because then we will always warn whenever people use ID. but we should have some option for that.

radumereuta: Well... we're already giving them warnings for every inferred variable. One more for the fact that we're adding Ids to the program parser shouldn't be that big of a deal.
In my last message, I was talking about a way to avoid doing the automatic thing (not include Ids). Your message says about letting the users know about what happened.

grosu: Warnings of that kind, which are visible only by request, yes.  However, note that nothing should be generated in those -PROGRAM-PARSING modules if the users do not use our predefined sugar, like List{Sort,","}.  The use of IDs in one's definition should not generate anything.  And the users are able to add more stuff to them by simply defining a module with that name and putting their additional stuff there.  That is exactly what happens with the IDs, for example: our ID module comes with another module, also visible in domains.k, called ID-PROGRAM-PARSING, which shows exactly what is "added" to the parser.  So there is nothing hidden generated for IDs, so no warning needed in that case.

Once we have parametric modules, my previous example with lists will consist of three parametric modules, which we will simply provide as predefined in domains.k, too, and then the use of List{Sort,","} will automatically desugar into appropriate instances and imports of those modules.  So again, nothing will be hidden there, too, except desugaring; but yes, I agree that in that case it is good to generate a hidden warning.  The principle here would be that we should generate a hidden warning whenever we add anything in any M-PROGRAM-PARSING, except for the inclusions of the other parsing module corresponding to the modules that M imports.

cos: Regarding warnings, I don't think the default should be to show warnings that something has been automatically generated. It will show warnings for even the simplest definition, confusing people that they might have done something wrong.

On the other hand, I agree there should be a way for the users to see what gets generated. But that can be achieved in other ways, e.g., with some meta-level hooks:

```
#print(#unparse(#up(#programParser(IMP-SYNTAX))))
```

i.e., get the program parser for IMP-SYNTAX, lift it to meta-level so that it can be unparsed by standard means, unparse it, and print it out.

In any case, considering our deadline, I am comfortable with the current partial solution.

grosu: @cos, that's what I understood initially, too, that the warnings would be visible, but @radumereuta was talking about hidden ones, like the ones for inferred sorts of variables.  You'd have to use an option -w to see them.

Also, as I said above, there is nothing to report for ID, because the module ID-PROGRAM-PARSING is explicitly added in the .k file, so nothing is generated.  There are only things to add for List{Sort,","} to -PROGRAM-PARSING, but lists need to be done properly anyway. So we can the warnings then.

Since we spent a lot of time with these exchanges above, I do want to take this opportunity to see if you guys agree with how we should deal with lists, as I explained above.  We don't have to do this until Jan 18, but I'd like to add these to the manual, to make sure we don't, again, waste time rediscussing the same thing over and over again.

radumereuta: I asked for a hidden warning, because we want simple definitions (like lots of our regression tests) to pass with out any warnings.

Regarding lists. Right now I'm using the fact that the users wrote `List{Id, ","}`, and store a kore representation in the definition. Then when time comes to generate a rule parser, or program parser, I extract that seed into multiple productions representative for the parser. See `kast.k` for that. There are some more detailed explanations of how it works, but everything is done inside the parser at the moment, we don't need the rewriter.

P.S. Merry Christmas!

grosu: @cos: it sort of makes sense to report something when new modules are generated, for the reason @radumereuta mentioned: we'd like to write tests where nothing is reported and test that (i.e., in this case, we may want to add tests where all the modules are written by hand).  However, since we didn't have this before and it is not in our intention to add new functionality right now. I would not delay the merge because of that.  @cos, if you think it needs a bit more than just a line or two of code to add that, then just add a one line TODO comment and let
s move on.

@dwightguth , can you please give pull requests towards the Jan 18 deadline high priority?  We want to merge this ASAP, because it is on the critical path: without it, @laurayuwen and myself are blocked, we cannot migrate tutorial definitions.

cos:
Jenkins: test this please.

cos:
I've added hidden warnings for both the case the user provides a `*-PROGRAM-PARSING` and for the case he doesn't. We cannot really consider either of them as the default.

PR #1950, opened 2015-12-23
fix klabel change from previous pr

dwightguth:
@cos please review

We need klabels to have only a single arity.


cos:
@dwightguth, please assign me to the PRs so that I don't lose track of what's on my plate. Thanks!

PR #1911, opened 2015-12-14
Fix for issue regarding function evaluation

msaxena2:
Proposed Fix for #1902 . WIP, not yet ready for review.

PR #1952, opened 2015-12-23
ksequence normalization

dwightguth:
Remaking Manasvi's pull request, I have fixed the comments from the other PR. @cos please re-review.

I tested the performance on IMP sum to 10000 and it's not any slower with this change. I can test with a more complex semantics also if you'd like.


dwightguth:
See #1911 for previous discussion.

cos:
thanks @dwightguth for picking this up!

cos:
Jenkins: test this please

cos:
the performance test should be enough -- just wanted to make sure there isn't something we're not seeing.

cos:
Jenkins: test this please

cos:
I think that failure was non-determinism

PR #1953, opened 2015-12-23
minor fixes to java backend to improve correctness of #KSequence

dwightguth:
Fixes an issue where KSequences inside a KApply inside a KSequence were not converted correctly, and a case where KSequences were incorrectly computed as sort KItem.


andreistefanescu:
lgtm

PR #1955, opened 2015-12-27
Feature/java backend assoc2

andreistefanescu:


andreistefanescu:
@cos @dwightguth please review

andreistefanescu:
@cos the remaining 2 failing tests are on strategies; I will fix them tomorrow if you don't get a chance before you do.

java-backend/src/main/java/org/kframework/backend/java/compile/KOREtoBackendKIL.java
```diff
@@ -111,20 +111,51 @@ public Term KApply1(org.kframework.kore.KLabel klabel, org.kframework.kore.KList
...
-        if (AbstractUnifier.isKSeq(kItem)) {
-            return stream(Assoc.flatten(kSeqLabel, Seq(kItem), kDotLabel).reverse())
-                    .map(Term.class::cast)
-                    .reduce((a, b) -> KItem.of(kSeqLabel, KList.concatenate(b, a), global))
-                    .get();
-        } else {
-            return kItem;
+        // associative operator
+        if (effectivelyAssocAttributes(definition.kLabelAttributesOf(klabel.name()))) {
+            // this assumes there are no KLabel variables
```
dwightguth, line 119: these assumptions here only affect associative operators though right?

java-backend/src/main/java/org/kframework/backend/java/indexing/IndexingPair.java
```diff
@@ -73,23 +73,7 @@ else if (kSequence.concreteSize() == 1) {
...
-        if (!instream.isConcreteCollection()) {
-            fstIndex = instream.elementsLeft().isEmpty() ? definition.indexingData.TOP_INDEX : getIndex(instream.get(0), definition);
-            sndIndex = instream.elementsRight().isEmpty() ? definition.indexingData.TOP_INDEX : getIndex(instream.get(-1), definition);
-        } else {
-            fstIndex = instream.isEmpty() ? definition.indexingData.BOTTOM_INDEX : getIndex(instream.get(0), definition);
-            sndIndex = instream.concreteSize() < 2 ? definition.indexingData.BOTTOM_INDEX : getIndex(instream.get(-1), definition);
-        }
-
-        return new IndexingPair(fstIndex, sndIndex);
+        throw new AssertionError("Should not reach this place anymore");
```
dwightguth, line 76: why can't these methods just be removed if they don't apply anymore?

java-backend/src/main/java/org/kframework/backend/java/kil/BuiltinList.java
```diff
@@ -1,140 +1,188 @@
...
+
+        elementTailSplits = new ElementTailSplit[children.size()];
     }
 
-    private static Variable computeFrame(List<Term> baseTerms) {
-        if (baseTerms.size() == 1 && baseTerms.get(0) instanceof Variable) {
-            return (Variable) baseTerms.get(0);
-        } else {
-            return null;
+    public static class ElementTailSplit {
```
dwightguth, line 59: needs javadoc

java-backend/src/main/java/org/kframework/backend/java/kil/BuiltinList.java
```diff
@@ -1,140 +1,188 @@
...
 
-    private BuiltinList(ImmutableList<Term> elementsLeft, GlobalContext global) {
-        this(elementsLeft,
-             ImmutableList.<Term>of(),
-             ImmutableList.<Term>of(),
-             ImmutableList.<BaseTermType>of(),
-             ImmutableList.<Variable>of(),
-             global);
-    }
+    public ElementTailSplit splitElementTail(int index, int bitSetLength) {
```
dwightguth, line 79: needs javadoc

java-backend/src/main/java/org/kframework/backend/java/kil/BuiltinList.java
```diff
@@ -256,245 +261,83 @@ public GlobalContext globalContext() {
...
+                    unitKLabel,
                     global);
-            return builtinList.baseTerms().size() == 1 && builtinList.concreteSize() == 0 ?
-                   builtinList.baseTerms().get(0) :
-                   builtinList;
+            return builtinList.size() == 1 ? builtinList.get(0) : builtinList;
+        }
+    }
+
+    public BuiltinList upElementToList(Term element) {
```
dwightguth: needs javadoc

java-backend/src/main/java/org/kframework/backend/java/kil/LocalRewriteTerm.java
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) 2015-2016 K Team. All Rights Reserved.
+package org.kframework.backend.java.kil;
+
+import org.apache.commons.lang3.tuple.Pair;
+import scala.collection.immutable.List;
+
+
+public class LocalRewriteTerm extends SMTLibTerm {
```
dwightguth, line 12: needs javadoc

java-backend/src/main/java/org/kframework/backend/java/rewritemachine/GenerateRHSInstructions.java
```diff
@@ -20,29 +20,13 @@
 
     @Override
     public void visit(BuiltinList node) {
-        if (node.isGround() && node.isNormal() && !node.isMutable()) {
+        if (node.isGround() && node.isNormal()) {
```
dwightguth, line 23: is this code even being used anymore?

java-backend/src/main/java/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
```diff
@@ -140,15 +140,9 @@ public static Term construct(List<RHSInstruction> rhsInstructions,
                 Constructor constructor = instruction.constructor();
                 switch (constructor.type()) {
                 case BUILTIN_LIST:
-                    BuiltinList.Builder builder = BuiltinList.builder(global);
+                    BuiltinList.Builder builder = BuiltinList.builder(constructor.assocListSort, constructor.assocListOperator, constructor.assocListUnit, global);
```
dwightguth, line 143: are we still even calling this?

java-backend/src/main/java/org/kframework/backend/java/compile/KOREtoBackendKIL.java
andreistefanescu, line 119: Yes. This means that if `+` is assoc, then L(L(0, 1), 2) does not match 0 + (1 + 2).

java-backend/src/main/java/org/kframework/backend/java/indexing/IndexingPair.java
andreistefanescu, line 76: I need to do a more extensive cleanup after this (especially of dead code that does not show up in this pull request).

java-backend/src/main/java/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
```diff
@@ -305,13 +305,13 @@ public ASTNode visit(org.kframework.kil.Bag node, Void _void) {
     public ASTNode visit(org.kframework.kil.ListBuiltin node, Void _void)  {
```
dwightguth, line 305: can this class be deleted yet?

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicUnifier.java
```diff
@@ -130,88 +125,89 @@ void add(Term left, Term right) {
     }
 
     void unifyList(BuiltinList list, BuiltinList otherList) {
```
dwightguth, line 127: if this is no longer used can we delete this method?

java-backend/src/main/java/org/kframework/backend/java/rewritemachine/GenerateRHSInstructions.java
andreistefanescu, line 23: It is used by `Term#substituteAndEvaluate`.

java-backend/src/test/java/org/kframework/backend/java/kil/BuiltinListTest.java
```diff
@@ -19,29 +19,23 @@
...
+                        .addAll(IntToken.of(9), IntToken.of(10)).build(),
                 IntToken.of(2),
                 IntToken.of(1),
                 termContext);
 
-        Assert.assertEquals(new Variable("L", Sort.LIST), builtinList.frame());
-        Assert.assertEquals(2, builtinList.concreteSize());
-        Assert.assertEquals(IntToken.of(9), builtinList.get(-1));
-        Assert.assertEquals(IntToken.of(9), builtinList.get(-2));
+        Assert.assertEquals(builtinList.children, ImmutableList.of(new Variable("L", Sort.LIST), IntToken.of(9), IntToken.of(9)));
```
dwightguth, line 39: this doesn't look to me like we are distinguishing between a variable of sort list and an element in a list containing a variable of sort list? that doesn't sound right to me.

java-backend/src/main/java/org/kframework/backend/java/rewritemachine/KAbstractRewriteMachine.java
andreistefanescu, line 143: It is, see above.

k-distribution/include/builtin/domains.k
```diff
@@ -170,7 +170,7 @@ module LIST
   syntax List ::= ".List"                 [function, hook(LIST.unit), smtlib(smt_seq_nil), latex(\dotCt{List})]
                 //| "."
   /*@ Construct a singleton List (a list with only  one element) */
-  syntax List ::= ListItem(K)             [function, hook(LIST.element), smtlib(smt_seq_elem)]
+  syntax List ::= ListItem(K)             [smtlib(smt_seq_elem)]
```
dwightguth, line 173: this is going to break the OCAML backend. We need to figure out a way to compromise somehow if you need ListItem to be a constructor in order for this to work.

java-backend/src/main/java/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java
andreistefanescu, line 305: I think so, but I did not check. I want to split that in a separate pull request.

kernel/src/main/java/org/kframework/kore/compile/ConvertDataStructureToLookup.java
```diff
@@ -243,6 +243,8 @@ public K apply(KApply k) {
                         }
                     } else {
                         // List
+                        if (!att.contains(Attribute.HOOK_KEY))
```
dwightguth, line 246: as written, again, this is going to screw with the OCAML backend.

kernel/src/main/java/org/kframework/kore/compile/ConvertDataStructureToLookup.java
```diff
@@ -452,7 +459,7 @@ private K convertSet(KApply k, KLabel collectionLabel, List<K> components) {
                         // TODO(dwightguth): choose better between lookup and choice.
                         Multiset<KVariable> vars = HashMultiset.create();
                         gatherVars(element, vars);
-                        if (vars.isEmpty()) {
+                        if (vars.isEmpty() || element instanceof KVariable && varConstraints.count(element) != 1) {
```
dwightguth: Can you explain to me the case where this code's behavior has changed?

kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
```diff
@@ -41,8 +42,13 @@ public ResolveContexts(KompileOptions kompileOptions) {
...
     }
 
+    public Definition resolve(Definition d) {
+        Module transformedMainModule = resolve(d.mainModule());
+        return Definition.apply(transformedMainModule, add(transformedMainModule, minus(d.mainModule(), d.entryModules())), d.att());
+    }
+
     public Module resolve(Module input) {
-        Set<Sentence> rulesToAdd = stream(input.localSentences())
+        Set<Sentence> rulesToAdd = stream(input.sentences())
```
dwightguth, line 51: this doesn't look right to me. If you call sentences then you're going to be duplicating rules left and right in the resulting module.

kernel/src/main/java/org/kframework/kore/compile/ConvertDataStructureToLookup.java
andreistefanescu: The rule for `join` in `simple-untyped`:

```
rule <k> join T:Int; => .K ...</k>
       <terminated>... SetItem(T) ...</terminated>
```

This generates a choice instead of a membership operation.

dwightguth:
My two major concerns here are that the code in the Java backend seems to contain a large amount of impenetrable algorithms that are hard to follow without more documentation, and that your change to the way ListItem works will almost certainly break RV-Match. We will need to have a discussion and include @grosu in that discussion to figure out the right way forward on the latter of those issues.

java-backend/src/test/java/org/kframework/backend/java/kil/BuiltinListTest.java
andreistefanescu, line 39: Yes, we do. This infrastructure is generic and handles `~>` as well as the builtin list and any other associative operations. We want to distinguish between `K1:KItem  ~> K2:K` (head/tail split) and `K1:K ~> K2:K` (any associative split).

kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
andreistefanescu, line 51: Sentences are going to be duplicated. This is a quick workaround the issue that strictness rules are generated in a module that does not see the configuration declaration, and thus `_ThreadCellBag_` is not flagged as assoc/comm.

@cos do you want to elaborate further on this?

grosu:
I'm following the discussion.  We can continue to discuss here as long as it works.

Regarding the algorithm, it is expected to be tricky.  So my suggested approach here would be to add a bit of documentation here and there to make things clearer for @dwightguth and quickly remind @andreistefanescu and @cos what was going one there when they look back at it in a couple of months, and then it is OK as far as the examples pass.  We'll have to fix bugs most likely as things break, but we do that anyway, don't we?

Regarding the ListItem thing, since the OCAML backend is proprietary, maybe @dwightguth can quickly explain what the expectations are for the OCAML backend for it?  We'll have to find a compromise here, but I'm not sure I understand the entire context well.  The best compromise may be one where both parts make a little change and things get overall clearer.

andreistefanescu:
The question is whether `ListItem` is a constructor (or a wrapper) or a function. In other words, is `ListItem(0) ListItem(1)` the list with elements `ListItem(0)` and `ListItem(1)` (with ListItem just a wrapper to allow lists of lists), or is it the list with elements  0 and 1? The first choice treats uniformly `~>` and the builtin list. I'm not sure this should break the ocaml backend (or more precisely, that it's hard to fix).

kernel/src/main/java/org/kframework/kore/compile/ConvertDataStructureToLookup.java
dwightguth: can you add parentheses around the inner boolean clause to make that clearer?

andreistefanescu: Sure.

java-backend/src/test/java/org/kframework/backend/java/kil/BuiltinListTest.java
dwightguth, line 39: Okay then what would the children field look like in each of those two cases?

kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
dwightguth, line 51: In general duplicating sentences is not sound. That is to say, I have in the past seen correctness bugs where terms do not rewrite correctly that trace themselves back to the fact that we had something like this happen. Therefore, this isn't really an acceptable workaround to a bug.

java-backend/src/test/java/org/kframework/backend/java/kil/BuiltinListTest.java
andreistefanescu, line 39: The children look the same in both cases. The associative matching algorithm differentiates between the two cases based on sorts.

dwightguth:
The OCAML backend works by treating ListItem as a function and mapping it to a hook that transforms a term into a list with a single element. I believe at some point in the past I tried it the way you are proposing and simply could not get it to work correctly in all cases. If I recall, it traces back to the fundamental dissimilarity between ListItem(1) which isn't a list at all under your approach, and `_List_`(ListItem(1), .List), which is a list with a single element. I was only able to resolve this ambiguity with the KSequence by making all terms of sort K be lists. However, that does not really work for ListItem because I cannot make all terms of sort KItem be lists as well without some serious difficulties and likely a significant loss of performance. I strongly suspect if I were to try to make anything work with things the way you propose I would run into the same issue. Indeed, it's possible that you would run into this issue as well if you attempted to execute the C semantics with this diff. I know that that was something I ran into with the code immediately prior to this diff in the Java backend as well.

java-backend/src/test/java/org/kframework/backend/java/kil/BuiltinListTest.java
dwightguth, line 39: But the sort isn't different in all cases... What if I have:

```
ListItem(L:List) ListItem(9) ListItem(9)
```

How are you distinguishing such a term from the term in the test above?

andreistefanescu:
@dwightguth This goes back to the discussion about subsorting the element to the list. That is not the case for the builtin list, and you want to compile that away for `~>`. Unfortunately, that is very tricky in general, and @cos and I decided the let the backend deal with it.

java-backend/src/test/java/org/kframework/backend/java/kil/BuiltinListTest.java
andreistefanescu, line 39: In the first case the children are  `L:List ListItem(9) ListItem(9)`, while in the second they are `ListItem(L:List) ListItem(9) ListItem(9)`.

k-distribution/include/builtin/domains.k
dwightguth, line 173: is it possible we could resolve this discrepancy by making two ListItem klabels? the first and the one used in definitions would be a function and hooked in the Java backend to return the second, which would only be used by the Java backend to represent an actual list constructor? And then you could write a transformer that transforms instances of the first on the LHS into the second? I would prefer to come up with some kind of solution that allows us to have a single domains.k without the burden of having to completely rework the OCAML backend.

andreistefanescu, line 173: We can go that way, but in that case I think it's the ocaml backend that could have a transformation that adds back the `function`` attribute, since this is the public, main backend.

dwightguth, line 173: Yeah, that solution would be satisfactory to me. It just feels strange to me to have ListItem be a constructor but SetItem and _|->_ not be.

grosu, line 173: Let me remind you that `ListItem` was nothing but a hack that we came up with long ago due to annoying parsing reasons.  I also proposed, repeatedly, to get rid of the `List` sort completely and replace it to either user-defined list sorts or, alternatively, with `KList`.  For example, since the `<in>` and `<out>` cells are just `KLabel`s now, it makes a lot of sense to just throw a `KList` in them.  Could we take advantage of this situation now and go for this, or you prefer to wait until after Jan 18?

kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
cos, line 51: no, this doesn't duplicate rules. The transformation pass is now only applied on the main module -- see the `resolveDefinition` above. Contexts eventually generate references to `<k>` cells but the `<k>` syntax is only available in the main module in most cases. This is definitely not right and is evident when trying to enforce stricter coherence rules (such as in #1900), but it also manifested itself here. I don't remember the exact details, but can try to refresh my memory if you want to know more.

k-distribution/include/builtin/domains.k
andreistefanescu, line 173: `SetItem` could also become a constructor. I'm not sure about `_|->` though. The collections need a more unified approach, but this pull request is already big enough.

andreistefanescu, line 173: @grosu conceptually, what you propose is easy. However, it would require a major change to all the tutorial examples, so I think it's risky to do it before 18.

kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
dwightguth, line 51: no, that's fine, you are correct and this makes sense to me. But you realize this is not a problem specific to contexts, yes? This would apply to any rule that has an implicit k cell in a module that doesn't import the configuration.

k-distribution/include/builtin/domains.k
dwightguth, line 173: I would be inclined to let this move forward as is and when I merge the latest K back into RV-Match, to implement the simple solution Andrei proposes.

cos, line 173: @grosu, while I think we do have the functionality to get rid of it (@andreistefanescu, do correct me if I'm missing some case), I think it would be better to wait until after Jan 18.

grosu, line 173: OK, then let's wait.  Thanks for your quick reaction on this, guys!  I promise you that we'll take the problem of lists seriously in our hands after Jan 18 and we'll come up with a solution that will make us all happy.  I just suffered for too many years in my life, maybe more than half of my life, due to ordered -sorted lists: first with OBJ2, then with CafeOBJ, then with OBJ3, then with Maude, and now with K.  There must be a clean solution to this problem.  Or better say a good compromise.  We should be able to find it, based on all the experience and sufferance we've been through.

cos:
I just want to note this was not a light decision. @andreistefanescu and I had long discussions regarding the whole "element as a list" problem. We initially considered solving it via compilation (the way I think it was done before, with the injection), but we kept finding corner cases. We eventually decided the safest path is to make the backend truly `assoc` and get over this problem once and for all. This essentially remove the issue altogether for the Java backend compiler, but backends where implementing it directly may be tricky (e.g., OCAML or Coq, or even the translation to Z3 eventually) will have to find their own ways to deal with it in the compiler.

kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
cos, line 51: @dwightguth, yes, I've encountered quite a few cases when working on sort modularization.

grosu:
This makes sense.  If a backend had support for Assoc (like Maude, ASF, Stratego, Elan) then they would not like any unnecessary treatment of the Assoc labels that they would have to undo in their backend.  I would even say the same for other "rewriting modulo" things, such as rewriting modulo commutativity, rewriting modulo heating/cooling, rewriting modulo structural operations, etc.  If a backend has, is capable of, or wants to provide support for any of these, it may not want the kompiler to touch those parts.  Anyway, I am happy with merging this if you guys are OK with the compromise you reached.

cos:
While we could definitely spend more time cleaning up, I agree it is better to merge quickly now. We should probably set down a few days in the next month or so (before things get out of cache) to clean up a few things, delete dead code, etc., both here and in other parts of the code base.

PR #1957, opened 2015-12-28
Parser @API interface

cos:
This PR:
- adds the @API annotation which will annotate all classes that we consider as part of the K API, i.e., classes that third-parties can rely on when using the K framework as a library
- create `org.kframework.Parser` class as the entry point for K's parsing engine

@radumereuta, I know you meant `org.kframework.parser.concrete2kore.ParseInModule` to fill this role, and I know we discussed its public interface at the time. Unfortunately, in time, its interface has started to accumulate "stuff" that is being used elsewhere, and we should keep the K-library-user-facing interfaces, i.e., `@API`, as small as possible. Also, it may be good to separate the classes used as `@API` from the rest of the implementation (via composition) so that we have the freedom to play with the implementation 

Once we merge this, I will make an announcement on the k-list for people to be extra careful when touching anything that's `@API`. Also, slowly, we should go back to improving the documentation of all classes that should be part of the `@API`. I propose to only add the `@API` when a class/interface passes muster in terms of both cleanliness and documentation.

@dwightguth, @bmmoore, @andreistefanescu, @grosu, you might also want to have a look, in case you have something against it.

### DIFF
--- a/kernel/src/main/java/org/kframework/Parser.java
+++ b/kernel/src/main/java/org/kframework/Parser.java
@@ -30,15 +30,15 @@ public class Parser {
     /**
      * Parses a string with a particular start symbol/sort.
      *
-     * @param withSort the start symbol/sort
+     * @param startSymbol the start symbol/sort
      * @param toParse  the String to parse
      * @param fromSource  the Source of the String toParse
      * @return a pair: the left projection is a parsed string as a K, if successful;
      * the right projection is the set of issues encountered while parsing
      */
     @SuppressWarnings("unchecked")
-    public Tuple2<Option<K>, Set<Warning>> apply(Sort withSort, String toParse, Source fromSource) {
-        Tuple2<Either<Set<ParseFailedException>, K>, Set<ParseFailedException>> res = parseInModule.parseString(toParse, withSort, fromSource);
+    public Tuple2<Option<K>, Set<Warning>> apply(Sort startSymbol, String toParse, Source fromSource) {
+        Tuple2<Either<Set<ParseFailedException>, K>, Set<ParseFailedException>> res = parseInModule.parseString(toParse, startSymbol, fromSource);
 
         Set<Warning> problemSet = new HashSet<>();
         problemSet.addAll((Set<Warning>) (Object) res._2());
@@ -50,13 +50,13 @@ public class Parser {
     /**
      * Parses a string with a particular start symbol/sort.
      *
-     * @param withSort the start symbol/sort
+     * @param startSymbol the start symbol/sort
      * @param toParse  the String to parse
      * @return a pair: the left projection is a parsed string as a K, if successful;
      * the right projection is the set of issues encountered while parsing
      */
-    public Tuple2<Option<K>, Set<Warning>> apply(Sort withSort, String toParse) {
-        return apply(withSort, toParse, Source.apply("generated"));
+    public Tuple2<Option<K>, Set<Warning>> apply(Sort startSymbol, String toParse) {
+        return apply(startSymbol, toParse, Source.apply("generated"));
     }
 
     public static Parser from(Module module) {
--- a/kernel/src/main/java/org/kframework/Parser.java
+++ b/kernel/src/main/java/org/kframework/Parser.java
@@ -1,0 +1,67 @@
+// Copyright (c) 2015 K Team. All Rights Reserved.
+package org.kframework;
+
+import org.kframework.attributes.Source;
+import org.kframework.definition.Module;
+import org.kframework.kore.K;
+import org.kframework.kore.Sort;
+import org.kframework.parser.concrete2kore.ParseInModule;
+import org.kframework.utils.errorsystem.ParseFailedException;
+import scala.Option;
+import scala.Tuple2;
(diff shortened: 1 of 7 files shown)


cos:
Jenkins: test this please.

kernel/src/main/java/org/kframework/Parser.java
```diff
@@ -0,0 +1,65 @@
...
+     * Parses a string with a particular start symbol/sort.
+     *
+     * @param withSort the start symbol/sort
+     * @param toParse  the String to parse
+     * @param fromSource  the Source of the String toParse
+     * @return a pair: the left projection is a parsed string as a K, if successful;
+     * the right projection is the set of issues encountered while parsing
+     */
+    @SuppressWarnings("unchecked")
+    public Tuple2<Option<K>, Set<Warning>> apply(Sort withSort, String toParse, Source fromSource) {
```
radumereuta: Why don't you just call it `startSymbol`? It is something standard.

kernel/src/main/java/org/kframework/Parser.java
```diff
@@ -0,0 +1,65 @@
...
+    private Parser(Module module) {
+        this.parseInModule = new ParseInModule(module);
+    }
+
+    /**
+     * Parses a string with a particular start symbol/sort.
+     *
+     * @param withSort the start symbol/sort
+     * @param toParse  the String to parse
+     * @param fromSource  the Source of the String toParse
```
radumereuta, line 35: I would call this `annotation`, and say annotate each term with the source information. Or something like that. I would like it to be clear that this parameter will be used to annotate each term with the source information.

radumereuta:
lgtm, except for the documentation issues I pointed out.

kernel/src/main/java/org/kframework/Parser.java
cos: changed

kernel/src/main/java/org/kframework/Parser.java
cos, line 35: Will leave as is. The current name is more expressive, and I don't want to go here into details about AST annotations.

cos:
@radumereuta, feel free to adjust the documentation directly if you think of improvements.

cos:
please don't delete the branch after merging... more to come.

radumereuta:
lgtm

radumereuta:
@cos, you mentioned other people in the PR, so I will let you merge when you feel you had enough feedback.

grosu:
I think this @API is a good idea.

dwightguth:
This looks good but I think you should probably make the Error and Warning API a bit more expressive. I would add fields to the Warning interface that allow users to obtain information about the warning, and I would come up with a similar structure for errors and treat a function as returning an error if and only if it throws an exception that implements Error. But we can discuss if you have other ideas for some of this, this is just my offhand thoughts.

cos:
@dwightguth, yes, that's the idea. I just started from the minimal working interface. We should eventually integrate this with `KException` somehow, but that's a larger change and didn't want to deal with it just yet. So in the meantime I'm trying to keep the signature of anything `@API` as small as possible.

cos:
In case anybody is wondering why I bother with this now, I'm trying to implement the `kdoc` functionality by using exclusively `@API` stuff. It's a good opportunity to smoothen K's frontend so that it can be used easily as a library by third parties .

cos:
@radumereuta, thanks, let's wait for further feedback until tomorrow. You should still be the one to merge and I'll remind you by flipping the flag and assigning.

edgar-pek:
This is a great idea, @cos!

radumereuta:
@cos, I'm traveling today, so don't wait for me. I can have another look tomorrow if you really need me to.

kernel/src/main/java/org/kframework/parser/concrete2kore/ParseInModule.java
```diff
@@ -46,7 +46,7 @@
     private final Module parsingModule;
     private volatile Grammar grammar = null;
     private final boolean strict;
-    ParseInModule(Module seedModule) {
+    public ParseInModule(Module seedModule) {
```
grosu, line 49: Quick javadoc, both for this but especially for the main one underneath?  We want all public methods to have a short description.  The names of the method and/or of its arguments do not make it crystal clear what everything means.

grosu:
@radumereuta , I'm OK with this.

@daejunpark , you may also want to take a quick look, because this might interfere with your error messages.

@bmmoore and @andreistefanescu, please take a look and give @radumereuta your blessing asap.

kernel/src/main/java/org/kframework/parser/concrete2kore/ParseInModule.java
cos, line 49: @radumereuta, could you add this documentation? Even after working on the code generating them with you for a while, I'm still a bit confused what each of the four input modules represents exactly.

cos:
@andreistefanescu, I would prefer to not wait for the documentation for `ParseInModule`. I'll add an issue for @radumereuta if you merge.

andreistefanescu:
lgtm

PR #1958, opened 2015-12-28
Definition @API interface

cos:
- adds an `API` module which will be an easy entrypoint for third-parties into the K project -- if they want more control, they can still do injection, etc. instead
- adds a `Definition` class for dealing with loading `Definition`s.


cos:
Jenkins: test this please.

cos:
I'm not sure about using another class named "Definition". I could also move this to `org.kframework.Parser` and have `loadDefinition` methods there.

cos:
@dwightguth, I'm adding a module so you may also want to have a look.

dwightguth:
this looks okay to me but bear in mind that so long as you aren't able to set more advanced flags on the parsing process, this is not going to be able to completely replace calls to kompile.

cos:
Yes, it's definitely far from a replacement of kompile as it has only very limited configurability. I'm trying to retrain myself from adding configurable inputs. Only adding them as needed.

cos:
@daejunpark, could you have a look at this? Mostly refactoring and adding another @API interface.

kernel/src/main/java/org/kframework/Definition.java
```diff
@@ -0,0 +1,86 @@
...
+
+@API
+public class Definition {
+
+    /**
+     * Parses the text to create a {@link Definition} object.
+     * The main module of the definition will be last module defined in the text file.
+     */
+    public static org.kframework.definition.Definition from(String definitionText) {
+        Pattern pattern = Pattern.compile("module ([A-Z_]*)");
```
andreistefanescu: Could this interfere in unexpected ways with the contents of the module, e.g. some syntax names `foo_module`?

cos: it might... I'll try to restrict it a bit

cos: the cleanest way would be to not require a "main" module, and only select one when need, but I'm not going into that large a change now.

andreistefanescu:
lgtm

cos:
Jenkins: test this please.

cos:
made the regex a bit more specific. merge if it ok.

PR #1961, opened 2015-12-29
kdoc with meta

cos:


cos:
Jenkins: test this please

k-distribution/include/builtin/e-kore.k
```diff
@@ -35,7 +35,7 @@ module OUTER-KORE
   imports ATTRIBUTES
 
   syntax KDefinition   ::= KRequireList KModuleList
-                               [klabel(#KDefinition), hook(org.kframework.definition.Definition)]
+                               [klabel(#KDefinition), hook(org.kframework.definition.Definition), latex(\begin{document} #1 #2 \end{document})]
```
cos: @grosu, we need some representation for new lines in latex -- afaik, there was none needed before because the outer was not done at the meta-level.

@radumereuta, could we simply use `\n`?

grosu: Can't you just use conventional latex?

```
latex(\begin{document}{#1}\newline{#2} \end{document})
```

Or are you taking about something else?

radumereuta: @cos, anything you write in that comment, will be put in the latex file. So `\n` in the attribute will be `\n` in latex. I'm not sure what that means in latex.

cos: well, that is the issue, will `\n` be a new line or the `\` and `n` characters? In any case, as long as some form of `\n` is allowed by `e-kore.k`, I can adjust what it means in the latex output.

cos:
@radumereuta, the `e-kore.k` definition does not parse correctly the simple definition in `KDocTest`. For some reason it does not reject the "configuration" string when parsing a bubble.

cos:
Jenkins: test this please

grosu:
@cos: I would say that we can also merge this one as soon as we merge #2198.  Or we can have one of the new students who joined the FSL work on it?

cos:
this PR contains changes to the K API and progress towards an API-based KDoc implementation. @andreistefanescu, can you review? 

@grosu, as far as I remember, the latex annotation style is very similar to before. We can make any adjustments when getting back to it after the pending PRs are merged.

andreistefanescu:
Looks good at a high level.

PR #1962, opened 2015-12-29
trying to clean up Kompile a bit

cos:
very large refactoring which will ultimately enable a clean `@API` implementation of `kdoc`


radumereuta:
I hope you will ask me to review when this is ready.

cos:
sure

cos:
@andreistefanescu, please review
@dwightguth and @radumereuta, you may want to have a look
Context:
I am not happy about the result of this PR but I think it is a significant improvement over what was there before. So please review this as a delta, not judging whether the end result is very good by any means.

I spent more than a day trying to disambiguate what was there. @dwightguth and @radumereuta, if you particularly do not like something, feel free to change it yourself.

cos:
review after #1958

kernel/src/main/java/org/kframework/Compiler.java
```diff
@@ -0,0 +1,42 @@
...
+ * Definition x Module -> Module // when only changing a module, but require information from the entire definition for the change
+ * Sentence -> Sentence
+ * Module x Sentence-> Sentence // when only changing a sentence, but require information from the entire module for the change
+ * Definition x Sentence -> Sentence // when only changing a sentence, but require information from the entire defintion for the change
+ */
+
+@API
+public class Compiler {
+    /**
+     * Generates the definition containing the modules appropriate for generating rule parsers.
```
radumereuta, line 26: I'm having a hard time wrapping my head around this sentence.
I don't understand why you need an entire definition to generate a parser for a single module.
The signature should be `SomeKindOfParserWrapper gen(Definition baseK, Module whichModule)`.

Please explain.

cos, line 26: I don't want to deal with the parser here. These are just a la carte compiler passes. See above.

cos, line 26: I think you know better what `getRuleGrammar` so please feel free to improve the documentation. The documentation from the `RuleGrammarGenerator` doesn't make much sense:

```
    /**
     * Creates the seed module that can be used to parse rules.
     * Imports module markers RULE-CELLS and K found in /include/kast.k.
     * @param mod The user defined module from which to start.
     * @return a new module which imports the original user module and a set of marker modules.
     */
```

kernel/src/main/java/org/kframework/Definition.java
```diff
@@ -0,0 +1,87 @@
...
+
+@API
+public class Definition {
+
+    /**
+     * Parses the text to create a {@link Definition} object.
+     * The main module of the definition will be last module defined in the text file.
+     */
+    public static org.kframework.definition.Definition from(String definitionText) throws IOException {
+        Pattern pattern = Pattern.compile("module ([A-Z_]*)");
```
radumereuta: What? But we have parsers for things like this...
This is so error prone. This needs fixing.

kernel/src/main/java/org/kframework/Compiler.java
cos, line 26: Why is it a "seed" module? Does the results actually parse the cells or just puts some non-functional markers? I know it's the latter, but then the name is misleading. It has mislead me at least, and, after figuring it out, I spent one day just trying to carve out of `Kompile` the code dealing with generating the appropriate syntax for parsing rules with abstract configuration.

kernel/src/main/java/org/kframework/Definition.java
cos: I agree this is not optimal. Feel free to pick it up and fix it.

kernel/src/main/java/org/kframework/Definition.java
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * Parses the text to create a {@link Definition} object.
+     */
+    public static org.kframework.definition.Definition from(String definitionText, String mainModuleName, Source source, List<File> lookupDirectories) throws IOException {
+        File tmpDir = Files.createTempDir();
+        File tempDir = new File(tmpDir.getAbsolutePath() + File.pathSeparator + "tempDir");
+        File definitionDir = new File(tmpDir.getAbsolutePath() + File.pathSeparator + "definitionDir");
+        File workingDir = new File(tmpDir.getAbsolutePath() + File.pathSeparator + "workingDir");
+        File kompiledDir = new File(tmpDir.getAbsolutePath() + File.pathSeparator + "kompiledDir");
```
radumereuta: `tmpDir` and `tempDir`. The names of variables are really important for code readability.
Maybe the first one should be `tempDirBase`?

cos: running into this issue highlights a larger issue with the whole `tempDir` concept :), but anyhow... I'll rename.

cos:
Jenkins: test this please.

kernel/src/main/java/org/kframework/Definition.java
```diff
@@ -0,0 +1,87 @@
...
+                Providers.of(definitionDir),
+                workingDir,
+                Providers.of(kompiledDir),
+                globalOptions,
+                System.getenv());
+        ParserUtils parserUtils = new ParserUtils(fileUtil::resolveWorkingDirectory, kem, globalOptions);
+
+        org.kframework.definition.Definition definition = parserUtils.loadDefinition(
+                mainModuleName,
+                mainModuleName,
```
radumereuta, line 75: But the syntax module can be different, and we have a convention to determine which one it is, why aren't you handling this case in the API?

kernel/src/main/java/org/kframework/Parser.java
```diff
@@ -23,15 +25,21 @@
 public class Parser {
     private final ParseInModule parseInModule;
 
-    private Parser(Module module) {
-        this.parseInModule = new ParseInModule(module);
+    private Parser(Module module) throws IOException {
+        // TODO: remove hack once the frontend is cleaner; also remove the IOException once the hack is cleared
```
radumereuta, line 29: What's not clean in the front-end? Please be more specific.
We usually put a name on TODOs, so we know who they belong to.

kernel/src/main/java/org/kframework/Parser.java
```diff
@@ -23,15 +25,21 @@
...
+        if (module.name().endsWith(RuleGrammarGenerator.RULE_CELLS)) {
+            org.kframework.definition.Definition definitionWithBuiltins = Definition.from("require \"domians.k\"", "K");
```
radumereuta, line 31: To generate parsers only `kast.k` is needed. Why are you including ALL the builtins?

kernel/src/main/java/org/kframework/Definition.java
cos, line 75: because a) which module we use for syntax should have nothing to do with parsing a definition object, and b) generally, I want to keep the `@API` as restrictive as possible.

kernel/src/main/java/org/kframework/Parser.java
cos, line 29: The fact that `ParseInModule` takes three different modules as input; modules which, even after spending time refactoring that parse, I still don't know what they do.

Roughly I understand they contains different syntax for disambiguation, etc., but the clean thing would be to do all that at the KORE level, not as a black box using different data structures.

kernel/src/main/java/org/kframework/Parser.java
cos, line 31: Ok. This is what I mean by the frontend not being "clean". What does the `RuleGrammarGenerator` do? Is it rule for generating the parser grammar for rules, or does it contain a mismatch of 5 other things? Actually, does it even generate anything for the parser, or does it simply prepare definitions for the parser, without having anything to do with the parser grammar? (I know definitions could be considered grammar, but that's not how we usually refer to them). Why do I need to know that only the file names `kast.k` is needed for instantiating it?

cos, line 31: "clean" means that I don't need to know how the entire thing is wired together before using a feature as simple conceptually as parsing. If you want SEish terms, it also means low coupling and high cohesion, which is not a characteristic of this code.

cos, line 31: I'm rambling a bit, but I'm pissed because, after we built this from scratch, I spent a few days just disentangling things. I feel like every time I'm not watching something like a hawk, it quickly becomes all mangled.

dwightguth:
I haven't gone over this in much detail but already I can tell you that if you are calling Definition.from anywhere in the main pipeline this is worse than before, and you simply can't do what you did with checked exceptions, it doesn't scale. If you want to bubble an exception more than one conceptual level up the call stack it needs to inherit from RuntimeException. This **cannot** be merged until you fix the issue with checked exceptions, and I don't have time to fix it for you. Sorry if you don't like that, but them's the breaks.

I will look into this deeper on another day, but it should be enough to get you started.

cos:
@dwightguth, I'm not a big fan of checked exceptions either, but it is the Java way. `IOException`s are actually one of the few cases where checked exceptions make some sense. Code doing IO is more fragile so you want to keep it relatively shallow and only use its results. Having many places that throw IOExceptions is usually a sign that something went wrong with the design. Our code is relatively decent in this respect.

In conclusion, I think it's best to keep the IO checked exceptions as they are after this PR.

grosu:
Guys, this feels like painful dejavu to me.  In the past, like 2 years ago or so, we had checked exceptions everywhere.  Everything was fine at the beginning, but as we added more and more stuff things started to fall apart, to an extent that it would take a lot of time and boilerplate code to add a new feature.  @radumereuta probably still remembers those times, @traiansf for sure.  They just grew organically into a mess that hampered modularity and code clarity.

Then we decided it was time to take the problem of exceptions seriously into our hands and we've redone the entire thing using unchecked exceptions.  This took quite a bit of time and it was a deliberate decision, motivated by a lot of pain in the past.  So if it is to learn from the past, I would say that we should stay with unchecked exceptions.

Yes, unchecked exceptions may hide program logic errors, but they give us more modularity and flexibility, that we ended up desperately needing.  Propagating checked exceptions across abstraction layers breaks abstraction by forcing the callers of an API to be concerned about the details of how the implementation of the API might fail.  That, in particular, makes developing K modules by third parties, say a group which wants to develop a push-down model checker, more painful and dependent upon our particular implementation of the API.

I guess @cos was not in the team when we faced those problems, so those who were should speak up.  In what regards me, I would rather not change the current approach unless we have a very very very strong motivation to, and a very very very strong guarantee that we will not run into the same problems that we ran into the past.

cos:
Again, I'm not a strong supporter of checked exceptions. I think they are, like many other concepts (injections, polymorphism, etc.), a good feature unless misused or abused. I have worked on projects where they are not an issue, but I agree they can become one if you're not careful.

In this particular case, we're talking about the API of our tool. I think it makes more sense for users to receive back a well-known `java.io.IOException` than some custom exception we make. And `IOException` is a checked exception and we cannot do anything about that.

In any case, if you feel strongly about this, I will use some unchecked exception -- not worth a very long discussion.

grosu:
@cos, please trust me, I would be all for checked exceptions if we hadn't gone through what we've gone in the past.  Maybe things are cleaner now and we would not run into the same problems, and maybe it is indeed a good methodology to give our users an IOException as you said, but it is a risk that I'm afraid of taking, based on prior bad experience.  This is all about risk management here, not about gut feeling nor about being right or wrong.  I really appreciate your flexibility and understanding.

cos:
ok, I removed the new checked `IOException` annotations, and a few old ones.

cos:
Jenkins: test this please

kernel/src/main/java/org/kframework/Parser.java
```diff
@@ -24,14 +26,20 @@
     private final ParseInModule parseInModule;
 
     private Parser(Module module) {
-        this.parseInModule = new ParseInModule(module);
+        // TODO: remove hack once the frontend is cleaner; also remove the IOException once the hack is cleared
```
grosu, line 29: IOException?  There seems to be none here.  Also, for the sake of the future, can you very briefly write what the hack to be removed is?

kernel/src/main/java/org/kframework/kdep/KDepFrontEnd.java
```diff
@@ -82,12 +84,13 @@ protected int run() {
...
-        List<org.kframework.kil.Module> modules = parser.slurp(prelude + FileUtil.load(options.mainDefinitionFile()),
+        List<org.kframework.kil.Module> modules = null;
+        modules = parser.slurp(prelude + FileUtil.load(options.mainDefinitionFile()),
                 Source.apply(options.mainDefinitionFile().getAbsolutePath()),
                 options.mainDefinitionFile().getParentFile(),
                 ListUtils.union(options.includes.stream()
-                        .map(files::resolveWorkingDirectory).collect(Collectors.toList()),
-                Lists.newArrayList(Kompile.BUILTIN_DIRECTORY)));
+                                .map(files::resolveWorkingDirectory).collect(Collectors.toList()),
+                        Lists.newArrayList(Kompile.BUILTIN_DIRECTORY)));
```
grosu, line 94: the indentation in the two lines above does not look right to me (as displayed by github at least).

grosu:
Although not asked, I've also looked through the code in this PR.  Looks good enough to me, considering the deadline.  I wished we didn't hardwire naming conventions everywhere in the code, such as the "-PROGRAM-PARSING" naming convention, to make them easy to change in the future, but that is not urgent, so let's not worry about it.  So if the others are happy with it, we should merge this asap.  Thanks again, @cos.

kernel/src/main/java/org/kframework/kompile/KompileFrontEnd.java
```diff
@@ -67,7 +67,8 @@ public int run() {
 
         Kompile kompile = new Kompile(options, files, kem, sw);
         //TODO(dwightguth): handle start symbols
-        CompiledDefinition def = kompile.run(options.mainDefinitionFile(), options.mainModule(), options.syntaxModule(), Sorts.K(), koreBackend.get().steps(kompile));
```
andreistefanescu: debugging leftover?

andreistefanescu:
The high-level design is a step in the right direction towards a clean(er) front-end. However, I did not look over the low-level details, as they are quite hard to follow.

PR #1966, opened 2015-12-30
Lambda all lessons

laurayuwen:


k-distribution/tutorial/1_k/1_lambda/lesson_1/lambda.k
```diff
@@ -1,8 +1,16 @@
...
+module LAMBDA-SYNTAX
+  imports DOMAINS-SYNTAX
   syntax Val ::= Id
                | "lambda" Id "." Exp
   syntax Exp ::= Val
                | Exp Exp      [left]
                | "(" Exp ")"  [bracket]
 endmodule
+
+module LAMBDA
```
grosu: Lesson 1 should all stay into only one module, for padagogical reasons.

k-distribution/tutorial/1_k/1_lambda/lesson_1/tests/closed-variable-capture.lambda.out
```diff
@@ -1,5 +1,4 @@
-Search results:
-
-Solution 1:
-V:K -->
-(lambda z . (z z)) (lambda x . (lambda y . (x y)))
+Solution: 0
+V
+---> 
+` lambda z . ` z z ` ` lambda x . lambda y . ` x y `
```
grosu: We really need to get back to parentheses instead of backticks asap.  @radumereuta , @cos: please!

k-distribution/tutorial/1_k/1_lambda/lesson_2.5/lambda.k
```diff
@@ -1,15 +1,20 @@
...
                | "lambda" Id "." Exp  [binder]
   syntax Exp ::= Val
                | Exp Exp              [left]
                | "(" Exp ")"          [bracket]
-  syntax Variable ::= Id
+  syntax KVariable ::= Id
+endmodule
 
+module LAMBDA
+  imports LAMBDA-SYNTAX
```
grosu: Same this one, for pedagogical reasons it needs to stay in one module.  @cos, do we have any technical problems if we don't have a -SYNTAX module at this moment?  We should not.

k-distribution/tutorial/1_k/1_lambda/lesson_2/lambda.k
```diff
@@ -1,15 +1,20 @@
...
                | "lambda" Id "." Exp  [binder]
   syntax Exp ::= Val
                | Exp Exp              [left]
                | "(" Exp ")"          [bracket]
-  syntax Variable ::= Id
+  syntax KVariable ::= Id
+endmodule
 
+module LAMBDA
+  imports LAMBDA-SYNTAX
```
grosu: Same here, should still be one module.  In fact, to keep it simple, let's not change the module structure at all in any of our definitions.

k-distribution/tutorial/1_k/1_lambda/lesson_1/tests/closed-variable-capture.lambda.out
```diff
@@ -1,5 +1,4 @@
-Search results:
-
-Solution 1:
-V:K -->
-(lambda z . (z z)) (lambda x . (lambda y . (x y)))
+Solution: 0
```
grosu: Guys, why do we number the solutions starting with 0?  We should start with 1.

k-distribution/tutorial/1_k/1_lambda/lesson_3/lambda.k
```diff
@@ -1,15 +1,21 @@
...
                | "lambda" Id "." Exp  [binder]
   syntax Exp ::= Val
                | Exp Exp              [strict, left]
                | "(" Exp ")"          [bracket]
-  syntax Variable ::= Id
+  syntax KVariable ::= Id
+endmodule
+
+module LAMBDA
+  imports LAMBDA-SYNTAX
```
grosu: Same here, there should be only one module.  We only discuss modules and start separating the syntax from the semantics in IMP.  See the tutorial movies/READMEs.  So we keep only one module across the entire lambda tutorial.

grosu:
@laurayuwen , I'm done with reviewing it.  Thanks!  The three main things are:

1) the separation of syntax from semantics, which we do not do until IMP

2) the use of backticks instead of parentheses, and

3) starting to number the solutions with 0 instead of 1.

I think you can take care of 1 and 3 yourself, but for 2 you probably need help from @cos and @radumereuta .

k-distribution/tutorial/1_k/1_lambda/lesson_2.5/lambda.k
laurayuwen: got it. I tried yesterday, it works with one module, you just need to specify --syntax-module LAMBDA in order to use itself as syntax module as well. I used two modules because last lesson was reviewed this way and I talked to @cos, we both thought it is clear to have separate modules. Should have asked you, will change it.

k-distribution/tutorial/1_k/1_lambda/lesson_1/tests/closed-variable-capture.lambda.out
laurayuwen: Sure

k-distribution/tutorial/1_k/1_lambda/lesson_2.5/lambda.k
grosu: Yes, the tutorial readmes and modules have priority here, for teaching reasons.  Students are not ready for modules here.

Also, we should incorporate the option `--syntax-module LAMBDA` as default whenever there is no `-SYNTAX` 
module, same like in K 3.6.  Again, for pedagogical reasons: we don't want to explain something complicated and which will change anyway eventually to students in the first lesson.

laurayuwen:
@grosu Is it the case that we want module structures of every lesson is the same as before? I looked at the changes made by Daejun in lambda++, all are separate into syntax and semantics. But I find that in the old tutorial, only the last lesson of lambda++ has two modules, the first five lessons only have one module.
As I changed the counter in search, tests in lambda++ and 5_types are going to fail before I change them.

grosu:
Yes, we want exactly the same module structure like before.  If not sure, take a look at the tutorial movies and you the definition there.  The point is that these movies will be used by Elsa to teach CS422 next semester and we do not have time to amend them too much.

laurayuwen:
OK, I updated many expected outputs because I changed the solution indexing. 
I also did all lambda++ lessons in commit 1457867. Basically, reverting ` and .K and module separation for lesson 2-6 and adding changes to lesson 1(it was left out because it's substitution based). Now we actually don't have much difference between what we have in K3.6. The main difference is to do with domains.
@grosu is concerned with including domains automatically, I agree with that. But here I didn't wait for the decision, I can change that altogether later if we reach a consensus. 
I will do another PR for 5_types so I just changed the .out in its last lesson to make Jenkins pass check.

k-distribution/tutorial/1_k/1_lambda/lesson_1/tests/closed-variable-capture.lambda.out
```diff
@@ -1,5 +1,4 @@
-Search results:
-
-Solution 1:
-V:K -->
-(lambda z . (z z)) (lambda x . (lambda y . (x y)))
+Solution: 1
```
grosu: But @laurayuwen, why did you change `Solution 1:` with `Solution: 1`, what is your rationale?  In general, we've spent a lot of time designing the user visible syntax in the past, so any change must be very very strongly supported with solid arguments.  This applies to everybody and everything in the user-visible syntax, of course.

k-distribution/tutorial/1_k/1_lambda/lesson_1/tests/closed-variable-capture.lambda.out
```diff
@@ -1,5 +1,4 @@
...
+V
+---> 
+` lambda z . ` z z ` ` lambda x . lambda y . ` x y `
```
grosu: Can somebody replace backticks in the unparser ASAP, so @laurayuwen can go ahead and update definitions?  @laurayuwen, you can also do that yourself, in fact.  We are not going to do anything smart in the unparser for now.  The only thing which we might do, but only if it does not take longer than 1h of additional coding, is to output the defined bracket instead of parentheses.

k-distribution/tutorial/1_k/1_lambda/lesson_9/lambda.k
```diff
@@ -141,6 +133,7 @@ attributes declared as annotations to their syntax declarations (below).  */
 /*@ \section{Conditional}
 Note that the \texttt{if} construct is strict only in its first argument. */
 
+  syntax Exp ::= "if" Exp "then" Exp "else" Exp    [strict(1)]
```
grosu, line 133: @laurayuwen , it is a bit hard for me to review this, because the diff is against an intermediate version and not the original one we had in K 3.6, which is the one we care about the most (because we used it in the movies).  Can you do what @daejunpark did with kool, namely to create a diff against the original definition only for revision reasons?  I know  all these look minor, but consistency with the old tutorial and the movies is really the most important thing for this migration, and minimal changes visible to users means _victory_ for us.

k-distribution/tutorial/1_k/1_lambda/lesson_1/tests/closed-variable-capture.lambda.out
laurayuwen: This is not my change, I just changed one place where 0 is change to 1. The solution syntax has been look like this for a long time.

grosu: Well, I have not used the new version for a long time.  Whoever reviewed the PR did a poor job by allowing that unjustified change.  So we need to change it back.  If you can easily change it please go ahead and do it.  Otherwise please use the blame feature and find the one who should change it back.  The point is that it is bad as is.

laurayuwen: I can have a look. What about the line below solution? There is also slight difference between the original and now.

k-distribution/tutorial/1_k/1_lambda/lesson_9/lambda.k
laurayuwen, line 133: I can do that, but since you decided to automatically import domains, Is it better to do it after that change, so that you can review minimal changes? I actually reverted to k3.6 and then made changes, so literally there is only a few changes besides domains. In the mean time, I can have a look at other tutorials, and unparsing issue.

k-distribution/tutorial/1_k/1_lambda/lesson_1/tests/closed-variable-capture.lambda.out
grosu: Yeah, that line should not change either.  Again, what was wrong with the `Var:Sort ---> ...` notation?  Whoever changed this, what was your rationale for doing it?

k-distribution/tutorial/1_k/1_lambda/lesson_9/lambda.k
grosu, line 133: thanks, @laurayuwen .  please don't do more tutorials right now, let's first make the ones we have look good.

> literally there is only a few changes besides domains. 

perfect!  i can review it when we are done with these changes, too, but just please keep this in mind: minimal changes.

laurayuwen:
@cos I tried to merge master into this branch, but when I try to kompile a lambda definition with one module, it seems that domains is not automatically imported. It reports the error of could not find sort Id. The definition kompiles normally when I add back domains import manually. 

Does the automatic import of domains only work with two modules?

cos:
hm... try adding `imports DOMAINS-SYNTAX` to the `DOMAINS` module in `domains.k`.

laurayuwen:
@cos that solved the problem.

cos:
lgtm

grosu:
I'd still like to the see the diff against 3.6, to make sure that the tutorial is still consistent, especially all the movie contents.  Looks good if I look at it here, but it is easy to miss something.

laurayuwen:
@grosu I copied the entire lambda and lambda++ folder from this branch to 3.6base to create those two artificial PR for you to review. I don't understand what am I missing? Why would movie changed?
I cannot create a PR directly from this branch to 3.6Base, that would show more than 1000 files changed.

The definitions you see in those two PR are the exact contents from this branch when I copied them (I did made one more commit here to remove require domains after you reviewed the artificial lambda PR). 

The tests update are exactly what you see in the artificial PR for lambda because the output is pretty nice now. for lambda++, we did not change tests folder, we added tests-kore folder, so in the artificial PR, instead of letting you look at all those tests in tests-kore, I update .out in tests folder to show you only the difference, which is that we haven't achieved search right now.

grosu:
> I don't understand what am I missing?

See my last comment in #1988.  The require domains.k are still there.  I specifically sent that comment first and then the one above.  Haven't you received them in that order?

> Why would movie changed?

Because if we end up with requires that were not there in 3.6, or with a different module structure, or with different outputs of krun (including with search), then I need to change the tutorial and the movies to be consistent.  I'd like to not have to do that, or if I do have to, I need to know exactly what the changes are.  I'm going to be very careful with that during the entire transition.  Every character matters.

> The definitions you see in those two PR are the exact contents from this branch when I copied them

No, they are not.  This one has no requires while the artificial one has requires, as you noted.  Why do things partially and not perfectly?

> for lambda++ ...

Let's first merge lambda and then we worry about lambda++.  Looks like we are very close now.

PR #1967, opened 2015-12-30
fix bubbling of rewrite out of cells

andreistefanescu:


kore/src/main/scala/org/kframework/kore/compile/RewriteToTop.scala
```diff
@@ -30,13 +30,26 @@ object RewriteToTop {
...
+    } else {
+      false
+    }
+    case rw: KRewrite => nonCell(rw.left) && nonCell(rw.right)
+    case _ => true
   }
 
-  private def makeRewriteIfNeeded(k: K): K = if (toLeft(k) != toRight(k)) ADT.KRewrite(toLeft(k), toRight(k)) else k
+  private def isCell(kapp: KApply): Boolean = {
+    kapp.klabel.name.startsWith("<") && kapp.klabel.name.endsWith(">")
```
dwightguth, line 52: is there any chance we could get like, a real implementation of this method? it shouldn't be any more complicated than checking the attributes of the production for the cell attribute.

andreistefanescu, line 52: I am just adapting the code to work for AC matching. I have enough problems already, so I won't touch existing code.

PR #1969, opened 2015-12-30
fixes to java backend for c semantics

dwightguth:


dwightguth:
Based on conversation with @andreistefanescu I am going to close this because it looks like KSequence is handled differently in his upcoming PR. @andreistefanescu you should take a look at this diff and see if any of these fixes can be applied to your copy though.

andreistefanescu:
Made obsolete by the A/AC feature of the java backend.

Issue #1932, opened 2015-12-19
unparsing issue with bracket


daejunpark:
Adding a bracket for Int sort (f3ba3bd94563d49112a0d4d9ef11b8ab5f26560e) affects unparsing lambda term (8a24fe8143a921108f0b67930d5c97bf3902ce9f).

daejunpark:
@dwightguth do you have an idea why it happened?

radumereuta:
Guys. Stop adding brackets for every sort. Only one is needed for K.

grosu:
And why is that, @radumereuta?  These domains should have brackets anyway, regardless of K.  If somebody wants to import INT in the syntax of their language, they would expect to have parentheses there when parsing programs, too, not only in the semantics.  I understand your concern, ambiguities, but we have to deal with those ambiguities anyway.  Remember that we now have the module qualification @ notation, too, which can be used when _real_ ambiguities are formed.  By _real_ ambiguity I mean one where the resulting KASTs are distinct.  For example, if one defines parentheses as brackets in their language, then that creates no ambiguities with the K parentheses, as those have the same "semantics" of brackets.  But if one has parentheses as function application in one's language, like in Scheme/LISP, then one as to disambiguate, say, the term `(f)`: in Scheme that means invoking the no-argument function `f`, while in K that means just `f`.

laurayuwen:
I don't understand why we need to add brackets to every sort? I see PR#1930 add bracket for Int, Set and Bag in order to use brackets instead of backticks in kool semantics, but it doesn't seem to be the right way. We didn't have backticks for very sort, and yet can use it everywhere.

grosu:
I just explained it above.  I can repeat:

These domains should have brackets anyway, regardless of K. If somebody wants to import INT in the syntax of their language, they would expect to have parentheses there when parsing programs, too, not only in the semantics...

laurayuwen:
They are two separate concerns, one is If somebody wants to import INT in the syntax of their language, they would expect to have parentheses there when parsing programs, too, not only in the semantics?  This one I am not sure, somebody many and somebody may not. 
The other one is we want to replace where we used backticks with parenthesis. And it doesn't convince me that it needs to be done separately for every sort.

cos:
It may be good, instead of adding brackets to every sort, to simply make a module:

```
module BRACKETS
  syntax HAS_BRACKETS ::= "(" HAS_BRACKETS ")" [bracket]
endmodule
```

[...]

PR #1970, opened 2015-12-30
Add parenthesis as brackets for sort K. #1932

radumereuta:
Let's see what fails.

### DIFF
--- a/k-distribution/include/builtin/kast.k
+++ b/k-distribution/include/builtin/kast.k
@@ -50,6 +50,7 @@ module KSEQ
                  | K "~>" K  [klabel(#KSequence), left, assoc, unit(#EmptyK), hook(org.kframework.kore.KSequence)]
   syntax left #KSequence
   syntax KBott     ::= "`" K "`"    [bracket]
+                     | "(" K ")"    [bracket]
 endmodule
 
 
--- a/k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
+++ b/k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k
@@ -903,7 +903,7 @@ invocations see the entire object, as needed for dynamic method dispatch. */
     => lookupMember(<envStack> ListItem((Class,EnvC)) EStack </envStack>, X)
     [structural]
   rule objectClosure(<crntClass> Class:Id </crntClass>
-                     <envStack> ListItem((Class':Id,_)) => .List ...</envStack>)
+                     <envStack> ListItem((Class':Id,_::EnvCell)) => .List ...</envStack>)
        . X:Id
     when Class =/=K Class'  [structural]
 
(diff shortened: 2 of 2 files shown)


radumereuta:
@grosu please review.
I've added the parenthesis you asked in #1932, and other places.
I had to patch kool-untyped to make it work, since the type inferencer is not smart enough to disambiguate between `KItem ::= (Id,EnvCell)` and brackets around a list of `Exps` or `Ids`. Personally, I find this error the expected behavior. It is just ambiguous if you take into consideration only types, like I do in the type inferencer. If you don't like the result, then we have to use extra information in the disambiguation algorithm.

I don't have good internet this week, so don't wait for me to merge it if you need these changes.

grosu:
The patch is OK, you are basically only casting _ to _::EnvCell to avoid the ambiguity.  That is acceptable, especially considering that we do not have module qualification yet.  So I'm going to merge this.  Thanks, @radumereuta .

laurayuwen:
Just to make sure, this PR makes sure that we can write write parenthesis instead of back-ticks in definitions; but it does not deal with unparsing, right? I merged this into the tutorial branch I am working on, it does not change the output like this

```
lambda y . ` ` lambda x . lambda y . ` x y ` ` y `
```

radumereuta:
Just parsing, haven't looked over unparsing.

daejunpark:
@radumereuta just to make sure, you're considering the followings as well, right?

https://github.com/kframework/k/blob/80ed28cd8d962e8e347ad3ee16da3a90b87fa46b/k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k#L461

https://github.com/kframework/k/blob/80ed28cd8d962e8e347ad3ee16da3a90b87fa46b/k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k#L493

https://github.com/kframework/k/blob/80ed28cd8d962e8e347ad3ee16da3a90b87fa46b/k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k#L556

https://github.com/kframework/k/blob/80ed28cd8d962e8e347ad3ee16da3a90b87fa46b/k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k#L572

radumereuta:
I've only added `( )`. I didn't remove the backtick ones. That will be left for a later date. I don't want to spend time now to fix the updated definitions.

grosu:
@radumereuta : there should be no bacticks in any definitions by Jan 18.  Otherwise we are not done.  Because it means that we need to modify the readmes, the movies, etc.  I hope those backticks can be eliminated simply because we add the parentheses as brackets to the various domains.

radumereuta:
The backticks can be eliminated right now. It's just not my top priority.
I made this PR so the original parenthesis based brackets can be used alongside backticks.
If someone has the spare cycles, can go into `kast.k` and eliminate the line where backticks are defined, and modify every definition that fails... It's just a semi-mechanical transformation.

daejunpark:
@radumereuta No, it is not that simple. I already did what you suggested, but it didn't work. Why don't you try to fix one of the followings? You'll find hidden issues.

https://github.com/kframework/k/blob/80ed28cd8d962e8e347ad3ee16da3a90b87fa46b/k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k#L461

https://github.com/kframework/k/blob/80ed28cd8d962e8e347ad3ee16da3a90b87fa46b/k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k#L493

https://github.com/kframework/k/blob/80ed28cd8d962e8e347ad3ee16da3a90b87fa46b/k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k#L556

https://github.com/kframework/k/blob/80ed28cd8d962e8e347ad3ee16da3a90b87fa46b/k-distribution/tutorial/2_languages/2_kool/1_untyped/kool-untyped.k#L572

Issue #1945, opened 2015-12-22
[em] nice error messages


daejunpark:
@grosu said:

> Also, in terms of what else to work on, @daejunpark , what is really really needed is to make sure that, once we make the transition, we will not need to look back at 3.6. Otherwise, by definitions, we are not ready to transit. In addition to making the tutorial examples work, which I am 100% sure that we will before the deadline, we also need to make sure that we do not lose any of the nice error messages that we had in 3.6, which we've spent many tens of hours to design. Some of them are unnecessary now, because we do stronger consistency checks on K modules, but many of them are probably still necessary. Without them, users/students will stumble into errors that they have no idea what they mean (e.g., uncaught exceptions). I would say that that is probably the best thing you can really help us with, especially while you travel in Korea. How does this sound? Thanks!

daejunpark:
@grosu just to make sure, by the nice error messages, do you mean a kind of static checks for K definitions? Where and how can I find the full list of those checks? Who should I ask for this?

cos:
@daejunpark, your best bet would be to look at instantiations of `KEMException` and figure out what they checked for. If it's not obvious on a first read, do a blame and bug the people who implemented most of that code.

daejunpark:
@cos thanks! I expected it as well. But, what K version should I look for? The one before killing KIL, or someone even before?

cos:
I guess 3.6 would be best. I don't think anything significant was done afterwards.

daejunpark:
got it 28ea17d, thanks!

daejunpark:
@grosu I'll work on it, as you suggested.

grosu:
Thanks, @daejunpark.  One thing which is nice about the new KORE-based approach, is that consistency checks are being performed at KORE module built time.  If any of the old checks can be performed that way, then it would be nice to do it.  This way, we move those checks at the place where they are first needed, and do not allow junk to be propagated through the kompile pipeline.

daejunpark:
Progress Report:

I reviewed all of error messages that had been reported in the version 3.6 but not in the current KORE pipeline, and brought them back if (I think) necessary (see the referenced PRs in the above).

The error messages were identified by `grep`ing the following keywords from all of Java source files:
- `KEMException.`
- `KExceptionManager.`
- `new KException`
- `register*Warning`

Among the error messages identified as above (498 in total), I considered only what had been removed by `Kill KIL` PR #1819 (177 in total).

[...]

PR #1975, opened 2016-01-01
[em] Unexpected isBag predicate not of arity 1 found

daejunpark:
from k-3.6/kernel/src/main/java/org/kframework/kore/compile/SortCells.java:322-324


daejunpark:
Jenkins: test this please

daejunpark:
@bmmoore please review

bmmoore:
Looks good for adding back that error message. I don't think we have any cases that trigger it unless you just explicitly write an invalid isBag application in a condition, and I'm not sure what testing we need for error messages, so I'm just merging this change now. 

If there was an actual bug with some definition where this message would have been helpful we can reconsider testing, but if you are just restoring old error behavior it looks fine to me.

bmmoore:
oops, need to merge then close.

daejunpark:
#1945

PR #1976, opened 2016-01-01
auto-import domains

cos:


kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
```diff
@@ -235,14 +246,14 @@ public Rule apply(org.kframework.kil.Rule r) {
...
-            default:
-                throw new AssertionError("Incorrect assoc string: " + assocOrig);
+        case "left":
+            return Associativity.Left();
+        case "right":
+            return Associativity.Right();
+        case "non-assoc":
+            return Associativity.NonAssoc();
+        default:
+            throw new AssertionError("Incorrect assoc string: " + assocOrig);
```
cos, line 256: just formatting above 9 lines

cos:
@dwightguth, please have a look. The significant change for you is that `K-IO` is not imported in `DOMAINS` anymore. `K-IO` requires some resources that are not available in any rewriting context, e.g., testing, so importing it only by choice is safer.

k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
```diff
@@ -148,7 +147,8 @@ endmodule
 
 module IMP
   imports IMP-SYNTAX
-  imports DOMAINS
+  imports K-IO
```
grosu: Is this explicit import of K-IO temporary only?

kernel/src/main/java/org/kframework/kompile/KompileOptions.java
```diff
@@ -120,7 +120,7 @@ public String syntaxModule() {
     @Parameter(names="-I", description="Add a directory to the search path for requires statements.", variableArity = true)
     public List<String> includes = new ArrayList<>();
 
-    @Parameter(names="--no-prelude", description="Do not implicitly require prelude.k.")
+    @Parameter(names="--no-prelude", description="Do not implicitly require prelude.k. It also deactivates automatic inclusion of the DOMAINS and DOMAINS-SYNTAX modules.")
```
grosu, line 123: This is good enough for now, no need for a separate option, indeed.

grosu:
LGTM, except for the explicit import of K-IO, which I'd like to not have to explain to students as early as IMP.  How big of a problem if that?  We can also just merge this for now, and then fix the K-IO thing later.

cos:
@grosu, it can be solved. I'm just not particularly happy about auto-importing something that will only work in contexts where IO is available. Still, we can discuss this at another time. I reverted the explicit imports, and adapted the offending tests.

grosu:
Merged, thanks @cos.  You may want to delete the branch now, not sure if I'm supposed to delete it for you.

PR #1978, opened 2016-01-02
sort-ordered bracket selection

cos:
@grosu, @laurayuwen, this is a fix for unparsing brackets.

The algorithms for unparsing with brackets seems quite decent. The only problem was that it selected a bracket randomly among those available for a sort (including from supersorts). The only change here is that we're now selecting the most specific bracket.


cos:
note that the PR targets the lambdaAllLessons branch, not master

laurayuwen:
The result looks good to me, thanks @cos .

grosu:
Perfect!  Merging.

Eventually, we are going to probably have attributes for unparsing specifically.  But hopefully we'll have K implemented in K by then.

grosu:
So @laurayuwen , we are now ready for your lambdaAllLessons.  Probably you need to first make a PR on the solution display thing, so when we merge, we have a minimal diff to 3.6.

PR #1979, opened 2016-01-02
[em] Make sure you give the correct stream names

daejunpark:
from k-3.6/kernel/src/main/java/org/kframework/compile/transformers/DesugarStreams.java:66-70


daejunpark:
@cos please review

daejunpark:
#1945

k-distribution/tests/regression/error-messages/incorrect-stream-name/b.k
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2015 K Team. All Rights Reserved.
+require "domains.k"
+
```
grosu, line 2: No explicit domains anymore: #1976

PR #1980, opened 2016-01-02
[em] Unsupported matching pattern in stdin stream cell

daejunpark:
from k-3.6/kernel/src/main/java/org/kframework/compile/transformers/ResolveBlockingInput.java

### DIFF
--- a/k-distribution/tests/regression/error-messages/unsupported-io-unblock-patterns/b.k.out
+++ b/k-distribution/tests/regression/error-messages/unsupported-io-unblock-patterns/b.k.out
@@ -1,0 +1,24 @@
+[Warning] Compiler: Unsupported matching pattern in stdin stream cell.
+The currently supported pattern is: <in> ListItem(V:Sort) => .List ... </in>
+	Source(./b.k)
+	Location(19,10,19,45)
+[Warning] Compiler: Unsupported matching pattern in stdin stream cell.
+The currently supported pattern is: <in> ListItem(V:Sort) => .List ... </in>
+	Source(./b.k)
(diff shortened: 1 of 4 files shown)


daejunpark:
@cos please review #1945

k-distribution/tests/regression/error-messages/unsupported-io-unblock-patterns/b.k.out
```diff
@@ -0,0 +1,24 @@
+[Warning] Compiler: Unsupported matching pattern in stdin stream cell.
+The currently supported pattern is: <in> ListItem(V:Sort) => .List ... </in>
```
grosu, line 2: Took me a while to see the difference between the currently supported pattern and the used pattern (the `...`).  This is fine now, but in the future I think we should require an associative list collection to be put in IO cells.  But that's anther discussion.

kernel/src/main/java/org/kframework/kompile/Kompile.java
```diff
@@ -189,7 +189,7 @@ private void checkDefinition(Definition parsedDef) {
     }
 
     public Definition resolveIOStreams(Definition d) {
-        return DefinitionTransformer.from(new ResolveIOStreams(d)::resolve, "resolving io streams").apply(d);
+        return DefinitionTransformer.from(new ResolveIOStreams(d,kem)::resolve, "resolving io streams").apply(d);
```
cos: formatting :)... a space before `kem`. This will be easier once we auto-format the entire project. I'll get to it in the next few days.

daejunpark:
#1945

PR #1981, opened 2016-01-02
[em] context-validity

daejunpark:
from k-3.6/kernel/src/main/java/org/kframework/compile/transformers/ContextsToHeating.java

### DIFF
--- a/kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
+++ b/kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
@@ -158,7 +158,18 @@ public class ResolveContexts {
                 }
                 super.apply(k);
             }
+            // return true when k is either HOLE or #SemanticCastToX(HOLE)
             private boolean isHOLE(K k) {
+                if (k instanceof KApply) {
+                    KApply kapp = (KApply) k;
+                    return kapp.klabel().name().startsWith("#SemanticCastTo") &&
+                            kapp.klist().size() == 1 &&
+                            isHOLEVar(kapp.klist().items().get(0));
+                } else {
+                    return isHOLEVar(k);
+                }
+            }
+            private boolean isHOLEVar(K k) {
                 return k instanceof KVariable && ((KVariable) k).name().equals("HOLE");
             }
         }.check(body);
--- a/kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
+++ b/kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
@@ -130,7 +130,17 @@ public class ResolveContexts {
                 Rule(KRewrite(KSequence(heated, frozen), cooled), requiresCool, BooleanUtils.TRUE, context.att().add("cool")));
     }
 
-    private void checkContextValidity(Context context) {
+    /**
+     * Check validity of context.
+     *
+     * Currently the following conditions are checked:
+     * - Contexts must have at least one HOLE.
+     * - Contexts must have a single rewrite.
+     * - Only the HOLE can be rewritten in a context definition.
+     *
+     * @param context to be checked
+     */
+    public static void checkContextValidity(Context context) {
         K body = context.body();
 
         int cntHoles = new FindK() {
--- a/kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
+++ b/kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
@@ -67,6 +67,7 @@ public class ResolveContexts {
     }
 
     private Stream<? extends Sentence> resolve(Context context, Module input) {
+        checkContextValidity(context);
         final SortedMap<KVariable, K> vars = new TreeMap<>((v1, v2) -> v1.name().compareTo(v2.name()));
         K body = context.body();
         K requiresHeat = context.requires();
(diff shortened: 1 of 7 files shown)


daejunpark:
@cos please review. #1945

kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
```diff
@@ -130,4 +128,50 @@ public void apply(KApply k) {
...
+                if (!isHOLE(k.left())) {
+                    throw KEMException.compilerError("Only the HOLE can be rewritten in a context definition", context);
+                }
+                super.apply(k);
+            }
+            // return true when k is either HOLE or #SemanticCastToX(HOLE)
+            private boolean isHOLE(K k) {
+                if (k instanceof KApply) {
+                    KApply kapp = (KApply) k;
+                    return kapp.klabel().name().startsWith("#SemanticCastTo") &&
```
cos, line 182: the visitor should descend down the semantic cast, so it doesn't seem necessary to consider it in a special way

kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
```diff
@@ -130,4 +128,50 @@ public void apply(KApply k) {
                 Rule(KRewrite(cooled, KSequence(heated, frozen)), requiresHeat, BooleanUtils.TRUE, context.att().add("heat")),
                 Rule(KRewrite(KSequence(heated, frozen), cooled), requiresCool, BooleanUtils.TRUE, context.att().add("cool")));
     }
+
+    private void checkContextValidity(Context context) {
+        K body = context.body();
+        new VisitK() {
```
cos: Let's shrink this. First, this can be split into a counter for holes and one for rewrites. Then, instead of using a visitor and counting manually, either:
1. use `FindK` and count the result, or
2. create a `CountK` that only counts without remembering the results.

kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
```diff
@@ -130,4 +128,50 @@ public void apply(KApply k) {
                 Rule(KRewrite(cooled, KSequence(heated, frozen)), requiresHeat, BooleanUtils.TRUE, context.att().add("heat")),
                 Rule(KRewrite(KSequence(heated, frozen), cooled), requiresCool, BooleanUtils.TRUE, context.att().add("cool")));
     }
+
+    private void checkContextValidity(Context context) {
```
cos: could be split into a static public method taking the K inside the `Context` so we could use it to check the `Context` without expanding it to rules.

kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
daejunpark, line 182: I don't think so. I need to check the lhs of KRewrite is either HOLE or #SemanticCastToX(HOLE).

kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
daejunpark: I didn't fully understand what you mean. Do you want to have an wrapper method that takes Sentence?

kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
cos, line 182: oh, ok then.

kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
cos: no, simply make the method `public static` and add some documentation.

daejunpark:
Jenkins: test this please

kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
```diff
@@ -130,4 +129,66 @@ public void apply(KApply k) {
...
+    /**
+     * Check validity of context.
+     *
+     * Currently the following conditions are checked:
+     * - Contexts must have at least one HOLE.
+     * - Contexts must have a single rewrite.
+     * - Only the HOLE can be rewritten in a context definition.
+     *
+     * @param context to be checked
+     */
```
cos, line 142: great!

kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
```diff
@@ -130,4 +129,66 @@ public void apply(KApply k) {
...
+            }
+        }.apply(body).size();
+        if (cntHoles < 1) {
+            throw KEMException.compilerError("Contexts must have at least one HOLE.", context);
+        }
+
+        int cntRewrites = new FindK() {
+            @Override
+            public scala.collection.Set<K> apply(KRewrite k) {
+                return this.merge(org.kframework.Collections.Set(k), super.apply(k));
```
daejunpark, line 163: @cos which one is correct? this.merge or super.merge?

cos, line 163: As long as you don't override merge, either works.

daejunpark, line 163: I mean, which one will be best practice?

cos, line 163: The best practice would be to use `this.merge` to have a uniform merging algorithm. Still, as this class doesn't override `merge` and the class is very small, it doesn't really matter. It would be very easy to spot that you need to update the `super` to `this` if adding a special `merge`.

daejunpark:
#1945

Issue #1883, opened 2015-12-04
Parsing errors off by +1 line


radumereuta:
Example:

```
odule Test
endmodule
```

@cos thinks that this comes because @dwightguth added `requires "domains.k"` by prepending it to the text.
I think it should be added programmatically. The original text should not be changed before sending it to the parser.

dwightguth:
You are correct, that seems to be the cause of this, because it does not affect anything other than the very entry-level file. Unfortunately, that means it doesn't really affect me in any way. I will only be able to get to this after I no longer have higher priority tasks to accomplish, which may be some time (my guess is on the order of weeks). You may want to fix this yourself.

radumereuta:
Can you at least pinpoint the place where you are doing this?

dwightguth:
https://github.com/kframework/k/blob/master/kernel/src/main/java/org/kframework/kompile/Kompile.java#L252

daejunpark:
@radumereuta was it fixed via #2019?

radumereuta:
It should be, yes.

PR #1982, opened 2016-01-03
[em] ktest hack

daejunpark:


daejunpark:
@cos please review.

This is a hack for negative tests of kompilation, exploiting `posixInitScript` to compare outputs of kompile. Note that ktest currently doesn't support it; it doesn't compare outputs of kompile, but check its exit code only.

I'll open an issue for this.

k-distribution/tests/regression/error-messages/unsupported-io-unblock-patterns/b.k.out
```diff
@@ -1,24 +1,12 @@
...
-	Source(./b.k)
-	Location(25,10,25,36)
 [Warning] Compiler: Unsupported matching pattern in stdin stream cell.
 The currently supported pattern is: <in> ListItem(V:Sort) => .List ... </in>
-	Source(./b.k)
-	Location(28,10,28,66)
 [Warning] Compiler: Unsupported matching pattern in stdin stream cell.
 The currently supported pattern is: <in> ListItem(V:Sort) => .List ... </in>
-	Source(./b.k)
-	Location(31,10,31,75)
```
cos, line 20: why is the location removed?

cos:
also, please mark this with the right label, milestone, and reviewer :)

k-distribution/tests/regression/error-messages/unsupported-io-unblock-patterns/b.k.out
daejunpark, line 20: Jenkins spoils the source information, which becomes hard to compare. The file paths are output in an absolute path in Jenkins machine. Also, the line numbers are added +1 due to #1883.

cos, line 20: could you display at least the Location?
if it's in any way hard, nevermind

cos, line 20: don't worry about the bug, the assertions will be updated when the bug is fixed

daejunpark, line 20: Location will be added back once #1883 is fixed. I opened an issue #1984 to keep track of it.

daejunpark:
#1945

PR #1986, opened 2016-01-03
Em/context validity

daejunpark:


daejunpark:
This is the same with #1982 that was already reviewed.

daejunpark:
#1945

Issue #1985, opened 2016-01-03
Module LANGUAGE-PARSING not found


daejunpark:
@cos do you have any idea?

```
require "domains.k"

module B-SYNTAX
    syntax Pgm ::= "run"
endmodule

module B
    imports B-SYNTAX

    configuration <T>
        <k> $PGM:Pgm </k> <a> $XA:Int </a>
    </T>

    rule run => 0
endmodule
```

```
$ krun -cXA=1 1.b
[Error] Critical: Parser returned a non-zero exit code: 113
Stdout:

Stderr:
[Error] Inner Parser: Module LANGUAGE-PARSING not found. Specify a module with
-m.
```

```
$ cat 1.b
run
```

cos:
`LANGUAGE-PARSING` is added by the compilation pipeline. Could you --debug this?

daejunpark:
```
$ krun  --debug -cXA=1 1.b
org.kframework.utils.errorsystem.ParseFailedException: [Error] Critical: Parser returned a non-zero exit code: 113
Stdout:

Stderr:
[Error] Inner Parser: Module LANGUAGE-PARSING not found. Specify a module with
-m.

    at org.kframework.krun.KRun.externalParse(KRun.java:317)
    at org.kframework.krun.KRun.parseConfigVars(KRun.java:262)
    at org.kframework.krun.KRun.run(KRun.java:70)
    at org.kframework.krun.KRunFrontEnd.run(KRunFrontEnd.java:91)
    at org.kframework.main.FrontEnd.main(FrontEnd.java:52)
    at org.kframework.main.Main.runApplication(Main.java:110)
    at org.kframework.main.Main.runApplication(Main.java:100)
    at org.kframework.main.Main.main(Main.java:52)
[Error] Critical: Parser returned a non-zero exit code: 113
Stdout:

Stderr:
[Error] Inner Parser: Module LANGUAGE-PARSING not found. Specify a module with
-m.
```

daejunpark:
https://github.com/kframework/k/blob/b422c31f1f10ba4fa028198b9da05eff7b2c4ca3/kernel/src/main/java/org/kframework/krun/KRunOptions.java#L117

Clue: If I drop the option `-m mainModuleName` for kast, then it works.

daejunpark:
BTW, do you know why we have duplicate configVars methods? I think the first one should be removed:
[...]

PR #1987, opened 2016-01-03
[em] config var not exist or missing

daejunpark:


daejunpark:
This PR should NOT be merged until 0fc7ed6 is reverted, and 0fc7ed6 cannot be reverted until #1985 is fixed.

@cos Could you review except 0fc7ed6?

k-distribution/tests/regression/error-messages/config-var-not-exist-or-missing/b.k
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) 2015 K Team. All Rights Reserved.
+require "domains.k"
```
grosu: We shouldn't need to require domains.k anymore, should we?

laurayuwen: It is better we don't need to require it, but currently we still do.

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -157,6 +158,20 @@ class Module(val name: String, val imports: Set[Module], unresolvedLocalSentence
...
+    val transformer = new FoldK[Set[KToken]] {
+      override def apply(k: KToken): Set[KToken] = {
+        if (k.sort.name == "KConfigVar") Set(k) else unit
+      }
+      def unit = Set()
+      def merge(set1: Set[KToken], set2: Set[KToken]) = set1 | set2
+    }
+    initRules.map(r => transformer.apply(r.body))
+             .fold(transformer.unit)(transformer.merge)
+  }
```
cos: I know we haven't done this well in the past (I wasn't always asked to review unfortunately), but I would prefer if we avoid adding anything to `Module` when it is not very clearly part of the "K language". Configuration variables are just a convention that are representable in multiple ways, so it may be good to put them elsewhere.

Maybe a special object/class dealing with configuration or configuration variables?

daejunpark: @cos do you satisfy with this?

cos: it's very close. could you rename the class to something more specific, e.g., ConfigVars, and move it to its own file?

daejunpark:
@cos please re-review

daejunpark:
#1945

PR #1989, opened 2016-01-03
[em] Found circularity in module imports

daejunpark:
from k-3.6/kernel/src/main/java/org/kframework/kil/loader/DefinitionContext.java:59


daejunpark:
@cos please review

daejunpark:
#1945

PR #1990, opened 2016-01-03
updates of all lambda++ lessons (only lesson_1 changed in definition)

laurayuwen:
This is the artificial PR of 3_lambda++ for @grosu to review. 
For definitions, only lesson 1 changed a little bit as it is substitution based, other lessons have no difference. 
We have been using tests-kore folder for testing, but here I updated tests directly for the ease of review. Basically, we cannot produce more than one search result now. Other tests not showing here have exactly the same output as before.
In the branch to be really merged, I keep the original tests folders unchanged and new outputs are in tests-kore.

### DIFF
--- a/k-distribution/tutorial/1_k/3_lambda++/lesson_4/tests/callcc-env1.lambda.out
+++ b/k-distribution/tutorial/1_k/3_lambda++/lesson_4/tests/callcc-env1.lambda.out
@@ -2,8 +2,4 @@ Search results:
 
 Solution 1:
 V:K -->
-3
-
-Solution 2:
-V:K -->
 4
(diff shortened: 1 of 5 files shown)


k-distribution/tutorial/1_k/3_lambda++/lesson_4/tests/callcc-env1.lambda.out
```diff
@@ -2,8 +2,4 @@ Search results:
 
 Solution 1:
 V:K -->
-3
-
-Solution 2:
-V:K -->
 4
```
grosu, line 5: We cannot merge this.  Please read the README.  The whole point of this is that we have two solutions here.  So we must be able to use search and see multiple solutions.  We cannot transit to K 4.0 before we are able to do that.  Search is needed in several other lessons in the tutorial.

PR #1993, opened 2016-01-04
resolve current directory for processes

cos:


dwightguth:
can you explain to me why this was needed?

cos:
I was trying to fix the issue @grosu raised. I'm wasn't sure windows correctly recognizes the "." as the current directory. As the issue was the extension, this is likely not a fix.

radumereuta:
Yeah, that's not the issue.
The problem is when you are trying to run a command line from a java process. On linux and macs you can write `kast a.txt`, but on windows, you have to write `kast.bat a.txt`.

cos:
@radumereuta, thanks! I know, I saw the other issue.

radumereuta:
For documentation purposes.

cos:
@dwightguth, please either merge or delete this PR.

dwightguth:
It's possible this change might affect the PWD or CD environment variables but it would likely not have any other effect since the working directory of new processes would be effectively the same both before and after. Given this, and the fact that any effect if one exishs would be subtle and potentially non obvious, I'm going to close this without merging. If some aspect of the current behavior can be found to be undesirable and more accurately traced back to this, we will reevaluate

PR #1994, opened 2016-01-04
use absolute path to os-specific script for default krun parser

dwightguth:
@grosu can you please test this on your machine and let me know if it fixes the issue?


grosu:
yes, this fixed the issue.

PR #1997, opened 2016-01-05
Types lesson9

laurayuwen:
This is an artificial PR for @grosu to review the definition changes and changed expected output for last lesson of 5_types.


k-distribution/tutorial/1_k/5_types/lesson_9/lambda.k
```diff
@@ -46,7 +47,7 @@ module LAMBDA
   syntax Type ::= MetaVariable
   rule <k> let X = T:Type in E => E ~> tenv(TEnv) ...</k>
        <tenv> TEnv
-        => TEnv[X <- (forall #metaVariables(T) -Set #metaVariables(tenv(TEnv))) #freezeVariables(T, tenv(TEnv))]
+        => TEnv[X <- (forall #metaVariables(T) -Set #metaVariables(tenv(TEnv))) #freezeVariables(T, tenv(TEnv)):>Type]
```
grosu: What does #freezeVariables do?  @andreistefanescu , can you please add a NOTE.md and explain it?  Maybe also change the name, because this name is not intuitive at all?

k-distribution/tutorial/1_k/5_types/lesson_9/lambda.k
```diff
@@ -46,7 +47,7 @@ module LAMBDA
...
        </tenv>
   rule <k> X:Id => #rename(T, Tvs) ...</k>
```
grosu: For uniformity, I would rename #rename to #renameVariables.

grosu: Actually, we now call them KVariables, don't we?  Then I would suggest the following names for the operations above:

#metaVariables => #KVariables
#rename => #renameKVariables
#freezeVariables => #whateverKVariables // I don't know what this does.  What it should do is to solve the constraints and "flush" them.

@andreistefanescu , what do you think?

andreistefanescu: @grosu, we want to have both `#metaVariables` and `#variables`, based on whether we want the returned variables to be further affected by unification or not. Same for the rename operations. Here we want the meta versions, and in logik we want the non-meta version. Of course, we can rename them in `renameKVariables` and `renameMetaKVariables`.

k-distribution/tutorial/1_k/5_types/lesson_9/lambda.k
andreistefanescu: @grosu, the name `#freezeVariables` is not the best; `#freezeVariables(t1, t2)` will replace all the variables of `t1` which are not variables of `t2` with meta-variables.

grosu: @andreistefanescu , I see and I remember now.  Please correct me, you too, @cos, but I do believe that we can elegantly and mathematically cleanly define all these operations using the metalevel.  For example, if we have an operation `#meta(t)` that takes term `t` to the meta-level, then we can define all these operations there, including generating fresh variables for some of its meta-variables when we down it, using rules with ? variables in their RHS and keeping track of what fresh variable corresponds to what meta-variable. We should discuss this in a separate meeting after the release, but I'm optimistic that this may very elegantly solve what we before though would be a "hack".  Of course, all this assuming that the meta-level capability itself is not a hack :)

Regarding naming, how about

```
#metaVariables(Term) => #metaKVariables(Term)
#rename(Term, MetaKVariableSet) => #freshKVariables(Term, MetaKVariableSet)
```

`#freezeVariables(t1,t2)` has a semantics which is kind of hard to explain.  How did we end up with it?  I'm not very concerned with it right now, because I think that the _right_ way to deal with this example is to symply declare `(forall_)_` to be a `[binder]` and then just write `(#KVariables(T) - #KVariables(TEnv)) T`.  But we do not have time for this now.  So I woul;d suggest to just rename `#freezeVariables(t1,t2)` to `#freezeKVariables(t1,t2)` and move on for now.

laurayuwen:
I am going to change all lessons of 5_types, have we reached a consensus on names?
This is a list of `#` in unification.k, correct me if needed. @grosu @andreistefanescu 

```
#renameVariables => #renameKVariables
#rename => #renameMetaKVariables
#variables => #kVariables
#variablesMap => #kVariablesMap
#metaVariables => #metaKVariables
#freezeVariables => #freezeKVariables
#unifiable (the same)
```

grosu:
Go ahead with these.  We'll see later how to change them.

laurayuwen:
@grosu Please have a look at the revised definition or the last commit.

grosu:
looks good.  you can merge it.

PR #1999, opened 2016-01-05
avoid using cell specific sets

dwightguth:
Fixes a bug where `<thread>...</thread> <thread>...</thread>` is parsed incorrectly leading  to lack of correct cell completion.


dwightguth:
@bmmoore please review

bmmoore:
This code seems fine for adding the avoid attribute, but the changes and commit messages are a bit light on explanations, and a test that involves actual parsing.

 What is the incorrect parse? I'm guessing the issue is ending up with the label for the cell-specific operation instead of `#cells`, but I'm not sure whether other productions with no terminal could be ambiguous too.
- Please put a comment in `GenerateSentencesFromConfigDecls` explaining why `avoid` is needed.
- For test coverage, adding a test that would prevent a regression on this is ideal (if it's already covered by some definition you can just mention that in the commit message or comment), but I would also be satisfied with just having a new issue specifically for getting test coverage on this (if you make an issue, please include an example input and incorrect parse that resulted before this pull request fixes things.).

bmmoore:
Okay, looks good.

Issue #2005, opened 2016-01-06
Unparsing: get rid of dot empty List


laurayuwen:
An example can be found in PR #1998.

cos:
this looks like low priority.

PR #2006, opened 2016-01-07
imp++ last lesson, diff with k3.6 for review

laurayuwen:
@grosu please have a look at this artificial PR. I already revert many changes back to K 3.6 in this commit, will create real PR only after you are ok with this.

### DIFF
--- a/k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
+++ b/k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
@@ -133,12 +133,12 @@ of statements surrounded by curly brackets.  */
                    Block "else" Block         [strict(1)]
                  | "while" "(" BExp ")" Block
                  | "int" Ids ";"
-                 | "print" "(" AExps ")" ";"  [strict]
+                 | "print" "(" AExps ")" ";"
                  | "halt" ";"
                  > "join" AExp ";"            [strict]
 
   syntax Ids   ::= List{Id,","}               [strict, klabel(ids)]
-  syntax AExps ::= List{AExp,","}             [strict]
+  syntax AExps ::= List{AExp,","}
   syntax Stmts ::= List{Stmt,""}
 endmodule
 
@@ -397,8 +397,8 @@ count as a computational step. */
 
   syntax Printable ::= Int | String
   syntax AExp ::= Printable
-  syntax Printables ::= List{Printable,","}
-  syntax KResult ::= Printables
+
+  context print(HOLE:AExp, AEs:AExps);
 
   rule <k> print(P:Printable,AEs => AEs); ...</k>
        <out>... .List => ListItem(P) </out>  [print]
--- a/k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
+++ b/k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
@@ -137,7 +137,7 @@ of statements surrounded by curly brackets.  */
                  | "halt" ";"
                  > "join" AExp ";"            [strict]
 
-  syntax Ids   ::= List{Id,","}               [strict, klabel(ids)]
+  syntax Ids   ::= List{Id,","}
   syntax AExps ::= List{AExp,","}
   syntax Stmts ::= List{Stmt,""}
 endmodule
@@ -396,6 +396,8 @@ all behaviors.  The second rule is structural because we do not want it to
 count as a computational step. */
 
   syntax Printable ::= Int | String
+/* currently it is necessary to subsort Printable to AExp,
+but future K should be able to infer that automatically. */
   syntax AExp ::= Printable
 
   context print(HOLE:AExp, AEs:AExps);
--- a/k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
+++ b/k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
@@ -1,4 +1,4 @@
-// Copyright (c) 2014-2015 K Team. All Rights Reserved.
+// Copyright (c) 2014-2016 K Team. All Rights Reserved.
 
 /*!
 \setlength{\parindent}{1em}
@@ -137,7 +137,7 @@ of statements surrounded by curly brackets.  */
                  | "halt" ";"
                  > "join" AExp ";"            [strict]
 
-  syntax Ids   ::= List{Id,","}               [strict]
+  syntax Ids   ::= List{Id,","}               [strict, klabel(ids)]
   syntax AExps ::= List{AExp,","}             [strict]
   syntax Stmts ::= List{Stmt,""}
 endmodule
@@ -182,7 +182,7 @@ buffers as lists of items. */
                       <id color="black"> 0 </id>
                     </thread>
                   </threads>
-                  <br/>
+//                <br/>
                   <store color="red"> .Map </store>
 //                  <in color="magenta"> .List </in>
 //                  <out color="Orchid"> .List </out>
@@ -397,6 +397,9 @@ count as a computational step. */
 
   syntax Printable ::= Int | String
   syntax AExp ::= Printable
+  syntax Printables ::= List{Printable,","}
+  syntax KResult ::= Printables
+
   rule <k> print(P:Printable,AEs => AEs); ...</k>
        <out>... .List => ListItem(P) </out>  [print]
   rule print(.AExps); => .  [structural]
@@ -444,7 +447,7 @@ with the declared configuration cell multiplicity.  The unique identifier
 of the new thread is generated using the ``fresh'' side condition. */
 
   rule <k> spawn S => !T:Int ...</k> <env> Rho </env>
-       (. => <thread>... <k> S </k> <env> Rho </env> <id> !T </id> ...</thread>)
+       (.Bag => <thread>... <k> S </k> <env> Rho </env> <id> !T </id> ...</thread>)
 
 /*@ \subsubsection{Join thread}
 A thread who wants to join another thread \texttt{T} has to wait until


k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
```diff
@@ -137,7 +137,7 @@ of statements surrounded by curly brackets.  */
                  | "halt" ";"
                  > "join" AExp ";"            [strict]
 
-  syntax Ids   ::= List{Id,","}               [strict]
+  syntax Ids   ::= List{Id,","}               [strict, klabel(ids)]
```
grosu: I've seen this klabel attribute in KOOL, too.  We need to get rid of it.

k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
```diff
@@ -397,6 +397,9 @@ count as a computational step. */
 
   syntax Printable ::= Int | String
   syntax AExp ::= Printable
+  syntax Printables ::= List{Printable,","}
+  syntax KResult ::= Printables
+
```
grosu, line 404: Can somebody explain why we need these two syntax declarations now?  I remember during the 3.6 discussions we were even thinking of getting rid of L399, but now instead we added two more.

k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
```diff
@@ -444,7 +447,7 @@ with the declared configuration cell multiplicity.  The unique identifier
 of the new thread is generated using the ``fresh'' side condition. */
 
   rule <k> spawn S => !T:Int ...</k> <env> Rho </env>
-       (. => <thread>... <k> S </k> <env> Rho </env> <id> !T </id> ...</thread>)
+       (.Bag => <thread>... <k> S </k> <env> Rho </env> <id> !T </id> ...</thread>)
```
grosu, line 452: So we still call these bags?  @andreistefanescu , is this going to change once we have the A/AC stuff?

k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
laurayuwen, line 404: The two lines were added because .AExps is not recognized as KResult.

k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
andreistefanescu, line 452: The name `.Bag` is change by the frontend into `.ThreadCellBag` or something like that. As far as I am concerned, we can keep the name `.Bag`.

grosu, line 452: OK, so then let's not worry about it for the Jan 18 deadline.

k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
grosu, line 404: This is very strange.  Both that `.AExps` is not recognized as or needs to be recognized as a `KResult`, and the way above to state it, indirectly by means of `Printables`.  I suspect the problem is more general, namely that lists of `Printable` cannot be plugged back into the argument of `print`, because K is not smart enough to figure out to us that they are meant to be a `KResult`.  And then `.AExps`, which is the same as `.Printables`, is a special case.

If that is the case, then we should probably choose one of the following:

1) keep `AExps ::= List{AExp,","}` as `strict`, in which case we expect the argument of `print` to be `Printables`, so we should get rid of L399 and replace `AEs` with `Ps` and `.AExps` with `.Printables` in L403 and L405.  Does that work?

2) get rid of the attribute `strict` for  `AExps ::= List{AExp,","}` and of L399-401, and instead add an evaluation context `context print(HOLE:AExp, AEs:AExps)`, so that `print` evaluates and then prints each of its arguments in order.  Can you see if this works?

laurayuwen, line 404: I agree the solution is indirectly strange, but given the place K get stuck, I still feel it is a problem of list evaluation itself, not print. I still tried your solution 1, it also got stuck at `<k> .AExps ~> #freezer5 ( "\n" ) ~> #freezer5 ( x ) ~> #freezer5 ( "Line  3: x = " ) ~> #freezer12 ( ) ~> ...` for locals.imp to `print("Line  3: x = ", x, "\n");`  freezer5 is strictness of AExps, and freezer12 is for print. Haven't tried the second solution because we can't use HOLE right now.

grosu, line 404: But without HOLE, how are we going to deal with KOOL?  @daejunpark ?

In the first approach I recommended, of course it is a matter of list evaluation.  I did not imply otherwise.  My point was that L399 subsorting Printable to AExp should not be necessary.  Can you make the other changes and make sure they are not the problem, and the problem is indeed adding vs not adding L399?  If that is indeed necessary, please add a comment next to it saying that it is currently necessary to subsort Printables to AExps, but that in the future K should be able to infer that automatically.

laurayuwen, line 404: Yes, the stuck info I gave above is from strict AExps and the following code where we print printables (because we did not sate the subsorting, a klabel is added for Printables to alleviate `Cannot compute least sort of term: .List{"'_,_"}(.KList)` and a syntax for print Printables is added in order to change the rules).

```
  syntax Printable ::= Int | String
  syntax Printables ::= List{Printable,","} [klabel(printables)]
  syntax KResult ::= Printables
  syntax Stmt ::= "print" "(" Printables ")" ";"
  rule <k> print(P:Printable,Ps:Printables => Ps); ...</k>
       <out>... .List => ListItem(P) </out>  [print]
  rule print(.Printables); => .  [structural]
```

If we add subsorting back and of course remove the klabel, things are fine.

```
 syntax Printable ::= Int | String
  syntax AExp ::= Printable
  syntax Printables ::= List{Printable,","} //[klabel(printables)]
  syntax KResult ::= Printables
  //syntax Stmt ::= "print" "(" Printables ")" ";" //We don't need the extra `syntax Stmt` also.
  rule <k> print(P:Printable,Ps:Printables => Ps); ...</k>
       <out>... .List => ListItem(P) </out>  [print]
  rule print(.Printables); => .  [structural]
```

Either add `syntax AExp ::= Printable` or `syntax AExps ::= Printables`, the subsorting seems essential here.

grosu, line 404: OK, then lets go ahead with `syntax AExp ::= Printable` and with the comment I made before.

laurayuwen, line 404: Oh my misunderstandings, last time we cannot use HOLE in the rule in the last lesson of 5_types doesn't mean we cannot use HOLE in context.

laurayuwen, line 404: We can also remove the strictness of both `AExps` and `print` and the following does the trick

```
  syntax Printable ::= Int | String
  syntax AExp ::= Printable

  context print(HOLE:AExp, AEs:AExps);

  rule <k> print(P:Printable,AEs => AEs); ...</k>
       <out>... .List => ListItem(P) </out>  [print]
  rule print(.AExps); => .  [structural]
```

This way rules don't need to any change.

grosu, line 404: I like this!  This was the second variant I proposed above and I am glad it works.  This also gives me a nice opportunity to explain context with holes earlier in the tutorial.  Everything turned out to be nice, after all.  Thanks, @laurayuwen

laurayuwen, line 404: Yes, I also prefer this solution. It cleverly avoids the problem of list evaluation using strictness. I should thank you for your forgiveness @grosu .

grosu, line 404: Let me know when ready for final review.  We have to rush with these simple languages, we still have a lot of work to do.

laurayuwen:
@grosu Please review

k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
```diff
@@ -133,12 +133,12 @@ of statements surrounded by curly brackets.  */
...
                  | "while" "(" BExp ")" Block
                  | "int" Ids ";"
-                 | "print" "(" AExps ")" ";"  [strict]
+                 | "print" "(" AExps ")" ";"
                  | "halt" ";"
                  > "join" AExp ";"            [strict]
 
-  syntax Ids   ::= List{Id,","}               [strict]
-  syntax AExps ::= List{AExp,","}             [strict]
+  syntax Ids   ::= List{Id,","}               [strict, klabel(ids)]
```
grosu: you should remove both attributes of this production.

k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
```diff
@@ -397,6 +397,9 @@ count as a computational step. */
 
   syntax Printable ::= Int | String
   syntax AExp ::= Printable
```
grosu, line 401: You should not need this production either.

grosu, line 401: Or if you need it because the parser is not smart to figure it out, please add the comment I suggested before back.

laurayuwen:
@grosu I noticed that the tutorial started talking about print in lesson 1, e.g. readme:
`we therefore define a new list syntactic category,`AExps`, which we pass
  as an argument to print; moreover, we declare print`strict`, to evaluate
  its`AExps`-list argument, and then the`AExps`list construct`seqstrict`,`
If we change to use HOLE instead, then we want to be consistent and never use strict, right?
If I remove the strictness from start, the expected output would also need to change.

k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
laurayuwen, line 401: yes it is still needed because of the same reason, I will add a comment

laurayuwen, line 401: Is such comment just needed in one lesson, or all lessons? Maybe just the last lesson? I don't see much comments in previous lessons.

grosu:
don't worry about the readmes, @laurayuwen, i will modify those accordingly.

> If we change to use HOLE instead, then we want to be consistent and never use strict, right?

I don't understand this.  We only use the HOLE for the semantics of print, not others.

> If I remove the strictness from start, the expected output would also need to change.

It shouldn't.  Just change it and let me take a look at it.

k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
grosu, line 401: Not only the last one, maybe only the first.  But for now let's add it to all of them to make sure we don't forget, and I'll clean up as I revise the readmes.

laurayuwen:
By never, I mean never use strict for print in ALL lessons. not talking about other construct.
For lesson1, removing strictness of print and AExps would result in the following difference:

```
Unexpected program output:
Expected:
<T>
    <k>
        "Input two numbers: " ~> HOLE, .AExps ~> print ( HOLE ) ; ~> int x
          , y ; ~> x = read ( ) ; ~> y = read ( ) ; ~> print ( 
          "Their sum is: ", x + y, "\n" ) ;
    </k>
    <state>
        .Map
    </state>
</T>


but found:
<T> <k> print ( "Input two numbers: " , .Ids ) ; ~> int x , ( y , .Ids ) ; ~> ( x = read ( ) ; ) ~> ( y = read ( ) ; ) ~> print ( "Their sum is: " , ( ( x + y ) , ( "\n" , .Ids ) ) ) ; ~> . </k> <state> .Map </state> </T>
```

grosu:
I see.  Yes, please propagate the same approach to print everywhere, but let us first agree on and merge the last one.  Then modify the output files accordingly wherever needed.  On an unrelated note, can we not output the `.Ids`, etc., unit in the unparser?  If not trivial, please add an issue at least, if not already, to take care of it asap after jan 18.

laurayuwen:
@grosu  Yes, I updated the last lesson already. For previous lessons, because they are unchanged previously, still the same as K3.6. Is it OK that you review real PR directly? Also, I am not sure for tests, do I create tests-kore or update tests directly?

laurayuwen:
I already created an issue #2005 for unparsing when @grosu cared about it last time in another PR.

grosu:
LGTM now.  @laurayuwen , I cannot know when you updated your PR, you have to tell me explicitly each time you want me to review something.

PR #2007, opened 2016-01-07
Binary format for KAST

dwightguth:
@cos @bmmoore please review

This seems to show a decent performance improvement for kcc when operating over large terms.


dwightguth:
I would also like to get @bmmoore 's approval on this before it's merged.

dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

dwightguth:
actually, hold off on this one, I want to try to optimize it further.

cos:
just curious, why have you decided to implement this from scratch instead of using one of the serialization libraries out there?

dwightguth:
If you know of a serialization framework that uses the same serialization format across both Java and OCAML and you believe would have as good or better performance, let me know. I'm pretty satisfied with this solution though because it wasn't too hard to implement

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -195,17 +208,20 @@ public static Rule parsePattern(FileUtil files, KExceptionManager kem, String pa
         return compiledDef.parsePatternIfAbsent(files, kem, pattern, source);
     }
 
-    public static void prettyPrint(CompiledDefinition compiledDef, OutputModes output, Consumer<String> print, K result) {
+    public static void prettyPrint(CompiledDefinition compiledDef, OutputModes output, Consumer<byte[]> print, K result) {
```
cos, line 211: the `.getBytes()` conversions are a bit ugly; how about the consumer itself converting the strings to bytes?

kernel/src/main/java/org/kframework/krun/RunProcess.java
```diff
@@ -1,22 +1,39 @@
...
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.PrintStream;
 import java.util.Map;
+import java.util.function.Supplier;
 
 // instantiate processes
 public class RunProcess {
 
+    public static Thread getOutputStreamThread(Process p2, Supplier<InputStream> in, PrintStream out) {
```
cos: I don't understand this part of the codebase well -- it seems very involved for its purpose. If you do intend to keep it clean, this method could use javadoc.

kernel/src/main/java/org/kframework/parser/binary/BinaryParser.java
```diff
@@ -0,0 +1,167 @@
...
+ * self-contained and composable. A client can take the output of two KAST terms and combine them into a single term
+ * simply by concatenating the terms together after stripping their MAGIC prefix and suffix. This will not be as
+ * space-compact as if the term was outputted all at once, but can be done in constant time without requiring the terms
+ * to be modified internally, and will still deserialze correctly.
+ */
+public class BinaryParser {
+
+    private static final byte[] MAGIC = {0x7f, 'K', 'A', 'S', 'T'};
+
+    public static final int BEGIN = 0, KTOKEN = 1, KAPPLY = 2, KSEQUENCE = 3, KVARIABLE = 4, KREWRITE = 5,
```
cos, line 58: could these be private?

kernel/src/main/java/org/kframework/parser/binary/BinaryParser.java
```diff
@@ -0,0 +1,167 @@
...
+            }
+            String s = new String(buf);
+            interns.add(s);
+            return s;
+        } else {
+            return interns.get(interns.size() - idx);
+        }
+    }
+
+    public static boolean isBinaryKast(byte[] bytes) {
```
cos, line 140: javadoc

cos, line 140: actually, nevermind, it's obvious from context

kernel/src/main/java/org/kframework/unparser/ToBinary.java
```diff
@@ -0,0 +1,122 @@
...
+
+    }
+
+    public static byte[] apply(K k) {
+        ByteArrayOutputStream out = new ByteArrayOutputStream();
+        apply(out, k);
+        return out.toByteArray();
+    }
+
+    private static class Traverse {
```
cos: the outer class only has static methods -- could the inner class contents be lifted to the outer class?

kernel/src/main/java/org/kframework/utils/koreparser/KoreParser.java
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2015-2016 K Team. All Rights Reserved.
+package org.kframework.utils.koreparser;
+
+import org.kframework.attributes.Source;
+import org.kframework.kore.K;
+
+/**
+ * Created by dwightguth on 11/17/15.
+ */
```
cos: javadoc

kernel/src/main/java/org/kframework/utils/koreparser/KoreParser.java
```diff
@@ -0,0 +1,16 @@
...
+@Deprecated
+public class KoreParser {
```
cos: why is this class deprecated on being created?

kernel/src/main/java/org/kframework/utils/koreparser/KoreParser.java
```diff
@@ -0,0 +1,16 @@
...
+
+    public static K parse(String s, Source src) {
```
cos: javadoc

kernel/src/test/java/org/kframework/unparser/BinaryKASTTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.kframework.utils.file.FileUtil;
+
+import java.io.File;
+
+import static org.junit.Assert.*;
+import static org.kframework.kore.KORE.*;
+
+/**
+ * Created by dwightguth on 11/16/15.
+ */
```
cos: please remove

kore/src/main/java/org/kframework/kore/ListIterable.java
```diff
@@ -0,0 +1,36 @@
...
+package org.kframework.kore;
+
+import scala.collection.immutable.List;
+
+import java.io.Serializable;
+import java.util.Iterator;
+
+/**
+ * Created by dwightguth on 11/19/15.
+ */
```
cos: remove

kore/src/main/java/org/kframework/kore/ListIterable.java
```diff
@@ -0,0 +1,36 @@
...
+public class ListIterable<T> implements Iterable<T>, Serializable {
```
cos: if this is intended as K's List-type interface, let's just call it List -- it could also live better directly under `org.kframework`, or maybe in `org.kframework.collection`

kore/src/main/scala/org/kframework/kore/ADT.scala
```diff
@@ -22,11 +22,15 @@ object ADT {
...
+    def size = klist.size
+    def asIterable = klist.asIterable
   }
 
   class KSequence private(val elements: List[K], val att: Att = Att()) extends kore.KSequence {
     val items: java.util.List[K] = elements.asJava
-    val kApply: kore.KApply = items.asScala reduceRightOption { (a, b) => KLabel(KLabels.KSEQ)(a, b) } getOrElse { KLabel(KLabels.DOTK)() } match {
+    val size: Int = elements.size
+    val asIterable: java.lang.Iterable[K] = new ListIterable(elements)
+    lazy val kApply: kore.KApply = items.asScala reduceRightOption { (a, b) => KLabel(KLabels.KSEQ)(a, b) } getOrElse { KLabel(KLabels.DOTK)() } match {
```
cos, line 33: nothing to do with this PR, but I think we should eventually get rid of `KSequence` in favor of just having a `KApply`

kore/src/main/scala/org/kframework/kore/ADT.scala
```diff
@@ -40,6 +44,11 @@ object ADT {
   }
 
   object KSequence {
+    private val emptyAtt = Att()
```
cos, line 47: emptyAtt is not specific to KSequence -- is there a problem with simply using `Att()`?

kore/src/main/scala/org/kframework/kore/Constructors.scala
```diff
@@ -17,11 +17,13 @@ trait Constructors[KK <: K] {
 
   val injectedKListLabel = "INJECTED-KLIST"
 
+  private val emptyAtt = Att()
```
cos, line 20: same, this is not shorter

kore/src/main/scala/org/kframework/kore/KORE.scala
```diff
@@ -22,21 +22,23 @@ object KORE extends Constructors[K] with ScalaSugared[K] {
 
   val constructor = this
 
+  lazy val emptyAtt = Attributes()
```
cos, line 25: same

kore/src/main/scala/org/kframework/kore/interface.scala
```diff
@@ -1,6 +1,7 @@
 package org.kframework.kore
 
 import org.kframework.attributes._
+import org.kframework.unparser.Unparse
```
cos, line 4: unused

kernel/src/main/java/org/kframework/krun/KRun.java
dwightguth, line 211: sorry to be clear you're proposing a `Consumer<String>`? How am I supposed to print bytes in that case? One of the properties  of raw binary data is that you would need some kind of quite space-inefficient and time-inefficient encoding like Base64 in order to encode it as a string.

kernel/src/main/java/org/kframework/parser/binary/BinaryParser.java
dwightguth, line 58: It needs to be accessed also by ToBinary.

kore/src/main/scala/org/kframework/kore/ADT.scala
dwightguth, line 47: Yes, it creates a new object for every KSequence created which is too slow and uses too much memory.

kore/src/main/scala/org/kframework/kore/Constructors.scala
dwightguth, line 20: same here

kore/src/main/scala/org/kframework/kore/KORE.scala
dwightguth, line 25: and here

kore/src/main/scala/org/kframework/kore/interface.scala
dwightguth, line 4: it is actually. It's used on line 14. The reason it wasn't needed before was that they used to be in the same package.

kore/src/main/scala/org/kframework/kore/ADT.scala
cos, line 47: Att() is immutable, so we can make it return the same object. Add this to the Att object in Att.scala:

```
  val empty = Att()
  def apply(atts: Set[K]): Att = if(atts.isEmpty) empty else Att(atts)
```

kernel/src/main/java/org/kframework/krun/KRun.java
cos, line 211: nevermind, now I see the cases include the binary version; the `prettyPrint` name is confusing.

kore/src/main/scala/org/kframework/kore/ADT.scala
dwightguth, line 47: That won't work because the time involved is actually the time spent constructing the atts set.

PR #2008, opened 2016-01-07
update copyright on all *.java

dwightguth:
So that we don't have to keep editing things all the time.


dwightguth:
Jenkins: test this please

PR #2010, opened 2016-01-07
Imp++Lesson_8

laurayuwen:
To merge into master, its diff with K3.6 already reviewed.

### DIFF
--- a/k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
+++ b/k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
@@ -1,5 +1,4 @@
-// Copyright (c) 2014-2015 K Team. All Rights Reserved.
-require "domains.k"
+// Copyright (c) 2014-2016 K Team. All Rights Reserved.
 
 /*!
 \setlength{\parindent}{1em}
@@ -78,7 +77,6 @@ of a program using the \texttt{search} option of \texttt{krun}.
 */
 
 module IMP-SYNTAX
-
 /*@ \section{Syntax}
 IMP++ adds several syntactic constructs to IMP\@.  Also, since the
 variable declaration construct is generalized to be used anywhere a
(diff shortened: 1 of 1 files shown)


k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
```diff
@@ -397,15 +395,16 @@ different threads may compete on the output buffer and we want to capture
...
+but future K should be able to infer that automatically. */
   syntax AExp ::= Printable
   syntax Printables ::= List{Printable,","}
   syntax KResult ::= Printables
-  syntax AExps ::= Printables
+
   rule <k> print(P:Printable,AEs => AEs); ...</k>
        <out>... .List => ListItem(P) </out>  [print]
-  rule print(.AExps); => .K  [structural]
+  rule print(.AExps); => .  [structural]
```
grosu, line 407: .Printables

k-distribution/tutorial/1_k/4_imp++/lesson_8/imp.k
```diff
@@ -397,15 +395,16 @@ different threads may compete on the output buffer and we want to capture
...
-  syntax KResult ::= Printable
   syntax Printable ::= Int | String
+/* currently it is necessary to subsort Printable to AExp,
+but future K should be able to infer that automatically. */
   syntax AExp ::= Printable
   syntax Printables ::= List{Printable,","}
   syntax KResult ::= Printables
-  syntax AExps ::= Printables
+
   rule <k> print(P:Printable,AEs => AEs); ...</k>
```
grosu, line 405: But @laurayuwen , this is not the version I reviewed and we agreed upon!!!  It should be Ps instead of AEs and .Printables instead of .AExps.

laurayuwen, line 405: Sorry! I thought we agreed subsorting is essential and we want minimal changes.

grosu, line 405: We want it done right.  What was there before was fine in that context, where the arguments of print were evaluated in order and printed, so the rest of the arguments were always an AExps.  Now they are all evaluated before they are printed.  These two are different semantics.

PR #2011, opened 2016-01-08
Imp++

laurayuwen:


laurayuwen:
Since we have different tests results, Should I update existing tests or add tests-kore and keep tests unchanged?@grosu

grosu:
@laurayuwen : update existing tests, why waste time anymore.

laurayuwen:
Could somebody review this PR of the definition change for all lessons in imp++? @grosu @cos 
There will be quite some test config and test result updates, so I feel it better not mix with definition updates. So the first seven lessons don't have tests enabled in this PR.

grosu:
Looks good, you can merge in what regards me.  I let you merge it, because I'm not sure in what order you want to merge this PR and the one modifying the .out files.

PR #2018, opened 2016-01-09
1_lambda revised

grosu:
@cos, @laurayuwen please revise.

I initially wanted to make these revisions final, that is, to incorporate all the comments in and then remove all the `NOTES.md` files, but I realized that it way too much to achieve by Jan 18.  Consequently, I decided to only update the tutorial enough to be actual, leaving some other refactorings for a later stage.  In fact, several users requested to put the entire tutorial into one document, like a book, which is probably a good idea.  I'm going to eventually write a program that does that automatically.  When I do that, I'll also incorporate all the remaining comments and notes in the NOTES.md files.


grosu:
@cos, @laurayuwen : can you guys merge this asap?  Why is it failing?  I haven't touched any code that ktest executed successfully.

grosu:
OK, they passed now.  I had a few mistakes, though, which I fixed and then everything passed.

@cos , can you please merge if you find everything OK?  @laurayuwen , if you have any comments you can let me know also after merging, and I'll take care of them later.

k-distribution/tutorial/1_k/1_lambda/lesson_5/README.md
```diff
@@ -80,9 +73,9 @@ rules, because the K rewriting machinery is not used for evaluating the side
 conditions.
```
cos: This may be confusing for a user. In practice, functions can be user-defined syntactical constructs whose semantics is defined using K rules, and K's rewriting machinery is actually used to evaluate them even when they appear in side conditions.

k-distribution/tutorial/1_k/1_lambda/lesson_7/README.md
```diff
@@ -50,10 +51,10 @@ factorial of 10:
 This program follows a common technique to define fixed points in untyped
 lambda calculus, based on passing a function to itself.
 
-We may not like to define fixed points following the approach above, because
+We may not like to define fixed-points following the approach above, because
 it requires global changes in the body of the function meant to be recursive,
 basically to pass it to itself (`f f` in our case above).  The approach below
-isolates the fixed point aspect of the function in a so-called *fixed point
+isolates the fixed-point aspect of the function in a so-called *fixed-poi1nt
```
cos: typo

k-distribution/tutorial/README.md
```diff
@@ -63,12 +67,14 @@ such as abrupt termination, exceptions, call/cc, concurrency, etc.
...
 here on, is a prototype implementation of the K Framework written in
-Java and Maude.  The K prototype is developed by a joint team of
+Java and Maude.  The K prototype is implemented by a joint team of
 faculty and students from the University of Illinois at
 Urbana-Champaign, USA (the FSL group, led by professor Grigore Rosu),
 the University Alexandru Ioan Cuza, Iasi, Romania (the FMSE group, led
-by professor Dorel Lucanu), the University of Bucharest (professor
-Traian Florin Serbanuta), as well as several individual enthusiasts.
+by professor Dorel Lucanu), and the University of Bucharest, Romania
+(professor Traian Florin Serbanuta), of developers from
```
cos: the "of" sounds strange here because it binds all the way to the top of the sentence. Maybe "a joint academia-industry team of faculty and students from [...], developers from [...], and several individual enthusiasts.". It makes the bindings less explicit but it is easier to read.

grosu:
Thanks, @cos, for the detailed reading.  Good points.  Fixed them.

Issue #1984, opened 2016-01-03
[em] add Location back to kompile output


daejunpark:
https://github.com/kframework/k/pull/1982#discussion_r48689602

daejunpark:
#1945

daejunpark:
fixed via #2042

PR #2032, opened 2016-01-13
[em] configuration abstraction validation

daejunpark:


daejunpark:
@cos please review

daejunpark:
#1945

k-distribution/tests/regression/error-messages/configuration-validity/b.k.out
```diff
@@ -0,0 +1,3 @@
+[Error] Compiler: Attempting to concatenate cells not of multiplicity="*" into
+a cell collection.
+  while executing phase "concretizing configuration" on sentence at
```
cos, line 3: Could it also print which cells could not be concatenated? Also, there's nothing after "at".

cos, line 3: oh, the part after "at" is not part of the assert because of Jenkins testing...

k-distribution/tests/regression/error-messages/configuration-validity/c.k.out
```diff
@@ -0,0 +1,2 @@
+[Error] Compiler: Had 1 parsing errors.
+[Error] Inner Parser: Parse error: unexpected character '<'.
```
cos, line 2: We should make sure this does display the location, in case it is not removed just for the assertion.

k-distribution/tests/regression/error-messages/configuration-validity/d.k.out
```diff
@@ -0,0 +1,2 @@
+[Error] Critical: Can't mix cells at different levels under a rewrite
+  while executing phase "concretizing configuration" on sentence at
```
cos, line 2: Again, we should display which cells are involved.

k-distribution/tests/regression/error-messages/configuration-validity/e.k.out
```diff
@@ -0,0 +1,4 @@
+[Error] Critical: All cells on the left and right of a rewrite must have the
+same parent:
+=>(<a5>(#noDots(),#token("1",Int),#noDots()),<b5>(#noDots(),#token("2",Int),#noDots()))
```
cos, line 3: It is not very nice that we're leaking `#doDots` in the display to the user. Not sure it is easy to fix. Maybe add an issue for us to fix it later?

daejunpark:
Since #1883 was fixed, I'll bring back the location information (line and column numbers only) as mentioned in #1984, in a separate PR.

k-distribution/tests/regression/error-messages/configuration-validity/e.k.out
daejunpark, line 3: @cos do you mean it in this particular output, or in general?

cos, line 3: In general. I don't think users know or should know what `#noDots` is.

daejunpark, line 3: Ok, I'll open an issue to fix it later.

daejunpark, line 3: Done #2033

daejunpark:
@cos please merge if your questions are answered.

k-distribution/tests/regression/error-messages/configuration-validity/e.k.out
dwightguth, line 3: Well first of all, we should be using ToKast  instead of K.toString because the latter uses a fairly unsound textual representation that is different from any syntax we should ever be expecting the user to understand. Although maybe we want K.toString to call ToKast, I don't know...

Second, probably the simplest solution is to use pretty printing and to assign the noDots and Dots klabels the correct concrete syntax when we pretty print. Then instead of what we have now we would have:

```
<a5> 1 </a5> <b5> 2 </b5>
```

Issue #2039, opened 2016-01-14
Caching rules does not work


radumereuta:
**How to reproduce:**  
`kompile -v lambda.k` any definition should be the same though  
**Expected behavior:**  
After running kompile the second time, the number of parsed rules should be `[0/100]`.  
**What happenes:**  
This is what kompile with verbose mode prints:  
`Parse definition [104/108 rules]                             =  2574`   

This is a regression after merging #2036.

dwightguth:
@cos if you could make this a high priority that would be great because while it doesn't break anything per se, it does really really slow down k development if I have to wait an extra 6 minutes every time I build the C semantics.

cos:
@radumereuta, thanks for highlighting it. I'll solve it tonight.

cos:
addressed by #2045

radumereuta:
Caching works right now. Sort of. I found another bug which you can trigger after you move a rule up or down in the file. The warnings are not updated to the new location.
Try 1: `kompile -v -w all test.k`

```
module TEST
  rule A => .K
endmodule
```

prints out:

```
Parse definition [109/109 rules]                             =  5900
[Warning] Compiler: Variable 'A' was not declared. Assuming sort K.
    Source(D:\work\test\.\test.k)
    Location(2,8,2,15)
```

Try 2: `kompile -v -w all test.k`

```
module TEST

  rule A => .K
endmodule
```

prints out:

```
Parse definition [0/109 rules]                               =  3476
[Warning] Compiler: Variable 'A' was not declared. Assuming sort K.
    Source(D:\work\test\.\test.k)
    Location(2,8,2,15)      // this should have been line 3.
```

This used to work in K3.6.

dwightguth:
@radumereuta I think that is less urgent because it probably didn't used to work in my version of the code either. We should probably fix it at some point but I'm not sure it should be part of #2045.

cos:
thanks, @dwightguth, right.
@radumereuta, please open a separate issue for the new problem.

radumereuta:
Opened #2047.

PR #2045, opened 2016-01-15
fix caching issue + refactoring

cos:


radumereuta:
Updated #2039 with the description of another caching related bug. Please fix it before merging.

dwightguth:
It's more urgent in this case that we don't call saveCaches more than once because that causes a significant performance penalty. Can you please go through this and make sure it's only called once on each code path while at the same time ensuring that it's always called after completion of parsing?

dwightguth:
@radumereuta can we merge this if it fixes the issue where rules are not cached at all? I would really like to go back to having second runs of  compiling the C semantics not take 6 minutes.

cos:
@dwightguth, the cache saving is only invoked once now.

radumereuta:
I haven't spent too much time trying to understand the original refactoring, so @dwightguth if you are happy with the PR, go ahead and merge. I tested it and it appears to report the correct number of cached rules.

radumereuta:
Tested again, after the latest fix, and the first run on an empty definition shows [108/108] parsed rules and 3 seconds. The second time [0/180], and 1.9 seconds.
I guess it may be fixed.

PR #2048, opened 2016-01-15
Fix #2000.

### LINKED ISSUE #2000: do priorities work?
I spent 15 minutes trying to make priorities work for lambda, lesson 1:

```
syntax priority lambda_._ > __
```

etc.  Tried lots of variants and nothing worked as I wanted with `krun`.  The default priority is the right one in krun, that is, `lambda x . x y` is displayed as `lambda x . (x y)` in the k cell by krun, but I cannot make it do the other parsing, `(lambda x . x) y`.  I tried putting the labels in backticks and that didn't work either.  I tried reversing > into <, nothing.  It appears that I can even use a wrong label name, say _._, and get no error.  On the other hand, kast reports an ambiguous parsing warning on the same input.  How comes krun silently ignores the parsing warnings that kast reports?  That should not happen.  Anwyay, I threw the towel, I have no idea how to make this work.  What's going on here?  We need at least to report errors when people misspell the KLabels, and to have krun report the warnings/errors reported by kast.

@radumereuta , please fix/elucidate.  Also, once done, please add an explanation in the manual wiki.  If this is indeed a bug and it is hard to fix, it is OK to postpone until after Jan 18, provided it does not interfere with the other tutorial examples.

radumereuta:
Do not review.


radumereuta:
@daejunpark please review. This should fix some of the issues found in kool and simple related to extra parenthesis.

There were several issues in the disambiguation filters, but the main change here was to modify the way we generate klabels automatically. More precisely, we removed the quote `'` in front of the labels.

This should fix at least #2000 and #1935.

PR #2049, opened 2016-01-15
update copyright for all k files

cos:


radumereuta:
lgtm, but I worked with @cos, so someone else needs to review it too.

PR #2050, opened 2016-01-15
5 types all lessons

laurayuwen:


laurayuwen:
@grosu Finally you can review this. I will create a separate PR to update tests after Andrei's PR #2077 and #2073 has been merged into master.

k-distribution/tutorial/1_k/5_types/lesson_1/imp.k
```diff
@@ -81,6 +81,7 @@ module IMP
   rule <k> int (X:Id,Xs:Ids => Xs); ...</k> <tenv> M:Map => M [ X <- int ] </tenv>
   rule int .Ids; => stmt
 
+  context print(HOLE:AExp, AEs:AExps);
```
grosu: I replaced this with

```
context print(HOLE:AExp, _);
```

in 4_imp++.  It should be able to infer the sort of _.  If not, let it like this.

k-distribution/tutorial/1_k/5_types/lesson_2/lambda.k
```diff
@@ -48,5 +47,5 @@ module LAMBDA
                | "mu" Id ":" Type "." Exp                             [binder]
   rule letrec F : T1  X : T2 = E in E'
     => let F : T1 = mu F : T1 . lambda X : T2 . E in E'               [macro]
-  rule mu X : T . E => (T -> T) (E[T/X])
+  rule mu X : T . E => (T -> T) (E[T/X]):>Exp
```
grosu, line 50: Interesting that you need this cast here.  I would expect that the result of applying the substitution is K, which should go everywhere.  No need to worry about this now, just an observation.  I'm going to post an issue about it tomorrow morning when I'll revise the readmes in 5_types.

k-distribution/tutorial/1_k/5_types/lesson_4/lambda.k
```diff
@@ -34,15 +34,14 @@ module LAMBDA
   rule T1:Type  + T2:Type => T1 = int ~> T2 = int ~> int
   rule T1:Type <= T2:Type => T1 = int ~> T2 = int ~> bool
 
-  //TODO(AndreiS): replace with strictness when java unification algorithm
-  //correctly handles sorts
-  rule lambda X . E:Exp => E[?T/X] ~> ?T:Type -> HOLE
-  rule T2:Type ~> T1:Type -> HOLE => T1 -> T2
+  syntax Exp ::= Exp "->" Exp                  [strict]
+  rule lambda X . E => ?T:Type -> E[?T/X]
```
grosu, line 38: Related to my previous comment above, it is even more interesting that you did not need a cast in the rule above.  They seem to be in the same category: a substitution applied to an Exp, which we want to be an Exp in the context.

grosu:
LGTM.  My previous comments not crucial.  Feel free to merge when you are happy with it.  I will take this over tomorrow morning and hope to finish it before 9am or so.

k-distribution/tutorial/1_k/5_types/lesson_4/lambda.k
laurayuwen, line 38: Yes, I also found it strange, only the mu rule had parsing error.

k-distribution/tutorial/1_k/5_types/lesson_1/imp.k
laurayuwen: addressed.

PR #2052, opened 2016-01-15
better freezer names

cos:


kernel/src/main/java/org/kframework/kore/compile/ResolveContexts.java
```diff
@@ -120,7 +120,7 @@ public void apply(KApply k) {
         K cooled = RewriteToTop.toLeft(body);
         // TODO(dwightguth): generate freezers better for pretty-printing purposes
         List<ProductionItem> items = new ArrayList<>();
-        KLabel freezerLabel = getUniqueFreezerLabel(input);
```
dwightguth, line 123: pretty sure cooled is not necessarily guaranteed to be a KApply. It will be in most cases, but you need to fall back to the old behavior if it's not.

cos, line 123: The context should be a HOLE surrounded by something, so I assume it has to have at least a KApply.

andreistefanescu:
lgtm

cos:
@andreistefanescu, cherry-picked the changes from the superstrict branch to get them out of that diff. merge if ok.

andreistefanescu:
lgtm

cos:
@andreistefanescu, please merge if you're ok with it :)

cos:
that was fast

PR #2059, opened 2016-01-17
Logik

laurayuwen:
DO NOT MERGE!


laurayuwen:
@cos for example, krun tests/append-1.logik --depth 2, both step should apply rule 

```
  rule <k> C:Clause Pgm => Pgm </k>
       (.Bag => <clause> #renameVariables(C) </clause>)
```

however variables in <clause> is not renamed, kept its original name as in program.

```
<generatedTop> <T> <k> ?- #token ( append , Literal ) ( #token ( cons , Literal ) ( #token ( a , Literal ) , ( #token ( nil , Literal ) , .Terms ) ) , ( #token ( cons , Literal ) ( #token ( b , Literal ) , ( #token ( nil , Literal ) , .Terms ) ) , ( #token ( V , Term ) , .Terms ) ) ) , .Terms . </k> <fresh> . </fresh> <clauses> <clause> #token ( append , Literal ) ( #token ( cons , Literal ) ( #token ( H , Term ) , ( #token ( T , Term ) , .Terms ) ) , ( #token ( L , Term ) , ( #token ( cons , Literal ) ( #token ( H , Term ) , ( #token ( Z , Term ) , .Terms ) ) , .Terms ) ) ) :- #token ( append , Literal ) ( #token ( T , Term ) , ( #token ( L , Term ) , ( #token ( Z , Term ) , .Terms ) ) ) , .Terms . </clause> <clause> #token ( append , Literal ) ( #token ( nil , Literal ) , ( #token ( L , Term ) , ( #token ( L , Term ) , .Terms ) ) ) . </clause> </clauses> <mgu> . </mgu> </T> .SolutionCell </generatedTop>
```

cos:
@andreistefanescu, you understand this better than me. What is this #renameVariables suppose to do? Give fresh names to all variables in a term?

grosu:
@cos: yes.  You need that in order to make sure that each use of a clause comes with different variables available for unificiation.

cos:
@grosu, I see. It's been some time since I looked at how logik works and didn't want to switch contexts fully.

andreistefanescu:
This problem seems to be that the variables are not identified properly (i.e. `#token ( V , Term )` does not become a `Variable` in the backend). There use to be some transformation doing that, but probably is not implemented in kore.

cos:
those are `#token` labels, not actual KTokens, right?

cos:
hmm..., actually no, because Term cannot be a K...

cos:
so they are actually tokens
@andreistefanescu, what is the rule by which they become variables in the backend? I.e., how do we distinguish "variable" tokens from the others? Shall we look at the sort, and if there is any production for that sort which is not `[token]`, make the Token is a KVariable of that sort?

cos:
If so, it seems a bit flaky...

andreistefanescu:
Tokens with the `[variable]`tag become variables. This is just to circumvent parsing issue.

cos:
Are they represented as regular `Variable`s in the backend?

andreistefanescu:
Yes, of the corresponding declared sort (i.e. Term).

laurayuwen:
So `Token{ }` when it has `[variable]` attribute should be regarded as kVariables? Like when we define `syntax KVariable ::=`?

andreistefanescu:
Yes, except the sort `KVariable` is not visible at definition level, hence the tag.

cos:
@laurayuwen, please have a look if this works as you expect.

It is not the cleanest solution, but the clean solution would involve making the meta-level work, and we don't have time for that until tomorrow.

laurayuwen:
@cos looks good, does rename. Thanks!

cos:
@laurayuwen, please either merge or take the changes from this branch to yours and close.

PR #2060, opened 2016-01-17
Em/fix location info

daejunpark:

PR #2061, opened 2016-01-17
Em/check configuration cells

daejunpark:


daejunpark:
@cos please review

daejunpark:
should be merged after #2060

kore/src/main/scala/org/kframework/definition/outer.scala
```diff
@@ -137,6 +137,12 @@ class Module(val name: String, val imports: Set[Module], unresolvedLocalSentence
     }).fold(Map())(mergeMultiset)
   }
 
+  lazy val cellProductionsFor: Map[Sort, Set[Production]] =
+    productions
+      .collect({ case p if p.att.contains("cell") => p })
+      .groupBy(_.sort)
+      .map { case (s, ps) => (s, ps) }
+
```
cos, line 145: we don't want the KORE classes to "know" about cells. Could this be moved somewhere else?

daejunpark, line 145: Hmm, why is that? Isn't it an outer definition that includes those syntactic sugars beyond the core part of KORE?

cos, line 145: no, both the new outer and the new inner classes are KORE. We call stuff beyond that e-kore and it is now loosely encoded in helper classes, meta-level, and remaining KIL. Some e-kore stuff has sneaked into outer.scala but we generally try to avoid that.

daejunpark, line 145: Ok, I'll fix it, but in separate PR. I'm a bit confused with these series of PRs that are partially merged in out of orders, and I want to clear these PRs first.

PR #2065, opened 2016-01-17
Em/check list decl

daejunpark:


daejunpark:
@cos please review

kernel/src/main/java/org/kframework/kore/convertors/KILtoKORE.java
```diff
@@ -99,6 +100,7 @@ public KILtoKORE(org.kframework.kil.loader.Context context) {
                                                   Map<String, org.kframework.definition.Module> koreModules,
                                                   scala.collection.Seq<Module> visitedModules) {
         checkCircularModuleImports(mainModule, visitedModules);
+        CheckListDecl.check(mainModule);
```
cos, line 103: could this be added to the compile pipeline instead of KILtoKORE?

daejunpark, line 103: I'm not positive. The user lists are attempted to translate into assoc(?) lists during parsing modules, and I think the check should be done before the translation, otherwise the translation may lead to unexpected behaviors (or exceptions) for invalid user list declarations.

cos, line 103: my bad, yes, the check is done of kil classes.

cos:
Jenkins: test this please.

PR #2072, opened 2016-01-17
Em/check sort top uniqueness

daejunpark:


daejunpark:
@cos please review

kernel/src/main/java/org/kframework/kore/compile/checks/CheckSortTopUniqueness.java
```diff
@@ -0,0 +1,51 @@
...
+import org.kframework.definition.SyntaxSort;
+import org.kframework.kore.Sort;
+import org.kframework.utils.errorsystem.KEMException;
+
+import static org.kframework.kore.KORE.Sort;
+
+import java.util.Set;
+
+/**
+ * Checks if multiple top sorts KList and Bag are found for
```
cos: doesn't read well. could you rephrase a bit?

kernel/src/main/java/org/kframework/kore/compile/checks/CheckSortTopUniqueness.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+public class CheckSortTopUniqueness {
+    private final Set<KEMException> errors;
+
+    private final Module module;
+
+    public CheckSortTopUniqueness(Set<KEMException> errors, Module module) {
```
cos, line 32: add documentation that `errors` is a mutable Set which gets populated by this pass.

daejunpark:
@cos please review again

PR #2073, opened 2016-01-17
Fix search pattern

andreistefanescu:


andreistefanescu:
@cos please review

laurayuwen:
First, it introduced an AssertionError when search. I commented that assertion and tested it, we still have the same issue with pattern.

```
java.lang.AssertionError
    at org.kframework.backend.java.symbolic.SymbolicRewriter.addSearchResult(SymbolicRewriter.java:534)
    at org.kframework.backend.java.symbolic.SymbolicRewriter.search(SymbolicRewriter.java:613)
    at org.kframework.backend.java.symbolic.InitializeRewriter$SymbolicRewriterGlue.search(InitializeRewriter.java:156)
    at org.kframework.krun.modes.KRunExecutionMode.execute(KRunExecutionMode.java:54)
    at org.kframework.krun.KRun.run(KRun.java:82)
    at org.kframework.krun.KRunFrontEnd.run(KRunFrontEnd.java:97)
    at org.kframework.main.FrontEnd.main(FrontEnd.java:52)
    at org.kframework.main.Main.runApplication(Main.java:110)
    at org.kframework.main.Main.runApplication(Main.java:100)
    at org.kframework.main.Main.main(Main.java:52)
```

cos:
the diff is empty...

andreistefanescu:
I was hungry and I did not notice.

java-backend/src/main/java/org/kframework/backend/java/symbolic/FastRuleMatcher.java
```diff
@@ -122,6 +126,25 @@ public FastRuleMatcher(GlobalContext global, int ruleCount) {
...
+                    .map(c -> c.simplify(context))
+                    .filter(c -> !c.isFalse())
+                    .map(c -> c.orientSubstitution(pattern.variableSet()))
+                    .filter(c -> c.isMatching(pattern.variableSet()))
+                    .map(c -> c.substitution())
+                    .collect(Collectors.toList());
+        } else {
+            return Collections.emptyList();
+        }
+    }
```
cos, line 149: what is different about this overloaded function?

andreistefanescu, line 149: It only does matching (it does not use rules), and it checks the result is matching (instead of unification).

cos, line 149: may be good to have a different name and a bit of documentation

laurayuwen:
this patch solves the problem for 5_types, lesson6

PR #2077, opened 2016-01-18
Fixed binder

andreistefanescu:


andreistefanescu:
@laurayuwen please see if this fixes the issue with substitution.

laurayuwen:
very good! fixed.

PR #2082, opened 2016-01-18
fix cos's bug

dwightguth:
@cos please review

And next time, when I comment on a diff saying something is going to break things, don't just assume I'm wrong and merge without talking to me.


cos:
@dwightguth, I did answer your comment within 15 min and then waited for a couple of days for any other reply from you.

Sorry for going forward with this, but we are on a deadline.

PR #2089, opened 2016-01-18
mlgmp float to int conversion fix.

chathhorn:
@dwightguth please review.

cf. https://github.com/kframework/c-semantics/issues/211


dwightguth:
@chathhorn when you rebuild your OCAML with this, the C semantics test suite passes as well as Robby's program?

chathhorn:
Yes, this fixes it. Note that I added that assertion in the test_suite.ml and it wasn't passing with the old version. It was clearly a bug.

PR #2099, opened 2016-01-18
bottom sort for lists with identical klabels

cos:


daejunpark:
Issue still there: https://github.com/kframework/k/pull/2113/files#r50074467

@cos Could you check why is it?

cos:
It may be related to #2102. I'll have a look in the morning.

PR #2103, opened 2016-01-19
when => requires + broken static simple

grosu:
@cos, just pushing this.  You can merge.  See the simple-typed-static.k.  Thanks, Grigore.


cos:
ups, didn't finish the check

grosu:
should be fine

cos:
will push a revert if there's an issue

PR #2106, opened 2016-01-19
this is a bit too easy... 

cos:


cos:
@grosu, it seems the problem was quite easy to fix, at least in the version with the explicit isLValue.

grosu:
@cos looks like it also works to replace the rule with the production, which I kind of prefer.  But it is nice that both work!

PR #2120, opened 2016-01-19
release preparation

cos:


dwightguth:
It would be nice to include a K 3-to-4 migration guide in this.

cos:
@radumereuta, could you add a bit of information about the parser here? I can do it otherwise.

radumereuta:
@cos, done, have a look.

cos:
@radumereuta, looks good. only made a small change.

radumereuta:
You are right, it sounds better.

cos:
return-object.kool is flaky... it consistently works on my computer but fails intermittently on codeship. As this PR does not touch code, disregard that error for now.

grosu:
Looks good.  Merging.

Issue #1903, opened 2015-12-12
`Production` vs `SyntaxSort`


cos:
Right now, in KORE, we define new sorts by either a `Production` or a `SyntaxSort`. `SyntaxSort` can only introduce a sort, not define a subsort relation.

This is inconsistent as it is an attempt to separate sorts from productions, but at the same time using `Production` is required to encode sort-only properties like subsorting. The duality ripples into a lot ugly "case"-driven code.

I think we ended up with this situation because subsorting and productions share the same syntax in `kore.k`. 

I propose to either:
1. Eliminate `SyntaxSort` all together and use a `Production` with no production items instead.
2. Add a "supersort" field to `SyntaxSort` such that it will encode all subsort relations. We could also enforce that all sorts descend from the some explicit top sort, i.e., `syntax Foo` becomes `syntax Foo ::= Top`. Not sure if that top sort should be `K` or a newly introduced one. Considering `K` has other implications, we might want to go for `Top`.

@grosu, @radumereuta, @bmmoore, @dwightguth, what do you think?

It's not a huge deal and I won't argue strongly for it but it struck me when skimming through the code in `Module`.

radumereuta:
Here are a few thoughts:
- the kernel of the parser doesn't care if a production is a subsort or a 'regular' production. Everything gets transformed into states for the kernel to interpret. It so happens that the what we call subsort, has only one non-terminal in the entire production. (in the parsing community this is also often called chain production, maybe to differentiate it from other meanings of subsort)
- it is only later on when we try to do sort inference we need such information. Because the subsort production doesn't consume characters, I need to include the subsorts into sort inference also.
- we have two types of productions: with KLabel and w/o KLabel
  `Exp ::= Int [klabel(Int2Exp)]` and `Exp ::= Int`. In this case, the front-end considers this a subsort, because my criteria is 'has only one non-terminal`, but the back-end doesn't, because it produces a node in the AST.
- we want to allow for a method to have annotations for sorts. I think the `SyntaxSort` was originally designed for this case: `syntax Exp [my, tags, here]`.

bmmoore:
The current role of `SyntaxSort` is for declaring sort attributes, or sometimes just to introduce sorts without giving them any content yet.

I'm very much against mixing that back together with subsorting declarations. For just one complication, if there was an optional subsort field, would the attributes of the SyntaxSort switch between describing the parent sort or the subsorting relationship based on whether that subsort field is set?

[...]

PR #2130, opened 2016-01-29
Fix ac bug

cos:


cos:
@radumereuta, could you have a look? @andreistefanescu  and I made some changes to the syntax to avoid a bad subsort, and the program parser doesn't work anymore.

cos:
the program not parsing is `constructors.kool`
with
`/Users/cos/workspaces/k/k-distribution/tutorial/2_languages/2_kool/2_typed/2_static`

k-distribution/tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.k
```diff
@@ -200,11 +205,16 @@ the wrapper in the generated documentation, we associate it an
...
-  rule class C:Id S => class C extends Object S
-
 endmodule
 
 
+module KOOL-TYPED-STATIC-PROGRAM-PARSING
+    imports KOOL-TYPED-STATIC-SYNTAX
+    imports ID-PROGRAM-PARSING
+//    syntax TypeId ::= r"(?<![A-Za-z0-9\\_])[A-Za-z\\_][A-Za-z0-9\\_]*"    [notInRules, token, autoReject, avoid]
+    syntax TypeId ::= "C"
```
grosu: How can this work?

k-distribution/tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.k
```diff
@@ -277,6 +287,8 @@ in the class extends relation. */
 
 /*@ \subsection{Unchanged semantics from statically typed SIMPLE}
 
+  rule class C:TypeId S => class C extends Object S
```
grosu: was it necessary to move this rule here?  and this can safely be a macro.

k-distribution/tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.k
```diff
@@ -364,7 +376,7 @@ typed SIMPLE, so we do not discuss them much here. */
 //@\subsubsection{Unchanged auxiliary operations from dynamically typed SIMPLE}
 
   syntax Stmts ::= mkDecls(Params)  [function]
-  rule mkDecls(T:Type X:Id, Ps:Params) => T X; mkDecls(Ps)
+  rule mkDecls((T:Type X:Id), Ps:Params) => T X; mkDecls(Ps)
```
grosu: the parentheses shouldn't be needed if we had a perfect parser, should them?

k-distribution/tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.k
```diff
@@ -434,11 +446,11 @@ at the end of this module). */
 
   rule <k> T:Type X:Id; => stuck(T X;) ...</k>
        <ctenvT>... X |-> _ ...</ctenvT>
-       <inClass> C:Id </inClass>
+       <inClass> C:TypeId </inClass>
 //       <br/>
        <out>... .List => ListItem("Member \"" +String Id2String(X)
                               +String "\" declared twice in class \""
-                              +String Id2String(C) +String "\"!\n") </out>
+                              +String TypeId2String(C) +String "\"!\n") </out>
```
grosu: Wouldn't it be nicer, easier, and more powerful from here on to start using the pretty-printer and the unparser?  I'm assuming that they would be hooked KLabels of the form

```
syntax String ::= #pretty-printer(K)  [hook(...)]
                | #unparse(K)  [hook(...)]
```

On token K terms, these should be as efficient as anything else.

k-distribution/tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.k
cos: The restriction to "C" is just a temporary hack to make sure ambiguities on `TypeId` are not the problem.

k-distribution/tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.k
cos: We can remove them. The problem was different.

k-distribution/tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.k
cos: Yes, it should work. Still, I wouldn't go into adding any new features before we have a beta release.

cos:
The tests for kool are not passing yet. The green tick above is not right: some reason, `mvn verify -DskipTests` runs ktests but does not mark the `mvn` run as a failure -- I have reverted to not splitting the mvn tasks. 

Regarding kool, if anybody can have a look, that would be great. Otherwise, the earliest I can pick it up is Monday afternoon.

radumereuta:
I don't get it, what failed?
I miss the old, easy to read, Jenkins reports.

cos:
the Jenkins reports were very likely a custom (made by Dwight) post-processing of our ugly output. The real solution would be to remove large swats of the unnecessary stuff ktest outputs, and use the proper stream (i.e., `err`) for printing errors.

Let me have a look in the meantime.

cos:
Yea... the problem I had before pushing was that one of the tests (constructors.kool) was not terminating.
I think what happens is that ktest times out but mvn does not report it correctly.

grosu:
Any changes to the design of the K tools need to be discussed with me as well.  Please do not just assume that whatever was there before and you don't like was wrong and thus needs to be changed to your liking.  If we decide to change it after the discussion then fine, we'll go ahead and change it and document the changes.  But do not make design changes without discussions.

Now that there are two implementations of K around, even more importantly than before we need to document the "K standard", so all implementations agree.  Otherwise we end up in chaos.

grosu:
@radumereuta : tried to navigate my way through the large output searching for "[ERROR]", and if none was found, I assumed it was a timeout.

It also seems to take a lot of time now, 15min or so, while it took only 2-3 min with Jenkins+naigun.  Can we still enable nailgun there?  I know the long term solution is to implement ktest so it does not need nailgun, but we are not there yet.

radumereuta:
I think nailgun is still active, and the biggest issue with the current run times is that we use remote servers with just one execution thread. On my laptop, for example (4 cores, 8 threads), 'mvn verify` takes 5'50". On the previous RV server used to take 5'25" (aprox).

Since we are discussing new designs, I want to invoke `ktest`. Something needs to change here because we had so many issues with this part of the framework. I've been discussing this with @cos before, and we (back me up here @cos) think we would be better off with something off the shelf. Even if we have to write more to set up tests, we would have more flexibility, better ways of testing, better ways of integrating with other tools.

cos:
@grosu, @radumereuta, this is about a test that does not terminate. It is not about Codeship being slow. Codeship is actually only slightly slower than the pervious Jenkins implementation. Assuming the individual cores on the RV machine are quite powerful, this would not be possible unless Codeship also makes multiple cores available. Furthermore, I watched the output and it seems to be scrambled in the way multi-threaded execution would make it. Bottom line, Codeship's performance is not the problem; a test not terminating is.

grosu:
We are not discussing new designs of K.

ktest is a tool that is part of the kframework, which is useful for people who write K definitions, who do exercises, projects, etc.  It was not designed with developers of the K framework in mind.  You can go ahead and write unit tests as a developer.  Anyway, ktest needs to accept incremental attributes, as discussed in several issues.  And so far nobody proposed a better solution, showing that he UNDERSTANDS ktest and what needs to do.  Until that happens, ktest stays.

cos:
@grosu, regarding not changing features, we are not adding or removing features with this PR. What it does do it adjust kool to avoid what might be seen as a rather ugly of the `klabel` attribute:

https://github.com/kframework/k/blob/a9c2c9ba802cd9733898b296fc242e0d074d6bbd/k-distribution/tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.k#L113

It seems the `klabel` trick here is a shortcut to get the class name injected to a type. @andreistefanescu  and I have discussed this and it seems a clean solution should create the injection semantically (heating/cooling etc), not through parsing. Another decent solution is to have a special production for the class name as type from the beginning, not through attribute-based injection. That is what the new `TypeId` achieves. We went for the latter solution because it seemed easier than figuring out the semantics of the former. At least for me as I'm not familiar with all the details of kool.

But we didn't make the change for stylistic reasons. It was actually wrong before. That trick also has a side effect: it makes `Id` a subtype of `Type`. As `Type` is a `KResult`, it inhibits the heating of `Id`-based lookups. Actually, I don't know how it even worked before... maybe it was relying on a bug or some convention.

The best thing to do would probably be to remove our `TypeId` and actually give the semantics cleanly (and without using the `klabel` attribute). As this is part of the tutorial, you probably know best how to approach this.

grosu:
When I said we are not discussing new designs of K here I meant it as a reply to @radumereuta's comment:

> Since we are discussing new designs, I want to invoke ktest. Something needs to change here ...

Now that you mentioned the issue with `Id` vs `TypeId`, it surprises me that @andreistefanescu didn't detail you the issue with the subsort injections.

We have oscillated several times between adding automatic injection klabels for all subsorts, vs not adding them.  There are advantages and disadvantages either way.  In the end, what was decided as a group, with @andreistefanescu there part of the decision committee, was to allow the users to decide themselves whether to include an injection klabel or not. Specifically, if the subsort production has no klabel attribute, then there is no injection; otherwise, the declared klabel is the injection.

So if we want to continue to have both subsorts with explicit injections and subsorts without any injection klabel, then the above is as good as it can be.  If you think that we should pick one of the two and stick to it, then we are having a different discussion, which I am glad to have (but be prepared, it will open a huge can of worms).  If we want to continue to support both, then I think the previous definition was OK, that is, not a parsing trick.  It depends on how you interpret its semantics.  The way I think it should be interpreted is that the injection klabel `class` should _always_ be there, so `Id` is _not_ a subsort of `Type` anymore, so you don't need to add the `TypeId` production.  All is needed is to fix the membership predicates to special-case when subsorts have injection klabels.  But I don't fight this.  If you and @andreistefanescu discussed this and agreed that an explicitly injected subsort should continue to be a normal subsort in spite of never appearing as a normal subsort anymore, then it is fine with me.  But in that case, since it is borderline counter-intuitive, can you please add a sentence or two in the manual, so we don't forget to discuss it?

grosu:
> The best thing to do would probably be to remove our TypeId and actually give the semantics
> cleanly (and without using the klabel attribute).

I don't know how to do that cleanly, which is the reason things were as they were.  Once I declare `Id` a subsort of `Type` without an explicit injection, which is what you need when you parse programs, I already have the problem you described (`Id` becomes a `KResult`).  So I need to do that in the special module used for parsing programs, where I also need to rewrite any such Id-as-type into something else (an injection?), which should be defined in an imported syntax that is also imported in the semantics.  Seems difficult.  Why don't you show me?

cos:
Regarding the injection klabels, I see what you mean. Still, it seems like a subtle issue that would be hard to explain to users. My opinion is that we should avoid it altogether if at all possible.

I'm not that familiar with kool typed, but the way I see this, we could simply allow the class name to be an Id, but not make the Id a type. Then, when processing the class (I think [here](https://github.com/kframework/k/blob/a9c2c9ba802cd9733898b296fc242e0d074d6bbd/k-distribution/tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.k#L546-L558)), we register it like this:

```
  rule <task> <k> class C:Id extends C':Id { S:Stmts } => stmt ...</k> </task>
       (.Bag => <class>...
               <className> class(C) </className>
               <extends> class(C') </extends>
             ...</class>)
//       <br/>
       (.Bag => <task>
                <k> checkType(class(C')) ~> S </k>
                <inClass> class(C) </inClass>
                <ctenvT> .Map </ctenvT>
             </task>)
    [structural]
```

Essentially, the name of the class is `Id` in the syntax, but `class(C)` of sort `Type` when it goes into the semantics. Am I missing something?

grosu:
How about program declarations of the form

```
    int x;
    A y;   // A is a class name
```

You need to parse both `int` and `A` as a type, same like in Java.  So you need them both to be subsorts of `Type` to parse programs.  And then you have all kinds of rules like

https://github.com/kframework/k/blob/a9c2c9ba802cd9733898b296fc242e0d074d6bbd/k-distribution/tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.k#L408-L409

which you want to work the same way, no matter whether the type is `int` or `A`.

cos:
I see three possible solutions:
1.  continue with the idea I started above: have all of them as `Id`s and only consider them `Type`s when interpreting them semantically. 
2. the solution we we went with in this PR: have a `TypeId` which is distinct from the regular `Id`.
3. do `Type ::= Id [token]` -- this would achieve the same purpose as the original definition, but the subsort erasure is more clear. I would be more inclined to read this as "parse whatever you like there, but the end result will be a `Type`". Still, this last solution still feels a bit like cheating.

grosu:
I don't see how 1 can work.  As I said, in the syntax that you use for parsing programs, you need to write `Id`s in programs where any type can occur.  So you will need some `TypePgm` sort for programs which has exactly the syntax we have now for `Type` but with `Id` a subsort of it.  Then, in the semantics, you want a different `Type`, which does not subsort `Id` but injects it.  You can do that with a translation from `TypePgm` to `Type`, but that is verbose and ugly.

I can live with 3 if you make it work, although `Id2String` will likely not work anymore and not clear how you can match on an `Id` type in rules anymore using concrete syntax.  Also, I don't see 3 as superior to or cleaner than the existing semantics of injected klabels; they both erase the original meaning of the subsort.  And to me as a teacher, your 3 is harder to explain.

Your 2 is the same as the original approach, except that you duplicate code and complicate the overall definition and the teacher's job, who now has to explain the -PARSING-PROGRAMS module name convention (which will change anyway in the near future), all because for some reason you prefer to keep the subsort relation between `Id` and `Type` in spite of the explicit injection.  In fact, I think that it is a bad idea for `isType(Id)` to be true regardless, because it may hide errors from users; for example, if they work with the meta-level representation and forget to use the explicit injection klabel.

> I see three possible solutions:

There is fourth one: keep the existing meaning.  All you have to do is to _fix_ the subsort membership: when `Supersort ::= Subsort [klabel(injection)]`, exclude the case `rule isSupersort(X) => true requires isSubsort(X)`.  The resulting membership predicate will be faster and safer for users, because it will not work when they forget to add the klabel injection.  As a bonus, eliminate also all this code 

```
module KOOL-TYPED-STATIC-SYNTAX-PROGRAM-PARSING
    imports KOOL-TYPED-STATIC-SYNTAX
    imports ID-PROGRAM-PARSING
    syntax TypeId ::= r"(?<![A-Za-z0-9\\_])[A-Za-z\\_][A-Za-z0-9\\_]*"    [token, autoReject]
endmodule
...
  syntax String ::= TypeId2String ( TypeId )    [function, hook(STRING.token2string)]
  syntax Id ::= TypeId2Id(TypeId)                                   [function]
  rule TypeId2Id(X:TypeId) => String2Id(TypeId2String(X))
```

which makes the instructor's life harder in the class, having to go into implementation details of K which will change soon.

So you have only benefits keeping things as they were.

cos:
> I don't see how 1 can work. As I said, in the syntax that you use for parsing programs, you need to write Ids in programs where any type can occur. So you will need some `TypePgm` sort for programs which has exactly the syntax we have now for `Type` but with Id a subsort of it. Then, in the semantics, you want a different Type, which does not subsort Id but injects it. You can do that with a translation from TypePgm to Type, but that is verbose and ugly.

I don't think I made myself understood. Solution 1 is the following: 
- all kool language syntax uses `Id` instead of `Type` -- no reference to `Type` whatsoever
- all semantic stuff (like the interpretation of `class_extends__` above) injects the right `Id`s to `Type` when encountering them. 

Yes, more verbose, but also very clear.

> I can live with 3 if you make it work, although Id2String will likely not work anymore and not clear how you can match on an Id type in rules anymore using concrete syntax. Also, I don't see 3 as superior to or cleaner than the existing semantics of injected klabels; they both erase the original meaning of the subsort. And to me as a teacher, your 3 is harder to explain.

I'm not arguing 3 is particularly good. Just an alternative that seems cleaner to me. `Id2String` is not a problem -- we can either hook the `Type` or use `#unparse`, etc.

> Your 2 is the same as the original approach, except that you duplicate code and complicate the overall definition and the teacher's job, who now has to explain the -PARSING-PROGRAMS module name convention (which will change anyway in the near future),

This solution is already implemented in this PR, but still buggy (not sure if because of this). And it is nothing like the original approach. Yes, it does use `-PARSING-PROGRAMS`, but so does the default solution of any definition -- we're just hiding it. Aside from that, considering that solution 1 still requires the injection, I'm starting to think it actually is the cleanest solution.

> all because for some reason you prefer to keep the subsort relation between Id and Type in spite of the explicit injection. 

I've been working with K for a few years now and this still tricked me. And it tricked Andrei -- or at least he didn't consider it ok. And it tricked Radu -- he said it looks like a bug. Maybe this would trick users too.

> In fact, I think that it is a bad idea for isType(Id) to be true regardless, because it may hide errors from users; for example, if they work with the meta-level representation and forget to use the explicit injection klabel.

Exactly, I fully agree. `Id` should have nothing to do with `Type`. And the code should make this crystal-clear as well. Having a "side" attribute have such a dramatic effect on syntax and semantics is misleading (at least it mislead me).

cos:
(ups, sent the above message to early -- editing it now)

cos:
> which makes the instructor's life harder in the class, having to go into implementation details of K which will change soon.

I don't think it's healthy to keep pretending that we're using the same syntax for both parsing programs and the semantics. It's not true, it will not be true for realistic cases in the future, and we make many other things ugly just to maintain this fake appearance. We could argue whether we want naming conventions for `-PROGRAM-PARSING`, explicit hooks, or something completely different. But it is there and, realistically, it will continue to be there.

I understand we can hide it for simple cases, but I hope we can try to avoid making important choices regarding our definitions (and, worse, about the features we add, or still keep, in K) guided by the need to hide it.

grosu:
> I don't think I made myself understood. Solution 1 is the following:

You made yourself very clear, but your solution does not work.  It requires, as I said, two different `Type` sorts, one for the syntax and one for the semantics.  This is the syntax of types:

https://github.com/kframework/k/blob/fix/ac-bug/k-distribution/tutorial/2_languages/2_kool/2_typed/2_static/kool-typed-static.k#L110-L115

A type is a term of that syntax, not an `Id`.  A term of that syntax can occur anywhere a type can occur  in a KOOL program.  How can such a term be an `Id`?

> This solution is already implemented in this PR, but still buggy (not sure if because of this).
> And it is nothing like the original approach. 

It is like the original approach in that it still uses an explicit injection added at parse time, and then it 
refers to the injection in the semantics.  And whatever you added to the `PARSING-PROGRAMS` is pretty much the same as what the original solution added, but with `Id` instead of `TypeId`.  So it is exactly the same, except that you preferred to duplicate the code and name it `TypeId` instead of `Id` to eliminate the subsorting of `Id` to `Type`.  So you eliminated the subsorting by duplicating it.  Ugly solution in my view.

And additionally, you have to expose the readers to all that adhoc convention to call the parsing modules in a certain way.  That will change soon and whatever we'll end up with will be explained in later tutorials, not here.  It is too early for that here.

> It's not true, it will not be true for realistic cases in the future, and we make many other
> things ugly just to maintain this fake appearance.

The timing is not right for explaining that, as I said.  And having to do all what is being done in this current PR is tons uglier than what was there before, again, only to maintain a _bug_ in the current implementation, that an injected subsort is a normal subsort.  It is not.

cos:
> A type is a term of that syntax, not an Id. A term of that syntax can occur anywhere a type can occur in a KOOL program. How can such a term be an Id?

By simply not having `Type` at all in the kool program. The program would only have `Id` in all places a type should be. If `_->_` can appear in programs (I don't know kool well enough, can it?) then, yes, we need a `PgmType` to cover this. And, at least to me, that would convey better the actual meaning.

> It is like the original approach in that it still uses an explicit injection added at parse time, and then it 
> refers to the injection in the semantics. 

I had another look over the implementation and, yes, we didn't take it all the way. We should have removed the `klabel` on `Type ::= TypeId [klabel(class)]` and adapted the rest. The injection is not needed anymore -- it's encoded by the sort.

>  And having to do all what is being done in this current PR is tons uglier than what was there before, again, only to maintain a bug in the current implementation, that **an injected subsort is a normal subsort. It is not.**

My high level point is that subtleties like this make K hard to use, even for presumably experts like us. @andreistefanescu and @radumereuta, I'm giving up here. If you have anything to add, please do so.

grosu:
What is particularly hurting us is the lack of a manual that explains all these things well.  The injected subsorts is probably the trickiest feature of K, particularly how they interact with the `List{...}` construct.  This is for expert users and indeed, KOOL typed is the trickiest language we have in the tutorial, combining OO with functional features.  Take a look at this program, to see why you cannot just have Ids instead of types:

https://github.com/kframework/k/blob/fix/ac-bug/k-distribution/tutorial/2_languages/2_kool/2_typed/programs/function-types.kool

The convention with `-PROGRAM-PARSING` is not there to stay, it was a compromise that we agreed upon in order to move on without having to add new features.  This will need to be discussed, but the idea is to eventually pass `$PGM` as a string to the configuration, and then one can/should explicitly use whatever module they want to parse it: `#parse(Module, $PGM, Sort)`.  Once that is nailed down and implemented, then we will be ready to explain the sugar that we are using now.  It would be a strategic mistake to jump into explaining adoc current implementation details that will soon change anyway.

To make this simple, let's take it step by step:
1. Do you agree that we want to allow both injected and normal subsorts, the difference being the latter have no klabel in the kasts, while the former have the user-defined klabel?
2. If the answer above is YES, do you agree that injected subsorts should _not_ be considered as normal subsorts in the membership predicates?  If for nothing else, at least for preventing users from having nasty undetected errors when they use the meta-level and forget to add the injection klabel?

If the answers to the above is YES, then we can keep the definition as is and thus delay making public to the users the precise decision on module naming conventions, which we don't have yet nailed down anyway, and buy ourselves time to do it after the release.

radumereuta:
I'm not adding new opinions, I want to clarify how K3.6 works.
We had implemented what @grosu mentioned in 2. in the last comment, namely, we had two types of subsorts: one for the parser, and one for the backend, depending on whether the subsort production had a klabel or not.

The parser will return ambiguities regardless of the subsort productions having a klabel or not, so this is my reasoning for considering injections into subsorts.
I was also fine with having the second view on subsorts (as predicates). But maybe one of the names has to change, since it's creating so much confusion (and I agree it's confusing, since it took me so long to remember why it works in K3.6).

grosu:
> we had two types of subsorts: one for the parser, and one for the backend, depending
> on whether the subsort production had a klabel or not.

Remember that I was a strong supporter for having only injected subsorts, and then have the kompile pipeline and the backends take care of adding the injection klabels where they were needed.  That is also what happens in functional languages when you deal with "subsorts".  However, I was ruled out.  I specifically remember @andreistefanescu and @bmmoore putting all sorts of (valid) arguments on the table why we don't want to do that, at least not now.  So I gave up and we ended up with allowing the user to decide whether a subsort is injecting or not.  And that being the case, then I agree that the current approach, to have a klabel attribute associated to the subsort production, which was proposed by @radumereuta from what I remember, is the best way to go.

> But maybe one of the names has to change, since it's creating so much confusion

What name are you talking about that should change?

> and I agree it's confusing, since it took me so long to remember why it works in K3.6

I remembered this quickly because I suffered because of it when I did kool typed static.  Quite a lot, in fact.  We lack a manual explaining these most advanced features of K, which is a major problem that we have to fix asap.  The tutorial is the only source of information right now, and in this case, the paragraph in kool typed static right before the syntax module describing this thing is old, since before we added the feature in K3.6.  So the only witness for how this should work is K3.6 right now.

radumereuta:
I was talking about the names of the subsort objects in the implementation: `isSubsorted` and `isSyntacticSubsorted`.

Are you proposing a solution? Or should we reproduce what we have in K3.6?

grosu:
Oh, I see what you meant by a different name.

I thought that this would be handled by the `-PROGRAM-PARSING` module approach that we have now.  That is, the `-PROGRAM-PARSING` to include both the subsort and the injection, while the normal semantics module to only include the injection.  That way, you'd need only one membership predicate, wouldn't we?

cos:
> Do you agree that we want to allow both injected and normal subsorts, the difference being the latter have no klabel in the kasts, while the former have the user-defined klabel?

My preference would be to figure out a way around this, i.e., not have both. At least in this case, there are several alternatives.

If we do really want to have them, then I think we should bite the bullet and have special K outer syntax (in KORE) that differentiates them. Even in the way we refer to them, we've started to blur the lines: the former is definitely not a subsort. I think we should make it crystal clear this is not just some side annotation, but a fundamental feature of K -- it is not something we can desugar easily to KORE. I see several such syntactic alternatives:
1. have a completely separate syntax for subsorting, e.g. (`subsort A < B`). `syntax A ::= B` would from now on not generate a subsort, but create a (possibly new) klabel.
2. Make the `klabel` version have a clear special syntax, e.g., `syntax [foo] A ::= B`. This would lift the `klabel` out of the attributes, which you might not like.

(For the record, I think we should also have special syntax for the `token` attribute.)

> If the answer above is YES, do you agree that injected subsorts should not be considered as normal subsorts in the membership predicates? If for nothing else, at least for preventing users from having nasty undetected errors when they use the meta-level and forget to add the injection klabel?

If we really need to do this, then yes, I agree with this. And again, I think treating and referring to them as "subsorts" in any way is at the root of the problem.

cos:
Regarding the rest of the discussion, I agree with @grosu that we shouldn't need a "syntactic subsort". If I understand this correctly, conceptually, the "syntactic subsort" is nothing more than a subsort in a version of the definition early in the parsing pipeline, before, conceptually, we apply rules which flatten some of the structure. We do not need the "syntactic subsort" later in the compiler pipeline, and I really dislike that it sneaked into the KORE classes.

cos:
### Proposal A

I gave this a bit more thought and, actually, having both behaviors is consistent with a change to KORE that I've been advocating for a long time (the latest iteration is #1903), but there is always someone that doesn't like something about it for some reason.

It would be great if we could settle this!

Right now, in KORE, we have:

```
SyntaxSort(s: Sort, att: Att) // declares a sort
Production(s: Sort, items: List[ProductionItem], att: Att) // for both productions and subsorting
```

When the `Production` is not a subsort, it is also required to have a `klabel` attribute. Maintaining the klabel as an attribute is ugly as it is not enforced statically and structurally. It's fundamental functionality hidden in attributes.

My proposal is to have, instead:

```
SyntaxSort(s: Sort, superSort: Sort, att: Att) // both declares a sort and gives it an upper bound
Production(klabel: KLabel, s: Sort, items: List[ProductionItem], att: Att) // strictly for productions
```

The previous `SyntaxSort` will become `SyntaxSort(s, K)`. 

If we had the above structure, the previous bug regarding "injected subsorts" would have been very likely avoided.

Let's call the above proposal "A".

### Proposal B

I would also lift the "token" attribute to the level of KORE. Again, in order to avoid having to special case analyses that look at productions based on a side attribute -- if a production has the `token` attribute, it should not be considered that it can have children. We don't special case for this now, which is very likely a bug.

There a number of ways in which we can do this. Here are a couple but there might be other better solutions:
1. Add a new `TokenProduction`
2. Have a special `ProductionItem` denoting a token -- this essentially extends the current functionality and will likely require some new outer syntax

radumereuta:
I have a few problems with both your proposals.
For **A**, what about `bracket`? It's a production that doesn't have a klabel. Also @bmmoore had some valid points at the end of #1903, which I don't find in your solution.

For **B**, have a look at how `Bubble` is defined in [e-kore.k](https://github.com/kframework/k/blob/master/k-distribution/include/builtin/e-kore.k#L6)
`Bubble ::= Bubble BubbleItem [token]`. I want to make a single token out of the entire bubble, but I want to look at individual `BubbleItems` and make sure they are not keywords. So saying that you are not allowed to have children under a token production seriously limits the parsing capabilities (from tokens that can be context free, to only regex).

From a parsing point of view, the current implementation actually makes a lot of sense. We use only productions to define syntax, which is consistent with BNF (with the exception of priorities and associativity which are often found in eBNF). Adding semantic influences to the syntax, will only move us further away from the simplicity and clarity of defining the syntax.

P.S. I agree that we have some semantic issues with interpreting the syntax, but we also have to be careful not to break something that is well thought of and works so nicely.

cos:
> For A, what about `bracket`? It's a production that doesn't have a klabel. 

From a syntactic perspective (or concrete syntax tree, if you like), it does have a klabel. From a semantic perspective, you can omit the production all-together. 

> Also @bmmoore had some valid points at the end of #1903, which I don't find in your solution.

@radumereuta, @bmmoore's main concern was attributes merging. But that is a more general problem that we have to solve anyway. Keeping the extra fluff of having a special declaration that "introduces" a sort is only a patch solution for a problem we have for all other productions, for rules, contexts, ... essentially everything.

> For **B**, have a look at how `Bubble` is defined in [e-kore.k](https://github.com/kframework/k/blob/master/k-distribution/include/builtin/e-kore.k#L6)
> `Bubble ::= Bubble BubbleItem [token]`. I want to make a single token out of the entire bubble, but I want to look at individual `BubbleItems` and make sure they are not keywords. So saying that you are not allowed to have children under a token production seriously limits the parsing capabilities (from tokens that can be context free, to only regex).

You could still do that. The only difference is that there would be a clear different at KORE level between the two. Nobody would be able to simply traverse productions and assume that their children are "regular". They will encounter a different type and they will have to make a choice of how they handle it.

That being said, I am not against other solutions. I just want to make this crystal clear at KORE level and not leave its interpretation, again, dependent on side attributes. One other solution (that @grosu might have proposed the high-level idea in the past, not sure) would be to translate the token attribute (or special syntax) to KORE level, maybe like this:

```
syntax A ::= B C [token]
```

goes to

```
syntax A ::= B C // moved to the parsing modules
rule bcLabel(b,c) => #down(#ktoken(#unparse(bcLabel(bc)), #Sort("A")))
```

> From a parsing point of view, the current implementation actually makes a lot of sense. We use only productions to define syntax, which is consistent with BNF (with the exception of priorities and associativity which are often found in eBNF). Adding semantic influences to the syntax, will only move us further away from the simplicity and clarity of defining the syntax.
> P.S. I agree that we have some semantic issues with interpreting the syntax, but we also have to be careful not to break something that is well thought of and works so nicely.

I agree that we should keep things as simple and as close to standard notation as possible. But we're not gaining anything by hiding the semantics of the K framework in side attributes. We are only fooling ourselves and others by saying that K is as simple as plain BNF.

grosu:
I think we should discuss this in person, @cos.  We are mixing too many things in this discussion, and you want to change more than is needed for the problem that we ran into.  Even if we decide to drop the injections all together, or to have injections all the time, we still have to do something special for this definition.  Similarly, even if we decide to keep them both around and have a different syntax for them, we will still have to do something special for this definition (e.g., include the subsort in the module used for parsing programs, and the injection in the semantics).  Or alternatively we have to leave with the current PR duplication of code and exposure of the adhoc convention with module `-PARSING-PROGRAMS`, which I think is the worst scenario.  Even if we decide to have different syntax for particular attributes that seem more important than others, like `token` or `klabel` for some productions, we will still have to do something special for this definition; besides, then we will have other attributes that will yell at us for similar special treatment (strict, left, assoc, etc., as well as combinations of them---e.g, how about a subsort which is a token? etc., etc.).

So why complicate ourselves _now_ adding _new_ problems on our plate, when we already have a simple solution to the problem at hand, without having to add any special membership predicates?

Let's meet and discuss tomorrow.  @andreistefanescu, can you also join?  10am?

grosu:
@kframework-bot , who are you?

cos:
it's Cosmin. I replied from the wrong account. Sorry for the confusion.

I created @kframework-bot to limit codeship's permissions to just the k repository.

cos:
Reverted `TypeId` and fixed the injection subsorting issue, but we end up with exactly the same non-termination situation. I tried to debug it a bit and it seems to be creating en ever-increasing sequence of `_._`:

```
<generatedTop>(<T>(<tasks>(_TaskCellBag_(<task>(<k>(#KSequence(_._(this(.KList),, _._(this(.KList),, _._(this(.KList),, _._(this(.KList),, _._(this(.KList),, _._(this(.KList),, _._(this(.KList),, _._(this(.KList),, ...
```

Before going into debugging this step by step, @grosu, could you have a look if this can be, in any way, a bug in the definition itself? I see some annotations like "OLD approach:" with commented code that seem related to this, so I'm wondering if we did anything different now that could cause this infinite loop.

grosu:
Which example creates that non-terminating sequence?  I just tried it on `field.kool` and got the following SMT-related (?) error:

```
D:\github\kframework\k\k-distribution\tutorial\2_languages\2_kool\2_typed\2_static>krun ..\programs\field.kool
(error "line 2 column 30: unknown constant null")
<generatedTop> <T> <tasks> <task> <k> stuck ( class ( Object ) . ( class ( Main ) . Main ) ) ~> #freezer_(_)1 ( .Exps ) ~> discard ~> class ( Main ) ~> #freezer_;0 ( ) </k> <tenv> .Map </tenv> .CtenvTCell <return> void </return> <inClass> Main </inClass> </task> <task> <k> stuck ( class ( Object ) . ( class ( Main ) . a ) ) ~> #freezer_,_1 ( string , .Exps ) ~> #freezer_,_0 ( "a = " ) ~> #freezerprint(_);0 ( ) </k> <tenv> .Map </tenv> .CtenvTCell <return> void </return> <inClass> Main </inClass> </task> </tasks> <classes> <class> <className> Main </className> <extends> Object </extends> <extendsAll> SetItem ( Object ) </extendsAll> <ctenv> Main |-> ( ( void , .Exps ) -> void ) a |-> int </ctenv> </class> </classes> </T> <out> ListItem ( #ostream ( #write ( 1 , "" +String ( "Member \"" +String Id2String ( class ( Main ) . a ) +String "\" not declared! (see class \"" +String "Main" +String "\")\n" ) ) ~> 1 ) ) ListItem ( "on" ) ListItem ( #buffer ( "" ) ) ListItem ( "Member \"" +String Id2String ( class ( Main ) . Main ) +String "\" not declared! (see class \"" +String "Main" +String "\")\n" ) </out> </generatedTop>
[Warning] Critical: failed to translate smtlib expression:
(declare-fun _340 () Bool)
(assert (and (= _340 true) (= null true)))
```

cos:
I'm getting non-termination locally for `constructors.kool` and codeship gets stuck at `tree-sum.kool`.

@andreistefanescu would know more about the SMT error message.

grosu:
See also #2135, I think the SMT error message is related to that.

grosu:
I looked over the kool typed static def and everything looks alright to me.  I have meetings today, so I'm wondering if @daejunpark can look deeper into the definition to see if there is any problem there.

daejunpark:
@grosu  I found the problem #2138 for the non-terminating issue, but have no idea how to fix it. @cos @andreistefanescu Could you take a look?

cos:
we made some changes to the definition. @grosu, have a look.

cos:
also, @andreistefanescu and I worked together on the last few issues, so he already reviewed those changes

grosu:
lgtm.  thanks, guys!

PR #2131, opened 2016-01-29
codeship

cos:


k-distribution/tutorial/2_languages/2_kool/2_typed/1_dynamic/tests/config.xml
```diff
@@ -5,7 +5,7 @@
   <test definition="kool-typed-dynamic.k"
         programs="../../programs/"
         extension="kool"
-        exclude="cycle"
+        exclude="cycle cast-2"
```
grosu, line 8: Why exclude things instead of using search and transitions on them?  In general, we should not start crippling the tutorial examples without prior discussions, because it took us a lot of time to come up with it.

cos, line 8: we'll add the test back when we address the flakiness -- tracked by #2133

PR #2140, opened 2016-02-08
activate logik

cos:


cos:
@laurayuwen, it seems logik is getting closer. could you have a look?

cos:
#2143 tracks the flaky ignored kool test 

#2144 tracks the ugly logik outputs

cos:
@grosu, please review and merge if ok -- even if not everything is passing yet, it helps to have a clear picture of what works and what doesn't. 

@andreistefanescu and I will try to fix the remaining tests today. Regardless if we succeed, I think we should still release today -- we have an issue tracking this (#2145).

PR #2151, opened 2016-02-22
avoid concretization bug

cos:


daejunpark:
@cos I found the problem, but not sure what will be the best solution.

The thing is that the following two modules of `domains.k` are required for `ResolveIOStreams` path, but are not explicitly imported by any other modules.

```
STDIN-STREAM
STDOUT-STREAM
```

Indeed, these modules are a kind of templates, and will be instantiated according to the user-provided configuration. For example, `STDIN-STREAM` contains `<stdin>` cell, which will be replaced with the actual cell name that a user provided for the input cell, e.g., `<in>`.

A quick solution will be to make these two modules be touched by `DefinitionTransformer` in the beginning, so that `moduleTransformer.memoization` can always have them. But, I'm not sure if it is the best way to go. Any better idea?

Issue #1959, opened 2015-12-29
Flaky tests


cos:
The `ktest`s are flaky, i.e., they fail non-deterministically -- true failures, not printing issues. For now, the problem is not that bad and in most cases a retry via "Jenkins: test this please" solves the problem. Still, as we add more of the tutorial, the problem will likely get worse.

Let's use this issue to mark the tests that we find flaky so that someone can inspect them at some point. I forgot to note down the tests as I encountered them but will do so from now on. Hopefully the problems are in the definitions and can be easily identified via `search`.

If you feel like looking into this problem, feel free to pick up this issue.

notifying active developers: @andreistefanescu, @daejunpark, @laurayuwen, @bmmoore, @dwightguth

List of non-deterministic tests (edit this comment to add more): 
- http://office.runtimeverification.com:8080/job/k-framework-pull-request/423/ 
  `
  /var/lib/jenkins/jobs/k-framework-pull-request/workspace/k-distribution/tutorial/1_k/3_lambda++/lesson_5.callcc-not-jump
  `
- http://office.runtimeverification.com:8080/job/k-framework-pull-request/448/
  `
  /var/lib/jenkins/jobs/k-framework-pull-request/workspace/k-distribution/tutorial/1_k/3_lambda++/lesson_6.callcc-env1
  `
- http://office.runtimeverification.com:8080/job/k-framework-pull-request/497/
  `
  /var/lib/jenkins/jobs/k-framework-pull-request/workspace/k-distribution/tutorial/1_k/3_lambda++/lesson_5.callcc-with-let
  `
- http://office.runtimeverification.com:8080/job/k-framework-pull-request/521/
  `
  /var/lib/jenkins/jobs/k-framework-pull-request/workspace/k-distribution/tutorial/1_k/3_lambda++/lesson_5.fibbo
  `
- http://office.runtimeverification.com:8080/job/k-framework-pull-request/531/
  `
  /var/lib/jenkins/jobs/k-framework-pull-request/workspace/k-distribution/tutorial/1_k/3_lambda++/lesson_5.callcc-env2
  `

grosu:
Two things that I can think about:

1) We should run RV-Predict on K following the instructions at https://runtimeverification.com/blog/?p=88, to see at least if there are data races in RV-Predict.  @dwightguth , @cos , @yilongli : would it make sense to add a jenkins job to do that with each PR?

[...]

PR #2163, opened 2016-04-07
Fixing start symbol for command line arguments.

radumereuta:
@cos or @daejunpark please review.

### DIFF
--- a/kernel/src/main/java/org/kframework/krun/KRun.java
+++ b/kernel/src/main/java/org/kframework/krun/KRun.java
@@ -296,6 +296,7 @@ public class KRun {
             String parser = entry.getValue().getRight();
             Sort sort = compiledDef.configurationVariableDefaultSorts.get("$" + name);
             assert sort != null: "Could not find configuration variable: $" + name;
+            
             K configVar = externalParse(parser, value, sort, Source.apply("<command line: -c" + name + ">"), compiledDef);
             output.put(KToken("$" + name, Sorts.KConfigVar()), configVar);
         }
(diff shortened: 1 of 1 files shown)


daejunpark:
@radumereuta Does this fix the IO problem?

radumereuta:
No, it's unrelated.
But I think I'm hitting flaky tests, because I tested the failing tests on my computer and they worked.
The second time I run the tests, different examples failed.

cos:
@radumereuta, comment out the flaky tests and document them here: #1959

radumereuta:
I don't think the tests are flaky. I think there is something in krun which fails from time to time.
Over time I noticed a wide variety of tests that fail, and all of them at krun.

cos:
Yes, there are several tests failing, but generally in lambda/lambda++. I think there is some unwanted non-determinism there. If you could spare some time to look into the non-determinism, that would be great. Otherwise, my suggestion is to comment out the tests and document the fact.

grosu:
Again: First and foremost, make sure that RV-Predict does NOT report any data-race in K.  How can we claim that we have a program verifier if it has random behaviors, in particular data-races?

Please: do not comment definition/tests out only because you feel like they are flaky.  Unless you guys bring evidence that the lambda definitions or tests are indeed non-deterministic due to mistakes in their design, I would say that it is a WRONG approach to remove them only because they are inconvenient.  And if they have indeed non-deterministic (but equivalent) results, which may well happen in the case of environment-based definitions of lambda calculi because the closures can get different locations, then why don't you run those tests with the search option and thus get even more powerful tests out of them?

radumereuta:
I decided to merge this into the module qualified sorts. It will be reviewed with the rest.
It is more likely the bug that made these tests flaky will be fixed there.

PR #2165, opened 2016-04-11
add assertion for unique module names in a definition

andreistefanescu:
This assertion fails on `k-distribution/samples-kore/kernelc`.


cos:
addressed in #1900

PR #2174, opened 2016-05-02
MInt not considered when loading configuration into spec rules

daejunpark:

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/kil/Token.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/kil/Token.java
@@ -1,6 +1,7 @@
 // Copyright (c) 2013-2016 K Team. All Rights Reserved.
 package org.kframework.backend.java.kil;
 
+import org.kframework.backend.java.builtins.BitVector;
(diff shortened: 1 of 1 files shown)


daejunpark:
@andreistefanescu Do I need to consider this as well?
https://github.com/kframework/k/blob/master/java-backend/src/main/java/org/kframework/backend/java/symbolic/KILtoBackendJavaKILTransformer.java#L174

andreistefanescu:
`KILtoBackendJavaKILTransformer.java` is dead code, so don't worry about it.

Are you trying to add syntax for machine integers?

daejunpark:
No, this fixes a bug that occurs when K prover loads and replaces configuration cells with place-holder variables in spec rules. For example, suppose we have `<memory> MEMORY:Map </memory>` in a spec rule, and `MEMORY` is supposed to be replaced with a map containing machine integers. In this case, the machine integers are loaded as UninterpretedToken instead of BitVector.

cos:
is this ready for review/merging?

daejunpark:
yes, it is.

PR #2188, opened 2016-07-20
Fix io

andreistefanescu:


java-backend/src/main/java/org/kframework/backend/java/symbolic/FastRuleMatcher.java
```diff
@@ -109,6 +110,11 @@ public FastRuleMatcher(GlobalContext global, int ruleCount) {
         List<RuleMatchResult> transitionResults = new ArrayList<>();
         for (int i = theMatchingRules.nextSetBit(0); i >= 0; i = theMatchingRules.nextSetBit(i + 1)) {
             Rule rule = global.getDefinition().ruleTable.get(i);
+            // skip over IO rules when in prove rules
+            if (proveFlag && rule.containsAttribute("stream")) {
+                continue;
+            }
```
cos, line 116: added #2189 to track

k-distribution/tutorial/1_k/2_imp/lesson_4/imp.k
```diff
@@ -54,4 +54,8 @@ module IMP
   rule <k> int (X,Xs => Xs);_ </k> <state> Rho:Map (.Map => X|->0) </state>
     requires notBool (X in keys(Rho))
   rule int .Ids; S => S  [structural]
+
+// verification ids
+  syntax Id ::= "n"     [token]
+              | "sum"   [token]
```
cos, line 60: could we declare these somewhere else?

andreistefanescu, line 60: it should be possible to declare them in the same module as the specification rules, but then that would require the creation of a new parser form that module, so for now I would leave them here.

cos, line 60: ok

Issue #1423, opened 2015-03-26
Detect module circularity


dwightguth:
We need to print an error message if modules import themselves rather than to get a stack overflow

PR #2191, opened 2016-07-28
Added dependency cycle detection as per #1423

adrianherrera:
Mimics a recursive DFS approach for detecting cycles in module requirements. Use canonical paths to ensure that file comparisons are correct.

This is my first time contributing/using K, so please let me know what you think!

### DIFF
--- a/kernel/src/main/java/org/kframework/parser/concrete2kore/ParserUtils.java
+++ b/kernel/src/main/java/org/kframework/parser/concrete2kore/ParserUtils.java
@@ -22,7 +22,9 @@ import org.kframework.utils.file.FileUtil;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Deque;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


cos:
Hi Adrian, thank you for the contribution! It looks good. Catching the checked exception inside `anyMatch` is not great, but there's not much to do considering the bug in the Java API. And in this case, the exception should never happen anyway.

Also, thanks for fixing the typo :-)

adrianherrera:
Hi Cosmin,

I agree with your comment on catching the exception, however I don't know if moving the try/catch block elsewhere helps much. If `getCanonicalFile` is failing then there are probably problems to worry about :) If you have any other suggestions I'd be happy to hear them.

daejunpark:
Just in case @adrianherrera is not allowed to see the Codeship log, here is the reason why the build failed:

```
[INFO] --- maven-checkstyle-plugin:2.13:check (checkstyle-java) @ kernel ---
[INFO] Starting audit...
/home/rof/src/github.com/kframework/k/kernel/src/main/java/org/kframework/parser/concrete2kore/ParserUtils.java:164:35: Catching 'Exception' is not allowed.
Audit done.
```

cos:
thanks @daejunpark 

checkstyle doesn't like the catching of `Exception`. @adrianherrera, could you fix this in your branch, or shall we take it from here?

sorry for not being able to see the build failure reason -- we'll try to figure out a solution for the future

adrianherrera:
Thanks guys. Yeah I'm unable to view the CI details :-( I'm happy to it up!

cos:
one simple check is to do `mvn verify -DskipKTest` locally. It runs in a couple of minutes and makes sure the basic parts are working.

adrianherrera:
Fixed. In hindsight, I have no idea why I caught `Exception` there and not `IOException`. Sorry for the brain fart!

daejunpark:
@cos the codeship build wasn't triggered for the new commit. How to trigger it?

adrianherrera:
Oops, that could be because I squashed the commits!

adrianherrera:
Oops, that could be because I squashed the commits

cos:
@daejunpark, Codeship only triggers for branches of the main repo, not for forks. To make it trigger, check out this branch, and push it to the main repo. It will trigger a build for that branch, and github will notice this branch is identical and also annotate it with the results.

daejunpark:
@cos thanks, now it seems to pass all the tests. You can merge if you want.

Issue #2194, opened 2016-08-06
error with klabel \\\_.\_ in samples/agent


andreistefanescu:
KLabel \_._ causes krun to crush. In the backend, the label becomes ._, and that seems to be the problem.

traiansf:
@radumereuta I've created a minimal example to ease your task.
Interestingly enough, compilation works just fine, the error arises only when attempting to run a program.

[minimal-example.zip](https://github.com/kframework/k/files/407619/minimal-example.zip)

radumereuta:
I think I found the problem. The output of kast is incorrect. The klabel of the production `Exp ::= "\\" Id "." Exp` should be ``\\_._`` because backslash is the escape character, and when pretty printed like that, should be doubled.

The problem is that I don't know how to proceed with the fix because the escape function is in `kernel`  `StringUtil.escapeKoreKLabel`, and the unperser is written in scala, in the `kore` module, which doesn't import the first one.

@cos, can I get some assistance with, what seems to be, a refactoring issue?

cos:
Moved the relevant functions to the kore module:
https://github.com/kframework/k/tree/hotfix/2194

Could you do the rewiring to use them?

radumereuta:
@traiansf please review.

PR #2196, opened 2016-08-11
Hotfix/2194

radumereuta:
Fixing #2194 
There was a problem with pretty printing, where klabels wouldn't be properly escaped.


traiansf:
I tested it and it works on the other agent examples.
looks good to me

cos:
we don't need to be extra strict about this but the convention is for the reviewer to merge.

regardless, thank you for the fix!

traiansf:
Sorry, I did not know about the convention... Maybe I should rebecome
active in this project :)

PR #2197, opened 2016-08-15
Remove DI + Cleanup API + EquivChecker

daejunpark:
Do not merge. Only for the review purpose.


shell/src/main/java/org/kframework/main/Main.java
```diff
@@ -0,0 +1,422 @@
...
+        System.exit(1);
+    }
+
+    private static volatile boolean isNailgun;
+
+    public static boolean isNailgun() {
+        return isNailgun;
+    }
+
+    public static void nailMain(NGContext context) {
```
grosu, line 410: Should we try to not depend on Nailgun anymore?

grosu:
@daejunpark : so you want us to review this as a diff with branch feature/kdoc-with-meta?  Either way, I think you should merge as soon as you can (provided all tests pass).

shell/src/main/java/org/kframework/main/Main.java
daejunpark: @grosu I just leaved it there since it is nothing to do with the dependency injection. Removing (or replacing) Nailgun can be done in a separate PR once we all agree.

daejunpark:
@grosu I'll definitely make this PR reviewable and take care of merging it once the ancestors are merged.

API/src/main/java/org/kframework/Kapi.java
```diff
@@ -0,0 +1,481 @@
...
+        CompiledDefinition compiledDef2 = kapi.kompile(def2, mod2);
+
+        // krun
+        RewriterResult result1 = kapi.krun(pgm1, null, compiledDef1);
+        kprint(compiledDef1, result1);
+        RewriterResult result2 = kapi.krun(pgm2, null, compiledDef2);
+        kprint(compiledDef2, result2);
+
+        // kprove
+        kapi.kprove(prove, prelude, compiledDef0);
```
andreistefanescu, line 209: does it work without this line?

API/src/main/java/org/kframework/EquivChecker.java
```diff
@@ -0,0 +1,269 @@
...
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * Created by daejunpark on 8/15/16.
+ */
+public class EquivChecker {
+
+    public static boolean equiv(
+            java.util.List<ConstrainedTerm> startSyncNodes1,
```
andreistefanescu, line 16: you could combine these into a record class, since all the lists have the same number of elements

cos:
@daejunpark, after integrating Andrei's feedback, could you make another PR towards master?

API/src/main/java/org/kframework/EquivChecker.java
```diff
@@ -0,0 +1,269 @@
...
+            java.util.List<ConstrainedTerm> startSyncNodes2,
```
andreistefanescu, line 17: minor thing, can `java.util.List` just be `List`?

API/src/main/java/org/kframework/EquivChecker.java
```diff
@@ -0,0 +1,269 @@
...
+            if (nt1.isEmpty()) continue;
+             */
+
+            currSyncNodes1.add(new SyncNode(i, null, t1, null)); // TODO: check if null is safe
+            currSyncNodes2.add(new SyncNode(i, null, t2, null));
+        }
+
+        while (!currSyncNodes1.isEmpty() && !currSyncNodes2.isEmpty()) {
+
+            java.util.List<Set<SyncNode>> nextSyncNodes1 = getNextSyncNodes(currSyncNodes1, targetSyncNodes1, rewriter1);
```
andreistefanescu, line 57: in general, I feel that using `org.apache.commons.lang3.tuple.Pair`, or specialized classes rather then lists with the same length would improve the code.

API/src/main/java/org/kframework/EquivChecker.java
```diff
@@ -0,0 +1,269 @@
...
+            SymbolicRewriter rewriter1,
+            SymbolicRewriter rewriter2
+    ) {
+
+        assert targetSyncNodes1.size() == ensures.size();
+        assert targetSyncNodes2.size() == ensures.size();
+
+        int numSyncPoints = ensures.size();
+
+        java.util.List<Set<SyncNode>> allSyncNodes1 = newListOfSets(numSyncPoints);
```
andreistefanescu, line 32: it seems to me you are simultaneously proving that all the sync points ca reach the sync points; I would suggest  a method for just one sync point, and then calling this method for each sync point

andreistefanescu (commented):
Overall it looks good. I made some style suggestions. I don't want to delay this any longer then necessary, so we should merge first and then we can discuss the algorithm for program equivalence further.

PR #2200, opened 2016-08-25
Fixing issue 2182 

radumereuta:
Where multiple list productions for the same type of cells would interfere and use too much memory.

### DIFF
--- a/k-distribution/tests/regression/config.xml
+++ b/k-distribution/tests/regression/config.xml
@@ -80,6 +80,10 @@
            directory="issue1724"
            skip="pdf" />
 
+  <include file="issue2182/config.xml"
+           directory="issue2182"
+           skip="pdf" />
+
   <include file="issue2194/config.xml"
            directory="issue2194"
            skip="pdf" />
--- a/k-distribution/tests/regression/issue2182/config.xml
+++ b/k-distribution/tests/regression/issue2182/config.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright (c) 2015-2016 K Team. All Rights Reserved. -->
+<tests>
+  <test
+      definition="test.k"
+      programs="."
+      extension="test"
+      results="." >
+  </test>
+</tests>
+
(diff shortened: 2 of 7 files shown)


dwightguth:
This is not really a good solution if type="Set", "Map", or "List". It won't actually target that production effectively in that case.

radumereuta:
I found this problem only on parsing cells that have multiplicity="*" where the there were some productions added during configuration concretzation (I think). They were only using `Bag` and I didn't want to remove those since most likely they are used later, I just removed them from the parser. They were already tagged with `avoid`, so they weren't intended for parsing anyway. What we failed to anticipate, was how much of a problem they can cause at parsing time.

I'm not sure if it's the best solution, but it seems the least upset right now.

radumereuta:
Ah, and its ready for review. Since you already started, @dwightguth please review.

dwightguth:
It looks fine other than the comment that I included. There's actually a bug in ConfigurationInfoFromModule that got fixed in the rv-k by adding a cellCollection attribute on all cell collection productions. You might consider doing the same and then using that attribute here, the way I did. Otherwise you're going to have to be doing increasingly desperate guesswork in order to try to decide whether something is a cell collection...

radumereuta:
@dwightguth please have another look. Those are flaky tests. Tested twice, and on my computer, and the tests work as intended.

dwightguth:
looks fine, I'm not going to comment on the flaky test issue. someone else can take over the review from here.

radumereuta:
@daejunpark please review.

radumereuta:
Can someone else review this? Maybe @cos.

cos:
Why are cell cell collections were interfering with `Bag` parsing? Is there some extra subsorting somewhere between `CellBag` and `Bag`?

radumereuta:
I haven't looked too deeply into that, but most likely yes, through some indirect subsorts.
I just noticed those productions weren't supposed to be in the parser, and just removed them.

daejunpark:
@radumereuta sorry for the delay. It looks good to me mainly because the change is small and seems natural, though I did not fully understand what's going on.

Let me try to use this branch to check it solved my original problem in addition to the simplified version I reported.

daejunpark:
It seems to work for my original problem as well. @cos you can merge if you have no more concern with this.

PR #2210, opened 2016-10-03
Ability to print constraints in search mode

msaxena2:
Opened PR from branch in the main repo to enable quicker building. Carried forward from #2209 .


cos:
you can update the outputs with `--update-out`

msaxena2:
@cos @ehildenb @lucaspena Please review. Also, the constraints are now only printed when search is called with the --print-constraints flag.

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
```diff
@@ -33,10 +36,12 @@
 
 import java.lang.invoke.MethodHandle;
 import java.math.BigInteger;
+import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
```
ehildenb: @msaxena2 It builds fine on my system without the extra import statements.

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
```diff
@@ -129,33 +134,30 @@ public RewriterResult execute(K k, Optional<Integer> depth) {
...
             Term backendKil = MacroExpander.expandAndEvaluate(termContext, kem, converter.convert(k));
-            this.rewriter = new SymbolicRewriter(rewritingContext,  kompileOptions, new KRunState.Counter(), converter);
+            this.rewriter = new SymbolicRewriter(rewritingContext, kompileOptions, new KRunState.Counter(), converter);
             JavaKRunState result = (JavaKRunState) rewriter.rewrite(new ConstrainedTerm(backendKil, termContext), depth.orElse(-1));
             return new RewriterResult(result.getStepsTaken(), result.getJavaKilTerm());
         }
 
         @Override
-        public List<? extends Map<? extends KVariable,? extends K>> match(K k, org.kframework.definition.Rule rule) {
+        public List<Tuple2<? extends Map<? extends KVariable, ? extends K>, ? extends K>> match(K k, org.kframework.definition.Rule rule) {
```
cos: it may not be clear what the second projection of the tuple is. a new class holding the map and the remaining constraints, or some javadoc to this method would help.

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
```diff
@@ -129,33 +134,30 @@ public RewriterResult execute(K k, Optional<Integer> depth) {
...
             return search(k, Optional.of(0), Optional.empty(), rule, SearchType.STAR);
         }
 
 
         @Override
-        public List<? extends Map<? extends KVariable, ? extends K>> search(K initialConfiguration, Optional<Integer> depth, Optional<Integer> bound, Rule pattern, SearchType searchType) {
+        public List<Tuple2<? extends Map<? extends KVariable, ? extends K>, ? extends K>> search(K initialConfiguration, Optional<Integer> depth, Optional<Integer> bound, Rule pattern, SearchType searchType) {
```
cos: same, class or javadoc

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
```diff
@@ -129,33 +134,30 @@ public RewriterResult execute(K k, Optional<Integer> depth) {
...
-                    .search(javaTerm, javaPattern, bound.orElse(NEGATIVE_VALUE), depth.orElse(NEGATIVE_VALUE), searchType, termContext)
+            this.rewriter = new SymbolicRewriter(rewritingContext, kompileOptions, new KRunState.Counter(), converter);
+            return rewriter.search(javaTerm, javaPattern, bound.orElse(NEGATIVE_VALUE), depth.orElse(NEGATIVE_VALUE), searchType, termContext)
                     .stream().collect(Collectors.toList());
-            return searchResults;
         }
 
 
-        public Tuple2<RewriterResult, List<? extends Map<? extends KVariable, ? extends K>>> executeAndMatch(K k, Optional<Integer> depth, Rule rule) {
+        public Tuple2<RewriterResult, List<Tuple2<? extends Map<? extends KVariable, ? extends K>, ? extends K>>> executeAndMatch(K k, Optional<Integer> depth, Rule rule) {
```
cos: same, class or javadoc.

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -402,7 +401,7 @@ public static ConstrainedTerm buildResult(
      * up to the bound as were found, and returns {@code true} if the bound has been reached.
      */
     private boolean addSearchResult(
-            HashSet<Substitution<Variable, Term>> searchResults,
+            HashSet<Tuple2<Substitution<Variable, Term>, K>> searchResults,
```
cos: same as above, class or javadoc

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -430,7 +433,7 @@ private boolean addSearchResult(
      * @param searchType  defines when we will attempt to match the pattern
      * @return a list of substitution mappings for results that matched the pattern
      */
-    public Set<Substitution<Variable, Term>> search(
+    public Set<Tuple2<Substitution<Variable, Term>, K>> search(
```
cos: same as above, class or javadoc

cos: also, the javadoc seems to be obsolete -- might help to update it anyhow

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -439,7 +442,7 @@ private boolean addSearchResult(
             TermContext context) {
         stopwatch.start();
 
-        HashSet<Substitution<Variable, Term>> searchResults = Sets.newHashSet();
+        HashSet<Tuple2<Substitution<Variable, Term>, K>> searchResults = Sets.newHashSet();
```
cos: same

kernel/src/main/java/org/kframework/debugger/DebuggerMatchResult.java
```diff
@@ -13,12 +14,12 @@
  * Represents the result of a match on a configuration
  */
 public class DebuggerMatchResult {
-    private final List<? extends Map<? extends KVariable, ? extends K>> substitutions;
+    private final List<Tuple2<? extends Map<? extends KVariable, ? extends K>, ? extends K>>  substitutions;
```
cos: same -- it looks like a separate class would be the best choice. as we want to work with K conjunctions in the future, it may be best to use that already, and document the methods.

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -125,27 +126,39 @@ public int run(CompiledDefinition compiledDef, KRunOptions options, Function<Mod
     }
 
     private void printSearchResult(SearchResult result, KRunOptions options, CompiledDefinition compiledDef) {
-        Set<Map<? extends KVariable, ? extends K>> searchResult = ((SearchResult) result).getSearchList().stream()
-                .map(subst -> filterAnonymousVariables(subst, result.getParsedRule()))
-                .collect(Collectors.toSet());
+        Set<Map<? extends KVariable, ? extends K>> searchResult = new HashSet<>();
+        ArrayList<K> constraints = new ArrayList<>();
+        result.getSearchList().forEach(tuple -> {
```
cos: "tuple" is not informative

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -125,27 +126,39 @@ public int run(CompiledDefinition compiledDef, KRunOptions options, Function<Mod
...
+            constraints.add(tuple._2());
+        });
         outputFile("Search results:\n\n", options);
         if (searchResult.isEmpty()) {
             outputFile("No search results \n", options);
         }
-        int i = 1;
-        List<String> results = new ArrayList<>();
+        int i = 0;
+        List<Tuple2<String, String>> results = new ArrayList<>();
```
cos: what's the first and the second string?

kernel/src/main/java/org/kframework/krun/KRunOptions.java
```diff
@@ -160,6 +160,9 @@ public OutputModeConverter(String optionName) {
     @Parameter(names="--search", description="In conjunction with it you can specify 3 options that are optional: pattern (the pattern used for search), bound (the number of desired solutions) and depth (the maximum depth of the search).")
     public boolean search = false;
 
+    @Parameter(names={"--print-constraint", "--print-constraints"}, description="If specified with search, prints the constraints for each solution.")
+    public boolean printConstraint = false;
```
cos: if there is no other problem, let's always print the constraints (unless they are simply true)

kore/src/main/scala/org/kframework/rewriter/Rewriter.scala
```diff
@@ -23,7 +23,7 @@ trait Rewriter {
   def execute(k: kore.K, depth: Optional[Integer]): RewriterResult
 
 
-  def `match`(k: kore.K, rule: Rule): java.util.List[_ <: java.util.Map[_ <: kore.KVariable, _ <: kore.K]]
+  def `match`(k: kore.K, rule: Rule): java.util.List[Tuple2[_ <: java.util.Map[_ <: kore.KVariable, _ <: kore.K], _ <: kore.K]]
```
cos: again, considering that this is the K API, we really should go with the conjunction. I know we discussed this before, but I didn't realize that the tuple will be leaking all the way to the API.

kore/src/main/scala/org/kframework/rewriter/Rewriter.scala
```diff
@@ -34,9 +34,9 @@ trait Rewriter {
    * @param pattern The rule (pattern + side condition) that we're trying to find a substitution for.
    * @return A list of substitutions, denoting all the configurations matching the given rule.
    */
-  def search(initialConfiguration: kore.K, depth: Optional[Integer], bound: Optional[Integer], pattern: Rule, searchType: SearchType): java.util.List[_ <: java.util.Map[_ <: kore.KVariable, _ <: kore.K]]
+  def search(initialConfiguration: kore.K, depth: Optional[Integer], bound: Optional[Integer], pattern: Rule, searchType: SearchType): java.util.List[Tuple2[_ <: java.util.Map[_ <: kore.KVariable, _ <: kore.K], _ <: kore.K]]
```
cos: same

kore/src/main/scala/org/kframework/rewriter/Rewriter.scala
```diff
@@ -34,9 +34,9 @@ trait Rewriter {
...
 
-  def executeAndMatch(k: kore.K, depth: Optional[Integer], rule: Rule): Tuple2[RewriterResult, java.util.List[_ <: java.util.Map[_ <: kore.KVariable, _ <: kore.K]]]
+  def executeAndMatch(k: kore.K, depth: Optional[Integer], rule: Rule): Tuple2[RewriterResult, java.util.List[Tuple2[_ <: java.util.Map[_ <: kore.KVariable, _ <: kore.K], _ <: kore.K]]]
```
cos: same

cos (changes requested):
There are two main changes:
1. replace the Tuple2 with a conjunction
2. if there is no reason to keep it, remove the `--print-constraint` flag and have it always on

kore/src/main/scala/org/kframework/rewriter/Rewriter.scala
msaxena2: I agree that a conjunction should be the way to go. The rewriter returns a list of maps, where each map represents a substitution. I'm not sure how we can go from a list of maps to a conjunction, unless we make the rewriter itself return the map as a K term (which can then be used in a conjunction with the constraint). @cos  is that what you had in mind as well?

andreistefanescu:
The list of conjunctions can be represented by a disjunction.

kore/src/main/scala/org/kframework/rewriter/Rewriter.scala
cos: Yes. As @andreistefanescu said below, you want a disjunction of conjunctions. Each map becomes a conjunction, and the list becomes a disjunction.

msaxena2:
@cos Please have a look.

java-backend/src/main/java/org/kframework/backend/java/symbolic/InitializeRewriter.java
```diff
@@ -129,33 +128,29 @@ public RewriterResult execute(K k, Optional<Integer> depth) {
...
-        public List<? extends Map<? extends KVariable,? extends K>> match(K k, org.kframework.definition.Rule rule) {
-            return search(k, Optional.of(0), Optional.empty(), rule, SearchType.STAR);
+        public K match(K k, org.kframework.definition.Rule rule) {
+            return search(k, Optional.of(0), Optional.empty(), rule, SearchType.STAR, true);
         }
 
 
         @Override
-        public List<? extends Map<? extends KVariable, ? extends K>> search(K initialConfiguration, Optional<Integer> depth, Optional<Integer> bound, Rule pattern, SearchType searchType) {
+        public K search(K initialConfiguration, Optional<Integer> depth, Optional<Integer> bound, Rule pattern, SearchType searchType, boolean substitution) {
```
cos: rename `substitution` to something line `resultsAsSubstitutions`

k-distribution/tests/examples/symbolic/1.a.out
```diff
@@ -1,8 +1,6 @@
-Search results:
-
-Solution 1:
+Solution 1
 <k> 0 </k>
-
-Solution 2:
+ _andBool_ V0 ==K 0
```
cos: this should be the ML conjunction, not the bool one

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -124,30 +123,85 @@ public int run(CompiledDefinition compiledDef, KRunOptions options, Function<Mod
...
+        } else {
+            prettyPrint(compiledDef, options.output, s -> bs.write(s, 0, s.length), result);
+            sb.append(new String(bs.toByteArray()));
+            return sb;
         }
-        Collections.sort(results);
+    }
+
+    public void printK(K result, KRunOptions options, CompiledDefinition compiledDef) {
+        Set<String> patternVariables = new HashSet<>();
```
cos: `patternVariables` => `patternVariableNames`

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -124,30 +123,85 @@ public int run(CompiledDefinition compiledDef, KRunOptions options, Function<Mod
...
+            new VisitK() {
+                @Override
+                public void apply(KVariable k) {
+                    patternVariables.add(k.name());
+                    super.apply(k);
+                }
+            }.apply(parsePattern(files, kem, options.pattern, compiledDef, Source.apply("<command line: --pattern>")).body());
+        }
+        if (result instanceof KApply && ((KApply) result).klabel().equals(KLabel(KLabels.OR))) {
+            Some<Tuple2<KLabel, scala.collection.immutable.List<K>>> searchResults = KApply$.MODULE$.unapply((KApply) result);
```
cos: very involved. as we're in Java, simply cast to `KApply` and get its `KLabel` and `KList`, or I think it also works do to directly `items()` on it

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -124,30 +123,85 @@ public int run(CompiledDefinition compiledDef, KRunOptions options, Function<Mod
...
+            scala.collection.Seq<K> seq = Assoc.flatten(KORE.KLabel(KLabels.OR), searchResults.get()._2(), KORE.KLabel(KLabels.ML_FALSE));
```
cos: this should us the ML disjunction

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -124,30 +123,85 @@ public int run(CompiledDefinition compiledDef, KRunOptions options, Function<Mod
...
+        if (options.pattern != null) {
+            new VisitK() {
+                @Override
+                public void apply(KVariable k) {
+                    patternVariables.add(k.name());
+                    super.apply(k);
+                }
+            }.apply(parsePattern(files, kem, options.pattern, compiledDef, Source.apply("<command line: --pattern>")).body());
+        }
+        if (result instanceof KApply && ((KApply) result).klabel().equals(KLabel(KLabels.OR))) {
```
cos: ML `or`

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -124,30 +123,85 @@ public int run(CompiledDefinition compiledDef, KRunOptions options, Function<Mod
...
+            Some<Tuple2<KLabel, scala.collection.immutable.List<K>>> searchResults = KApply$.MODULE$.unapply((KApply) result);
+            scala.collection.Seq<K> seq = Assoc.flatten(KORE.KLabel(KLabels.OR), searchResults.get()._2(), KORE.KLabel(KLabels.ML_FALSE));
+            List<K> resultList = mutable(seq).stream().filter(x -> !x.getClass().toString().contains("BoolToken")).collect(Collectors.toList());
```
cos: do we end up with `true` and `false` bool tokens in the result? we shouldn't, for multiple reasons.

first, (unless they really are the final result of the computation; not the case ever in our examples) bool stuff should not appear in the disjunction directly. it should be `theBoolStuff ==K true`.

second, if talking about ML top and bottom (as opposed to true/false) we should either see just top, or a disjunction that has no bottom (normalized)

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -157,23 +211,29 @@ private void printSearchResult(SearchResult result, KRunOptions options, Compile
      * @param res The substitution from the match of the user specified pattern on the Final Configuration.
      * @return An int representing the error code.
      */
-    public static int getExitCode(KExceptionManager kem, List<? extends Map<? extends KVariable, ? extends K>> res) {
-        if (res.size() != 1) {
-            kem.registerCriticalWarning("Found " + res.size() + " solutions to exit code pattern. Returning 112.");
+    public static int getExitCode(KExceptionManager kem, K res) {
```
cos, line 216: what is this "exitCode"?

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -124,30 +123,85 @@ public int run(CompiledDefinition compiledDef, KRunOptions options, Function<Mod
...
-        List<String> results = new ArrayList<>();
-        for (Map<? extends KVariable, ? extends K> substitution : searchResult) {
-            ByteArrayOutputStream sb = new ByteArrayOutputStream();
-            prettyPrintSubstitution(substitution, result.getParsedRule(), compiledDef, options.output, v -> sb.write(v, 0, v.length));
-            //Note that this is actually unsafe, but we are here assuming that --search is not used with --output binary
-            results.add(new String(sb.toByteArray()));
+    private StringBuilder filterAnonVarsAndPrint(K result, Set<String> filterSet, CompiledDefinition compiledDef, KRunOptions options) {
+        StringBuilder sb = new StringBuilder();
+        ByteArrayOutputStream bs = new ByteArrayOutputStream();
+        if (result instanceof KApply && ((KApply) result).klabel().toString().equals(KLabels.AND)) {
```
cos: do not check string-based equality

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -525,15 +552,41 @@ private boolean addSearchResult(
...
 
-        Set<Substitution<Variable, Term>> adaptedResults = searchResults.stream().map(r -> {
-            RenameAnonymousVariables renameAnonymousVariables = new RenameAnonymousVariables();
-            Substitution<Variable, Term> subs = new HashMapSubstitution();
-            r.forEach((k, v) -> subs.plus(renameAnonymousVariables.getRenamedVariable(k), renameAnonymousVariables.apply(v)));
-            return subs;
-        }).collect(Collectors.toSet());
+    private void flattenList(List<K> unflat, List<K> flat) {
+        unflat.forEach(x -> {
+            if (x instanceof KItem && ((KItem) x).klabel().toString().equals(KLabels.AND)) {
```
cos: do not check equality based on `toString`

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -525,15 +552,41 @@ private boolean addSearchResult(
...
+                flattenList(((KItem) x).items(), flat);
+            } else {
+                flat.add(x);
+            }
+        });
+    }
+
+    private K processConjuncts(ConjunctiveFormula conjunct) {
+        if (conjunct.klabel().toString().equals(KLabels.AND)) {
```
cos: same, `toString`

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -124,30 +123,85 @@ public int run(CompiledDefinition compiledDef, KRunOptions options, Function<Mod
...
+            List<K> conjunctions = mutable(Assoc.flatten(KLabel(KLabels.AND), immutable(((KApply) result).items()), KLabel(KLabels.ML_TRUE)));
+            conjunctions = conjunctions.stream().filter(x -> {
+                if (x.getClass().toString().contains("BoolToken")) {
+                    return false;
+                }
+                if ((options.pattern != null) && x instanceof KApply && ((KApply) x).klabel().toString().equals(KLabels.EQUALS)) {
```
cos: `toString`

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -124,30 +123,85 @@ public int run(CompiledDefinition compiledDef, KRunOptions options, Function<Mod
...
-            prettyPrintSubstitution(substitution, result.getParsedRule(), compiledDef, options.output, v -> sb.write(v, 0, v.length));
-            //Note that this is actually unsafe, but we are here assuming that --search is not used with --output binary
-            results.add(new String(sb.toByteArray()));
+    private StringBuilder filterAnonVarsAndPrint(K result, Set<String> filterSet, CompiledDefinition compiledDef, KRunOptions options) {
+        StringBuilder sb = new StringBuilder();
+        ByteArrayOutputStream bs = new ByteArrayOutputStream();
+        if (result instanceof KApply && ((KApply) result).klabel().toString().equals(KLabels.AND)) {
+            List<K> conjunctions = mutable(Assoc.flatten(KLabel(KLabels.AND), immutable(((KApply) result).items()), KLabel(KLabels.ML_TRUE)));
+            conjunctions = conjunctions.stream().filter(x -> {
+                if (x.getClass().toString().contains("BoolToken")) {
```
cos: don't use a string for the check

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -124,30 +123,85 @@ public int run(CompiledDefinition compiledDef, KRunOptions options, Function<Mod
...
-            outputFile("No search results \n", options);
-        }
-        int i = 1;
-        List<String> results = new ArrayList<>();
-        for (Map<? extends KVariable, ? extends K> substitution : searchResult) {
-            ByteArrayOutputStream sb = new ByteArrayOutputStream();
-            prettyPrintSubstitution(substitution, result.getParsedRule(), compiledDef, options.output, v -> sb.write(v, 0, v.length));
-            //Note that this is actually unsafe, but we are here assuming that --search is not used with --output binary
-            results.add(new String(sb.toByteArray()));
+    private StringBuilder filterAnonVarsAndPrint(K result, Set<String> filterSet, CompiledDefinition compiledDef, KRunOptions options) {
```
cos, line 126: if possible, separate the filter from the print

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -157,23 +211,29 @@ private void printSearchResult(SearchResult result, KRunOptions options, Compile
...
+        K solution = flatList.apply(1);
         Set<Integer> vars = new HashSet<>();
-        for (K t : solution.values()) {
-            // TODO(andreistefanescu): fix Token.sort() to return a kore.Sort that obeys kore.Sort's equality contract.
-            if (t instanceof KToken && Sorts.Int().equals(((KToken) t).sort())) {
-                try {
-                    vars.add(Integer.valueOf(((KToken) t).s()));
-                } catch (NumberFormatException e) {
-                    throw KEMException.criticalError("Exit code found was not in the range of an integer. Found: " + ((KToken) t).s(), e);
+        new VisitK() {
```
cos, line 225: extract visitor to someplace else

kernel/src/main/java/org/kframework/krun/KRun.java
```diff
@@ -124,30 +123,85 @@ public int run(CompiledDefinition compiledDef, KRunOptions options, Function<Mod
...
+
+    public void printK(K result, KRunOptions options, CompiledDefinition compiledDef) {
+        Set<String> patternVariables = new HashSet<>();
         StringBuilder sb = new StringBuilder();
-        for (String solution : results) {
-            sb.append("Solution ").append(i++).append(":\n");
-            sb.append(solution);
-            sb.append("\n");
+        if (options.pattern != null) {
+            new VisitK() {
```
cos, line 164: separate visitor -- this method is kind of large

kore/src/main/scala/org/kframework/rewriter/Rewriter.scala
```diff
@@ -34,9 +34,9 @@ trait Rewriter {
    * @param pattern The rule (pattern + side condition) that we're trying to find a substitution for.
    * @return A list of substitutions, denoting all the configurations matching the given rule.
    */
-  def search(initialConfiguration: kore.K, depth: Optional[Integer], bound: Optional[Integer], pattern: Rule, searchType: SearchType): java.util.List[_ <: java.util.Map[_ <: kore.KVariable, _ <: kore.K]]
+  def search(initialConfiguration: kore.K, depth: Optional[Integer], bound: Optional[Integer], pattern: Rule, searchType: SearchType, subsitution:Boolean): kore.K
```
cos: I really don't like how this method behaves differently based on the substitution flag. Let's leave it like this for now so we can merge, but please add an issue and we should fix this with the very first opportunity.

cos (changes requested):

kernel/src/main/java/org/kframework/krun/KRun.java
msaxena2: Already existing code, seems like it matches some pre-existing pattern on the final configuration. Should probably get rid of it in a separate PR.

kernel/src/main/java/org/kframework/krun/KRun.java
msaxena2: Let's let this be as is for now. We'll probably get rid of this function in a later PR.

msaxena2:
@cos Please re-review.

cos (approved):

PR #2214, opened 2016-10-07
remove the format attribute in PRETTY output mode.

yzhang90:
remove the format attribute in PRETTY output mode (for cs427 purpose). Can recover it if necessary


cos (approved):
thanks!

PR #2216, opened 2016-10-07
Fix for requires with relative paths in the equivalence checker

theo25:
This is the fix for requires relative paths. I have rebased it on top of master.

@daejunpark

### DIFF
--- a/API/src/main/java/org/kframework/Kapi.java
+++ b/API/src/main/java/org/kframework/Kapi.java
@@ -1,6 +1,7 @@
 // Copyright (c) 2016 K Team. All Rights Reserved.
 package org.kframework;
 
+import com.google.common.collect.Lists;
 import org.kframework.RewriterResult;
 import org.kframework.attributes.Att;
 import org.kframework.attributes.Source;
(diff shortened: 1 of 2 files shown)


daejunpark:
This PR will be subsumed by #2217.

@theo25 could you please check if #2217 (or keq branch 6e40fb5f0ec35925aa1ef387636690e7cedcad9e)  fixed your problem?

theo25:
@daejunpark No, the problem is still there when trying with this branch/commit.

daejunpark:
@theo25 oh, sorry I forgot to say that you should use another command. `Kapi.main` is deprecated and will be deleted. Please use this command:

```
$ keq <prelude.smt2> <def0.k> <mod0> <def1.k> <mod1> <def2.k> <mod2> <spec.k>
```

For example,

```
$ cd k-distribution/tests/equiv/test2
$ keq ../../../../k-distribution/include/z3/basic.smt2 imp.k IMP imp1.k IMP imp2.k IMP if_spec.k
```

You can find more examples in https://github.com/kframework/k/blob/keq/k-distribution/tests/equiv/test2/run.sh

theo25:
@daejunpark I am actually using the org.kframework.main.Main class from the shell module with -keq as first argument and the rest of the arguments as you suggested. Is this also deprecated?

daejunpark:
@theo25 you're right. Now I see what's the problem. I'll incorporate your fix into the PR.

daejunpark:
@theo25 could you please check if #2225 fixed the original problem?

theo25:
@daejunpark Yes it did. Thanks!

daejunpark:
closed via #2225

PR #2217, opened 2016-10-11
add tests for keq

daejunpark:


daejunpark:
@cos please review. I added a ktest entry for keq testing. For this, I implemented a new functionality of ktest that allows to run other tools than krun, which is keq in this case.

But still I have an unresolved issue when maven running ktest under nailgun --- multiple instances of keq under ktest under kserver/nailgun lead to memory corruption error (freeing _un_allocated memory). So, I made a simple hack into maven pom.xml so that the ktest entries for keq are executed without nailgun. Fixing it correctly is todo.

cos:
it's very curious to get memory corruption. did it appear repeatedly?

daejunpark:
@cos yes, it is pretty much reproducible.

daejunpark:
To reproduce:

```
 $ kserver
 $ cd k-distribution/tests/equiv
 $ ktest config.xml
```

Then you will see:

```
$ kserver
K server started on 127.0.0.1:2113
java(989,0x13d9c6000) malloc: *** error for object 0x7fe8b604cd10: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
java(989,0x169c08000) malloc: *** error for object 0x7fe8b604cd18: incorrect checksum for freed object - object was probably modified after being freed.
*** set a breakpoint in malloc_error_break to debug
/Users/daejunpark/work/k/k-distribution/target/release/k/bin/kserver: line 11:   989 Abort trap: 6           java -Djava.awt.headless=true -Djansi.force=true $K_OPTS -ea -cp "$(dirname "$0")/../lib/java/*" org.kframework.main.Main -kserver "$@"
```

@cos have you seen one similar with this?

daejunpark:
I guess it was due to z3. --z3-executable option fixes this problem, but it becomes too slow, since keq calls z3 heavily.

cos:
is it somehow faster if you separate the tests as in this PR?

daejunpark:
No, I just wanted to run it without kserver/nailgun. All of the original tests are run over kserver.

k-distribution/tests/config.xml
```diff
@@ -18,5 +18,11 @@
   <include file="../samples-kore/kernelc/tests/config.xml"
            directory="samples-kore/kernelc"
            skip="pdf" />
+
+<!--
+  <include file     =      "equiv/config.xml"
+           directory="tests/equiv" />
+-->
+
```
cos, line 26: should we remove this?

cos (approved):

k-distribution/tests/config.xml
daejunpark: I want to leave it there. Ideally the test should be here.

PR #2224, opened 2016-10-20
refactorings and fixes throughout

cos:
- the `moduleQualifySortPredicates` pass is not reused so memoization doesn't create problems anymore; when reusing a pass, memoization will remember results from previous runs --  will eventually have to make sure this problem cannot appear in other cases.
- `GenerateSortPredicateSyntax` is applied to `localSorts` only so it doesn't generate redundant subsorts anymore, and `BOOL-SYNTAX` is only imported when not there
- removed a redundant recursion in `RuleGrammarGenerator` - @radumereuta, have a look if you can
- `localSorts` and `listProductions` in `Module` are trimmed down of inherited stuff


cos:
anyone, please review

daejunpark (approved):

daejunpark:
@cos could you explain why the second bullet matters? is there any problem due to that?

cos:
before it was spamming the modules with sort predicates that were already defined in imported modules. now each module only defines sort predicates for the sorts the are introduced by it.

for imports, it only imports the `BOOL-SYNTAX` module when it hasn't already been imported via some other import path.

daejunpark:
ok, you can merge if you want

cos:
I need a review through github to merge. Also, it would be great if you could merge once it passes CodeShip again -- there seems to be a problem with a Java core dump, bit strange.

daejunpark:
for the codeship fail problem, i just sent an email. we can discuss over it.

PR #2229, opened 2016-10-30
refactoring/clean up module transformers

cos:
Main changes:
- removed the use of "hybrid" from all but the front-end code
- most transformers now inherit from ModuleTransformer, which takes care of correct recursive application


cos:
@daejunpark, please review

cos:
@daejunpark is at OOPSLA. anyone else, please feel free to pick this up.

k-distribution/src/test/java/org/kframework/kore/convertors/BubbleParsing.java
```diff
@@ -95,8 +97,8 @@ public BubbleParsing(URL koreSyntax, String mainModule, String startSort) {
      * TODO(radu): generalize this function, and eliminate duplicates
```
radumereuta: Is this comment still valid?

kernel/src/main/java/org/kframework/kore/compile/ResolveAnonVar.java
```diff
@@ -43,7 +44,7 @@ private Context resolve(Context context) {
                 context.att());
     }
 
-    public synchronized Sentence resolve(Sentence s) {
+    public synchronized Sentence f(Sentence s) {
```
radumereuta: This should have a better name. At least JavaDoc.

kernel/src/main/java/org/kframework/kore/compile/ResolveHeatCoolAttribute.java
```diff
@@ -55,7 +56,7 @@ private K transform(K requires, Att att) {
         throw new AssertionError("unreachable");
     }
 
-    public Sentence resolve(Sentence s) {
+    public Sentence f(Sentence s) {
```
radumereuta: Again, better name/JavaDoc, since it's public.

kernel/src/main/java/org/kframework/kore/compile/ResolveSemanticCasts.java
```diff
@@ -124,7 +125,7 @@ public K apply(KVariable k) {
     }
 
 
-    public synchronized Sentence resolve(Sentence s) {
+    public synchronized Sentence f(Sentence s) {
```
radumereuta: dito

kore/src/main/scala/org/kframework/definition/transformers.scala
```diff
@@ -142,14 +120,38 @@ trait HybridModuleTransformer extends ModuleTransformer {
...
-  def processModule(input: Module): Module = wrapExceptions({
+  final def processModule(input: Module): Module = wrapExceptions({
     process(input, input.imports map this)
   })
 
   protected def process(inputModule: Module, alreadyProcessedImports: Set[Module]): Module
 }
 
-abstract class WithInputDefinitionModuleTransformer(inputDefinition: Definition) extends BasicModuleTransformer {
+abstract class SentenceBasedModuleTransformer extends BasicModuleTransformer {
```
radumereuta, line 134: Ah, this is where they come from.
This should have some JavaDoc. Maybe you can skip the previous ones if this class becomes easy to understand.

radumereuta (changes requested):
Overall looks as the code became a bit more readable, although it needs a bit more comments.

k-distribution/src/test/java/org/kframework/kore/convertors/BubbleParsing.java
cos: no, it looks like you already made it nicer at some point.

one thing that could be done here is have the class implement `SentenceBasedModuleTransformer` instead of `BasicModuleTransformer`. The body of the map would become the implementation of `f` and the rest of the code in this function would go away.

cos: no, it looks like you already made it nicer at some point.

one thing that could be done here is have the class implement `SentenceBasedModuleTransformer` instead of `BasicModuleTransformer`. The body of the map would become the implementation of `f` and the rest of the code in this function would go away.

cos:
@radumereuta, I addressed all your comments.

The only place with remaining "hybrid" transformers is the frontend. Would you have time to have a go at removing them from there as well at some point? Or we could work together on it if you like.

radumereuta:
lgtm

radumereuta (approved):

Issue #2213, opened 2016-10-06
 Uncaught exception thrown of type AssertionError


ZhuXiaoran07:
When kompiling the following definition:

```
module TEST-SYNTAX 
  syntax Stmt  ::= "COUNTER" "=" Id ";"   
endmodule
module TEST
  imports TEST-SYNTAX
  syntax KResult ::= Int | Bool | String
  configuration <T color="yellow">
                    <k> $PGM:Stmt</k>
                  <counters>
                  <counter multiplicity="*"> .K </counter>
                  </counters> 
                </T>

    rule <k> COUNTER=I; =>. ...</k>
         <counters> CT:Bag(.Bag =><counter> I </counter>) </counters> 
endmodule

```

it results in  **[Error] Internal: Uncaught exception thrown of type AssertionError.
Please rerun your program with the --debug flag to generate a stack trace, and
file a bug report at https://github.com/kframework/k/issues**.

But if modify the _counters_ cell in the rule as following:

```
module TEST-SYNTAX 
  syntax Stmt  ::= "COUNTER" "=" Id ";"   
endmodule
module TEST
  imports TEST-SYNTAX
  syntax KResult ::= Int | Bool | String
  configuration <T color="yellow">
                    <k> $PGM:Stmt</k>
                  <counters>
                  <counter multiplicity="*"> .K </counter>
                  </counters> 
                </T>

    rule <k> COUNTER=I; =>. ...</k>
         <counters> CT:Bag =>CT (<counter> I </counter>) </counters>
endmodule
```

works well when being kompiled.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/148313234996210/192842039496672)

cos:
it looks like a bug in concretization. thanks for submitting!

Issue #2057, opened 2016-01-16
I/O does not seem to work properly


grosu:
Somehow, the output buffer does not seem to be properly flushed to the screen:
```
D:\github\grosu\k\k-distribution\tutorial\1_k\4_imp++\lesson_4>kompile imp.k
D:\github\grosu\k\k-distribution\tutorial\1_k\4_imp++\lesson_4>krun ..\lesson_1\tests\io.imp
10
Input two numbers: 20
Their sum is: 30
<T> <k> . </k> <env> x |-> 1 y |-> 2 </env> <store> 1 |-> 10 2 |-> 20 </store> <
in> ListItem ( #buffer ( "" ) ) ListItem ( "on" ) ListItem ( #istream ( 0 ) ) </
in> <out> ListItem ( #ostream ( 1 ) ) ListItem ( "on" ) ListItem ( #buffer ( ""
) ) </out> </T>
D:\github\grosu\k\k-distribution\tutorial\1_k\4_imp++\lesson_4>krun ..\lesson_1\
tests\sum-io.imp


10
Add numbers up to (<= 0 to quit)? 100
Sum = 55
Add numbers up to (<= 0 to quit)? 1000
Sum = 5050
Add numbers up to (<= 0 to quit)? 0
Sum = 500500
Add numbers up to (<= 0 to quit)? <T> <k> halt ; ~> while ( true ) { print ( "Ad
d numbers up to (<= 0 to quit)? " , .Ids ) ; n = read ( ) ; if ( n <= 0 ) { halt
 ; } else { s = 0 ; while ( ! ( n <= 0 ) ) { s = s + n ; n = n + -1 ; } print (
"Sum = " , ( s , ( "\n" , .Ids ) ) ) ; } } </k> <env> s |-> 2 n |-> 1 </env> <st
ore> 1 |-> 0 2 |-> 500500 </store> <in> ListItem ( #buffer ( "" ) ) ListItem ( "
on" ) ListItem ( #istream ( 0 ) ) </in> <out> ListItem ( #ostream ( 1 ) ) ListIt
em ( "on" ) ListItem ( #buffer ( "" ) ) </out> </T>
D:\github\grosu\k\k-distribution\tutorial\1_k\4_imp++\lesson_4>
```
I'm using Windows 10.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/148313234996210/148313972294945)

daejunpark:
[...]

PR #2233, opened 2016-11-27
fix the bug #2213 and #2057, for bug 2057, it will not  appear on Uba…

JoeSun940303:
…ntu VM but will appear on OSX or windows system. We add [read] after rule <k> read and change and change _ to AEs:AExps so that the print can be flushed out both in OSX system and windows system

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/148313234996210/224288619295662)


k-distribution/src/test/java/org/kframework/krun/FlushTest.java
```diff
@@ -0,0 +1,95 @@
...
+                "    }\n" +
+                "    print(\"Sum = \", s, \"\\n\");\n" +
+                "  }\n" +
+                "}\n";
+        parsed = utils.getParsed(pgm, Source.apply("generated by " + getClass().getSimpleName()));
+    }
+
+
+    @Test
+    public void formatTest() throws IOException {
```
cos, line 79: what does this test?

k-distribution/src/test/java/org/kframework/kompile/KompileTest.java
```diff
@@ -0,0 +1,62 @@
...
+    private KRunOptions kRunOptions;
+    private String pgm;
+    private final ByteArrayOutputStream outContent = new ByteArrayOutputStream();
+
+
+    @Rule
+    public ExpectedException thrown= ExpectedException.none();
+
+    @Test
+    public void KompileTest() throws URISyntaxException, IOException, KEMException {
```
cos, line 48: what does this test? please add javadoc

k-distribution/src/test/resources/convertor-tests/PrintFlushTest.k
```diff
@@ -0,0 +1,87 @@
...
+                 | "halt" ";"
+                 > "spawn" Stmt
+                 > Stmt Stmt                  [left]
+
+  syntax Ids   ::= List{Id,","}
+  syntax AExps ::= List{AExp,","}
+endmodule
+
+
+module IMP
```
cos, line 33: Is this the copy-pasted from another place? If so, just refer to that file.

k-distribution/tutorial/1_k/4_imp++/lesson_4/imp.k
```diff
@@ -1,87 +1,93 @@
...
+ syntax Stmt  ::= Block
+                | Id "=" AExp ";"            [strict(2)]
+                | "if" "(" BExp ")"
+                  Block "else" Block         [strict(1)]
+                | "while" "(" BExp ")" Block
+                | "int" Ids ";"
+                | "print" "(" AExps ")" ";"
+                | "halt" ";"
+                > "spawn" Stmt
+                > Stmt Stmt                  [left]
```
cos, line 26: please do not reformat the files if not necessary -- it makes reviewing much harder and breaks file change history

cos (changes requested):

k-distribution/src/test/java/org/kframework/kompile/KompileTest.java
JoeSun940303: Sorry about the confusion, just add javadoc to these two tests

k-distribution/tutorial/1_k/4_imp++/lesson_4/imp.k
JoeSun940303: Sorry about this, when we tried to fix this bug, we add some spaces to the files but it hard for us to reformat it to the former version. The real change is labeled in dark green in the comparison and all other thing are kept the same.

k-distribution/src/test/resources/convertor-tests/PrintFlushTest.k
JoeSun940303: This, this is copied from k-distribution/tutorial/1_k/4_imp++/lesson_4/imp.k, the reason we add this file is that if we want to read .k file by using KoreUtils to test it, seems the file should only be placed in resources folder.

k-distribution/tutorial/1_k/4_imp++/lesson_4/imp.k
cos: I suggest you redo the changes and on a fresh fork of master. I know how github formatting works, but that doesn't help with breaking the change history. Thanks!

k-distribution/src/test/resources/convertor-tests/PrintFlushTest.k
cos: In this case, you should minimize the file to whatever is needed to test the fix.

daejunpark:

PR #2028, opened 2016-01-12
Fixing #1943. Parenthesis only as brackets for K.

radumereuta:
Remove backtick and leave only parenthesis as brackets for sort K.

Do not review yet.

Issue #1948, opened 2015-12-22
wrong rule for initKCell for configurations without $PGM


andreistefanescu:
The following module compile incorrectly, generating the wrong rule for ```initKCell```:
```
require "domains.k"

module TEST
  imports INT
  syntax S ::= S "*" S    [assoc, unit(.S)]
             | ".S"
             | "a"
             | "b"
  syntax KItem ::= foo(S)
  rule <k> 0 => foo(`b * b` * `a * b`) </k>
  rule <k> foo(`X * a` * Y) => 1 </k>

  configuration <k> 0 </k>
endmodule
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/148313234996210/148313972293368)

andreistefanescu:
This does not block the release of the tutorial.

xc93:
Hi, @cos Cosmin, I can't find what the symbol "\`" means in tutorials, so I don't know how to debug this issue. Could you provide me some clues? What does \` a * a \`  for example?

cos:
@hreada, it was a feature of K that we removed in the meantime. Simply replace the ` with parentheses, e.g., 
```
` a * a ` => (a * a)
```

xc93:
@cos As far as I understood, this is an old feature that has been removed from the current K. So how shall I "fix" this bug?

PR #2235, opened 2016-11-28
Revise error message for deprecated(old) style of k definitions.

xc93:
Revising the error message when parsing a K definition that uses old fashion of ' \` ' instead of parenthesis. The new error message looks the following.

[Error] Compiler: Had 2 parsing errors.
[Error] Inner Parser: Parse error: unexpected character '*'.Please use '(' or
')' instead of '\`'
        Source(./test.k)
        Location(10,29,10,30)
[Error] Inner Parser: Parse error: unexpected character '*'.Please use '(' or
')' instead of '\`'
        Source(./test.k)
        Location(11,24,11,25)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/148313234996210/224288625852417)

### DIFF
--- a/kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Parser.java
+++ b/kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Parser.java
@@ -423,23 +423,23 @@ public class Parser {
                 lines[i] = l;
                 columns[i] = c;
                 switch (input.charAt(i)) {
-                    case '\r' :
-                        if (i+1 < input.length()) {
-                            if (input.charAt(i+1) == '\n') {
-                                lines[i+1] = l;
-                                columns[i+1] = c + 1;
-                                i++;
-                            }
+                case '\r' :
+                    if (i+1 < input.length()) {
+                        if (input.charAt(i+1) == '\n') {
+                            lines[i+1] = l;
+                            columns[i+1] = c + 1;
+                            i++;
                         }
-                    case '\n'      :
-                    case  '\u000B' :
-                    case  '\u000C' :
-                    case  '\u0085' :
-                    case  '\u2028' :
-                    case  '\u2029' :
-                        l++; c = 1; break;
-                    default :
-                        c++;
+                    }
+                case '\n'      :
+                case  '\u000B' :
+                case  '\u000C' :
+                case  '\u0085' :
+                case  '\u2028' :
+                case  '\u2029' :
+                    l++; c = 1; break;
+                default :
+                    c++;
                 }
             }
             lines[input.length()] = l;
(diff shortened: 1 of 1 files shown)


kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Parser.java
```diff
@@ -574,9 +574,11 @@ public Term parse(NonTerminal nt, int position) {
             CharSequence content = s.input;
             ParseError perror = getErrors();
 
+            //Here we fix the error message for bug #1948
             String msg = content.length() == perror.position ?
                     "Parse error: unexpected end of file." :
                     "Parse error: unexpected character '" + content.charAt(perror.position) + "'.";
+            if (content.toString().indexOf('`')>=0) msg += "Please use '(' or ')' instead of '`'";
```
radumereuta: This can actually fail badly, and the error message can be misleading in some cases.
If the language actually includes a terminal backtick, and the user has a proper error at that location, now K would suggest using parenthesis instead of actually fixing the problem.

radumereuta:
The backtick modifications started after this issue was reported: #1970 and #2028.
I think for #1948 we should just manually fix the backticks and continue debugging from there.

Also, when doing pull request, try to keep the original alignment, because it makes code reviews a lot harder.

xc93:
@radumereuta I changed backtick to parenthesis and everything works fine under the latest version of K. So what is the real problem here?

cos:
I agree with @radumereuta. The backticks haven't been in use long enough to warrant a special error message. The alternate "fix" for this is to go through the existing definitions and check whether there are any leftover backticks. If there are none -- likely -- great. If there are some, we should simply replace them with parenthesis.

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Parser.java
```diff
@@ -668,22 +670,22 @@ private boolean finishStateReturn(StateReturn stateReturn) {
...
-                    stateReturn.key.stateEnd)).function);
+                    stateReturn.key.stateCall.function,
+                    s.stateReturns.get(new StateReturn.Key(
+                            s.stateCalls.get(new StateCall.Key(
+                                    s.ntCalls.get(new NonTerminalCall.Key(
+                                            ((Grammar.NonTerminalState) stateReturn.key.stateCall.key.state).child,
+                                            stateReturn.key.stateCall.key.stateBegin)),
+                                    stateReturn.key.stateEnd,
+                                    ((Grammar.NonTerminalState) stateReturn.key.stateCall.key.state).child.exitState)),
+                            stateReturn.key.stateEnd)).function);
```
cos, line 688: please do not reformat the files.

kernel/src/main/java/org/kframework/parser/concrete2kore/kernel/Parser.java
cos: maybe rephrase as: "Using backtick, '`', for K brackets is deprecated. If you are using it for this purpose, please use regular parenthesis, '(' and ')', instead."

cos (changes requested):

daejunpark:

PR #2241, opened 2016-11-30
Changed error message for issue 2095

brayou:
Copied Throwable stack trace but with new error message if the Throwable error message contains "Can't mix cells at different levels under a rewrite"
New error message is: "Critical: Incomplete rule assignment"

Done with partner @xianzezhan

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/148313234996210/224288562958499)

### DIFF
--- a/kore/src/main/scala/org/kframework/definition/transformers.scala
+++ b/kore/src/main/scala/org/kframework/definition/transformers.scala
@@ -52,8 +52,8 @@ object ModuleTransformer {
   }
 }
 
-class ModuleTransformerException(moduleTransformerName: String, cause: Throwable) extends Exception(cause) {
-  override def toString = moduleTransformerName + " : " + cause.toString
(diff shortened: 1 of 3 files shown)


kore/src/main/scala/org/kframework/definition/transformers.scala
```diff
@@ -66,7 +66,23 @@ abstract class ModuleTransformer extends (Module => Module) {
     f
   } catch {
     case e: ModuleTransformerException => throw e
-    case e: Throwable => throw new ModuleTransformerException(name, e)
+    case e: Throwable => {
+      if (e.toString().contains("Can't mix cells")) {
```
cos, line 70: Adjusting the exception based on the exception's description string is not a fix for this issue. Try to trace the exception to its source, see why the original message was wrong, and adjust there to throw the correct exception.

Also, this file contains generic transformer code, so it should not contain exception details regarding a particular pass.

cos (changes requested):

daejunpark:

PR #2243, opened 2016-12-02
Bugfix2219

HelenRouty:
The bug is: when using --no-prelude to generate compiledDefinition for an empty module, there will be a NoSuchElementException. The is due to a lack of checks for the Optional (Java) or Option(scala) types before extracting values from them. To fix the bug, we created an empty-module for unit test and added checks for Optional and Options.  After adding checks, we were able to create a CompiledDefinition with its topCellInitializer of null value.

To further avoid Exceptions when running Krun program caused by having no cells in compiledDefinition, we passed the noPrelude flag from kRunOption.output to CompiledDefinition. Then we disabled definition parsing and rename anonymous variables when generating K program result for the empty-module. Meanwhile, we added more checks to avoid exceptions for null value K and null KApply's klable value.

Finally, we added test cases for non-empty module without --no-prelude parameter for regression test. We also added more test cases for kast program, KAST and BINARY pretty print modes.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/148313234996210/224288564753197)


.idea/codeStyleSettings.xml
```diff
@@ -1,29 +0,0 @@
...
-          <option name="BLANK_LINES_AROUND_METHOD" value="0" />
-          <option name="BLANK_LINES_AROUND_METHOD_IN_INTERFACE" value="0" />
-          <option name="KEEP_SIMPLE_BLOCKS_IN_ONE_LINE" value="true" />
-        </codeStyleSettings>
-      </value>
-    </option>
-    <option name="USE_PER_PROJECT_SETTINGS" value="true" />
-  </component>
-</project>
-
```
cos, line 29: please don't delete this file

kernel/src/main/java/org/kframework/kompile/DefinitionParsing.java
```diff
@@ -292,29 +290,29 @@ private Rule upRule(K contents) {
...
+            case "#ruleNoConditions":
+                return Rule(items.get(0), BooleanUtils.TRUE, BooleanUtils.TRUE, ruleContents.att());
+            case "#ruleRequires":
+                return Rule(items.get(0), items.get(1), BooleanUtils.TRUE, ruleContents.att());
+            case "#ruleEnsures":
+                return Rule(items.get(0), BooleanUtils.TRUE, items.get(1), ruleContents.att());
+            case "#ruleRequiresEnsures":
+                return Rule(items.get(0), items.get(1), items.get(2), ruleContents.att());
+            default:
+                throw new AssertionError("Wrong KLabel for rule content");
```
cos: please don't reformat the file

kore/src/main/scala/org/kframework/compile/ConfigurationInfoFromModule.scala
```diff
@@ -69,8 +69,11 @@ class ConfigurationInfoFromModule(val m: Module) extends ConfigurationInfo {
   lazy val topCell: Sort = {
     if (topCells.size > 1)
       throw new AssertionError("Too many top cells:" + topCells)
-
-    topCells.head
+    // we should check whether topCell is empty before using it
+    if (topCells.size != 0)
+      topCells.head
+    else
+      null
```
cos: you could return the `K` sort here instead.

cos (changes requested):
This is nice work and shows a good understanding of the problem. Still, I think there is a way to handle this without propagating null checks all over the codebase. Instead of allowing `null`, use a new, made up, placeholder term. For the sort, see the comment below.

HelenRouty:
@cos  Thank you for your suggestions. Do you mean that we can create a placeholder Sort: 
E.g. `val KNull  = Sort("KNull")` as a built-in Sort in Sorts.scala and use Sorts.KNull to replace all null value check for Sort? 
Should we have null value placeholders for KLable and KApply as well?

cos:
You can use the top sort, i.e., `K`, as a "null" sort. For terms, yes, I'm suggesting to add a KLabel, named something like "<empty>", and make a term of arity 0 from it, i.e., `KApply(KLabel("<empty>"), List())` .

HelenRouty:
@cos Thank you. So we should use Sorts.K to replace "null" sort right?
Could we use the KORE.KLabel("") for "null" KLabel?

cos:
> So we should use Sorts.K to replace "null" sort right?

Yes.

> Could we use the KORE.KLabel("") for "null" KLabel?

No, use something like "<empty>", obtained by adding a production to the definition.

HelenRouty:
Hi @cos. Thank you for providing more details. However, I still have a hard time to understand what type of KLabel should be used. Could we schedule an appointment with you at any time on Dec.12(Mon) after 5:00pm or any time after Dec.12?

cos:
We can meet today at 5pm if you like.

HelenRouty:
@cos Thank you.

HelenRouty:
Thanks to @cos explanation, we found that we misunderstood the root cause of this bug. A better description and fix are in this newer pull request: [https://github.com/kframework/k/pull/2249](https://github.com/kframework/k/pull/2249).

PR #2244, opened 2016-12-06
Bypass the type check when parsing programs. 

radumereuta:
Bypass the type check when parsing programs. 
That's only needed when parsing rules.


kernel/src/main/java/org/kframework/parser/concrete2kore/ParseInModule.java
```diff
@@ -79,13 +78,26 @@ public Module getExtensionModule() {
...
+
+    /**
+     * Parse as input the given string and start symbol using the module stored in the object.
+     * All the type checks are omitted when parsing programs.
+     * @param input          the string to parse.
+     * @param startSymbol    the start symbol from which to parse.
+     * @return the Term representation of the parsed input.
+     */
+    public Tuple2<Either<Set<ParseFailedException>, K>, Set<ParseFailedException>>
+            parseProgramString(String input, Sort startSymbol, Source source) {
```
cos: rename to `parseStringWithoutTypecheck`, so that we don't mention "Program" in code which should be agnostic to what it parses?

cos (changes requested):

cos (approved):

PR #2246, opened 2016-12-09
fix issue 2215

hksskh:
fix the issue 2215 by adding boolean variable to iteratively record if there is any cell that is not leaf cell among all child cells of current cell. major modification in function genInternal in class GenerateSentencesFromConfigDecl.
pass the ConfigurationInheritanceTest mentioned in issue 1918 by @bmmoore, and also pass other tests.
stil fail in one assertion:
in class GenerateSentencesFromConfigDecl, function computeSentencesOfWellFormedCell, line 274, the "assert isLeaf" fails. Don't quite understand why this assertion exists. simply comment this assertion and pass all other tests.


kernel/src/main/java/org/kframework/kore/compile/GenerateSentencesFromConfigDecl.java
```diff
@@ -257,7 +271,8 @@ public K apply(KToken k) {
         Sort sort = Sort(sortName);
 
         if (cellProperties.contains("maincell")) {
-            assert isLeaf;
+            //assert isLeaf;
+            isLeaf = true;
```
cos: the `maincell` is contains a KSequence(`~>`), which is not part of the configuration, thus the cell should be a leaf.

kernel/src/main/java/org/kframework/kore/compile/GenerateSentencesFromConfigDecl.java
```diff
@@ -77,7 +78,7 @@
      * @param m The module the configuration declaration is in. Used to get the sort of leaf cells.
      * @return A tuple of the sentences generated, a list of the sorts of the children of the cell, and the body of the initializer.
      */
-    private static Tuple3<Set<Sentence>, List<Sort>, K> genInternal(K term, K ensures, Att cfgAtt, Module m) {
+    private static Tuple4<Set<Sentence>, List<Sort>, K, Boolean> genInternal(K term, K ensures, Att cfgAtt, Module m) {
```
cos, line 84: what does the `Boolean` encode? maybe explain in the javadoc above

cos (changes requested):

hksskh:
add new commit, please review it, thanks. @cos

daejunpark:

PR #2247, opened 2016-12-10
feature/better start performance

cos:
Before, executing sum upto 10 (essentially an empty program) on my laptop took, on average, 8.1s.

The main performance culprits are/were:

1. We were parsing programs by calling `kast` externally (it's Dwight's love for command line interfaces). This created a whole new, cold JVM just for this. Replacing it with same-JVM parsing dropped the time to 4.7s (40% improvement)

2. Scala's reflection mechanism is very fast in "server" mode but it seems to take a while to warm up -- my bad here, I didn't know. Replaced it with plain Java reflection and the time dropped to 3.7s (20% improvement)


cos:
@radumereuta or @daejunpark, could you have a quick look and merge?

daejunpark (approved):

PR #2249, opened 2016-12-14
Bugfix2219new

HelenRouty:
## Bug Description
A NoSuchElementException occurs when using --no-prelude to generate compiledDefinition for an empty module definition file (no syntax but possibly rules and configurations). E.g.
`kompile --no-prelude --debug empty-module.k`
This bug is due to the problem that "DEFAULT-DEFINITIONS", the built-in modules in k-distribution's kast.k and domains.k are **NOT LOADED**. 

Note that --no-prelude flag is intended to **NOT IMPORT** domains.k but **SHOULD LOAD** the domains.k. As parsing configurations requires many modules in domains.k, e.g. Map module, we would better **LOAD** all the built-in modules (kast and domain), and simply **NOT IMPORT** them.

## Change Log
1. ParserUtils.loadDefinition(), load Kompile.REQUIRE_PRELUDE_K to previousModules even with --no-prelude flag. This change fixes the bug2219.
2. KDepFrontEnd.run(), load Kompile.REQUIRE_PRELUDE_K to modules even with --no-prelude flag. This change is to keep the meaning of --no-prelude flag consistent throughout kframework: load prelude.k to modules but not import into k compiled definition.
3. change the misleading names of no-prelude (and noprelude) to notAutoImportDomain

## Tests
Create unit tests for kompile and kdep empty module in k-distribution: TstKompileOnKORE_IT.java and TstKdepOnKORE_IT.java

### DIFF
--- a/k-distribution/src/test/java/org/kframework/kdep/TstKdepOnKORE_IT.java
+++ b/k-distribution/src/test/java/org/kframework/kdep/TstKdepOnKORE_IT.java
@@ -44,7 +44,7 @@ public class TstKdepOnKORE_IT {
     /* *
      * kdepEmptyModuleTest runs the regression test for
      * `kdep --no-prelude k-distribution/src/test/resources/compiler-tests/empty-module.k`
-     * The empty-module.k is created with no imports, syntax, rules, and configuration in it.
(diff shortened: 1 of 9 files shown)


daejunpark:

PR #2250, opened 2016-12-29
Feature/kore of kore

daejunpark:
Do not merge. Just for a quick review.
@cos could you take a look to see if there is any serious mistake? Please note (and ignore) that I refactored the code you wrote in a way of minimizing OOP and Scala-specific features.


kore/src/main/scala/org/kframework/minikore/KoreToMini.scala
```diff
@@ -0,0 +1,130 @@
...
+
+import scala.collection.JavaConverters._
+import scala.collection._
+
+object KoreToMini {
+
+  // Outer
+
+  def apply(d: definition.Definition): Definition = {
+    var modules = d.modules.toSeq.map(apply)
```
daejunpark: @cos is it safe to use `modules` field to get all (resolved) modules?

kore/src/main/scala/org/kframework/minikore/KoreToMini.scala
```diff
@@ -0,0 +1,130 @@
...
+      Term("RegexTerminal", Seq(S(precedeRegex), S(regex), S(followRegex)))
+  }
+
+  def encode(s :definition.Sentence): Sentence = {
+    val att = s match {
+      case definition.ModuleComment(comment, att) => Term("ModuleComment", Seq(S(comment))) +: apply(att)
+      case definition.SyntaxPriority(priorities, att) => Seq() // TODO(Daejun): implement
+      case definition.SyntaxAssociativity(assoc, tags, att) => Seq() // TODO(Daejun): implement
+      case definition.Bubble(_, _, _) => Seq() // TODO(Daejun): find why it appears here
+      case definition.Context(_, _, _) => Seq() // TODO(Daejun): find why it appears here
```
daejunpark: @cos any idea why `Bubble` and `Context` appear here? For example, Bubble appears in DEFAULT-CONFIGURATION in kast.k, and Context appears in IMP.

kore/src/main/scala/org/kframework/minikore/KoreToMini.scala
```diff
@@ -0,0 +1,130 @@
...
+  def apply(k: K): Pattern = {
+    val p = k match {
+      case KApply(klabel, klist) =>
+        Term(klabel.name, klist.map(apply))
+      case kvar @ SortedKVariable(name, att) =>
+        assert(att == k.att)
+        Variable(name, kvar.sort.name)
+      case KVariable(name) =>
+        // TODO(Daejun): may need to distinguish against SortedKVariable
+        apply(SortedKVariable(name, k.att)) // from SortedADT in ADT.scala
```
daejunpark: @cos is it correct?

kore/src/main/scala/org/kframework/minikore/KoreToMini.scala
```diff
@@ -0,0 +1,130 @@
...
+  //
+  // NOTE: Since interface.scala is defined using traits,
+  // its companion object (Unapply.scala) is provided to support pattern matching.
+
+  def apply(k: K): Pattern = {
+    val p = k match {
+      case KApply(klabel, klist) =>
+        Term(klabel.name, klist.map(apply))
+      case kvar @ SortedKVariable(name, att) =>
+        assert(att == k.att)
```
daejunpark: @cos does it always hold?

kore/src/main/scala/org/kframework/minikore/KoreToMini.scala
```diff
@@ -0,0 +1,130 @@
...
+  def encodePatternAtt(p: Pattern, att: Att): Pattern = {
+    att.foldLeft(p)((z,a) => {
+      Term("#", Seq(z,a))
+    })
+  }
+
+  // encodeKSeq(Seq(p1,p2,p3,p4)) = #kseq(p1,#kseq(p2,#keq(p3,p4))) // TODO(Daejun): add test
+  def encodeKSeq(ps: Seq[Pattern]): Pattern = {
+    ps.reverse match {
+      case Seq(p) => p // NOTE(Daejun): `case p :: Nil` doesn't work (match failure). why??
```
daejunpark: @cos any idea of why `case p :: Nil` failed to match? what's the difference between `p :: Nil` and `Seq(p)`? just a type inference issue?

kore/src/main/scala/org/kframework/minikore/KoreToMini.scala
cos: yes, it should be the same object

kore/src/main/scala/org/kframework/minikore/KoreToMini.scala
```diff
@@ -0,0 +1,151 @@
...
+    case definition.Terminal(value, followRegex) =>
+      Term(iTerminal, S(value) +: followRegex.map(s => S(s)))
+    case definition.RegexTerminal(precedeRegex, regex, followRegex) =>
+      Term(iRegexTerminal, Seq(S(precedeRegex), S(regex), S(followRegex)))
+  }
+
+  def encode(s :definition.Sentence): Sentence = {
+    val att = s match {
+      case definition.ModuleComment(comment, att) => Term(iModuleComment, Seq(S(comment))) +: apply(att)
+      case definition.SyntaxPriority(priorities, att) => Seq() // TODO(Daejun): implement
```
cos: it's safer to use `???` for unimplemented parts

kore/src/main/scala/org/kframework/minikore/KoreToMini.scala
cos: it might be better to handle the variable directly, including getting its sort from attributes, instead of creating a sorted one

kore/src/main/scala/org/kframework/minikore/KoreToMini.scala
cos: no idea -- it also looks to me like it should match, but I generally don't use the `::` operator, so don't have experience with it

cos (approved):
looks good

Issue #2273, opened 2017-02-09
OneIntegerGenericBitSet.clear is not implemented


kheradmand:
I got the following assertion : 
```
java.lang.AssertionError: unimplemented
	at org.kframework.utils.OneIntegerGenericBitSet.clear(OneIntegerGenericBitSet.java:183)
	at org.kframework.backend.java.symbolic.FastRuleMatcher.addUnification(FastRuleMatcher.java:515)
	at org.kframework.backend.java.symbolic.FastRuleMatcher.match(FastRuleMatcher.java:228)
	at org.kframework.backend.java.symbolic.FastRuleMatcher.matchAssoc(FastRuleMatcher.java:400)
	at org.kframework.backend.java.symbolic.FastRuleMatcher.matchAssoc(FastRuleMatcher.java:441)
	at org.kframework.backend.java.symbolic.FastRuleMatcher.match(FastRuleMatcher.java:323)
	at org.kframework.backend.java.symbolic.FastRuleMatcher.match(FastRuleMatcher.java:271)
	at org.kframework.backend.java.symbolic.FastRuleMatcher.match(FastRuleMatcher.java:315)
	at org.kframework.backend.java.symbolic.FastRuleMatcher.match(FastRuleMatcher.java:315)
	at org.kframework.backend.java.symbolic.FastRuleMatcher.matchRulePattern(FastRuleMatcher.java:107)
	at org.kframework.backend.java.symbolic.SymbolicRewriter.fastComputeRewriteStep(SymbolicRewriter.java:164)
	at org.kframework.backend.java.symbolic.SymbolicRewriter.computeRewriteStep(SymbolicRewriter.java:105)
	at org.kframework.backend.java.symbolic.SymbolicRewriter.rewrite(SymbolicRewriter.java:86)
	at org.kframework.backend.java.symbolic.InitializeRewriter$SymbolicRewriterGlue.execute(InitializeRewriter.java:132)
	at org.kframework.krun.modes.KRunExecutionMode.execute(KRunExecutionMode.java:65)
	at org.kframework.krun.KRun.run(KRun.java:95)
	at org.kframework.krun.KRunFrontEnd.run(KRunFrontEnd.java:68)
	at org.kframework.main.FrontEnd.main(FrontEnd.java:34)
	at org.kframework.main.Main.runApplication(Main.java:420)
	at org.kframework.main.Main.runApplication(Main.java:271)
	at org.kframework.main.Main.main(Main.java:84)
```

[...]

PR #2275, opened 2017-02-10
remove unsupported map builtin function

daejunpark:


daejunpark:
@kheradmand please review and approve this pull request.

dwightguth:
FYI the C semantics uses this heavily for performance reasons.

kheradmand:
@daejunpark 
maybe it would be a good idea just to comment it instead of removing. 

@dwightguth 
the feature is not implemented anyway (I guess), please look at #2273 and the problems it caused for me.

daejunpark:
@dwightguth thanks for your comment. @kheradmand Now I keep `in_keys` in domains.k, but make java backend to throw exception when it is used.

daejunpark:
@kheradmand could you approve this? You can click the link above and check a radio box to approve it.

kheradmand (approved):
lgtm

PR #2295, opened 2017-04-19
Refactoring

msaxena2:
This PR achieves two goals - 
- Renames everything old/frontend kore related to legacy-kore.
- Makes K work with the new Kore (mini), which is referred to as kore.

There is now a complete separation. The old/frontend kore is legacy-kore, and the new/mini kore is just kore.


cos (changes requested):
I am only ok with changing the name to "legacykore" if there is a commitment to migrate all the code in the next few months -- i.e., make the frontend/legacykore/whatever data structures implement the new KORE interface. 

Otherwise, we end up in a situation where most of our compiler pipeline is deemed "legacy" because we have a new interface with 0 code written using it. The term "legacy" will discourage people from maintaining that code while there is no replacement code using the new interface.

The better course of action would probably be to rename this project to "frontend" and move the code from `kernel` which deals with compiler passes here. The kernel has lots of stuff we might not want to deal with, while `kore` is just data structures and Scala-implemented compiler passes.

PR #2324, opened 2017-07-24
Verification Related Fixes

msaxena2:
Helpful Debugging Info for `z3` + Missing Case Fix for Fast Rule Matcher.

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/symbolic/FastRuleMatcher.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/symbolic/FastRuleMatcher.java
@@ -329,6 +329,8 @@ public class FastRuleMatcher {
             return subject.equals(pattern) ? ruleMask : empty;
         } else if (subject instanceof BuiltinSet && pattern instanceof BuiltinSet) {
             return subject.equals(pattern) ? ruleMask : empty;
+        } else if (subject instanceof KList && pattern instanceof KList) {
+            return subject.equals(pattern) ? ruleMask : empty;
         } else {
             assert subject instanceof KItem || subject instanceof BuiltinList || subject instanceof Token || subject instanceof BuiltinMap : "unexpected class at matching: " + subject.getClass();
             assert pattern instanceof KItem || pattern instanceof BuiltinList || pattern instanceof Token : "unexpected class at matching: " + pattern.getClass();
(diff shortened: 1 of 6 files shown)


msaxena2:
@nishantjr - Your commits are here too. Please take a look.

msaxena2:
@cos @andreistefanescu Please review. The second last commit (abe14ab ) fixes issues with unexpected Z3 queries in the prover.

cos (approved):
lgtm but @andreistefanescu should take a look as well

java-backend/src/main/java/org/kframework/backend/java/symbolic/FastRuleMatcher.java
```diff
@@ -326,6 +327,10 @@ private BitSet match(Term subject, Term pattern, BitSet ruleMask, scala.collecti
         } else if (subject instanceof Token && pattern instanceof Token) {
             // TODO: make tokens unique?
             return subject.equals(pattern) ? ruleMask : empty;
+        } else if (subject instanceof BuiltinSet && pattern instanceof BuiltinSet) {
+            return subject.equals(pattern) ? ruleMask : empty;
+        } else if (subject instanceof KList && pattern instanceof KList) {
```
andreistefanescu: I don't think this `match` was ever supposed to be called with `KList` arguments...

andreistefanescu:
@msaxena2 your fix was a good attempt, but this should set the proper constraint just before use. Please review.

java-backend/src/main/java/org/kframework/backend/java/symbolic/FastRuleMatcher.java
msaxena2: @andreistefanescu It seems like EVM verification doesn't go through without this. Also, this may have broken the tests, but I can't see why - from the error messages. I haven't been able to come up with a reasonable fix for this. Any chance you can take a look?

grosu: So, @msaxena2 , are you saying that we can still not have one version of the semantics that passes all the tests and can be used for the verification tasks we planned for tomorrow's technical report?

I agree, intuitively, with @andreistefanescu that the match function should be called on proper terms only, not on `KList`s.  So instead of adding a case here, you should probably go back to where the match function is called and see why it was called on `KList` terms first place, shouldn't you?

msaxena2:
@cos @andreistefanescu I was able to get around the issue by slightly changing the semantics of EVM (one of the rules that had an `owise` attribute was causing the issue). I think we should merge these changes, and deal with issues with `owise` in a separate PR - using a much simpler definition than EVM.

PR #2329, opened 2017-08-17
Adding option to kast to keep remaining ambiguities after parsing.

ehildenb:
Old branch, let's get merged into master.

@radumereuta can you double-check that it still makes sense?

Also, can you add a test that ambiguities are being kept correctly?


radumereuta (approved):
@ehildenb I added a test, as you requested. All tests seem to pass now. The first time I committed this, tests failed randomly, so I didn't create a pull request, but apparently, the flaky tests got fixed in the mean time.

ehildenb:
@radumereuta I rebased on master to keep the history cleaner (as opposed to merging into master). Looks good to me, thanks for adding the test. @msaxena2 can you take a quick look as well then merge if you also think so?

ehildenb:
Hmmm, it would appear we have some flaky tests. I checked after rebasing that the diff between your version with the merge and this version was empty.

radumereuta:
We have this issue for a long time now. Usually, we just restart the build until it passes.

msaxena2:
I think the flakiness arises from codeship's load balancing - some tests non deterministically time out. This problem might go away after switching to our own dedicated server.

On 5 Sep 2017 11:30, Radu Mereuta <notifications@github.com> wrote:

We have this issue for a long time now. Usually, we just restart the build until it passes.

msaxena2:
Will take a look and merge before the end of the day today.

On 5 Sep 2017 10:05, ehildenb <notifications@github.com> wrote:

@radumereuta<https://github.com/radumereuta> I rebased on master to keep the history cleaner (as opposed to merging into master). Looks good to me, thanks for adding the test. @msaxena2<https://github.com/msaxena2> can you take a quick look as well then merge if you also think so?

kernel/src/test/java/org/kframework/parser/concrete2kore/RuleGrammarTest.java
```diff
@@ -451,14 +451,29 @@ public void test25() {
...
+        String def = "" +
+                "module TEST " +
+                "syntax One ::= \"1\" " +
+                "syntax Two ::= \"1\" " +
+                "syntax Exp ::= Two | One " +
+                "endmodule";
+        parseProgram("1", def, "Exp", 0,
+                KApply(KLabel("AMB"), KApply(KLabel("1")), KApply(KLabel("1"))),
+                true);
+
```
nishantjr, line 477: Should there be some sort of assertion here?

kernel/src/main/java/org/kframework/kast/KastOptions.java
```diff
@@ -84,6 +84,9 @@ public Sort convert(String arg) {
     @Parameter(names={"--module", "-m"}, description="Parse text in the specified module. Defaults to the syntax module of the definition.")
     public String module;
 
+    @Parameter(names={"--keepAmb", "-amb"}, description="Keep remaining ambiguities after applying disambiguation filter.")
```
nishantjr: Maybe `--keep-ambiguous` would be more standard?

nishantjr (commented):
Couple of minor improvements I think should go in.

kernel/src/test/java/org/kframework/parser/concrete2kore/RuleGrammarTest.java
ehildenb: `parseProgram` performs a check. Look at the other tests in the file.

kernel/src/main/java/org/kframework/kast/KastOptions.java
ehildenb: I agree that `--keepAmb` could be improved, but I think `--keep-ambiguities` would be better than `--keep-ambiguous`

msaxena2:
I've changed the `--keepAmb` to` --keep-ambiguites` by ammending the commit that originally introduced to the `--keepAmb` option. Lgtm otherwise - I'll merge tomorrow morning if tests pass.

PR #2330, opened 2017-08-17
Replace "kale" with "skala" as an argument to --backend

ehildenb:


ehildenb:
@cos  can you make sure this is still correct and lets get it merged?

cos (approved):
if easy, let's call it "scala" instead of "skala"

lucaspena:
We can change it from `skala` to `scala` here, but this looks like just a comment. It's not clear exactly what we'd have to change in order to correctly rename it to `scala`. For example, are there other repositories like `kale` or `kore` that depend on the name `skala`? Feel free to open up another pull request and change it to `scala` everywhere needed, but if it's ok with you we'd like to just get this merged as is.

ehildenb:
Merging, we can re-name it again later if needed. It looks like renaming it will require more work than just changing things in this repository.

PR #2331, opened 2017-08-17
fix freezerLabel proper sort

ehildenb:
@daejunpark we resurrected this branch from many months ago.

Can you make sure it's still relevant and makes sense?

If so we can add some tests and get it merged.


msaxena2:
@daejunpark @ehildenb Is this ready for review? I guess this isn't really blocking, but it'd be good to get this in.

ehildenb:
I've rebased on `master`. We need tests before this can be merged.

PR #2334, opened 2017-08-23
add ecdsa public key recovery function to KRYTPO module

dwightguth:


dwightguth:
Please don't merge this yet, it's computing incorrect senders. I will investigate further tomorrow.

dwightguth:
Okay, this is ready for review. I tested with the evm semantics. There seems to be another bug somewhere but it's unrelated to the crypto built-in. I'll investigate tomorrow.

msaxena2 (commented):
@dwightguth Could you add a test to BuiltinCryptoOperationsTest.java for this?

dwightguth:
I can do that tomorrow, yeah, but I'm not entirely sure what good it will do because I will have to derive the test input and outputs from running the test code in a debugger and inspecting the outputs, in which case I'm not really testing anything. I can tell you that the code correctly computes the senders of transactions in the evm semantics though.

msaxena2:
@dwightguth Is it possible to find a standard set of values that we can test against? Like some documentation with an example online?

dwightguth:
I'll look tomorrow, maybe ethereumj has tests.

dwightguth:
@msaxena2 I added a test for public key recovery by manually translating the test inputs and outputs for a test that came from ethereumj into K. Can we get this merged now that it's tested and documented?

msaxena2:
@dwightguth Lgtm. Will merge as soon as tests pass.

msaxena2 (approved):

PR #2342, opened 2017-09-13
add ripemd160 hash to KRYPTO module

dwightguth:
This is needed for the RIPEMD160 precompiled contract in the EVM semantics.


k-distribution/tests/builtins/ripemd160.crypto
```diff
@@ -0,0 +1,2 @@
+ripemd160DigestTest("")
+ripemd160DigestTest("foo")
```
msaxena2: Is it possible to add the sources that w can refer to - like a link to a document with examples? (like in the other test files)

msaxena2 (commented):
Looks good apart from the one minor comment. Thanks @dwightguth for fixing the documentation for the other functions.

msaxena2:
Also, can you squash the two commits into one. Seems like the second commit isn't necessarily an entity on its own.

dwightguth:
@msaxena2 please re-review

msaxena2 (approved):

PR #2344, opened 2017-09-22
parent/pom.xml: Enforce Plugin Version 1.4.1 => 3.0.0-M1

msaxena2:
The old version of the plugin had issues dealing with
versions of Java outside the range 1.8 range. The update
fixes the issue.


nishantjr:
Can you get codeship to rebuild? I don't seem to have permission.

nishantjr (approved):

PR #2346, opened 2017-10-05
add StringBuffer sort

dwightguth:
Note that this does not actually provide linear-time string concatenation; it is added for compatibility with RV-K, which has a linear-time implementation of StringBuffer.


cos (approved):

daejunpark (approved):

daejunpark:
@dwightguth This wasn't written having the verification in mind, right? In that case, it looks good to me.

PR #2348, opened 2017-10-18
add catch block for invalid point compression

dwightguth:


java-backend/src/main/java/org/kframework/backend/java/builtins/BuiltinCryptoOperations.java
```diff
@@ -94,7 +94,7 @@ public static StringToken ecdsaRecover(StringToken messageHash, IntToken v, Stri
         try {
             ECDSARecover key = ECDSARecover.signatureToKey(hashBytes, rBytes, sBytes, vByte);
             return StringToken.of(Arrays.copyOfRange(key.getPubKey(), 1, 65));
-        } catch (SignatureException e) {
```
msaxena2, line 97: @dwightguth Can't seem to remember, but why do we not want to use K Exception Manager to record this as a critical error and quit? So the function will return an empty Token in case of a recovery failure right? I'm ok with this behavior, but could you also update the `@return` attribute to say that an empty Token is returned in case of recovery failure?

msaxena2 (approved):

PR #2349, opened 2017-10-20
ARRAY module in UIUC K

dwightguth:


ehildenb:
Please add tests.

ehildenb:
What should the semantics of `makeArray(5, 0)[7]` be? That is, should it return the default value `0`, or some error term because `7` it outside the bounds of the array?

dwightguth:
`makeArray(5, 0)[7]` and `makeArray(5, 0)[7 <- 1][7]` should both be zero. The idea behind ARRAY is that an efficient implementation backs an array as an actual array with a fixed length, so any element not explicitly assigned or outside the length of the array is always equal to the default value.

k-distribution/include/builtin/domains.k
```diff
@@ -20,6 +20,38 @@ module DOMAINS
   imports SET
 endmodule
 
+module ARRAY
+  imports MAP
+  imports INT
+
+  syntax Array ::= array(Map, Int, K)
```
bmmoore, line 27: Is there any way to stop this from being directly called to create a value with keys set outside the expected range?

bmmoore (changes requested):
The array code looks good.

My only complaint is that the PR also moves the crypto tests into their own `tests/builtins/crypto` directory. That's a good change, but entirely unrelated to arrays and also unmentioned in the commit messages. Perhaps it belongs in #2350 instead? If you don't mind I can move that change into that PR instead and approve this.

bmmoore (approved):

bmmoore:
As @dwightguth pointed out I missed that we only support one test definition per directory, so making a subdirectory for the tests previously directly in `test/builtins` is a good idea before making `test/builtins/array`.

PR #2350, opened 2017-10-20
bn128 elliptic curve builtins in k

dwightguth:
For the EVM semantics of Byzantium.


daejunpark (approved):

daejunpark:
I'm ok to merge this. I haven't carefully read the actual implementation of the fields and curves, but it should be ok to add it for now as a builtin, since we're not going to symbolically reason about the crypto algorithms in the near future.

daejunpark:
@bmmoore Are you reviewing (or will you review) this? Otherwise, I'd like to merge this.

bmmoore (commented):
This is fine. Most of the code is just imported from ethereumJ. Having the middle commit refer to hook names before the last commit adds them is a bit out of order, but everything is included in the end.

PR #2355, opened 2017-10-29
Fix for #configuration builtin

theo25:


daejunpark:
@theo25 let's see if it passes the tests:
https://github.com/kframework/k/commits/theo-reflection-fix

daejunpark:
@theo25 says:

> So, TermContext has a private field called topTerm that was never set and kept its default null value. The #configuration builtin attempts to access this field to get the current configuration but since it’s null, a NullPointerException is thrown and never gets caught. So I added the correct (I think) initialization of the field when a ConstrainedTerm is built out of a Term and a TermContext.
>
> Also, I have changed the implementation of the builtin. I am not sure what a KLabelInjection is but it did not return a term with a KLabel corresponding to the top cell of the configuration (e.g. “\<T\>”). Instead it was returning term with a KLabel that somehow was made out of the whole configuration (e.g. “#\<T\>(\<k\> ...)”). I replaced that with simply returning the Term stored in the topTerm field of the context.

daejunpark:
The jenkins failure seems to be due to ng setup failure. @ehildenb could you confirm that we can ignore this failure to merge this?

ehildenb:
I'm looking into it, will try to re-trigger the test with corrected settings.

theo25:
@daejunpark FYI, I tried this fix on top of keq with our equivalence checking examples and everything works fine.

daejunpark (approved):

java-backend/src/main/java/org/kframework/backend/java/builtins/MetaK.java
```diff
@@ -200,6 +200,7 @@ public static KItem getKLabel(KItem kItem, TermContext context) {
     }
 
     public static Term configuration(TermContext context) {
-        return KLabelInjection.injectionOf(context.getTopTerm(), context.global());
+        //return KLabelInjection.injectionOf(context.getTopTerm(), context.global());
+        return context.getTopTerm();
```
daejunpark, line 204: `KLabelInjection` is somewhat legacy and kind of deprecated. It should be OK to drop.

java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
```diff
@@ -81,6 +81,7 @@ public boolean equals(Object obj) {
     public ConstrainedTerm(Term term, ConjunctiveFormula constraint, TermContext context) {
         this.data = new Data(term, constraint);
         this.context = context;
+        this.context.setTopTerm(term);
```
daejunpark, line 84: `TermContext.topTerm` is never used, so this change shouldn't affect others.

daejunpark (approved):

daejunpark:
The jenkins failure has nothing to do with this PR and I merge this.

PR #2356, opened 2017-10-31
add nowrap output option

yzhang90:
@daejunpark  please review.


daejunpark (approved):

msaxena2:
@daejunpark @yzhang90 Guys we're trying to follow much better guidelines for commit messages - in line with the evm-semantics. It would've been nicer if one of the other maintainers was also tagged for the review (so that we may have pointed the issue out).

daejunpark:
@msaxena2 What guideline are you talking about? I do not recall that I have discussed such a thing.

PR #2358, opened 2017-11-03
SMTOperations: using KExceptionManager for missing SMTLib translation…

lucaspena:
… output with prover

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/kil/GlobalContext.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/kil/GlobalContext.java
@@ -6,7 +6,6 @@ import org.kframework.KapiGlobal;
 import org.kframework.backend.java.kil.KItem.KItemOperations;
 import org.kframework.backend.java.symbolic.BuiltinFunction;
 import org.kframework.backend.java.symbolic.Equality.EqualityOperations;
-import org.kframework.backend.java.symbolic.JavaExecutionOptions;
 import org.kframework.backend.java.symbolic.SMTOperations;
 import org.kframework.backend.java.symbolic.Stage;
 import org.kframework.backend.java.util.Z3Wrapper;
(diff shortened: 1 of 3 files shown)


ehildenb:
@msaxena2 why does it say "Passed" above but when I actually login to Jenkins it says "FAILED"?

java-backend/src/main/java/org/kframework/backend/java/kil/GlobalContext.java
```diff
@@ -53,7 +52,7 @@ public GlobalContext(
         this.hookProvider = hookProvider;
         this.files = files;
         this.equalityOps = new EqualityOperations(() -> def);
-        this.constraintOps = new SMTOperations(() -> def, smtOptions, new Z3Wrapper(smtOptions, kem, globalOptions, files));
+        this.constraintOps = new SMTOperations(() -> def, smtOptions, new Z3Wrapper(smtOptions, kem, globalOptions, files), globalOptions);
```
andreistefanescu: maybe pass `kem` directly to `SMTOperations`, since the `globalOptions` are not used for anything else

andreistefanescu (approved):
lgtm, alternatively I guess you could just not log the warnings if `--debug` is not set

java-backend/src/main/java/org/kframework/backend/java/symbolic/SMTOperations.java
```diff
@@ -16,12 +19,17 @@
     private final SMTOptions smtOptions;
     private final Z3Wrapper z3;
 
+    private final KExceptionManager kem;
+    private HashSet<String> missingSMT;
```
ehildenb: @andreistefanescu one thing I wasn't sure about was if this should be `final` or `static` or something like that. Whichever one means "there is only one copy of this field for all the objects". Alternatively, if only one instance of the `SMTOperations` class is ever made, then I don't see doing that being necessary.

andreistefanescu: well, `static` means per class instead of per object. However, I don't remember whether it was possible/recommended to have multiple instances. Just pass the instance from the `GlobalContext` constructor to the `SMTOperations` constructor, and avoid this issue.

ehildenb:
Hey @andreistefanescu, given the discussion above, I think the correct thing to do is:

-   Modify the KEM to check if a given warning has already been produced (instead of having `SMTOperations` perform the local/specific check).
-   Pass the KEM to `SMTOperations` instead of `globalOperations`.

Does that sound correct?

andreistefanescu:
nit: I would pass the `kem` to `SMTOperations`, but I would keep the caching in `SMTOperations`, mostly because one may want to get all the instances for more interesting errors.

lucaspena:
@andreistefanescu we amended our original commit to pass `kem` directly to `SMTOperations` as you suggested.

ehildenb:
@lucaspena please review the most recent changes.

PR #2359, opened 2017-11-04
Improve user interaction with prover

msaxena2:
* Files don't need end with "spec.k".
 * "true" is not printed when no claims are found. A more appropriate
message is used instead.
 * Claims can be introduce by adding the "claim" attribute to a rule.
 * Modules now have ".claims" field for fetching Claims.


daejunpark:
Jenkins: test this please

PR #2369, opened 2017-11-09
Prover changes for EVM ABI verification

bmmoore:
Adds smtlib attributes to a few builtins, and changes the generated Z3 querys to something which is much faster in some proofs.


msaxena2:
add to whitelist

msaxena2:
test this please

msaxena2:
test this please

msaxena2:
test this please

msaxena2:
test this please

msaxena2:
test this please

msaxena2:
test this please

msaxena2:
test this please

msaxena2:
test this please

java-backend/src/main/java/org/kframework/backend/java/symbolic/KILtoSMTLib.java
```diff
@@ -194,9 +195,9 @@ public static String translateImplication(
         sb.append(leftTransformer.getConstantDeclarations(Sets.difference(
                 Sets.union(leftTransformer.variables(), rightTransformer.variables()),
                 rightHandSideOnlyVariables)));
-        sb.append("(assert (and ");
+        sb.append("(assert ");
         sb.append(leftExpression);
-        sb.append(" (not ");
+        sb.append(")\n(check-sat)\n(assert (not ");
```
andreistefanescu, line 200: normally, I would request having both options + a flag, since this changes behavior otherwise (e.g. when the first `check-sat` timeouts trying to find a model, it would have worked in the previous version). however, if it's too complicated and you only plan to use the engine for this project, you may do the change

andreistefanescu (approved):

java-backend/src/main/java/org/kframework/backend/java/symbolic/KILtoSMTLib.java
bmmoore: Do you think it's likely that the first `(check-sat)` would timeout but the other version would be proved in a reasonable amount of time?

andreistefanescu: Yes, especially if there are universally quantified axioms, in which case the `check-sat` that would report `sat` tries to prove that the axioms are consistent, which can be undecidable. However, the `check-sat` that would respond `unsat` just need to find one way to derive `false`.

This may not be an issue for `evm` verification, but it may be an issue for the standard c\java\javascript verification. That's why I defer to your judgment.

bmmoore:
The string builtin seems to be problematic for some tests, so I'm making a smaller version of this pull request.

PR #2371, opened 2017-11-10
cache z3 implication queries; do not add redundant constraints when a…

andreistefanescu:
…pplying a rule


andreistefanescu:
now `transferFrom-then` takes less than 5 minutes

ehildenb:
@andreistefanescu you're a hero!

java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
```diff
@@ -234,7 +235,16 @@ public Term term() {
...
+            if (candidate.substitution().keySet().equals(variables)
+                    && !candidate.isSubstitution()
+                    && subjectConstraint.implies(ConjunctiveFormula.of(context.global()).addAll(candidateConstraint.equalities()), Sets.newHashSet())) {
+                solutions.add(Triple.of(
+                        subjectConstraint.addAndSimplify(candidateConstraint.substitution(), context),
+                        true,
+                        pair.getLeft()));
+            } else {
+                solutions.add(Triple.of(solution, isMatching, pair.getLeft()));
+            }
```
bmmoore, line 257: What is this block doing with the new constraint?
It looks like we end up in the `then` case when the existing `subjectConstraint` implies the new candidate constraint.
If that happens only when the new constraint is redundant, why are we modifying the existing constraint at all?

Also, it looks like we are returning a list of distinct ways to satisfy a possibly-disjunctive constraint, but if multiple disjuncts are implied by current assumptions, shouldn't we just return one solution?

Furthermore, if some disjuncts are implied by the current assumptions, is there ever a reason to additional return additional solutions that add extra constraints? Wouldn't that be a subset of the states satisfying the current constraint?

bmmoore (changes requested):

bmmoore:
The caching part looks just fine, maybe we could merge that part immediately if there is anything to change in the other part.

I think in our proofs it is more common that some of the current constraints are *implied by* the new constraint, and can be removed to simplify the problem.

java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
daejunpark: @bmmoore for the first question, I think `subjectContraint` implies only the equalities of `candidateConstraint`, and we need to apply the substitution of `candidateConstraint` into `subjectContraint`.

daejunpark: This part is not super clear to me as well, but at least it doesn't appear to hurt the correctness, and I think we can go with it for now if it's urgent.

daejunpark (approved):

java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
andreistefanescu: re: why modify the existing constraint?
we are adding the substitution from the matching to the `subjectConstraint` in order to do the rewriting, and then the substitution is removed. This is what happens even if it's a syntactic matching.

re: why return multiple identical solutions?
That could happen, and they will be collapsed later when all the results are put in a set (unless for some other reason the results don't compare equal). I guess one can use a set here, but it's not a frequent occurrence.

re: current assumptions imply some of the disjunctions
Correct.

General comment: the disjunctions come from rule matching, and not from the `requires` clauses (those remain `orBool`, which was my way of avoiding the problem altogether). So I don't think disjunctions are a issue here.

andreistefanescu (commented):
I don't think that parts of the current constraint are implied by the new constraints in the case of `transferFrom-then` (I checked the debug output, with this change nothing is added to the formula, but the `then` branch is taken a lot). In general, one needs to implement more advanced formula simplification, but I don't have a quick hack right now.

bmmoore (approved):

PR #2372, opened 2017-11-15
add power-modulo operator to UIUC K

dwightguth:
This is needed to execute the power-modulo operation in the modexp precompiled contract in the EVM semantics.


bmmoore (approved):
Looks good.
Z3 has no builtin function for a powmod operator so I'm fine leaving out an smtlib attribute until needed.

dwightguth:
when we need it we can just represent it with a macro in the prelude that does power and modulo directly. but until we verify contracts that call precompiled contracts it won't be an issue.

daejunpark (approved):

PR #2378, opened 2017-11-28
Fix a bug in negMInt

daejunpark:
Rebase #2374


daejunpark:
@msaxena2 please review

PR #2379, opened 2017-11-30
encoding int map as smt array

daejunpark:
Add `--map-as-int-array` option, enabling to reason about symbolic maps (especially their equivalence) whose keys and values are of integers. Such int maps are common in low-level languages. See the added tests to see what are supported.


k-distribution/samples-kore/impa/2_spec.k
```diff
@@ -0,0 +1,12 @@
+module SPEC
+import IMP
+
+rule <T>
+  <k> store(Loc:Int, Key:Int, Val:Int) ~>
+      store(Loc2:Int, Key:Int, Val:Int) => .K ... </k>
+  <store> Loc |-> array(M11:Map => ?M12:Map)
+          Loc2 |-> array(M21:Map => ?M22:Map) ... </store>
+</T>
+  requires M11 ==K M21
```
andreistefanescu, line 10: nit: `M1` instead of `M11` and `M21` and `?M2` instead of `?M12` and `?M22`

k-distribution/samples-kore/impa/1_spec.k
```diff
@@ -0,0 +1,10 @@
+module SPEC
+import IMP
+
+rule <T>
+  <k> store(Loc:Int, Key:Int, Val:Int) => .K ... </k>
+  <store> Loc |-> array(M1:Map => ?M2:Map) ... </store>
+</T>
+  requires M1[Key] ==K Val
+  ensures M1 ==K ?M2
```
andreistefanescu, line 9: nit: `M` instead of `M1` and `?M2`

API/src/main/java/org/kframework/Kapi.java
```diff
@@ -311,6 +311,8 @@ public static Info getInfo(CompiledDefinition compiledDef, String proofFile, Str
 
         //// setting options
 
+        globalOptions.debug = true;
```
andreistefanescu, line 314: is the `--debug` flag not propagated properly?

andreistefanescu (approved):
keep in mind that the smt arrays are infinite, while the k maps are finite, so this needs to be applied with care...

k-distribution/samples-kore/impa/1_spec.k
daejunpark: True, but it is intentional to simulate keq-generated constraints.

k-distribution/samples-kore/impa/2_spec.k
daejunpark: Again, it is intentional to simulate keq-generated constraints.

API/src/main/java/org/kframework/Kapi.java
daejunpark: will fix in a separate PR.

daejunpark:
You're right. Indeed, we devised an encoding of the partial function using the array theory, but haven't incorporated into the java backend. In the meantime, this can be used in a restricted way as you noted.
> keep in mind that the smt arrays are infinite, while the k maps are finite, so this needs to be applied with care...

PR #2382, opened 2017-12-08
Add [concrete] attribute for [function] symbols

daejunpark:
A concrete function is supposed to be evaluated only when all of the given
arguments are concrete.


andreistefanescu (approved):
I'm curios, what is the context for this?

daejunpark:
A rule in the spec.k file has a higher priority than semantic rules, which can be used to abstract the semantics. But it doesn't work for function rules, since the functions are evaluated immediately. The [concrete] attribute allows to abstract the functional rules.

daejunpark:
Jenkins: retest this please

PR #2391, opened 2018-01-08
Cache function/anywhere symbol evaluation

andreistefanescu:
Check if the `function`/`anywhere` symbols have been evaluated before, and if the substitution does not contain any of the variables in the term, then do not attempt to evaluate again. This is not sound if we attempt to evaluate with a different set of rules. The substitution and term variables intersection is also more expensive than the previous check, so this change may result in reduce performance in other cases, but on the `evm` proofs is speeds things up.

I don't remember, how can I run all tests to make sure that this doesn't break anything?

### DIFF
--- a/java-backend/src/main/java/org/kframework/backend/java/kil/JavaSymbolicObject.java
+++ b/java-backend/src/main/java/org/kframework/backend/java/kil/JavaSymbolicObject.java
@@ -2,7 +2,9 @@
 
 package org.kframework.backend.java.kil;
 
+import com.google.common.collect.Sets;
 import org.kframework.backend.java.symbolic.BinderSubstitutionTransformer;
+import org.kframework.backend.java.symbolic.ConjunctiveFormula;
 import org.kframework.backend.java.symbolic.IncrementalCollector;
 import org.kframework.backend.java.symbolic.LocalVisitor;
 import org.kframework.backend.java.symbolic.SubstitutionTransformer;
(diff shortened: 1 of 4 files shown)


cos (approved):
Looks good. Probably @dwightguth will pick it up, move it to RV, and figure out the tests there.

daejunpark:
@andreistefanescu This is not the case when the semantics is fixed during the execution/verification, right?
> This is not sound if we attempt to evaluate with a different set of rules.

Regarding the viper erc20 proof, I'm running it and will let you know.
> I don't remember, how can I run all tests to make sure that this doesn't break anything?

andreistefanescu:
@daejunpark this should be correct, I'm just not entirely sure

daejunpark:
@andreistefanescu Um, I found ~10% speed up, but some proofs failed. Will look into it later.

andreistefanescu:
The problem with the previous version was that I forgot that symbol evaluation happens under a constraint, and if that changes (e.g. after a split), then the evaluation result may change.

I'm not sure about the second commit, it makes the verification a little slower, but might improve concrete execution. Feel free to merge it or not.

daejunpark:
@andreistefanescu Looks good to me. Note that I fixed a small space issue. I'll merge it.

Just to make sure, how much was the verification slowdown by the second commit? You mean the ERC20 verification, right? I'm thinking not to merge the second commit since the concrete execution will be anyway done by RV-K.

daejunpark:
I merged the first commit only for now, and leave the second commit in the branch. I'll keep this branch for the record.

PR #2394, opened 2018-01-12
add hook KREFLECTION.isConcrete

daejunpark:
Add `KREFLECTION.isConcrete` hook which checks if the given term has no variables and no unresolved functions.


daejunpark:
I merge this to release erc20 verification on time. We can revert it later if a problem found.

PR #2396, opened 2018-01-14
frontend/pom.xml: kore dependency version 1.1 => 0.1

msaxena2:
- K depends on a BETA version of kore(2.12-0.1). The latest release (2.12-1.1) is incompatible.
- This fixes the breaking master build.


msaxena2:
Since this is a small patch to fix a broken master, I'm merging it without a review. We can revert later, if the need arises.

PR #2399, opened 2018-01-25
fix top constraint

andreistefanescu:
patch for `transferFrom-failure-1-spec`. The specific issue here is that `chop(BAL_TO_1347:Int +Int VALUE_1344:Int) >Int BAL_TO_1347:Int))` is being evaluated to `BAL_TO_1347:Int +Int VALUE_1344:Int >Int BAL_TO_1347:Int` when that exact inequality is in the `topConstraint`. This is sound, and makes a call to z3, but results in information loss. I don't have a simple solution, but this patch should unblock @daejunpark 

more generally, this would require better control over how constraints are evaluated


daejunpark:
@andreistefanescu could you explain a bit more about the information loss and how this resetting the topconstraint solves the problem?

andreistefanescu:
Currently, we use the existing constraint ("top constraint") when we apply functions inside terms and formulae. If we set it before trying to take a rewrite step, and don't unset it, it will be used to simplify anything, including itself. In particular, when the constraint has the `chop(BAL_TO_1347:Int +Int VALUE_1344:Int) >Int BAL_TO_1347:Int` inequality (and also that both `BAL_TO_1347` and `VALUE_1344:Int` are between `0` and `2^256`), it will attempt to apply the `chop` function rule `chop(I) => I if 0 <= I <2^256`, and it will succeed, because `chop(BAL_TO_1347:Int +Int VALUE_1344:Int) >Int BAL_TO_1347:Int ==> 0 <= BAL_TO_1347:Int +Int VALUE_1344:Int < 2^256`. But then, the resulting constraint only has `BAL_TO_1347:Int +Int VALUE_1344:Int >Int BAL_TO_1347:Int`, which is only implied by the initial constraint, but not equivalent, hence the information loss. Because of this, the constraint for the execution path without exceptions becomes satisfiable, and the proof fails. This is sound but incomplete, so my fix is to avoid simplifying a constraint using itself.

java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
```diff
@@ -163,6 +163,7 @@ public ConjunctiveFormula matchImplies(ConstrainedTerm constrainedTerm, boolean
             return null;
         }
 
+        context.setTopConstraint(null);
```
daejunpark, line 166: no reset required?

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -195,6 +195,8 @@ private boolean strategyIsSave(ConstrainedTerm subject) {
                 theNew = theNew.substituteAndEvaluate(substitution, subject.termContext());
             }
 
+            subject.termContext().setTopConstraint(null);
+
```
daejunpark: no reset required?

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
```diff
@@ -382,6 +384,7 @@ public static ConstrainedTerm buildResult(
         Set<Variable> substitutedVars = Sets.union(rule.freshConstants(), rule.matchingVariables());
         constraint = constraint.orientSubstitution(substitutedVars);
         Term term = rule.rightHandSide().substituteAndEvaluate(constraint.substitution(), context);
+        context.setTopConstraint(null);
```
daejunpark: no reset required?

java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
andreistefanescu: ah, this is mostly as a best practice, it marks the end point after which there shouldn't be any simplification with this constraint. I think it adds clarity, and also reduces the risk that this constraint may be used in an unrelated context that does not set a top constraint.

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
daejunpark, line 387: No reset required? I mean, `result` below (line 396 or 399) is constructed with `context` and returned without the existing top constraint. Why is this safe?

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
daejunpark, line 199: No reset required? I mean, `result` below (line 211 or 213) is constructed with `subject.termContext()` and returned without the existing top constraint. Why is this safe?

java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
daejunpark: Is this safe because the top constraint will be set anyway later like the above (line 153)?

java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
```diff
@@ -221,7 +222,10 @@ public Term term() {
             Pair<Map<scala.collection.immutable.List<Pair<Integer, Integer>>, Term>, ConjunctiveFormula> pair = ConstrainedTerm.splitRewrites(candidate);
             ConjunctiveFormula candidateConstraint = pair.getRight();
 
+            context.setTopConstraint(null);
             ConjunctiveFormula solution = candidateConstraint.addAndSimplify(subjectConstraint, context);
```
daejunpark, line 226: I can see that `subjectConstraint` is not used to simply `subjectConstraint` itself, but it will not used to simply `candidateConstraint` as well. Would it be OK?

java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
daejunpark: I see. And it will be set again later like the above (line 153) when needed, right?

java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
andreistefanescu: `candidateConstraint` should already be simplified

daejunpark: I see. Then the below should be OK because of the same reason.

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
andreistefanescu: constructing the `ConstrainedTerm` only creates the object, without any further simplification; the pattern expansion afterwards is affected, but that is complicated, and not used in `evm`

daejunpark (approved):

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
andreistefanescu: same as above

java-backend/src/main/java/org/kframework/backend/java/symbolic/SymbolicRewriter.java
daejunpark: I see. Not it's clear. Thanks!

java-backend/src/main/java/org/kframework/backend/java/kil/ConstrainedTerm.java
andreistefanescu: yes. Ideally, we should be passing the constraint as an argument everywhere (instead of using `TermContext`), and that would clarify things, but that's quite a large change

daejunpark:
@andreistefanescu this change also improves the prover performance for the hkg proofs as well, which is great! But, unfortunately, it introduces an exiting proof failure: `viper/transferFrom-failure-2-spec.k`. Could you take a look?

PR #2400, opened 2018-01-27
context.setTopConstraint after constraint.orientSubstitution in build…

andreistefanescu:
…Result


andreistefanescu:
constraint before `orientSubstitution`: `B <- A /\ A > C /\ C > 0`
constraint after `orientSubstitution`: `A <- B /\ B > C /\ C > 0`
`rule.rightHandSide()`: `foo(A > 0)`
after applying the substitution: `foo(B > 0)`
The bug is that the prover tries to prove the implication `A > C /\ C > 0 ==> B > 0`, and obviously fails. Fixing it by setting the top constraint after orienting the substitution.

daejunpark (approved):

daejunpark:
I'll merge this once I confirm that all of the proofs are passed.

daejunpark:
All proofs are passed, so I merge.

